Issue #5, opened 2015-11-18
use of source code from tpm2.0-tss tests


flihp:
Resolving #3 will fix much of this. Still though there are bits of the tools that reference source files that aren't exposed as libraries. Specifically the use of code from the SAPI test programs here:

https://github.com/01org/tpm2.0-tools/blob/master/sapi-tools/Makefile.am#L75
https://github.com/01org/tpm2.0-tools/blob/master/sapi-tools/Makefile.am#L58

is very strange and unfriendly to packaging. If these source files are useful to code outside of the sapi tree they need to be exposed as a library.

flihp:
This issue is terribly general. It's a real issue but difficult to take meaningful action on due to the scope of the problem. It needs to be decomposed into something actionable. This is in progress.

gwei3:
@flihp, do you need more discussion or help on this?

flihp:
if you've got ideas I'd love to hear them. My previous comment is mostly a precursor to me closing this overly general issue and breaking it up into smaller more actionable tasks.

gwei3:
I would like to duplicate all the source code under TPM2.0-TSS/test/common/sample/ to tpm2.0-tools/sapi-tools/utilities/ folder, remove the source code not used in the new location, thus to decouple from source code of TPM2.0-TSS first.

flihp:
That's one way to do it. Alternatively we could move the tools into a 'tools' directory in the TPM2.0-TSS repo and eliminate the duplication all together. I'm sure there are drawbacks to this approach as well but with a little git magic I'm pretty sure we can preserve the short history of the tools repo. I haven't thought this through much though. Thoughts?

wcarthur1:
I’d rather not duplicate code if at all possible—maintenance nightmare.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

flihp:
I think that's a :+1: for moving the migrating the tools over to the TPM2.0-TSS repo.

gwei3:
We'd better resolve the dependency issue without migrating. We are regarding the tpm2.0-tools as the first app build upon TPM2.0-TSS, which is a really good example to showcase how to build apps based on TPM2.0-TSS, and can help finding and resolve the problem for how to prepare TPM2.0-TSS for using by standalone apps. We can migrate tpm2.0-tools, but we can't migrate all future apps developed by different people.

flihp:
[...]

PR #16, opened 2016-01-12
Copycommon

flihp:
Per request by @gwei3 in #5 the solution we're pursuing is to copy the common code from the TPM2.0-TSS into the tools repo.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -1,15 +1,5 @@
 #! /bin/sh
 
-TSS_DIR=./tss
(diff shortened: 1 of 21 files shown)


wcarthur1:
Is the issue cases where the tpmclient/tpmtest routines use internal non-specified data structures?

In that case, I’d just remove those tests from tpmtest.

In order to do some of the API tests, tpmclient uses some internal data structures, and in many, if not most, cases I documented that with “don’t do this in normal apps” or something to that effect.

Will Arthur
Intel Corporation
Server Security Firmware
803-216-2101

PR #20, opened 2016-01-24
common: Do not include tpm2sapi/tss2_sapi_util.h

### LINKED ISSUE #19: TCTI_MAGIC and VERSION
The header in the TSS2 repo that exposes these values is a non-standard header and it will go away soon. We need to get them from elsewhere, ideally from the TCTI headers that are implementation specific.

flihp:
This file no longer pulls in the TCTI_MAGIC / TCTI_VERSION. It is also
not a standard tss2 header so it will go away soon. These are values
come from a TCTI implementation and so we get them from the respective
TCTI headers instead now.

This resolves #19

### DIFF
--- a/sapi-tools/common.cpp
+++ b/sapi-tools/common.cpp
@@ -62,7 +62,6 @@
 
 #include <tpm2sapi/tpm20.h>
 #include "sample.h"
-#include <tpm2sapi/tss2_sysapi_util.h>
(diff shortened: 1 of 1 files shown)


flihp:
Please note that this PR depends on a PR from the TPM2.0-TSS repo. Please time acceptance of this PR with https://github.com/01org/TPM2.0-TSS/pull/110

PR #52, opened 2016-03-14
nvreadlock

atomsoftwarestudios:
I added the tpm2_nvreadlock tool to lock defined nv ram location for further reads until the machine is rebooted.


Makefile
```diff
@@ -0,0 +1,1398 @@
+# Makefile.in generated by automake 1.14.1 from Makefile.am.
```
gwei3: You should not change and check in the Makefile directly, instead, please add you changes into Makefile.am, then using ./configure to generate the Makefile.

src/tpm2_nvreadlock.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if(rval != TPM_RC_SUCCESS)
+    {
+        printf("\nFailed to lock NVRAM area at index 0x%x (%d).Error:0x%x\n", nvIndex, nvIndex, rval );
+        return -1;
+    }
+
+    return 0;
+}
+
+void showHelp(const char *name)
```
gwei3: Please update showHelp function to remove wrong options brought in by copy-paste, like -s -o etc.

src/tpm2_nvreadlock.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    if( argc > (int)(2*sizeof(sOpts)/sizeof(struct option)) )
+    {
+        showArgMismatch(argv[0]);
+        return -1;
+    }
+
+    while ( ( opt = getopt_long( argc, argv, "x:a:s:o:P:p:d:hv", sOpts, NULL ) ) != -1 )
+    {
```
gwei3: "o: " must be removed too.

gwei3:
BTW, please also update manual to include the usage for tpm2_nvreadlock.

Makefile
atomsoftwarestudios: I am sorry, I am quiet new to it, let me do it right :)

PR #70, opened 2016-05-26
Android Make file

erezgeva:
Android Make file to build both tpm2.0-tools and TPM2.0-TSS.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -1,0 +1,85 @@
+LOCAL_PATH := $(call my-dir)
+
+# This project
+TPM2_TPM_BASE=.
+# Download from https://github.com/01org/TPM2.0-TSS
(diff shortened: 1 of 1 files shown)


gwei3:
It is fine to check in the makefile for android, but can you help to maintain it?

erezgeva:
I was using it with AOSP, there you do not need to check. 
The AOSP build machine search for Android.mk by default, ignoring all other make files.
I built it minimalistic and simple as possible for easy to maintain.
I will help gladly with its maintenance.
Since I do not use Android NDK and don't have time for that. I'd be nice if someone could test that with Android NDK.
I saw that the tree structure was change so the Make file can be changed.
It is also possible to split the Android make file to 2, one for the TSS and one for the tpm tools.

PR #77, opened 2016-07-05
pkg-config

flihp:
Use pkg-config to locate the SAPI, TCTI, libcurl and libcrypto build flags.


flihp:
The merging of #218 in TPM2.0-TSS requires this. Merging this w/o review to keep the build functional.

Issue #71, opened 2016-06-14
Error codes


mikzaq:
Is there any way to decipher the error codes that get thrown by the commands? I am tryng to use tpm2_nvdefine and am getting a 0x14b error, but I have no idea what this means.

flihp:
This sounds like a feature request to me. Setting the label.

flihp:
Assigning to myself. I'm half way through a tool for this.

gwei3:
This was fixed by the recently merged PR #82.

PR #78, opened 2016-07-05
rc-decode: Add header with simple functions to decode TPM_RCs.

flihp:
Use cmocka to unit test it. Add .gitignore rules to ignore test logs. This is a small step toward resolving #71.


flihp:
replaced by #82

PR #89, opened 2016-07-18
Update tpm2_rc_decode

DamienMolina:
short argument for the version


flihp:
Good catch. Pushed with a few changes:

1. Added a more complete commit message
1. Removed unnecessary changes to the license header
1. Removed unnecessary changes to the help message

PR #95, opened 2016-08-10
configure.ac: Allow for the unit tests to be enabled selectively.

flihp:
This makes the dependence on libcmocka optional.


flihp:
this resolves #95

PR #96, opened 2016-08-14
common: Remove use of TeardownSocketTcti.

flihp:
This function has been removed from the socket TCTI upstream. All finalization
should be done through the finalize function pointer in the TCTI
structure. tss2_tcti_finalize is the right helper macro for this. We must
deallocate the TCTI context structure ourselves now too. Finally this patch
also removes the return value from the TeardownTctiResMgrContext function.
None of the finalization functions return any data and we were never checking
this value anyways.


flihp:
This resoves #97

PR #110, opened 2016-11-29
tpm-tools: convert all cpp to c files

williamcroberts:
Drop the dependency on C++ as we were not really using
any compelling language features.


williamcroberts:
FYI You will see some various whitespace changes, I fixed whatever git diff --check complained about.

gwei3:
Good idea, but need more careful review. I will review it.

williamcroberts:
The diffstat is relatively small (15 mins is at most what you would need to review), can I get you to look at this by 12/09/2016?

gwei3:
@williamcroberts, sure, I will review it before end of this week.

src/tpm2_akparse.c
```diff
@@ -137,38 +137,36 @@ int SaveEccKeyToFile(const char keyfile[], UINT16 algType, const TPMS_ECC_POINT
...
-    else
-    {
-        printf("\nThe algorithm type(0x%4.4x) is not supported\n", outPublic.t.publicArea.type);
-        return -6;
+        case TPM_ALG_RSA:
+            sizedBuffer = (TPM2B *)&outPublic.t.publicArea.unique.rsa.t;
+        break;
+        case TPM_ALG_ECC:
+            sizedBuffer = (TPM2B *)&outPublic.t.publicArea.unique.ecc;
+        break;
```
gwei3: ECC case should be handled specially since ecc is not a TPM2B compatible structure. we should use SaveEccKeyToFile() to save it instead of SaveKeyToFile().

src/tpm2_getmanufec.c
```diff
@@ -271,18 +272,24 @@ unsigned char *HashEKPublicKey(void)
...
         printf("File Open Error\n");
-    else 
+    else
     {
         fseek(fp, 0x66, 0);
-        fread(EKpubKey, 1, 256, fp);
+        size_t read = fread(EKpubKey, 1, 256, fp);
+        if (read != 256)
+        {
+            printf ("Could not read whole file.");
```
gwei3: Before return NULL, we should free the allocated memory for hash, and also need to close the opened file.

williamcroberts: Sure, but the old code didn't check this, but will fix this issue since were here.

williamcroberts: There's even more issues to fix here, if the alloc fails we will use a NULL ptr.

src/tpm2_akparse.c
```diff
@@ -138,38 +138,35 @@ int parseAKPublicArea()
...
+        case TPM_ALG_RSA:
+            sizedBuffer = &outPublic.t.publicArea.unique.rsa.b;
+            /* no break */
+        case TPM_ALG_KEYEDHASH:
+            sizedBuffer = &outPublic.t.publicArea.unique.keyedHash.b;
+            /* no break */
+        case TPM_ALG_SYMCIPHER:
+            sizedBuffer = &outPublic.t.publicArea.unique.sym.b;
+            rc = SaveKeyToFile(akKeyFile, outPublic.t.publicArea.type, sizedBuffer) ? -2 : 0;
+        break;
```
gwei3: You should not do it this way, this will not function as your expectation: no matter outPublic.t.publicArea.type is TPM_ALG_RSA, TPM_ALG_KEYEDHASH, or TPM_ALG_SYMCIPHER, the final result will be sizedBuffer = &outPublic.t.publicArea.unique.sym.b;

src/tpm2_getmanufec.c
gwei3: ok, let me merge your big PR and then create separate fix for other issues.

PR #112, opened 2016-12-05
travis: correct wget over http

williamcroberts:
Use a a protocol that is not susceptable to man-in-the-
middle attacks for downloading a binary to the CI server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ addons:
     - libcurl4-openssl-dev
 
 install: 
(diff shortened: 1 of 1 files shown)


flihp:
Thumbs up for the use of https but given Sourceforge's security record you may want to check the hash of the tarball too before you build it :stuck_out_tongue_winking_eye:

williamcroberts:
Yeah we should, not really sure how those travis files work.

gwei3:
Good catching! thanks for the fix.

Issue #87, opened 2016-07-16
Make tools configurable for TCTI interface selection


gwei3:
So far, tools can only use resourcemgr TCTI implementation. It would be a good direction to enhance it to be able to use other existing TCTI implementation such as device TCTI, simulator TCTI, etc, and also make it configurable. Further, it will even better to make the TCTI choices extensible without code change.

flihp:
Enabling TCTIs at configure time would be my preference here. autoconf supports options to the configure script like --with-tcti-device / --with-tcti-socket. This would give optimum flexibility at the expense of some complexity in the build.

flihp:
This is addressed in PR #113. Not applied to all tools yet but provides framework & a few new tools to model transition for remaining tools.

flihp:
Closing after merge of #113. Still need to port tools over though. Should have a new ticket.

PR #113, opened 2016-12-07
Dynamic TCTIs, man page framework, and 3 tools

flihp:
This is an uber-PR so I'll apologize in advance. Each commit has a pretty thorough description so I'll just give a quick summary here:

At it's core this PR adds the machinery necessary to allow dynamic selection and configuration of TCTI modules. I've added support for both the device and the socket TCTI. Either can be selected or configured through the environment or the command line. The details are documented in the manpages for the new tools this PR adds.

Additionally the configure.ac / Makefile.am have been instrumented to allow TCTI modules to be excluded @ configure / build time. These mechanisms should be used as a reference if / when support for new / additional TCTIs are added. Since all of the existing tools are hard coded to the socket TCTI it is the default. If the socket TCTI is disabled these tools will not be built.

Additionally this PR adds 3 new tools. All of which have been integrated into the test scripts. They are documented in their respective man pages.

Finally some notes on form:
It is my hope is that the existing tools will be ported over to use this framework in future PRs. I intend to take on this work but would gladly accept help :wink: As each tool is ported we should be making manpages etc.
I've also departed from the existing coding standard that we've been mimicking from the TSS repo, which in turn was mimicking the coding standard from the reference code in the TPM specifications. Though I don't have an specific issues with this format, it's not documented anywhere so it's very difficult to know whether or not one is following it faithfully. My plan is to migrate the TSS code over to use the format that I'm using here which is the Linux kernel format for C with one intentional deviation: 4 space indentation instead of 8.

Let me know what you think & thanks for the review!


flihp:
This will resolve #87

src/options.c
```diff
@@ -0,0 +1,303 @@
...
+ * retrieved from the environment will just be over written with whatever we
+ * get from argv.
+ */
+int
+get_common_opts (int                     argc,
+                 char                   *argv[],
+                 common_opts_t          *common_opts)
+{
+    int c = 0, option_index = 0, ret = 0;
+    char *arg_str = "t:d:n:p:hvV";
```
gwei3: n: should be a typo for a: above. Further, I would suggest using '-T' for "--tcti", '-R' for "--socket-address" to avoid conflicts with short options used in existing tools. Change short options in existing tools will cause confusion for users.

flihp:
Thanks for the review @gwei3. I should have time tomorrow to make the changes you've requested.

flihp:
Just realized that I forgot to include the license text in my new files. Will rewrite this PR again with the license added momentarily.

flihp:
Thanks @gwei3, interested to know what else is missing. If you're willing to share I'm happy to integrate whatever additional features you think are necessary. May be worth creating 'issues' for this stuff so we can keep track.

gwei3:
@flihp, please refer to #114. Just two typos for the short options in the switch clause.

PR #115, opened 2016-12-13
fix two code issues

gwei3:
issue 1: the switch clause in tpm2_akparse.c is problematic.
issue 2: some malloc was not followed by error handling check.

### DIFF
--- a/src/tpm2_akparse.c
+++ b/src/tpm2_akparse.c
@@ -141,29 +141,33 @@ int parseAKPublicArea()
     TPM2B *sizedBuffer;
     int rc = -1;
 
-    if ( loadDataFromFile(akDataFile, (UINT8 *)&outPublic, &size) )
+    if( loadDataFromFile(akDataFile, (UINT8 *)&outPublic, &size) )
     {
         return -1;
     }
 
-    switch ( outPublic.t.publicArea.type )
+    if( outPublic.t.publicArea.type == TPM_ALG_ECC )
     {
+        rc = SaveEccKeyToFile (akKeyFile, outPublic.t.publicArea.type, &outPublic.t.publicArea.unique.ecc) ? -2 : 0;
+    }
+    else
+    {
+        switch( outPublic.t.publicArea.type )
+        {
         case TPM_ALG_RSA:
             sizedBuffer = &outPublic.t.publicArea.unique.rsa.b;
-            /* no break */
+            break;
         case TPM_ALG_KEYEDHASH:
             sizedBuffer = &outPublic.t.publicArea.unique.keyedHash.b;
-            /* no break */
+            break;
         case TPM_ALG_SYMCIPHER:
             sizedBuffer = &outPublic.t.publicArea.unique.sym.b;
-            rc = SaveKeyToFile(akKeyFile, outPublic.t.publicArea.type, sizedBuffer) ? -2 : 0;
-        break;
-        case TPM_ALG_ECC:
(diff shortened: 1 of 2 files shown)


gwei3:
@flihp, please help double check my fixes.

flihp:
A few general comments on these commits:
1. In 0c0d34266c661bac9eaca80142752b60a2fed154 it's pretty clear what you're doing and why. It is however a bit of a mystery why you `return -1` in one case and `return -99` in the other.
2. In d5226a0262849bcb71275d8986dd630a0fbc9f4e you've change a bunch of the logic and a few return values and without some indication as to what the return values mean or what the purpose of the function is it's a bit more difficult to review this. The logic looks right to me but it would be much easier to review if I could verify the logic with a written description of what the function is intended to do and a description of why you're making these changes. Your commit message states: "Fix the switch case clause problem in tpm2_akparse.c" but I don't know what the "switch case clause problem" is, or how the change actually fixes it.
3. The spacing around `(` and `)` in your function calls and conditionals is inconsistent. Are we putting what space around all parentheses or just in conditionals? Line [49](https://github.com/gwei3/tpm2.0-tools/blob/d5226a0262849bcb71275d8986dd630a0fbc9f4e/src/tpm2_akparse.c#L149) mixes both forms.

gwei3:
@flihp, good comments!
1. for the exit code in tools, as you can see, I used to try to use different minor numbers to indicate where the error happened so that user can check the code when meet error. When people changes the code eventually, the exit code becomes a little bit messed now. It is the time to reconsider how to define the exit code in a more suitable style. I will open a issue for the error/exit code cleanup.
2. The original code uses if/elses to handle the dispatching, when PR#110 try to change this to switch clause, there are some wrong changes which didn't keep the original logic like below:

```
int parseAKPublicArea()
{
    TPM2B_PUBLIC outPublic;
    UINT16 size = sizeof(outPublic);
    if( loadDataFromFile(akDataFile, (UINT8 *)&outPublic, &size) )
    {
        return -1;
    }

    if( TPM_ALG_RSA == outPublic.t.publicArea.type )
    {
        if(SaveKeyToFile(akKeyFile, TPM_ALG_RSA, outPublic.t.publicArea.unique.rsa.t))
            return -2;
    }
    else if( TPM_ALG_ECC == outPublic.t.publicArea.type )
    {
        if(SaveEccKeyToFile(akKeyFile, TPM_ALG_ECC, outPublic.t.publicArea.unique.ecc))
            return -3;
    }
    else if( TPM_ALG_KEYEDHASH == outPublic.t.publicArea.type )
    {
        if(SaveKeyToFile(akKeyFile, TPM_ALG_KEYEDHASH, outPublic.t.publicArea.unique.keyedHash.t))
            return -4;
    }
    else if( TPM_ALG_SYMCIPHER == outPublic.t.publicArea.type )
    {
        if(SaveKeyToFile(akKeyFile, TPM_ALG_SYMCIPHER, outPublic.t.publicArea.unique.sym.t))
            return -5;
    }
    else
    {
        printf("\nThe algorithm type(0x%4.4x) is not supported\n", outPublic.t.publicArea.type);
        return -6;
    }

    return 0;
}
```
What I did it just try to back to the equal original logic as above, except the return code.

3. oh, that is my fault. I will fix it.

flihp:
Ah I fell into the same trap you did with the 'approve' button on the code review. I guess it doesn't apply to individual patches. Either way, thanks for the clarification. Looks good to me. Will merge in a second.

Issue #111, opened 2016-11-29
findChar


williamcroberts:
Replace occurrences of findChar usage strchr instead.

gwei3:
findChar can be used to find char in a sub string w/o NULL terminator, and was used for this purpose too in the current code. So it cannot be replaced by strchr.

williamcroberts:
Ok then, switch it to memchr()

gwei3:
That's good.

PR #123, opened 2016-12-15
numerous fixes

williamcroberts:
Each commit fixes a specific thing:

1. This resolves the usages of findChar as in issue #111
2. This refactors the parsePCR code to common block
    We should consider renaming common to util.c and util.h during another refactor
3. Fixes a warning on execve usage, improper header defined.
4. test_all.sh exits 0, causing travis to think it has passed even on failure cases, fix that and print a nice summary at the end. Stop dumping test summary to files now that it's printed nicely at the end.
5. Fix Issue #124 - Use snprintf instead of that broken safeStrNCpy routine


gwei3:
Most changes are ok for me, but I need more time to review ee0ed7f5517046b134d1c4f3770f30dfa952b8e5 & 65f54bdecd1eee8bb1cfccf0cdf0fdfdd4c56c86 commits.

PR #127, opened 2016-12-21
tests: gitignore addition

williamcroberts:
Only worry about shell scripts and ignore the test artifacts that
end up in this directory.

### DIFF
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,0 +1,9 @@
+# Ignore everything
+*
+
+# ...even if they are in subdirectories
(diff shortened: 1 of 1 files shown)


test/.gitignore
```diff
@@ -0,0 +1,9 @@
+# Ignore everything
+*
+
+# ...even if they are in subdirectories
+!*/
+
+# But not these files...
+!.gitignore
+!*.sh
```
williamcroberts: and .c files!

PR #129, opened 2016-12-21
tests: gitignore addition

williamcroberts:
Only worry about shell scripts, unit tests, test fixtures,
the readme and the gitignore file itself.  and ignore the
test artifacts that end up in this directory.

### DIFF
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,0 +1,8 @@
+# Ignore everything
+*
+# But not these files...
+!.gitignore
+!README.md
+!get-capability-tpm-prop-fixed.bin
+!*.sh
+!*.c


test/.gitignore
```diff
@@ -0,0 +1,12 @@
+# Ignore everything
+*
+
+# ...even if they are in subdirectories
+!*/
```
gwei3: I think you are trying to ignore anything in subdir, so shouldn't it be as below?
*/

gwei3:
@williamcroberts looks like this need to be redo after your test cleanup changes merged.

test/.gitignore
williamcroberts: Actually the * would get subdirs as well, so we can just delete the line and the incorrect comment.

williamcroberts:
gwei3 this is OK to pull now.

PR #132, opened 2016-12-22
Test cleanup

williamcroberts:


.travis.yml
```diff
@@ -39,6 +39,6 @@ script:
   - make -j$(nproc) check
   - sudo make install
```
williamcroberts: I don't think we need this "sudo make install", we can just set PATH to point to the src dir as the first search path in the test_all.sh script.

Makefile.am
```diff
@@ -161,13 +161,13 @@ src_tpm2_send_command_SOURCES = src/tpm2_send_command.c src/main.c
...
+test_unit_tpm2_rc_decode_unit_CFLAGS  = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
+test_unit_tpm2_rc_decode_unit_LDADD   = $(CMOCKA_LIBS)
+test_unit_tpm2_rc_decode_unit_SOURCES = test/unit/tpm2-rc-decode_unit.c
 
-test_tpm2_rc_entry_unit_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
-test_tpm2_rc_entry_unit_LDADD    = $(CMOCKA_LIBS)
-test_tpm2_rc_entry_unit_SOURCES  = src/rc-decode.c test/tpm2-rc-entry_unit.c
+test_unit_tpm2_rc_entry_unit_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
+test_unit_tpm2_rc_entry_unit_LDADD    = $(CMOCKA_LIBS)
+test_unit_tpm2_rc_entry_unit_SOURCES  = src/rc-decode.c test/unit/tpm2-rc-entry_unit.c
```
williamcroberts: At some point, not now, but these unit test should be auto generated, ie drop in a C file into test/unit and it should just get picked up. There should be no need to write a main, a script could easily scub all matching prototypes into main. Also, the run method for cmock is complaining about deprecated usage.

test/system/test_all.sh
```diff
@@ -30,13 +30,10 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
 
-####before this script, please make sure all test scripts copy here, TPM is initialized.
-
 #!/bin/bash
-new_path=`dirname $0`
```
williamcroberts: We never needed this, we can just execute the script with a relative path so no need to set the script dir into PATH

Makefile.am
flihp: Some efficiency here would be nice. Unfortunately if you start doing interesting things with cmocka (like using the `-Wl,--wrap`for mocking) then you'll need to provide LDFLAGS for the unit tests. Since we're not doing anything like this (yet?) we may be able to get away with what you describe but we may be placing constraints on ourselves that make future work more difficult.

Issue #54, opened 2016-03-16
Add new tools to convert between tpm2 data and PEM/DER


gwei3:

idesai:
PR#135 is intended as a start for this purpose.

williamcroberts:
https://github.com/01org/tpm2.0-tools/pull/138

PR #135, opened 2016-12-23
Add new tools to convert between tpm2 data and PEM/DER for Issue #54

idesai:
Adding a utility to convert an RSA public key generated per TCG structure to standard PEM and DER formats. (Thanks dguerri for the excellent post)
Also adding a folder utils at the base tree for keeping utilities as such in one place.


idesai:
Perhaps a better location for a utilities folder is:    src/utils
This would keep the top level cleaner.

PR #136, opened 2016-12-29
introduce coding standard

williamcroberts:
Intorduce a coding standard and style guide for the TPM2.0-tools.

### DIFF
--- a/coding_standard_c.txt
+++ b/coding_standard_c.txt
@@ -1,0 +1,149 @@
+TPM2.0-tools Coding Standard
+
+Golden rule
+Code should hold as close to the C99 standard as possible with the exception that GCC specific extensions are generally accepted. The code must compile without warnings (for the primary target compiler) if the compiler is instructed to report all warnings.  If some warning is unavoidable, the offensive line must be documented as to why it can not be changed to remove the warning. [1]  It may also be possible to suppress specific warnings dependent on compiler options – this is acceptable but could also be dangerous if the suppression has too large a scope.
+
+Commenting
+There are four acceptable commenting styles: block, critical-line, next-line, and line.  Block comments indicate many lines of code below are being covered by the comment.  These are useful as file, function, algorithmic, or multi-line equation level comments.  Next-line comments apply only to the line of code immediately beneath them.  Critical-line comments are equivalent to next-liners, but they are formatted like block comments to indicate the line of code following (or the comment itself) is more important than normal.  Each of these three comments are formatted so that the initial "/*" is left with the code/scope to which it applies.  Line comments apply to the line on which they appear, always to the right of all code on the line.  Line comments are preferred over next-liners because they do not interrupt the vertical flow of code... use the next-line style if the line of code would exceed 80 characters when a line comment is added.  Next-line and line comments should be used sparingly... since they are rarely complete sentences, it is often better in the long run to rewrite the line of code to be understandable alone or in the context of a block comment.
+
+/*
+ * This block comment could apply to some function and describe its inner
+ * workings.  Notice these sentences have traditional capitalization and
+ * punctuation... that's because it has to be read in a way completely
+ * unlike the Post-It style content of next-line and line comments.
+ */
+
+/*
+ * This line or comment is important... at least more than usual
+ */
+
+/* the next line is regular, but needs a comment */
+
+// This is also acceptable for a line comment.
+
+// This is not an acceptable block comment.
+// These read very imbalanced.
+// Don't do this.
+
+“The size of the comment should be proportional to the size of the code that it refers to. Consequently, properties of code that can fit within a single 24-line screen should not be commented unless they are not obvious. By contrast, even obvious global properties and invariants may need to be made explicit. This doesn't have to be through comments, though. The assert() macro is an excellent executable comment.” [2]  All source files and functions require leading block comments for consistency.
+
+Naming conventions
+Variable names should clearly imply what their values mean... do not rely on the types of variables to do this.  Multiple word names are descriptive and most easily read if words are separated with the underscore character, “_”.  DoNotUseCamelBackStyle because ItIsNotAs easy to read, especially when actual whitespace IsAddedFor syntactical purposes.  It is common practice to keep variable and function names completely lowercase.  Constants are usually lowercase as well, but some groups prefer the standard of all-caps so they have the “eye feel” of older #define macros.
+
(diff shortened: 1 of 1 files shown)


coding_standard_c.txt
```diff
@@ -0,0 +1,147 @@
...
+/* Proper variable scope */
+void foo() {
+
+	do_something();
+	unsigned i;
+	for(i=0; i < 100; i++) {
+		do_more(i);
+	}
+}
+
```
flihp: I don't see how the scope of the variable `i` here is any different. Both have "function scope". I think I see what you're getting at which seems to be declaring variables as close to their use as possible?

coding_standard_c.txt
```diff
@@ -0,0 +1,147 @@
...
+Variables should only be of a signed type if something should ever be negative. A common, incorrect use, is to declare loop indices as int instead of unsigned.
+
+Formatting
+Always use tab characters rather than spaces at the beginning of lines.  Make sure your editing environment uses at least five visual spaces to represent tabs.  This will make code push horizontal limits with very little depth, thus deterring extraordinary amounts of nested statements.  The large tab approach “makes poorly-written programs blatantly ugly”. [2]  It may periodically cause small pieces of well written code to seem unruly, but it is worth it overall.
```
flihp: Tab indentation instead of spaces is going to be hard to sell here I think. I'm pretty attached to indenting with an explicit number of spaces (4 typically). This will also cause a pile of space to tab whitespace churn that will be really weird. Are you flexible on this one?

coding_standard_c.txt
```diff
@@ -0,0 +1,147 @@
...
+     statement1;
+     statement2;
+}
+
+void some_function(arg1, arg2) {
+     statement1;
+     statement2;
+}
+
+Conditional statements (such as if, else, while, for, switch, etc) should always be followed by such block statements, even if they contain only one line statement.  These formatting conditions are contrary to Kernighan and Ritchie's “one true brace style”. [3]
```
flihp: I'm like the simplicity of single line with no braces. So long as we're consistent with indentation (something gcc 6 will complain about / warn) this shouldn't be a problem for readability. I'm willing to be flexible on this one though since what you're proposing is more explicit and you're allowing for the leading brace to be on the same line as the control statement.

coding_standard_c.txt
```diff
@@ -0,0 +1,147 @@
...
+
+Line Length should not exceed 100 characters and should be split on the nearest whitespace or delimiter character.
+
```
flihp: Is 80 characters too limited? I started sticking to 80 just on principle but found that much more than 80 was a strain. Scanning my eyes horizontal too far causes me to lose track of where I started. This isn't a sticking point for me though since we've got source here that wraps across my screen and getting down under 100 would be an improvement.

Also there are some lines in this document well over 100 characters long :smile:

coding_standard_c.txt
```diff
@@ -0,0 +1,147 @@
...
+Naming conventions
+Variable names should clearly imply what their values mean... do not rely on the types of variables to do this.  Multiple word names are descriptive and most easily read if words are separated with the underscore character, “_”.  DoNotUseCamelBackStyle because ItIsNotAs easy to read, especially when actual whitespace IsAddedFor syntactical purposes.  It is common practice to keep variable and function names completely lowercase.  Constants are usually lowercase as well, but some groups prefer the standard of all-caps so they have the “eye feel” of older #define macros.
+
+unsigned int table_index = find_index(jacket_table, “color”, COLOR_RED);
+
+Boolean variables should be named to reflect the true condition... even if the variable is used to detect false conditions.
```
flihp: How do you feel about using explicit boolean types? Should we encourage the use of stdbool instead of ints?

coding_standard_c.txt
```diff
@@ -0,0 +1,147 @@
...
+
+// This is not an acceptable block comment.
+// These read very imbalanced.
+// Don't do this.
+
+“The size of the comment should be proportional to the size of the code that it refers to. Consequently, properties of code that can fit within a single 24-line screen should not be commented unless they are not obvious. By contrast, even obvious global properties and invariants may need to be made explicit. This doesn't have to be through comments, though. The assert() macro is an excellent executable comment.” [2]  All source files and functions require leading block comments for consistency.
+
+Naming conventions
+Variable names should clearly imply what their values mean... do not rely on the types of variables to do this.  Multiple word names are descriptive and most easily read if words are separated with the underscore character, “_”.  DoNotUseCamelBackStyle because ItIsNotAs easy to read, especially when actual whitespace IsAddedFor syntactical purposes.  It is common practice to keep variable and function names completely lowercase.  Constants are usually lowercase as well, but some groups prefer the standard of all-caps so they have the “eye feel” of older #define macros.
```
flihp: Yes!

coding_standard_c.txt
williamcroberts: Yes.

coding_standard_c.txt
williamcroberts: The scope is reduced even within the function since you can only reference the variable after it's declaration.

coding_standard_c.txt
williamcroberts: Spaces are fine by me.

coding_standard_c.txt
williamcroberts: There have been bugs and critical issues introduced via not bracketing single line statements. This also avoids the whole, well if one part of the block uses them, the other needs to, etc. Consider if/else if/else statements.

coding_standard_c.txt
williamcroberts: I just changed this to a 100 as Google switched, but only in Java. So I don't have any strong opinion here, 80 is fine by me.

https://google.github.io/styleguide/cppguide.html#Line_Length

coding_standard_c.txt
williamcroberts: Having it local to use helps understand why the variable exists, and it also promotes people to declare variables for use instead of using a variable over and over for different things (since it's readily available declared at function entry and adding to that list might be "hard".) Let the compiler worry about optimizing memory and register usage of the variables...

coding_standard_c.txt
williamcroberts: 4 spaces agreed as well. However, its likely that almost every line will see some change.

flihp:
Works for me. No one else has showed up with concerns or comments and I think your logic around my previous comments is reasonable. None of the issues I've raised are worth delaying this any longer since most are preferences. Merging is a few seconds.

PR #176, opened 2017-01-12
tpm2_listpersistent: Port to use common infrastructure.

flihp:
This one was very straight forward. After removing the duplicate options
processing code ... there wasn't any left. The main function was reduced
to effectively a single function call. This commit just removes 'main'
outright and turns the 'listPersistent' into the 'execute_tool'
function. There are also a few cleanups made to the manpage.


flihp:
This resolves #176

PR #196, opened 2017-01-25
tpm2_quote.c: port to use common infrastructure

williamcroberts:
Fixes 174


src/tpm2_quote.c
```diff
@@ -361,21 +320,15 @@ int main(int argc, char *argv[])
...
         {
-        case 'h':
-            h_flag = 1;
-            break;
-        case 'v':
-            v_flag = 1;
-            break;
         case 'k':
-            if(getSizeUint32Hex(optarg,&akHandle) != 0)
+            if(!string_bytes_get_uint32(optarg,&akHandle) != 0)
```
gwei3: now string_bytes_get_uint32() return bool directly, we should remove  " != 0" above.

src/tpm2_quote.c
```diff
@@ -415,7 +368,7 @@ int main(int argc, char *argv[])
             l_flag = 1;
             break;
         case 'g':
-            if(getSizeUint16Hex(optarg,&pcrSelections.pcrSelections[0].hash) != 0)
+            if(!string_bytes_get_uint16(optarg,&pcrSelections.pcrSelections[0].hash) != 0)
```
gwei3: similarly, need remove  " != 0" above.

src/tpm2_quote.c
williamcroberts: Yep, thanks!

PR #200, opened 2017-01-30
tpm2_getpubek: port to use common infrastructure

williamcroberts:
Refactor to be inline with coding standard.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -53,7 +53,8 @@ sbin_PROGRAMS = \
     src/tpm2_verifysignature \
     src/tpm2_getmanufec \
     src/tpm2_quote \
-    src/tpm2_takeownership
+    src/tpm2_takeownership \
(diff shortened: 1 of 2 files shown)


flihp:
I dropped @idesai from the reviewer list. Making the executive decision that 1 review is sufficient.

williamcroberts:
Does the tool require you to wait for all reviewers to review?

flihp:
I don't think so. It may if you use the 'merge' button but the merge commits drive me nuts when looking at git history directly so I manually merge (--ff-only). I just removed the extra reviewer ... because it seemed like the right thing to do? Probably wasn't necessary though :man_shrugging:

williamcroberts:
@flihp they have the ability to "rebase and merge" on the "merge pull request" button if you use the drop down selection on the right hand side of that button. So we should just be using that.

PR #201, opened 2017-01-31
configure: use git describe for version number

williamcroberts:
Remove the hardcoded version number and use git describe. This makes
it impossible to release a version of the tools that does not agree
with the git tag.

Additionally, builds not on a tag will be marked dirty:

$ ./src/tpm2_listpcrs -v
./src/tpm2_listpcrs, version v1.1.0-81-ga5d2946

Once the current HEAD is a tag, it will only read the tag number.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,5 @@
-AC_INIT([tpm2.0-tools], [1.1.0])
+AC_INIT([tpm2.0-tools],
+    [m4_esyscmd_s([git describe --tags --always --dirty])])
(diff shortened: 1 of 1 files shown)


flihp:
Solid. We may want to rethink the 'v' prefix on our tags now. A version string like "version v1.0.0" seems redundant. We can defer that discussion till the next release though. Merging now.

williamcroberts:
I was thinking we would want to write up what our release process is.

PR #207, opened 2017-02-02
This should fix issue#206 for enforcing policies for authorization of created objects

idesai:
Please close this PR (...which is enforcing policy based authorization of the objects created with tpm2_create...) after you would have closed PR#205 (...which is creating policies...)


williamcroberts:
You should cherry-pick this onto the branch that is being submitted for PR #205 , I am going to close this.

An change that has a true dependency on another patch should just be on the PR that introduces that dependency.

PR #208, opened 2017-02-08
man: Fix comment command in manpage

grandpaul:
The comments in troff should be .\\" thus this commit replace
all ." to fix man warnings.

Signed-off-by: Ying-Chun Liu (PaulLiu) <paulliu@debian.org>

### DIFF
--- a/man/tpm2_activatecredential.8.in
+++ b/man/tpm2_activatecredential.8.in
@@ -1,31 +1,31 @@
-." Copyright (c) 2016, Intel Corporation
-." All rights reserved.
-."
-." Redistribution and use in source and binary forms, with or without
-." modification, are permitted provided that the following conditions are met:
-."
-." 1. Redistributions of source code must retain the above copyright notice,
-." this list of conditions and the following disclaimer.
-."
-." 2. Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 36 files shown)


flihp:
Thanks @grandpaul. I'm new to troff etc and pretty much built the man pages from some old tutorials. These confirm that your patch is right and that we should be using .\" but I'm not able to get `man` or `groff` to output any warnings about these comments.

Invoking man like:
```
MANPATH=$(pwd)/man man --warnings=all tpm2_create
```
does produce a few warnings but none related to these comments. I'm merging your patch as-is but I was hoping that you would provide instructions to reproduce these warnings. I'd like to integrate these into a test for our travis-ci script to catch similar mistakes in the future.

grandpaul:
Hi @flihp 
This comes from lintian (Debian check tool). For example,
man --warnings -E UTF-8 -l -Tutf8 -Z ./tpm2_akparse.8
And on stderr, we got
<standard input>:3: warning: macro `"' not defined

flihp:
Thanks @grandpaul. I'll create an 'issue' so we can catch these warnings in our CI setup. I pushed your patch last night in 995ea4fe8da065781b9af673a119c9e8a92551cb so I'm closing the PR.

PR #210, opened 2017-02-08
Add an example "default" environment file to minimize output binary s…

flihp:
…ize.

We're building most of the source into a "utility" library and linking
all tools against it. By default the linker includes everything in the
library, not just the code needed by the executable. Providing these
flags to the compiler and linker reduces the size of the output binaries
by more than a half. Using tpm2_create as an example...

Without this new CONFIG_SITE file:

$ stat -c "%s %n" src/tpm2_create
85080 ./src/tpm2_create

With this new CONFIG_SITE file:

$state -c "%s %n" src/tpm2_create
29680 ./src/tpm2_create

The .travis.yml file is updated to use these flags as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,15 +34,15 @@ before_script:
 script:
   - mkdir ./build
   - pushd ./build
-  - ../configure --enable-unit
+  - CONFIG_SITE=../lib/default_config.site ../configure --enable-unit
   - make -j$(nproc)
   - make -j$(nproc) check
   - make -j$(nproc) clean
-  - ../configure --enable-unit --without-tcti-socket
+  - CONFIG_SITE=../lib/default_config.site ../configure --enable-unit --without-tcti-socket
   - make -j$(nproc)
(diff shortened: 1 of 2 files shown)


flihp:
There is, but it may be a bit esoteric. It comes from the goal of making building and packaging as easy as possible. This means making as few assumptions on behalf of the user / packager as possible. So for any build environment stuff that isn't required to build the tools (librariy dependencies & include paths) I'm trying to keep from hard coding them into the automake file.

So while I can't come up with a reason anyone would want to build the tools w/o these flags it just feels safer to put them into a CONFIG_SITE file to make overriding the behavior as easy as possible.

I'm open to moving this into the Makefile.am directly if there's a compelling reason but keeping the environment separate just "feels right" to me.

williamcroberts:
I'm pretty OK with that, I know if they use a different Linker or build on Mac those things might need to change.

flihp:
It would be great to build up a collection of configurations for builds that we test.

PR #213, opened 2017-02-09
introduce file io routines with unit tests

williamcroberts:
Introduce routines in string-bytes.h for detecting endianess
and converting endianess.

Introduce generic file-io routines that save multibyte 16,32 and 64
bit values as big endian to a FILE stream as well as raw byte
read/write routines.  Additionally, have routines for reading and
writing a file specific header that includes a magic and version
field.

Include unit tests for these routines.


williamcroberts:
Changing the file layout on disk would be a major version bump, as old files would no longer be compatible. Tools themselves could detect this as an invalid file and fix it up on behalf of the user as a possibility, or we could make it a flag.

williamcroberts:
We need to think through on when and how we want to version file formats. Many of the files, notably the ones using SaveDataToFile() are just outputting raw single tpm structures that are just raw bytes. I think things like this don't need any format versioning as they are coupled directly to TPM formats and cannot change. I think any other file format, where we save things like algorithm with the key (SaveEccKeyToFile in tpm2_akparse) or writeCrtSecToFile() or saveTpmContextToFile() would need to place a header on it.

In summary, raw tpm byte arrays, no header and continue raw output. Complex structures of our own design need the header.

williamcroberts:
I think we would also want to kill off ChangeEndian.h during this PR as a separate commit as well.

src/files.c
```diff
@@ -144,3 +147,156 @@ int getFileSize(const char *path, long *fileSize)
...
+ * @param f
+ *  The file to read from.
+ * @param data
+ *  The data buffer to read into.
+ * @param size
+ *  The size of the buffer, which is also the amount of bytes to read.
+ * @return
+ *  True on success, False otherwise.
+ */
+static bool readx(FILE *f, UINT8 *data, size_t size) {
```
williamcroberts: newline

williamcroberts:
This isn't ready yet, outside of the code comments I left, and someone else may leave, we should take the time to kill of ChangeEndian.h

src/files.c
```diff
@@ -144,3 +147,156 @@ int getFileSize(const char *path, long *fileSize)
...
+    bool files_write_##size(FILE *out, UINT##size data) { \
+        BAIL_ON_NULL("FILE", out); \
+        BE_CONVERT(&data, size); \
+        return writex(out, (UINT8 *)&data, sizeof(data)); \
+    }
+
+#define FILE_READ(size) \
+    bool files_read_##size(FILE *out, UINT##size *data) { \
+	    BAIL_ON_NULL("FILE", out); \
+	    BAIL_ON_NULL("data", data); \
```
gwei3: wrong indent in above two lines.

src/string-bytes.c
```diff
@@ -125,3 +125,64 @@ UINT16 string_bytes_copy_tpm2b(TPM2B *dest, TPM2B *src) {
...
+
+    uint32_t test_word;
+    uint8_t *test_byte;
+
+    test_word = 0xFF000000;
+    test_byte = (uint8_t *) (&test_word);
+
+    return test_byte[0] == 0xFF;
+}
+
```
gwei3: maybe you shall implement string_bytes_endian_convert_x via Micros like what you did for files_read/write_x.

src/files.c
williamcroberts: its declaring a function, the lines are part of the body, so should be indented.

src/string-bytes.c
williamcroberts: The routine you posted the comment closest to is a run time test to determine endianess. As far as the others, they could be declared as macros if I did a loop, and the loop counter should be small enough for the loop-unroller to unroll them (thus preserving lack of branches). But I am not too worried about those being macro'd up right now.

williamcroberts: But I did it anyways +1

PR #259, opened 2017-03-17
tpm2_send_command: Fix bad length specifiers in LOG_* statements.

flihp:
Use 'z' for size_t, not 'l'. This problem only manifests on 32bit
systems where size_t is not a long int. This is why the 'z' length
modifier exists.


flihp:
this resolves #259

PR #271, opened 2017-04-04
Android Support

williamcroberts:
tools build and link, requires: 
https://github.com/01org/TPM2.0-TSS/pull/395
https://android-review.googlesource.com/#/c/364619/

### DIFF
--- a/lib/options.h
+++ b/lib/options.h
@@ -124,7 +124,15 @@ showArgMismatch (const char *name)
 static inline void
 showVersion (const char *name)
 {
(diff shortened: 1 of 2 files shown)


lib/options.h
```diff
@@ -121,10 +121,19 @@ showArgMismatch (const char *name)
     printf("Please type \"%s -h\" get the usage!\n", name);
 }
 
+
```
williamcroberts: erratic whitespace change

Issue #254, opened 2017-03-15
test_tpm2_encryptdecrypt test should probe Symmetric Encryption support using GetCapability


tonyespy:
If the test_tpm2_encryptdecrypt.sh test is run on a device with a TPM 2.0 implementation that doesn't support Symmetric Encryption, the test will fail.

It should be possible for the script to use the new tpm2_dump_capability (in master) command to request the attributes for the TPM2_EncryptDecrypt command (CC=0x164).  If no results are returned, the command isn't supported by the device, and the test shouldn't fail.

This is being reported as of tpm2.0-tools version 1.1.0.

flihp:
Good catch. Seems like the right thing to do here would be to have tpm2_encryptdecrypt.c check for command support before execution. Probably something we should do for each tool but I guess that all depends on what commands are required vs optional.

PR #283, opened 2017-05-10
Rework encryptdecrypt test

DomiNic5787:
Check whether the command encryptdecrypt is supported by the device before testing it


DomiNic5787:
This commit should fix issue #254

PR #284, opened 2017-05-15
tpm2_listpcrs: use TPM2_GetCapability to determine PCRs to read

snits:
Allow TPM to tell us PCRs that are available to be read in a bank
instead of assuming there will be 24. This resolves an issue on
a system where in the bios you can choose between hash functions
(sha1 and sha256 in this case). Both will show up as supported,
but when it tries reading the bank that isn't selected in the bios
it makes no progress, and then fails saying that there are too many
PCRs to read.

This consolidates the pcr_selections init code into one function.

Also a minor change to show_pcr_values to use sizeofSelect to
determine loop iteration instead of hardcoding 24.

### DIFF
--- a/tools/tpm2_listpcrs.c
+++ b/tools/tpm2_listpcrs.c
@@ -173,38 +173,35 @@ static bool read_pcr_values(listpcr_context *context) {
     return true;
 }
 
-static void init_pcr_selection_from_algorithm(TPMI_ALG_HASH alg_id,
-        TPML_PCR_SELECTION *pcr_selections) {
+static bool init_pcr_selection(TPMI_ALG_HASH alg_id, listpcr_context *context) {
 
-    pcr_selections->count = 1;
-    pcr_selections->pcrSelections[0].hash = alg_id;
-    set_pcr_select_size(&pcr_selections->pcrSelections[0], 3);
-    clear_pcr_select_bits(&pcr_selections->pcrSelections[0]);
+    TPMI_YES_NO moreData;
+    TPMS_CAPABILITY_DATA cap_data;
+    TPML_PCR_SELECTION *pcr_sel = &context->pcr_selections;
+    UINT32 rval, i, j;
 
-    UINT32 pcr_id;
-    for (pcr_id = 0; pcr_id < 24; pcr_id++) {
-        set_pcr_select_bit(&pcr_selections->pcrSelections[0], pcr_id);
+    rval = Tss2_Sys_GetCapability(context->sapi_context, 0, TPM_CAP_PCRS, 0, 1, &moreData, &cap_data, 0);
+    if (rval != TPM_RC_SUCCESS) {
+        LOG_ERR("GetCapability: Get PCR allocation status Error. TPM Error:0x%x......\n", rval);
+        return false;
     }
-}
 
-/* XXX Could this internally call init_pcr_selection_from_algorithm to reduce duplicate code? */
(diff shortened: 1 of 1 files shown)


snits:
Hi William,
I'm still trying to decide if this is the best way to handle this. I'm trying to figure out a decent way to push this down into get_banks, since TPM2_GetCapability is already being called there. As mentioned in the commit message, we've seen this problem where on certain systems you can select a hash function in the bios for the tpm. Both hash functions show up as supported with tpm2_listpcrs -s, but when it tries to read the function that isn't selected in the bios the pcr read calls succeed, but there is nothing there so the bitfields are getting cleared out. Eventually it iterates through 24 times, jumps out of the loop, sees that it still has pcrs to read, and complains about it.

That CI run failed because it couldn't find realpath. Not sure what happened there.

tools/tpm2_listpcrs.c
```diff
@@ -173,38 +173,36 @@ static bool read_pcr_values(listpcr_context *context) {
...
-    pcr_selections->count = 1;
-    pcr_selections->pcrSelections[0].hash = alg_id;
-    set_pcr_select_size(&pcr_selections->pcrSelections[0], 3);
-    clear_pcr_select_bits(&pcr_selections->pcrSelections[0]);
-
-    UINT32 pcr_id;
-    for (pcr_id = 0; pcr_id < 24; pcr_id++) {
-        set_pcr_select_bit(&pcr_selections->pcrSelections[0], pcr_id);
+static bool init_pcr_selection(TPMI_ALG_HASH alg_id, listpcr_context *context)
+{
```
williamcroberts: Currently this code base is a mixed mess of style, but adhere to the coding style under: tpm2.0-tools/misc/coding_standard_c.txt.

Functions have opening brace on the same line, variables use underscores, no camel case.

williamcroberts:
Thank you for this, I think this takes care of: https://github.com/01org/tpm2.0-tools/issues/119 and I miss diagnosed how to handle that issue.

snits:
Still failing to find realpath.

Sorry about the style. Hopefully it is good now. I'm dealing with 1.1.0 and this at the same time so get confused at times.

williamcroberts:
Dont be sorry, we like to see changes, ill look into the Travis issue.

williamcroberts:
Hopefully PR #285 fixes it.

williamcroberts:
That's actually not the problem. We just set PATH in the .travis.yml file. The problem is, that the resourcemanager in TSS has been removed, so were not actually starting the resourcemanager. So we need to actually grab the 1.0 branch, or better yet swicth to the new resource manager.

However, this highlights some issues in the test script, we have a test passing!

williamcroberts:
Sorry for the delay, can you re-push. I fixed the travis build.

tools/tpm2_listpcrs.c
```diff
@@ -271,14 +268,16 @@ static bool show_selected_pcr_values(listpcr_context *context) {
 
 static bool show_all_pcr_values(listpcr_context *context) {
 
-    init_pcr_selection_all(&context->algs, &context->pcr_selections);
+    if (!init_pcr_selection(0, context))
```
williamcroberts: normally we don't like to bury routines in conditionals, but ill let this slide.

williamcroberts:
Nevermind on re-pushing, I found a restart build button on Travis...

williamcroberts:
Moved to: https://github.com/01org/tpm2.0-tools/pull/286

PR #289, opened 2017-05-17
TCTI tabrmd

flihp:
Add support for the tabrmd TCTI, make it the default, and integrate it into the Travis CI stuff.


lib/context-util.c
```diff
@@ -124,6 +127,37 @@ tcti_socket_init (char const *address,
...
+TSS2_TCTI_CONTEXT*
+tcti_tabrmd_init (void)
+{
+    TSS2_TCTI_CONTEXT *tcti_ctx;
+    TSS2_RC rc;
+    size_t size;
+
+    rc = tss2_tcti_tabrmd_init(NULL, &size);
+    if (rc != TSS2_RC_SUCCESS) {
+        fprintf (stderr,
```
williamcroberts: LOG_ERR() here and throughout

.travis.yml
```diff
@@ -23,10 +25,16 @@ install:
     - popd && popd
     - git clone https://github.com/01org/TPM2.0-TSS.git
     - pushd TPM2.0-TSS
-    - git checkout 1.0
+    - git checkout 1.1.0_beta0
```
williamcroberts: Later i'll update this to master, unless you want to in this commit.

.travis.yml
```diff
@@ -23,10 +25,16 @@ install:
...
     - ./bootstrap && ./configure && make -j$(nproc) && sudo make install && popd
     - sudo ldconfig /usr/local/lib
-    - resourcemgr -sim -tpmhost 127.0.0.1 &
+    - git clone https://github.com/01org/tpm2-abrmd.git
+    - pushd tpm2-abrmd
+    - ./bootstrap && ./configure --with-dbuspolicydir=/etc/dbus-1/system.d && make -j$(nproc) && sudo make install && popd
```
williamcroberts: why doesn't configure attempt to look for dbuspolicydir in common locations or default to this path?

.travis.yml
```diff
@@ -23,10 +25,16 @@ install:
...
+    - sudo mkdir -p /var/lib/tpm
+    - sudo groupadd tss && sudo useradd -M -d /var/lib/tpm -s /bin/false -g tss tss
+    - sudo pkill -HUP dbus-daemon
+    - sudo -u tss tpm2-abrmd --tcti=socket &
```
williamcroberts: Jeepers dbus sucks....

.travis.yml
flihp: It mostly does. The default behavior according to GNU is for `${prefix}` to be set to `/usr/local` so w/o this `${dbuspolicydir}` ends up being under `/usr/local`. An alternative would be to use `--sysconfdir=/etc`. I went with using `--with-dbuspolicydir` directly since it makes the intent clear but I'm not particularly tied to either solution.

.travis.yml
flihp: In all fairness this is my doing. The dbus policy packaged with tpm2-abrmd will allow a process run as either root or the tss user to claim the Tabrmd name on the system bus. Setting up a tss user seemed like "the right thing to do" here. The only thing we can blame on dbus is having to send `SIGHUP` to the `dbus-daemon` so that it picks up the configuration installed by the `tpm2-abrmd` build.

I could get this down to two lines if the `tpm2-abrmd` is run as root.

lib/context-util.c
flihp: can do

flihp:
By default `./configure` will try to enable all 3 TCTI modules. If any are not installed on the system they will be disabled. They have a precedence: tabrmd > socket > device such that by default the tabrmd will be used by default *if* it's enabled. If it's not enabled then the tools fallback to the socket TCTI, then the device. Thinking about it now it may make more sense to have the precedence be: tabrmd > device > socket possibly with the socket TCTI being disabled by default since it's only useful for communicating with the simulator now.

flihp:
This v2 patch set fixes up the LOG_ERR stuff and switches us back to the master branch of the TSS.

williamcroberts:
Thanks, merged.

williamcroberts:
Weird, it said all checks passed and I just got a fail message. Revert here: https://github.com/01org/tpm2.0-tools/pull/290

williamcroberts:
The proposed order of tabrmd > device > socket does make the most sense to me as well.

Attempting to restore your changes here: https://github.com/01org/tpm2.0-tools/pull/291

PR #293, opened 2017-05-22
configure: Undefine FORTIFY_SOURCE first, then define it

PeterHuewe:
On my hardened gentoo system _FORTIFY_SOURCE is already defined,
thus it makes sense to undefine it first then define it, not vice
versa.

Signed-off-by: Peter Huewe <peterhuewe@gmx.de>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,8 +123,8 @@ AS_IF([test x"$hardening" != x"no"], [
   add_hardened_c_flag([-Wstack-protector])
   add_hardened_c_flag([-fstack-protector-all])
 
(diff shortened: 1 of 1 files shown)


PeterHuewe:
Don't really know why travis is failing - don't think it's an issue with the commit itself.

williamcroberts:
@PeterHuewe the failure is from "test_tpm2_getmanufec.sh Fail". Which has been flaky lately. It actually flunked a patch this morning for no reason. I re-kicked the travis build/test.

PR #300, opened 2017-06-21
Reenable getmanufec test

williamcroberts:


tools/tpm2_getmanufec.c
```diff
@@ -375,19 +374,23 @@ char *Base64Encode(const unsigned char* buffer)
...
     }
+
+    char *final_string = NULL;
+
     CURL *curl = curl_easy_init();
     if (curl) {
         char *output = curl_easy_escape(curl, b64text, len);
         if (output) {
-            strncpy(b64text, output, strlen(output));
+            final_string = strdup(output);
```
williamcroberts: we do this so callers can just use free() instead of being curl aware.

PR #301, opened 2017-06-22
Fix build issues with GCC7 and OpenSSL 1.1.0

martinezjavier:
Hello,

These commits fixes some compile errors I found when building tpm2.0-tools on a Fedora 26 machine.

The first commit is just a cleanup for a left over comment in INSTALL about the src directory that no longer exists, the second commit fixes some build warnings due GCC7 complaining about not having explicit comments for case blocks falling through the next one in switch statements and the last commit adds some #ifdefery to cope with the recent API changes in the OpenSSL 1.1.0 library.

Best regards,
Javier

### DIFF
--- a/lib/tpm_kdfa.c
+++ b/lib/tpm_kdfa.c
@@ -51,6 +51,34 @@ static const EVP_MD *tpm_algorithm_to_openssl_digest(TPMI_ALG_HASH algorithm) {
     /* no return, not possible */
 }
 
+static HMAC_CTX *hmac_alloc()
+{
+    HMAC_CTX *ctx;
+#if OPENSSL_VERSION_NUMBER < 0x1010000fL /* OpenSSL 1.1.0 */
+    ctx = malloc(sizeof(*ctx));
+#else
+    ctx = HMAC_CTX_new();
+#endif
+    if (!ctx)
+        return NULL;
+
+#if OPENSSL_VERSION_NUMBER < 0x1010000fL
+    HMAC_CTX_init(ctx);
+#endif
+
+    return ctx;
+}
+
+static void hmac_del(HMAC_CTX *ctx)
+{
+#if OPENSSL_VERSION_NUMBER < 0x1010000fL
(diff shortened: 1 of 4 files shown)


lib/tpm_kdfa.c
```diff
@@ -90,12 +90,26 @@ TPM_RC tpm_kdfa(TPMI_ALG_HASH hashAlg,
         return TPM_RC_HASH;
     }
 
-    HMAC_CTX ctx;
-    HMAC_CTX_init(&ctx);
-    int rc = HMAC_Init_ex(&ctx, key->buffer, key->size, md, NULL);
+    HMAC_CTX *ctx;
+#if OPENSSL_VERSION_NUMBER < 0x1010000fL /* OpenSSL 1.1.0 */
```
williamcroberts: I was waiting for this to be an issue. Rather than ifdef them inline, can we just create some routines (within tpm_kdfa.c is fine) that handle this for us. This function is already nasty enough and I don't want to make it worse.

maybe something like the following prototypes:

```
HMAC_CTX *hmac_new();
void hmac_free(HMAC_CTX *ctx);
```

martinezjavier: @williamcroberts thanks a lot for your feedback, I've re-spun the branch and hopefully now it looks OK for you.

lib/tpm_kdfa.c
```diff
@@ -51,6 +51,34 @@ static const EVP_MD *tpm_algorithm_to_openssl_digest(TPMI_ALG_HASH algorithm) {
     /* no return, not possible */
 }
 
+static HMAC_CTX *hmac_alloc()
+{
```
williamcroberts: FYI not aligned with coding standard here, see CONTRIBUTING. But I'll take it anyways.

martinezjavier: @williamcroberts sigh, I thought I checked it followed the coding standards but I may had missed when re-spinning the commit, sorry about that. I'll be more careful next time.

Thanks for picking them!

PR #306, opened 2017-06-29
Allow tpm2.0-tools to fallback to a different TCTI if creating context with the default fails

martinezjavier:
This is more of an RFC than a pull-request ready to be merged. I wondered if allowing the tools to fallback to another TCTI if the default fails would be a useful feature.

There's already support to choose the TCTI either at build time (--with-tcti-*) or at runtime (using the TPM2TOOLS_TCTI_NAME en var or the --tcti option) but this can be made more transparent to the users by allowing the tools to attempt using the different available TCTIs.

For example the tools (built with both tabrmd and device TCTI support) could be ran from initramfs and access the TPM directly and access using the tpm2-abrmd resource manager once the real rootfs has been mounted, without needing to set environment variables or passing additional options.

### DIFF
--- a/lib/context-util.c
+++ b/lib/context-util.c
@@ -217,23 +217,34 @@ sapi_init_from_options (common_opts_t *options)
 TSS2_TCTI_CONTEXT*
 tcti_init_from_options (common_opts_t *options)
 {
-    switch (options->tcti_type) {
+    TSS2_TCTI_CONTEXT *tcti_ctx;
+
+    do {
+        switch (options->tcti_type) {
 #ifdef HAVE_TCTI_DEV
-    case DEVICE_TCTI:
-        return tcti_device_init (options->device_file);
+        case DEVICE_TCTI:
+            tcti_ctx = tcti_device_init (options->device_file);
+            break;
 #endif
 #ifdef HAVE_TCTI_SOCK
-    case SOCKET_TCTI:
-        return tcti_socket_init (options->socket_address,
-                                 options->socket_port);
+        case SOCKET_TCTI:
+            tcti_ctx = tcti_socket_init (options->socket_address,
+                                         options->socket_port);
+            break;
 #endif
 #ifdef HAVE_TCTI_TABRMD
-    case TABRMD_TCTI:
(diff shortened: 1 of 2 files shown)


williamcroberts:
I don't hate this, and at a minimum I would want the first 2 changes. @flihp thoughts?

flihp:
This would make the default more likely to work "out of the box". But for the the use case described (fall back to device TCTI when tabrmd isn't available in initramfs) I'd still encourage people to select the device-TCTI explicitly. The initramfs is all about bringing up the root filesystem ASAP. Assuming you're using the TPM to protect some disk crypto key, you likely won't want a failing call to the tabrmd on your normal boot path since it will add some overhead (albeit unquantified). I'd even recommend building a statically compiled version of the tools with only the device TCTI to keep from pulling all of the tabrmd TCTI dependencies into the initramfs (glib, gio, dbus etc).

All that said, this isn't really an argument against the feature / proposed behavior. It may be useful elsewhere. The problem I'm having is that I can't come up with another situation where I'd find it useful.

martinezjavier:
@flihp thanks for your feedback. I agree with you and that's why the patch-set was an RFC. The use case was just an example and a little handwavy (although my final is to unseal a key in the initramfs to open a LUKS volume).

I had this patch in my local tree because I'm still in the learning about the TPM and exploring the different options phase, and so I'm testing with both the tabrmd user-space resource manager and the new in-kernel resource manager (/dev/tpmrm0) that landed in v4.12-rc1. So having the tools to fallback to access the TPM device directly was convenient to me so I could use the same binaries to test with both setups without additional options or environment variables.

@williamcroberts yes, I also think the other two patches have merits on their own. Probably I should had split these in two different pull requests...

williamcroberts:
I thought about this more over the weekend, the more I disliked it, I wouldn't want it to use a tcti that I didn't explicitly tell it to.

martinezjavier:
@williamcroberts fair enough.

I guess you still want the first patch though? I can create a new PR only with that one.

williamcroberts:
@martinezjavier Ill take the first 2 patches if you want to submit them.

PR #316, opened 2017-07-06
Policy enforcement on created object

idesai:
Adding option to enforce policy based authorization for the created object thus modifying the final object attribute.


idesai:
Added tests for this PR

test/system/test_tpm2_create_all.sh
```diff
@@ -61,3 +61,17 @@ for pCtx in `ls ctx.cpri*`
     done
 done
 
+return_val=$?
+echo "f28230c080bbe417141199e36d18978228d8948fc10a6a24921b9eba6bb1d988" \
+| xxd -r -p > policy.bin
+tpm2_createprimary -A o -g 0xb -C prim.ctx -G 0x1
+tpm2_create -c prim.ctx -g 0xb -G 0x1 -L policy.bin -o key.pub -O key.priv -E
+cmp -i 4:0 -n 32 key.pub policy.bin -s
```
williamcroberts: Same comment here as the others, should failures be reported this late in the game.

tools/tpm2_create.c
```diff
@@ -57,7 +57,7 @@ TPMS_AUTH_COMMAND sessionData = {
 
 bool hexPasswd = false;
 
-int setAlg(TPMI_ALG_PUBLIC type,TPMI_ALG_HASH nameAlg,TPM2B_PUBLIC *inPublic, int I_flag)
+int setAlg(TPMI_ALG_PUBLIC type,TPMI_ALG_HASH nameAlg,TPM2B_PUBLIC *inPublic, int I_flag, bool E_flag)
```
williamcroberts: E_flag needs a better name

test/system/test_tpm2_create_all.sh
idesai: added intermediate checks

tools/tpm2_create.c
idesai: changed to is_policy_enforced

test/system/test_tpm2_create_all.sh
```diff
@@ -61,3 +61,27 @@ for pCtx in `ls ctx.cpri*`
...
+    else
+        cmp -i 4:0 -n 32 key.pub policy.bin -s
+        if [ $? != 0 ];then
+            echo "tpm2_create_error: Policy digest did not match in auth structure" >> create.error.log
+            return_val=1
+        else
+            return_val=0
+        fi
+    fi
+fi
```
williamcroberts: Wow you made this incredibly convoluted. Just do a series of if fail then bail and cleanup statements. If it makes it to the end, then exit 0.
```
if [ $? != 0 ]; then
  echo "something failed"
  rm <file>
  exit 1
fi

... next thing
exit 0
```

PR #317, opened 2017-07-06
Adding option to create primary objects with policy based authorization

idesai:
This commit should enable creating primary objects with authorization based on policies with an option to enforce policy.

### DIFF
--- a/test/system/test_tpm2_createprimary_all.sh
+++ b/test/system/test_tpm2_createprimary_all.sh
@@ -36,7 +36,6 @@ gAlg=
 GAlg=
 
 rm createprimary.error.log rf
-
 for gAlg in 0x04 0x0B 0x0C 0x0D 0x12
     do 
         for GAlg in 0x01 0x08 0x23 0x25
@@ -52,3 +51,25 @@ for gAlg in 0x04 0x0B 0x0C 0x0D 0x12
         done
 done
 
+#test for createprimary objects with policy authorization structures
+echo "f28230c080bbe417141199e36d18978228d8948fc10a6a24921b9eba6bb1d988" \
+| xxd -r -p > policy.bin
+
+tpm2_createprimary  -A o -G 0x1 -g 0xb -E -C prim.ctx -L policy.bin
+if [ $? != 0 ];then
+ exit 1
+fi
+
+tpm2_readpublic -c prim.ctx -o testprim.pub
+if [ $? != 0 ];then
+ exit 1
+fi
+
+cmp -i 14:0 -n 32 testprim.pub policy.bin -s
+if [ $? != 0 ];then
+ echo "Failed: createprimary with policy authorization structure" >> createprimary.error.log
+ exit 1
+fi
+
+rm -f prim.ctx policy.bin testprim.pub
+exit 0
--- a/tools/tpm2_createprimary.c
+++ b/tools/tpm2_createprimary.c
@@ -58,7 +58,7 @@ TPMS_AUTH_COMMAND sessionData = {
 bool hexPasswd = false;
 TPM_HANDLE handle2048rsa;
 
-int setAlg(TPMI_ALG_PUBLIC type,TPMI_ALG_HASH nameAlg,TPM2B_PUBLIC *inPublic)
+int setAlg(TPMI_ALG_PUBLIC type,TPMI_ALG_HASH nameAlg,TPM2B_PUBLIC *inPublic, bool is_policy_enforced)
 {
     switch(nameAlg)
     {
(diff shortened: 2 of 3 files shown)


tools/tpm2_createprimary.c
```diff
@@ -51,7 +51,7 @@ TPMS_AUTH_COMMAND sessionData;
 bool hexPasswd = false;
 TPM_HANDLE handle2048rsa;
 
-int setAlg(TPMI_ALG_PUBLIC type,TPMI_ALG_HASH nameAlg,TPM2B_PUBLIC *inPublic)
+int setAlg(TPMI_ALG_PUBLIC type,TPMI_ALG_HASH nameAlg,TPM2B_PUBLIC *inPublic, int E_flag)
```
williamcroberts: bool type for this...

tools/tpm2_createprimary.c
```diff
@@ -71,13 +71,17 @@ int setAlg(TPMI_ALG_PUBLIC type,TPMI_ALG_HASH nameAlg,TPM2B_PUBLIC *inPublic)
     // First clear attributes bit field.
     *(UINT32 *)&(inPublic->t.publicArea.objectAttributes) = 0;
     inPublic->t.publicArea.objectAttributes.restricted = 1;
-    inPublic->t.publicArea.objectAttributes.userWithAuth = 1;
+    //check if auth policy needs to be enforced
+    if(E_flag) {
```
williamcroberts: no need for a branch userWithAuth = flag;

Once flag is a bool type, it should be 0 or 1 by language convention. you could do flag & 0x1 to mask off anything that could accidentally be added.

idesai:
@williamcroberts I will add some tests for them soon, cool with the PR pending until then. These changes should be in the main branch since these are simply defining object authorization with policy defined with tpm2_createpolicy and does not have session dependencies (that can be satisfied using tpm_shell)

idesai:
Added tests for this PR

test/system/test_tpm2_createprimary_all.sh
```diff
@@ -52,3 +52,15 @@ for gAlg in 0x04 0x0B 0x0C 0x0D 0x12
         done
 done
 
+#test for createprimary objects with policy authorization structures
+echo "f28230c080bbe417141199e36d18978228d8948fc10a6a24921b9eba6bb1d988" \
+| xxd -r -p > policy.bin
+tpm2_createprimary  -A o -G 0x1 -g 0xb -E -C prim.ctx -L policy.bin
+tpm2_readpublic -c prim.ctx -o testprim.pub
+cmp -i 14:0 -n 32 testprim.pub policy.bin -s
```
williamcroberts: Do we want this to fail this far, or should we check rc codes for each command? I would imagine each command, so we can have better output messages on error.

tools/tpm2_createprimary.c
```diff
@@ -291,6 +293,17 @@ execute_tool (int               argc,
         case 'X':
             hexPasswd = true;
             break;
+        case 'L':
+            inPublic.t.publicArea.authPolicy.t.size = sizeof(inPublic.t.publicArea.authPolicy) - 2;
```
williamcroberts: BUFFER_SIZE() macro here.

tools/tpm2_createprimary.c
```diff
@@ -160,7 +159,7 @@ int createPrimary(TSS2_SYS_CONTEXT *sysContext, TPMI_RH_HIERARCHY hierarchy,
     inSensitive->t.sensitive.data.t.size = 0;
     inSensitive->t.size = inSensitive->t.sensitive.userAuth.b.size + 2;
 
-    if(setAlg(type, nameAlg, inPublic))
+    if(setAlg(type, nameAlg, inPublic, E_flag))
```
williamcroberts: E_flag, when passing this into routines, change it to is_policy_enforced, so it has more meaning.

tools/tpm2_createprimary.c
```diff
@@ -216,6 +217,7 @@ execute_tool (int               argc,
         g_flag = 0,
         G_flag = 0,
         C_flag = 0;
+    bool E_flag = false;
```
williamcroberts: The name E_flag is ok here, but when passed along as a function argument, it should have a more meaningful name.

test/system/test_tpm2_createprimary_all.sh
```diff
@@ -52,3 +51,23 @@ for gAlg in 0x04 0x0B 0x0C 0x0D 0x12
...
+    if [ $? != 0 ];then
+        return_val=1
+    else
+        cmp -i 14:0 -n 32 testprim.pub policy.bin -s
+        return_val=$?
+        if [ $return_val != 0 ];then
+            echo "Failed: createprimary with policy authorization structure" >> createprimary.error.log
+        fi
+    fi
+fi
```
williamcroberts: same here if fail then cleanup and bail

Issue #260, opened 2017-03-17
optarg - stop copying


williamcroberts:
Stop copying optarg and just store the pointer for later usage. See https://www.gnu.org/software/libc/manual/html_node/Using-Getopt.html for more info.

PR #340, opened 2017-07-13
Fix remaining tools doing unnecessary options arguments copies

martinezjavier:
This contains fixes for all the remaining offenders of issue #260. The only file that I didn't fix was tpm2_createpolicy.c to avoid creating a conflict for @idesai, since he has a pending pull request refactoring that file.


williamcroberts:
We can solve any merge conflicts, I wouldn't worry about it to much.

williamcroberts:
Working on it now. I was actually planning on tackling this today, so I really appreciate your help.

PR #334, opened 2017-07-11
Refactoring tpm2_createpolicy tool to support:

idesai:
1. Trial and Real policy sessions.
2. PCR inputs accepted as raw bytes (output of tpm2_listpcrs). Default reads current pcrs based on selection.
3. PCR selection with set-list to include multiple pcrs in the policy
4. Keep policy session alive and output policy session handle at tool exit (device tcti usage only)

PR #344, opened 2017-07-13
tpm2_createprimary: Adding support for session handle input for policy based authorization

idesai:
The PR has a dependency on the PR #334, the new create policy tool. the new option is only relevant for device tcti or any other tcti that can sustain sessions between application runs.
(a test is pending for this PR.)


williamcroberts:
Anything that has a dependency on another PR, just submit on one PR and break it up into multiple commits....

Issue #296, opened 2017-06-12
tpm2 startup command


williamcroberts:
See issue #292 for the details.

send a startup command by default.

williamcroberts:
Won't Fix

PR #349, opened 2017-07-17
main: add tpm2_startup() command to all tools

williamcroberts:
A common issue when using a device directly, is that, potentially,
no startup command was issued. A tpm cannot be used until a
startup command is issued.

The tools do not call startup implicitly, so add that to the tools
and ignore initialized or session errors. However, log at the
warning level any other return codes.

Fixes: #296

### DIFF
--- a/tools/main.c
+++ b/tools/main.c
@@ -76,6 +76,23 @@ main (int   argc,
     sapi_context = sapi_init_from_options (&opts);
     if (sapi_context == NULL)
         exit (1);
+
+    /*
+     * Call Tpm2_startup() as not all TCTIs/Environments handle this for you.
+     * For instance, going at the device tcti with no resource manager, a
+     * common issue is to forget the startup command.
+     * Valid Errors:
+     * TPM_RC_SUCCESS (0x00)
+     * TPM_RC_INITIALIZE(0x100)
+     * TPM_RC_2 (0x2XX) - Session errors
+     */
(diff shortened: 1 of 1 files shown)


williamcroberts:
@martinezjavier - What do you think about this PR? Unfortunately I cant add you as a collaborator at this moment, but I will sort that out soon.

idesai:
@williamcroberts, I thought about this. This might actually be useful for simulator only. Typically BIOS does issue the command during boot on real devices. There exists a tpm2_startup tool for this very purpose that can be made of the test/install script. Doing it this way will cause startup to be issued at every tool run and that can break context dependencies.

williamcroberts:
@idesai - typical bios would. What about non-PC architectures? The boot code should do this, but is it guaranteed? Suppose Rasberry PI with a tpm added via serial bus?

idesai:
@williamcroberts even then you would issue a startup command just once per boot for which you can use the tpm2_startup tool.

williamcroberts:
@idsea - True, but you would have to know that you need to do it, which I am OK with. Considering this is part of the flow of how it works... ill drop this.

martinezjavier:
@williamcroberts sorry I was offline so I didn't see your message before, but FWIW I agree with @idesai on this.

webmeister:
If the BIOS does not issue the startup command, then the Linux kernel driver will. So it should be pretty rare to find a TPM device that has not seen a startup command (you would need to reset it without reloading the kernel driver).

PR #352, opened 2017-07-17
fix whitespace errors introduced by PR #334

williamcroberts:


williamcroberts:
@idesai watch that editor setting so you don't accidentally make more changes than required for a patch.

PR #354, opened 2017-07-18
Allow  tpm2_unseal to unseal an object using a PCR policy for authentication

martinezjavier:
This pull request adds support for tpm2_unseal to authenticate using a PCR policy so objects can be unsealed only when a PCR policy is satisfied.

The first commit is just a cleanup, the second is the one that adds the support and the third one extends the test_tpm2_unseal.sh test case to cover the object creation and unsealing using a PCR policy for auth.

### DIFF
--- a/tools/tpm2_unseal.c
+++ b/tools/tpm2_unseal.c
@@ -43,7 +43,10 @@
 #include "main.h"
 #include "options.h"
 #include "password_util.h"
+#include "pcr.h"
 #include "tpm2_util.h"
+#include "tpm_hash.h"
+#include "tpm_session.h"
 
 typedef struct tpm_unseal_ctx tpm_unseal_ctx;
 struct tpm_unseal_ctx {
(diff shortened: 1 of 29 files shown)


tools/tpm2_unseal.c
```diff
@@ -111,13 +117,17 @@ static bool init(int argc, char *argv[], tpm_unseal_ctx *ctx) {
...
         };
         UINT8 all;
     } flags = { .all = 0 };
 
     int opt;
     bool hexPasswd = false;
     char *contextItemFile = NULL;
+    TPML_PCR_SELECTION pcr_selections;
+    TPM2B_DIGEST pcr_hashes = TPM2B_TYPE_INIT(TPM2B_DIGEST, buffer);
     while ((opt = getopt_long(argc, argv, optstring, long_options, NULL)) != -1) {
```
williamcroberts: I'll submit a patch with optind = 0 here. Surprised we haven't hit any weird parsing errors.

PR #355, opened 2017-07-18
optind fixes

williamcroberts:
* Fix optind resetting
* Fix an frpintf for error output by using LOG_ERR


martinezjavier:
@williamcroberts your PR looks good to me.

PR #357, opened 2017-07-19
Making compile on GCC 4.7.2

kkrach:
I found some issues with GCC 4.7.2 (i686-pc-linux-gnu-gcc (Sourcery CodeBench Lite 2012.09-62) 4.7.2) and provided a solution. You are free to fix it in any other way...

Further more, I hit this compiler error:
> tools/tpm2_getmanufec.c: In function 'Base64Encode':
> tools/tpm2_getmanufec.c:366:5: error: value computed is not used [-Werror=unused-value]
> tools/tpm2_getmanufec.c:368:5: error: value computed is not used [-Werror=unused-value]
> cc1: all warnings being treated as errors

I used `make CFLAGS="-g -O2 -Wno-error=unused-value"` as workaround.

BR, Karl

### DIFF
--- a/tools/tpm2_getmanufec.c
+++ b/tools/tpm2_getmanufec.c
@@ -447,7 +447,7 @@ int RetrieveEndorsementCredentials(char *b64h)
         goto out_easy_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, respfile);
+    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, 0L);
     if (rc != CURLE_OK) {
         LOG_ERR("curl_easy_setopt for CURLOPT_VERBOSE failed: %s", curl_easy_strerror(rc));
         goto out_easy_cleanup;


tools/tpm2_getmanufec.c
```diff
@@ -447,7 +447,7 @@ int RetrieveEndorsementCredentials(char *b64h)
         goto out_easy_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, respfile);
+    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, 0L);
```
williamcroberts: Yeah looking at this again, we want a verbose log file if respfile is set. So either !!respFile to change it to an int type, or respfile != NULL, which will do the same thing.

lib/log.c
```diff
@@ -15,7 +15,7 @@ bool
 log_set_level (log_level value)
 {
     bool result = false;
-    if (value >= log_level_error || value <= log_level_verbose) {
+    if ((int)value >= log_level_error || value <= log_level_verbose) {
```
williamcroberts: I'd just kill this check and make the function return void. This is really superfluous for an internal routine.

kkrach:
`if (value >= log_level_error && value <= log_level_verbose)`: This line seems to check, whether the parameter is inside the allowed range - which implicitly expects that non-enum values can be passed. Since afaik the signedness of an enum is implementation dependend, `value` can be an `signed integer` or an `unsigned integer`. In case of an unsigned integer `value >= 0` will make no sense - and this is what the compiler complains about, independent of && or ||. With a different compiler (or changed enum values) where the enum is a signed integer, this check will make sense. So I will follow @williamcroberts's suggestion to remove this check and make it void - since I don't know any better solution, if you don't like the int-cast-hack (which I don't like either). If you have a better solution, it's highly welcome!

Since I don't know how to change the commit message after having it pushed, I will close the pull request and start a second try.

kkrach:
It was no blindly fix of a compiler warning - for me it looks like a copy paste error (since in this case `respfile` is always != NULL and the documentation of CURLOPT_VERBOSE states, that only 0L and 1L are allowed values). As you do, I wondered if verbose must be enabled, but I agree with you, that enabling would be the better fix. As stated above, I will close the pull request, fix it with `respfile == NULL ? 0L : 1L` according to your hint - even if I would change it to 1L - and will create a second pull request with a better commit message.

PR #359, opened 2017-07-20
Fix for nv writes the first 1024 bytes and repeats

idesai:


idesai:
Fix for issue#358

williamcroberts:
Yes lets get that commit message to show:
1. What was broken and why
2. What you did to fix it.

PR #360, opened 2017-07-20
tpm2_nvread: Adding a warning to the user about max read size

idesai:
tpm2_nvread: Adding a warning to the user about max read size

### DIFF
--- a/tools/tpm2_nvread.c
+++ b/tools/tpm2_nvread.c
@@ -165,6 +165,11 @@ static bool init(int argc, char *argv[], tpm_nvread_ctx *ctx) {
                         optarg);
                 return false;
             }
+            if (ctx->size_to_read > 1024) {
+                LOG_ERR("Maximum read size allowed is 1024."
+                    " Use offset option to continue reading data.\n");
+                return false;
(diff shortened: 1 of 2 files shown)


tools/tpm2_nvread.c
```diff
@@ -165,6 +165,11 @@ static bool init(int argc, char *argv[], tpm_nvread_ctx *ctx) {
                         optarg);
                 return false;
             }
+            if (ctx->size_to_read > 1024) {
```
williamcroberts: Couldn't we just keep reading in 1024 blocks for them?

williamcroberts: That seems to be what nv_write is doing...

williamcroberts:
Better commit message, and a bug ticket on next would be nice.

tools/tpm2_nvread.c
idesai: @williamcroberts correct, just need to test the nv index size prior with nv read public to avoid nv wear out.

williamcroberts:
This will be on #363

PR #362, opened 2017-07-20
Fixing compiler warnings with GCC 4.7.2

kkrach:
This is the correction of [pull request 357](https://github.com/01org/tpm2.0-tools/pull/357). Since the commit message had to be corrected, I ve decided to close the previous pull-request and create a new one:
- Considering remarks of @martinezjavier  and @williamcroberts 
- Correcting the commit message

### DIFF
--- a/tools/tpm2_getmanufec.c
+++ b/tools/tpm2_getmanufec.c
@@ -447,7 +447,7 @@ int RetrieveEndorsementCredentials(char *b64h)
         goto out_easy_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, respfile);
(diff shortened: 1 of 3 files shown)


tools/tpm2_getmanufec.c
```diff
@@ -447,7 +447,7 @@ int RetrieveEndorsementCredentials(char *b64h)
         goto out_easy_cleanup;
     }
 
-    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, respfile);
+    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, respfile == NULL ? 0L : 1L);
     if (rc != CURLE_OK) {
```
martinezjavier: You don't need to use an unary opeator here, just use respfile != NULL as @williamcroberts mentioned.

kkrach:
Done.

Issue #348, opened 2017-07-16
PCR policy and values do not seem to affect nvread


spmaniato:
Hello :slightly_smiling_face: 

I'm working with an Intel TPM 2.0 device in an embedded Linux environment. I have installed TPM2.0-TSS (`6e58f75`), tpm2-abrmd (`1003fcf`), and tpm2.0-tools (effc1fa).

I'm trying to set up secure boot based on a PCR policy. To this end, I want the TPM to prevent reading of a "secret" stored in its NV area when the value of a PCR (`09` in the example below) is not the one specified in the policy. However, it looks like I can read the secret regardless of the policy. Please see example below.

Here's the happy path, which I think works as expected:

```bash
$ dd if=/dev/urandom of="/tmp/secret.bin" bs=1 count=32
$ xxd -p /tmp/secret.bin 
24516beb6a0a9ac1366857b66435513d902817383fcbae7bb3e2dfed7ce8b799

$ tpm2_listpcrs -L 0x4:9 -o pcr_09.bin
$ xxd -p pcr_09.bin 
94a2cdb9512a9c288d1e21186f2f6ccbdd607d41

Bank/Algorithm: TPM_ALG_SHA1(0x0004)
PCR_09: 94 a2 cd b9 51 2a 9c 28 8d 1e 21 18 6f 2f 6c cb dd 60 7d 41

$ tpm2_createpolicy -P -i 9 -F pcr_09.bin -f policy.digest
$ xxd -p policy.digest 
65b786bb8a2edc6900d4b90ea6f20da30c3ea83da45a8c0fa323419c8f401f3e

$ tpm2_nvdefine -x 0x1500001 -a 0x40000001 -s 32 -t 0x2000A -r -L policy.digest -P <OWNER_PW>
$ tpm2_nvlist
1 NV indexes defined.

  0. NV Index: 0x1500001
  {
	Hash algorithm(nameAlg):11
 	The Index attributes(attributes):0xa0002
 	The size of the data area(dataSize):32
 	Authorization Policy for R/W/D: 65B786BB8A2EDC6900D4B90EA6F20DA30C3EA83DA45A8C0FA323419C8F401F3E
  }

$ tpm2_nvwrite -x 0x1500001 -a 0x40000001 -f /tmp/secret.bin -P <OWNER_PW>
$ tpm2_nvread -x 0x1500001 -a 0x40000001 -s 32 -o 0 -P <OWNER_PW>

The size of data:32
[...]

PR #363, opened 2017-07-20
Idesai tpm2 nvwrite size fix and add nice-name handling for nv attributes

williamcroberts:
This replaces PR #359 and adds to it.

This starts to get rid of specifying the attributes for nv indexes during nvdefine in raw numerical format.
I still need to modify tom2_nvdefine to use it, and update it's man page.

- Unit tests are in place for the new attribute keyword parser.
- System test is in place for writing a big nvindex.
- System tests in place for using nice names.
- System tests in place for verifying that nvread reads the same bytes as nvwrite.
- nvread output is in xxd -r compatible format.
- nvread -s is no longer needed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -93,7 +93,8 @@ lib_libcommon_a_SOURCES = \
     lib/tpm_hmac.c \
     lib/tpm_kdfa.c \
     lib/tpm_session.c \
-    lib/tpm2_util.c
+    lib/tpm2_util.c \
+    lib/tpm2_nv_util.c
 
 tools_tpm2_create_SOURCES = tools/tpm2_create.c tools/main.c
 tools_tpm2_createprimary_SOURCES = tools/tpm2_createprimary.c tools/main.c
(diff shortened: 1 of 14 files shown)


williamcroberts:
@martinezjavier - this helps coalesce the options for dealing with policy enforcement as well.

martinezjavier:
@williamcroberts yes, is just that I'm still not that familiar with the TPM2 NV features to properly review nv related code...

Once I've tpm2_unseal working properly (and AFAICT it's only missing my last PR #365), I'll dig into the NV support to fix issue #348 and so I should be able to review this.

williamcroberts:
@martinezjavier sounds good, im gonna leave this up through the weekend and look at landing this on Monday. It fixes the nvread and nvwrite issues, as well as putting tests in place at both the unit level, but also making sure that nvread/write behavior across large files works. It also is now verifying the results of read, that it matches the bytes written.

idesai:
@williamcroberts I very much like this concept. Thanks!

Issue #364, opened 2017-07-21
tpm2_unseal fails when using more than 3 PCRs for the auth policy session


martinezjavier:
The tpm2_unseal tool is not parsing correctly the raw PCR file that contains the hashes, only the first TPM2B_DIGEST buffer is filed.

This means that when using more than 3 PCRs for the auth policy session, it fails with the following error:

```
$ tpm2_unseal -c load.context -L 0x4:0,1,2,3 -F pcrs.bin
ERROR: File "pcrs.bin" size is larger than buffer, got 80 expected less than 64
```

PR #365, opened 2017-07-21
Refactor build_policy() and reuse it from tpm2_unseal tool (and also fix issue #364)

martinezjavier:
This pull request does some refactor on build_policy() so the function is generic and can be moved to PCR lib to be reused by other tools. For start it's used by tpm2_unseal but I believe it can also be used by other tools (e.g: tpm2_nvread that needs the same as mentioned in issue #348).

By using build_policy(), the raw PCR hashes file is parsed correctly so this set also fixes issue #364.

To easy reviewing, I've tried to split the changes in as many logical steps as possible so each patch does only one thing.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -93,6 +93,7 @@ lib_libcommon_a_SOURCES = \
     lib/tpm_hmac.c \
     lib/tpm_kdfa.c \
     lib/tpm_session.c \
+    lib/tpm2_policy.c \
     lib/tpm2_util.c \
     lib/tpm2_nv_util.c
 
--- a/lib/tpm2_policy.c
+++ b/lib/tpm2_policy.c
@@ -1,0 +1,235 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <errno.h>
+
+#include "files.h"
+#include "log.h"
+#include "tpm2_policy.h"
+
+static unsigned get_size_from_alg(TPMI_ALG_HASH hashAlg) {
+    switch (hashAlg) {
+        case TPM_ALG_SHA1:
+            return SHA1_DIGEST_SIZE;
+        case TPM_ALG_SHA256:
+            return SHA256_DIGEST_SIZE;
+        case TPM_ALG_SHA384:
+            return SHA384_DIGEST_SIZE;
+        case TPM_ALG_SHA512:
+            return SHA512_DIGEST_SIZE;
+        case TPM_ALG_SM3_256:
+            return SM3_256_DIGEST_SIZE;
+        default:
+            LOG_ERR("Unknown hashAlg, cannot determine digest size.\n");
+            return 0;
+    }
+}
+
+static bool evaluate_populate_pcr_digests(TPML_PCR_SELECTION pcr_selections,
+                                          char *raw_pcrs_file,
+                                          TPML_DIGEST *pcr_values) {
+    //octet value of a pcr selection group
+    uint8_t group_val=0;
+    //total pcr indices per algorithm/ bank. Typically this is 24
+    uint8_t total_indices_for_this_alg=0;
+    //cumulative size total of selected indices per hashAlg
+    unsigned expected_pcr_input_file_size=0;
+    //loop counters
+    unsigned i, j, k, dgst_cnt=0;
+    const uint8_t bits_per_nibble[] = {0,1,1,2,1,2,2,3,1,2,2,3,2,3,3,4};
+
+    //Iterating the number of pcr banks selected
+    for (i=0; i < pcr_selections.count; i++) {
+        //Looping to check total pcr select bits in the pcr-select-octets for a bank
+        for (j=0; j < pcr_selections.pcrSelections[i].sizeofSelect; j++) {
+            group_val = pcr_selections.pcrSelections[i].pcrSelect[j];
+            total_indices_for_this_alg += bits_per_nibble[group_val & 0x0f];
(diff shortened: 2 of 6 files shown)


lib/pcr.c
```diff
@@ -161,3 +163,225 @@ TPM_RC get_max_supported_pcrs(TSS2_SYS_CONTEXT *sapi_context, UINT32 *max_pcrs)
...
+                                                     policy_session_type,
+                                                     &symmetric,
+                                                     policy_digest_hash_alg);
+    if (rval != TPM_RC_SUCCESS) {
+        LOG_ERR("Failed tpm session start auth with params\n");
+    }
+    return rval;
+}
+
+TPM_RC build_policy(TSS2_SYS_CONTEXT *sapi_context,
```
williamcroberts: pcr_build_policy() since the file name is pcr.h

idesai: @williamcroberts i was suggesting it to be more like tpm2_policy_pcr by moving the function to tpm2_policy(new file)

martinezjavier:
@idesai thanks a lot for your review and feedback!

re: e3627ca - sorry but I didn't understand what you meant. Could you please elaborate on this and maybe provide an example? AFAICT the indirection level is never used.

re: 8052069 - there's no behavior change for tpm2_createpolicy aftir this commit. IOW, it will continue printing the EXTENDED_POLICY_SESSION_HANDLE and saving the file. The reason to take this out of build_policy() is to allow tpm2_unseal to use that function since we don't those actions to happen there.

For example, if I keep these in build_policy() then:

```
$ tpm2_unseal -c load.context -L 0x4:0,1,2,3 -F pcrs.bin
```

Will not only print to stdout the unsealed data but also the EXTENDED_POLICY_SESSION_HANDLE which makes scripts not able to use the tpm2_unseal output.

re: aba275c - yes, tpm2_unseal is an exception but we also need to do the same exception for tpm2_nvread (as mentioned in issue #348). So that would make three parts where the same logic is needed. I think there's a difference between the features that we add as an exception (and document these in the man pages) and how these features are implemented (and we should avoid code duplication as much as possible).

re: b6b7388 and 921e6c5 - Ok, I'll move these to a tpm2_policy.{c,h}

martinezjavier:
@williamcroberts thanks, I know from the coding_standard_c.txt file that functions should have a prefix for their modules. Is just that I forgot about that when moving to lib, sorry about that.

I'll make sure to comply with the coding standard on the next version.

martinezjavier:
@idesai @williamcroberts I've a pushed a rebased branch taking your comments into account.

Changes since v1:
- Move build_policy() and helper functions to lib/tpm2_policy.{c,h} instead of lib/pcr.{c,h}. Suggested by @idesai 
- Added module prefix to build_policy(). Suggested by @williamcroberts. Since I called the lib file tpm2_policy, I renamed it to tpm2_policy_build (tpm2_policy_build_policy sounded really bad...).

@idesai I didn't change anything with respect to your comments for e3627ca and 8052069, since as mentioned it was not clear to me what changes should I do for these (if any). Thanks!

idesai:
@martinezjavier, the level of indirection is necessary to make a policy value  have value X or Y. I know it is not obvious at the moment :) but gearing this one for the lua branch. So lets not merge e3627ca and 8052069 just yet. I understand the implication of unseal tool having to duplicate some minimal code for the idea about having this special use case of unseal enabled.  I will try and create a PR soon, so it can be obvious :) Thanks!

martinezjavier:
@idesai I still don't understand though why e3627ca would be an issue though, that's the only thing that prevents to reuse the build_policy() function. Why callers (like tpm2_createpolicy) can't just do it after calling build_policy() like in that commit?

AIso, it's seems cleaner for me that the function has no side effects (it's doing much more than just building a policy like printing and storing data).

idesai:
@martinezjavier from the commit, it appears you make the calls only when policy type is PCR

martinezjavier:
@idesai yes because that's the only user of the function in mainline :)

martinezjavier:
@idesai but got your point and I agree with you. I'll see to reuse build_pcr_policy() instead of the more generic build_policy() so doesn't affect the future code that implements the other EA policy.

Thanks a lot for your feedback!

williamcroberts:
I;m Ok with https://github.com/01org/tpm2.0-tools/commit/8052069a10b7deaeacf847832b92c27626dea4f5 as it lets us get build policy into a generic form that can be used to satisfy PCR policies in tools directly when using an RM. Lets just drophttps://github.com/01org/tpm2.0-tools/commit/8052069a10b7deaeacf847832b92c27626dea4f5 from the PR so we don't have a delete just to add it back in later (normally i'd say kill it, but if we have an in tree use by the next release, ill keep it till then).

idesai:
@martinezjavier , @williamcroberts sounds good. Thanks 👍

martinezjavier:
@williamcroberts @idesai Ok, I'll drop the patch removing the indirection level but keep the other one as you suggested then. Thanks a lot to both for the feedback!

martinezjavier:
@williamcroberts @idesai pushed a new branch addressing the issues you pointed out.

Changes since v2:
- Drop the patch that removed the indirection level on build_policy()
- Change test_tpm2_unseal test script to use a set of PCRs instead of a single one, this would had found the issue solved by this PR earlier.

martinezjavier:
@williamcroberts I've rebased this on top of 66430e173092 ("Treat file sizes as unsigned integers") since there were conflicts after PR #370 was merged.

PR #370, opened 2017-07-24
Use identical types in comparison

webmeister:
Without this change, the compiler complains about a comparison between signed and unsigned integer expressions (filesize != expected_pcr_input_file_size) and aborts the compilation (due to Werror). Since expected_pcr_input_file_size exists only to be compared to filesize, its type can simply be changed to match that of filesize.

### DIFF
--- a/tools/tpm2_createpolicy.c
+++ b/tools/tpm2_createpolicy.c
@@ -150,7 +150,7 @@ static bool evaluate_populate_pcr_digests(create_policy_ctx *pctx, TPML_DIGEST *
 
     //Check if the input pcrs file size is the same size as the pcr selection setlist
     if (pctx->pcr_policy_options.is_raw_pcrs_file) {
-        long filesize = 0;
+        unsigned long filesize = 0;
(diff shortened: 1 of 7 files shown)


tools/tpm2_createpolicy.c
```diff
@@ -112,7 +112,7 @@ static bool evaluate_populate_pcr_digests(create_policy_ctx *pctx, TPML_DIGEST *
     //total pcr indices per algorithm/ bank. Typically this is 24
     uint8_t total_indices_for_this_alg=0;
     //cumulative size total of selected indices per hashAlg
-    unsigned expected_pcr_input_file_size=0;
+    long expected_pcr_input_file_size=0;
```
williamcroberts: We would never expect this to go negative. The long for file_size seems to be from the ftell() call to get the file size, but it is checked for negative (negative ftell() return is error), so only values >= 0 are ever set. It looks like the interface for getting the file size should change to unsigned long.

webmeister: It slightly increased the size of the patch, but now everything is unsigned :)

williamcroberts:
@webmeister  Thanks

PR #371, opened 2017-07-24
Fix uninitialized inPublic.t.publicArea.authPolicy

gwei3:
In tpm2_create & tpm2_createprimary, if -L option was not specified
to initialize authPolicy, the authPolicy size of inPublic will be
random value and may cause sysapi context insufficient error.

Signed-off-by: Gang Wei <gang.wei@intel.com>


gwei3:
This issue was found when I run test/system/test_all.sh. 14 test items failed due to this issue.

PR #373, opened 2017-07-26
Algorithm nice names

williamcroberts:
This adds support for both PCR lists and algorithm specifiers to use human friendly names rather than hex. I also organized the man pages a little but so the redundant what algorithms are supported are grouped into common files. System tests were updated to use the new nice names.

TODO: Add unit tests.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -150,7 +150,8 @@ check_PROGRAMS = \
     test/unit/test_files \
     test/unit/test_tpm2_header \
     test/unit/test_tpm2_nv_util \
-    test/unit/test_tpm2_alg_util
+    test/unit/test_tpm2_alg_util \
+    test/unit/test_pcr
 
 test_unit_tpm2_rc_decode_unit_CFLAGS  = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
 test_unit_tpm2_rc_decode_unit_LDADD   = $(CMOCKA_LIBS) $(LIB_COMMON)
@@ -180,6 +181,10 @@ test_unit_test_tpm2_nv_util_SOURCES  = test/unit/test_tpm2_nv_util.c
 test_unit_test_tpm2_alg_util_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
 test_unit_test_tpm2_alg_util_LDADD    = $(CMOCKA_LIBS) $(LIB_COMMON)
 test_unit_test_tpm2_alg_util_SOURCES  = test/unit/test_tpm2_alg_util.c
+
+test_unit_test_pcr_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
+test_unit_test_pcr_LDADD    = $(CMOCKA_LIBS) $(LIB_COMMON)
+test_unit_test_pcr_SOURCES  = test/unit/test_pcr.c
 
 endif
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -223,7 +223,7 @@ man8_MANS = \
     man/man8/tpm2_dictionarylockout.8 \
     man/man8/tpm2_createpolicy.8
 
-man/man8/%.8 : man/%.8.in man/tcti-options.troff man/tcti-environment.troff
+man/man8/%.8 : man/%.8.in man/tcti-options.troff man/tcti-environment.troff man/alg-common.troff man/hash-alg-common.troff
 	rm -f $@
 	mkdir -p man/man8
 if HAVE_TCTI_DEV
@@ -238,6 +238,10 @@ endif
 	    -e '/@TCTI_OPTIONS_INCLUDE@/d' \
 	    -e '/@TCTI_ENVIRONMENT_INCLUDE@/r man/tcti-environment.troff' \
 	    -e '/@TCTI_ENVIRONMENT_INCLUDE@/d' \
+	    -e '/@ALG_COMMON_INCLUDE@/r man/alg-common.troff' \
+	    -e '/@ALG_COMMON_INCLUDE@/d' \
+	    -e '/@HASH_ALG_COMMON_INCLUDE@/r man/hash-alg-common.troff' \
+	    -e '/@HASH_ALG_COMMON_INCLUDE@/d' \
 	    < $< >> $@
 
 CLEANFILES = $(man8_MANS)
--- a/Makefile.am
+++ b/Makefile.am
@@ -95,7 +95,8 @@ lib_libcommon_a_SOURCES = \
     lib/tpm_session.c \
     lib/tpm2_policy.c \
     lib/tpm2_util.c \
-    lib/tpm2_nv_util.c
+    lib/tpm2_nv_util.c \
+    lib/tpm2_alg_util.c
 
 tools_tpm2_create_SOURCES = tools/tpm2_create.c tools/main.c
 tools_tpm2_createprimary_SOURCES = tools/tpm2_createprimary.c tools/main.c
@@ -148,7 +149,8 @@ check_PROGRAMS = \
     test/unit/test_string_bytes \
     test/unit/test_files \
     test/unit/test_tpm2_header \
-    test/unit/test_tpm2_nv_util
+    test/unit/test_tpm2_nv_util \
+    test/unit/test_tpm2_alg_util
 
 test_unit_tpm2_rc_decode_unit_CFLAGS  = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
 test_unit_tpm2_rc_decode_unit_LDADD   = $(CMOCKA_LIBS) $(LIB_COMMON)
@@ -174,6 +176,10 @@ test_unit_test_tpm2_header_SOURCES  = test/unit/test_tpm2_header.c
 test_unit_test_tpm2_nv_util_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
 test_unit_test_tpm2_nv_util_LDADD    = $(CMOCKA_LIBS) $(LIB_COMMON)
 test_unit_test_tpm2_nv_util_SOURCES  = test/unit/test_tpm2_nv_util.c
+
+test_unit_test_tpm2_alg_util_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
+test_unit_test_tpm2_alg_util_LDADD    = $(CMOCKA_LIBS) $(LIB_COMMON)
+test_unit_test_tpm2_alg_util_SOURCES  = test/unit/test_tpm2_alg_util.c
 
 endif
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -228,7 +228,7 @@ man8_MANS = \
     man/man8/tpm2_dictionarylockout.8 \
     man/man8/tpm2_createpolicy.8
 
-man/man8/%.8 : man/%.8.in man/common-options.troff man/tcti-options.troff man/tcti-environment.troff man/alg-common.troff man/hash-alg-common.troff man/object-alg-common.troff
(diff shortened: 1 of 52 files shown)


test/system/test_tpm2_createpolicy.sh
```diff
@@ -35,23 +35,21 @@
 
 declare -A expected_policy_digest=(["sha1"]="f28230c080bbe417141199e36d18978228d8948fc10a6a24921b9eba6bb1d988"
                                    ["sha256"]="33e36e786c878632494217c3f490e74ca0a3a122a8a4f3c5302500df3b32b3b8")
-declare -A hash_name=([0x4]="sha1" [0xb]="sha256")
 pcr_index=0
 
-for halg in 0x4 0xb
+for halg in sha1 sha256
 do
```
martinezjavier: In some tests I see that you use a mix of nice name and hex values and in other tests you convert all of them to nice names. I prefer the former to make sure that the tests exercise both options.

But I think there should be some consistency, either the tests should exercise both or all should be converted to nice names.

tools/tpm2_sign.c
```diff
@@ -289,8 +290,8 @@ static bool init(int argc, char *argv[], tpm_sign_ctx *ctx) {
         }
     }
 
-    if (!((flags.k || flags.c) && flags.g && flags.m && flags.s)) {
-        LOG_ERR("Expected options (k or c) and g and m and s");
+    if (!((flags.k || flags.c) && flags.m && flags.s)) {
+        LOG_ERR("Expected options (k or c) and m and s");
         return false;
```
martinezjavier: This is an unrelated fix (-g option isn't needed because sha1 is the default), so I think it should be in a separate patch with its own explanation why is needed.

tools/tpm2_verifysignature.c
```diff
@@ -321,7 +322,7 @@ int execute_tool(int argc, char *argv[], char *envp[], common_opts_t *opts,
 
     tpm2_verifysig_ctx ctx = {
             .flags = { .all = 0 },
-            .halg = TPM_ALG_SHA256,
+            .halg = TPM_ALG_SHA1,
             .msgHash = TPM2B_TYPE_INIT(TPM2B_DIGEST, buffer),
```
martinezjavier: I think changing the default should also be split in a separate commit.

martinezjavier:
@williamcroberts this is great! I've tested your branch and it works good for me. It also allowed me to remove code in a script that only had to convert from pretty names for users to the hex values to pass to the tools. Thanks!

I just had a few minor comments in case you want to take it into account.

Makefile.am
```diff
@@ -175,6 +177,10 @@ test_unit_test_tpm2_nv_util_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
 test_unit_test_tpm2_nv_util_LDADD    = $(CMOCKA_LIBS) $(LIB_COMMON)
 test_unit_test_tpm2_nv_util_SOURCES  = test/unit/test_tpm2_nv_util.c
 
+test_unit_test_tpm2_alg_util_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
+test_unit_test_tpm2_alg_util_LDADD    = $(CMOCKA_LIBS) $(LIB_COMMON)
+test_unit_test_tpm2_alg_util_SOURCES  = test/unit/test_tpm2_alg_util.c
+
```
williamcroberts: Yeah this is an artifact of me accidentally running git clean without adding the test files, I actually had them in place. So I want to re-create them and squash them into this commit.

lib/tpm2_alg_util.c
```diff
@@ -0,0 +1,143 @@
+#include <string.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
```
williamcroberts: done

lib/tpm2_alg_util.c
```diff
@@ -0,0 +1,143 @@
...
+#include <sapi/tpm20.h>
+
+#include "tpm2_util.h"
+#include "tpm2_alg_util.h"
+
```
williamcroberts: done

lib/tpm2_alg_util.c
```diff
@@ -0,0 +1,143 @@
...
+#endif
+#ifndef ALG_SHA3_384_VALUE
+#define ALG_SHA3_384_VALUE 0x28
+#endif
+#ifndef ALG_SHA3_512_VALUE
+#define ALG_SHA3_512_VALUE 0x29
+#endif
+
+void tpm2_alg_util_for_each_alg(tpm2_alg_util_alg_iteraror iterator, void *userdata) {
+
```
williamcroberts: Well it is iterating over the contents, what you choose to do with it is arbitrary. I was torn between just making the data extern vs encapsulating it like this. The encapsulation here insn't buying us much except for avoiding the writing of a for loop for what that's worth.

lib/tpm2_alg_util.c
```diff
@@ -0,0 +1,143 @@
...
+        { .name = "cbc", .id = ALG_CBC_VALUE },
+        { .name = "cfb", .id = ALG_CFB_VALUE },
+        { .name = "ecb", .id = ALG_ECB_VALUE },
+    };
+
+    size_t i;
+    for (i=0; i < ARRAY_LEN(algs); i++) {
+        const alg_pair *alg = &algs[i];
+        bool result = iterator(alg->id, alg->name, userdata);
+        if (result) {
```
williamcroberts: sure, that name is not very good.

test/system/test_tpm2_certify.sh
```diff
@@ -38,8 +38,9 @@ file_certify_key_name=name.load.B1_B8
 file_output_attest=attest.out
 file_output_signature=certify_signature.out 
   
-
+# do not update alg_hash without updating alg_hash_name
 alg_hash=0x000B
+alg_hash_name=sha256
```
williamcroberts: I wanted to mix and match the hex values and the nice names to get coverage on both paths.

man/object-alg-common.troff
```diff
@@ -0,0 +1,8 @@
+.br
+Supported public object algorithms are:
+  '0x1' or 'rsa' for TPM_ALG_RSA (default)
+  '0x8' or 'keyedhash' for TPM_ALG_KEYEDHASH
+  '0x23' or 'ecc' for TPM_ALG_ECC
+  '0x25' or 'sha512' for TPM_ALG_SHA512
+.br
```
williamcroberts: Good catch

lib/tpm2_alg_util.c
```diff
@@ -0,0 +1,143 @@
...
+typedef struct convert_data convert_data;
+struct convert_data {
+    const char *name;
+    TPM_ALG_ID id;
+};
+
+static bool convert_iterator(TPM_ALG_ID id, const char *name, void *userdata) {
+
+    convert_data *d = (convert_data *)userdata;
+
```
williamcroberts: Well if someone needs to pass other userdata along it makes it difficult. The fact that it's a convert_data structure is just because of how what is being done while iterating over the data elements.

tools/tpm2_verifysignature.c
williamcroberts: When I update the man page for it, the default is mentioned in the common file, so if I split it into a separate commit, there's a point in time where the man page doesn't match reality, which I don't think we would ever want to happen. Splitting it doesn't buy us anything, because if we revert it, we'd have to revert the other patch as well.

Makefile.am
martinezjavier: Ok.

lib/tpm2_alg_util.c
martinezjavier: Yes, it's iterating but I think that's an implementation detail. From the user POV is finding a match for a given algorithm name / ID. But I don't have a strong opinion on this so I'm OK on leaving it as is.

test/system/test_tpm2_certify.sh
martinezjavier: Ok.

tools/tpm2_verifysignature.c
martinezjavier: Ah, good point. I didn't think about the man page.

lib/tpm2_alg_util.c
martinezjavier: Yes, my point was that I don't see what other user data could be provided. Most likely alg_pair will be enough and if someone ever needs something different, then it could be made more generic.

lib/tpm2_alg_util.c
williamcroberts: You'll see how this indirection of a callback and void * is used in a more generic sense when you see the tests. In the tests, my callback for each element is searching a list and modifying it. Thus its iterator callback and userdata are unique to it.

I would like to modify a tool or tools so that they could pass an option to dump this map. Perhaps taking rc_decode, and modifying it to be able to decode algorithms and dump the map.

lib/tpm2_alg_util.c
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+}
+
+typedef struct convert_data convert_data;
+struct convert_data {
+    const char *name;
+    TPM_ALG_ID id;
+};
+
```
williamcroberts: Fair enough.

lib/tpm2_alg_util.c
williamcroberts: The tests will use a different one, and then eventually some tool to dump the map will use one, eahc with there own task to be done on each element, and all will have different userdata.

lib/tpm2_alg_util.c
```diff
@@ -0,0 +1,143 @@
...
+struct alg_pair {
+    const char *name;
+    TPM_ALG_ID id;
+};
+
+/*
+ * The TSS has a bug where it was missing algs 0x27 trough 0x29.
+ * see: https://github.com/01org/TPM2.0-TSS/issues/476
+ * per https://trustedcomputinggroup.org/wp-content/uploads/TCG_Algorithm_Registry_Rev_1.24.pdf
+ */
```
williamcroberts: FIXME with bug: https://github.com/01org/tpm2.0-tools/issues/375

lib/tpm2_alg_util.c
williamcroberts: done

lib/tpm2_alg_util.c
williamcroberts: done

williamcroberts:
@martinezjavier I think I got all your comments either addressed or fixed. Thank you for the review. I added the tests back in, that provide (I think) ample coverage. I'll leave this up for another day and see if any new feedback is left.

test/unit/test_tpm2_alg_util.c
```diff
@@ -0,0 +1,277 @@
...
+        assert_int_equal(value, found_id); \
+        assert_int_equal(value, from_hex_str); \
+        assert_int_equal(value, from_nice_str); \
+    }
+
+single_item_test(rsa, ALG_RSA_VALUE)
+/*
+ * sha sha1 is it's own test, as alg to string
+ * can return either, based on the map ordering.
+ *
```
williamcroberts: extra space, delete it.

test/unit/test_tpm2_alg_util.c
```diff
@@ -0,0 +1,277 @@
...
+    ALG_SYMCIPHER_VALUE,
+    ALG_CAMELLIA_VALUE,
+    ALG_SHA3_256_VALUE,
+    ALG_SHA3_384_VALUE,
+    ALG_SHA3_512_VALUE,
+    ALG_CTR_VALUE,
+    ALG_OFB_VALUE,
+    ALG_CBC_VALUE,
+    ALG_CFB_VALUE,
+    ALG_ECB_VALUE };
```
williamcroberts: formatting }; on last line.

lib/tpm2_alg_util.c
martinezjavier: @williamcroberts right, I looked and the tests and now I agree with the code. 

I just didn't want the indirection level if it would not be used (I didn't know the tests where using it since weren't pushed in the first version). Thanks for the clarification.

PR #377, opened 2017-07-27
tpm2_getmanufec: handle output more gracefully

williamcroberts:
Currently, the -E option will fail when -E is not specified:
$ tpm2_getmanufec -g 0x01 -O -N -U -f test_ek.pub -S https://ekop.intel.com/ekcertservice/
...
ERROR: Could not open file for writing: "(null)"

Make -E option, optional by making the default stdout. This requires
modifying the tool to stop outputing other information to sdout,
which is currently unused output in the test suite.

Make the information only output when verbose is enabled, this
includes curl verbose information.

Update errors and warnings to go to stderr.

Fixes: 361

### DIFF
--- a/tools/tpm2_getmanufec.c
+++ b/tools/tpm2_getmanufec.c
@@ -62,14 +62,14 @@ bool hexPasswd = false;
 TPM_HANDLE persistentHandle;
 UINT32 algorithmType = TPM_ALG_RSA;
 
-char *ECcertFile;
+static FILE *ECcertFile = NULL;
 char *EKserverAddr = NULL;
 unsigned int nonPersistentRead = 0;
 unsigned int SSL_NO_VERIFY = 0;
 unsigned int OfflineProv = 0;
 bool is_session_based_auth = false;
 TPMI_SH_AUTH_SESSION auth_session_handle = 0;
-
+static bool verbose = false;
 
 BYTE authPolicy[] = {0x83, 0x71, 0x97, 0x67, 0x44, 0x84, 0xB3, 0xF8,
                      0x1A, 0x90, 0xCC, 0x8D, 0x46, 0xA5, 0xD7, 0x24,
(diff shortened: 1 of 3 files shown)


williamcroberts:
@martinezjavier

tools/tpm2_getmanufec.c
```diff
@@ -332,11 +332,13 @@ unsigned char *HashEKPublicKey(void)
...
-    unsigned i;
-    for (i = 0; i < SHA256_DIGEST_LENGTH; i++) {
-        printf("%02X", hash[i]);
+    if (verbose) {
+        unsigned i;
+        for (i = 0; i < SHA256_DIGEST_LENGTH; i++) {
+            printf("%02X", hash[i]);
+        }
+        printf("\n");
     }
```
martinezjavier: Shouldn't this be changed to LOG_INFO() as well?

tools/tpm2_getmanufec.c
```diff
@@ -407,22 +409,16 @@ int RetrieveEndorsementCredentials(char *b64h)
     size_t len = 1 + strlen(b64h) + strlen(EKserverAddr);
     char *weblink = (char *)malloc(len);
     if (!weblink) {
-        LOG_ERR("Could not open file for writing: \"%s\"", ECcertFile);
-                return ret;
+        LOG_ERR("oom");
+        return ret;
```
martinezjavier: Can we make this a better error message?

tools/tpm2_getmanufec.c
williamcroberts: No, because then we end up with additional information that LOG_INFO prints out, filename and line number.... for each printf statement.

tools/tpm2_getmanufec.c
williamcroberts: LOG_ERR has filename and number, so a simple oom is  enough, considering the other data printed.

martinezjavier: Ok

tools/tpm2_getmanufec.c
martinezjavier: @williamcroberts oh, right. Sorry for the noise then.

PR #381, opened 2017-07-28
gcc7: Change two case fallthrough conditions that upset gcc7.

flihp:

### DIFF
--- a/tools/main.c
+++ b/tools/main.c
@@ -60,7 +60,7 @@ main (int   argc,
     case 1:
         execute_man (argv[0], envp);
         LOG_ERR ("failed to load manpage, check your environment / PATH\n");
-        /* fallthrough */
+        exit (1);
     case 2:
         exit (1);
     }
--- a/tools/tpm2_dump_capability.c
+++ b/tools/tpm2_dump_capability.c
@@ -600,7 +600,7 @@ dump_tpm_capability (TPMU_CAPABILITIES    *capabilities,
     case TPM_CAP_COMMANDS:
         dump_command_attr_array (capabilities->command.commandAttributes,
                                  capabilities->command.count);
-        /* fallthrough */
+        break;
     default:
         return 1;
     }


tools/main.c
```diff
@@ -60,7 +60,7 @@ main (int   argc,
     case 1:
         execute_man (argv[0], envp);
         LOG_ERR ("failed to load manpage, check your environment / PATH\n");
-        /* fallthrough */
+        exit (1);
```
pohly: There's a variety of ways to markup legitimate fallthrough cases (https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7) and the existing /* fallthrough */ works, but only when passing additional compile parameters. The toolchain where I ran into this problem enables the warning without using -Wimplicit-fallthrough=2, so the comment was not recognized.

Just calling plain exit() here would be simpler than expecting people compiling tpm2.0-tools to pick a suitable level for -Wimplicit-fallthrough, but if you prefer the current code, then I can also check whether the default warning level can be relaxed a bit.

tools/tpm2_dump_capability.c
```diff
@@ -600,7 +600,7 @@ dump_tpm_capability (TPMU_CAPABILITIES    *capabilities,
     case TPM_CAP_COMMANDS:
         dump_command_attr_array (capabilities->command.commandAttributes,
                                  capabilities->command.count);
-        /* fallthrough */
+        break;
```
pohly: This changes the semantic, now  dump_tpm_capability() no longer returns 1 for TPM_CAP_COMMANDS. Does that mean that gcc7 has found a legitimate bug by ignoring the "/* fallthrough */"?

webmeister: Yes, according to its documentation dump_tpm_capability should return 1 only when it fails to handle the parameters. But the only caller of this function ignores the return value, so nothing was actually broken.

tools/main.c
martinezjavier: @pohly those comments comes from commit 9466263446fb ("build: use supported comment to suppress GCC7 fallthrough warning") that fixed the issue for me. AFAIK -Wimplicit-fallthrough is the same than -Wimplicit-fallthrough=3.

Now, looking at the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html), it seems that indeed -Wimplicit-fallthrough=2 doesn't match this comment since uses the following regular expression to match .*falls?[ \t-]*thr(ough|u).*

Which seems weird for me since I would had expected -Wimplicit-fallthrough=2  to be more relaxed than -Wimplicit-fallthrough=3

It's a pity that I didn't use fallsthrough that's matched for level 2 and 3, but again -Wimplicit-fallthrough=4 only matches fallthrough...

tools/tpm2_dump_capability.c
martinezjavier: @pohly agreed, if that's a bug then it should be split in a different commit IMHO.

tools/main.c
williamcroberts: Why not change it to fallsthrough or set up the correct compiler settings in configure.ac?

flihp: IMHO the compiler settings you've got setup right now are correct. My reasoning went something like this: It's a legit fallthrough case but even then the original author (me?) felt it was necessary to document it with a comment. Might as well just fill that line with an explicit call and remove all ambiguity.

tools/tpm2_dump_capability.c
flihp: @webmeister nailed it: the bug never manifest due to the caller ignoring the return result. This means the errant behavior was never visible to users invoking this tool. I figured I'd save the git history an extra commit and just lump these two changes together since I found them both at the same time and they're fixing a common warning.

I'll defer to the maintainer / reviewer on this one.

PR #393, opened 2017-08-03
README.md: Update URLs to account for tpm2.0-tools -> tpm2-tools rename.

flihp:


martinezjavier:
BTW, I also like tpm2-tools more but I think that all the occurrences of the old project name should be replaced in the code base:

$ git grep tpm2.0-tools | wc -l
63

martinezjavier:
BTW, I also like tpm2-tools more but I think that all the occurrences of the old project name should be replaced in the code base:

$ git grep tpm2.0-tools | wc -l
63

martinezjavier:
@williamcroberts great.

williamcroberts:
On #394

Issue #404, opened 2017-08-15
tpm2_rc_decode should be install into bin, not sbin


liuqun:
Maybe we should install `tpm2_rc_decode` into /usr/local/bin, since it doesn't require any root permissions (such as reading /dev/tpm0 directly)

williamcroberts:
I don't even know why they are in sbin, /usr/bin or /usr/local/bin seem more appropriate for all tools.

williamcroberts:
@martinezjavier - How would this affect your packaging, and what input do you have?

If I move stuff, do I somehow need to instruct autotools to look in the old location for make uninstall,
or do we not care?

liuqun:
https://github.com/01org/tpm2-tools/blob/a635550b264bf4e2e51a69c9384b671363adcc40/Makefile.am#L43

Autotools provide both `sbin_PROGRAMS` and `bin_PROGRAMS` for projects.

Move tpm2_rc_decode into `bin_PROGRAMS=tpm2_rc_decode` should work. And"make uninstall" would also remove tpm2_rc_decode from "/usr/local/bin/" as expected.

martinezjavier:
@williamcroberts I don't maintain a tpm2-tools package, they are @yunyings and @snits so you should ask them how this may affect them.

But my understanding is that upstream shouldn't care about what packages do, instead packages should have an upgrade for cases like this when upstream changes an install path for binaries.

yunyings:
Changing tools installation location in upstream requires SPEC file update to specify the new location for Fedora package. As long as there's no API/ABI change, the SPEC change is quite simple and straight forward.

williamcroberts:
@liuqun my comment was around someone already doing an install from source, so the tool is in sbin, and then moves to bin, and they never did an uninstall between changes, they would have an orphaned sbin/<tool> file kicking around and then two instances of it on PATH. I didn't know if autotools had a way of saying, I moved stuff, so on install go poke around and remove dead ones.

With that said, I see no really good reason any of the tools should be in sbin, but I am inclined to either leave them as be, or move them all to bin.

martinezjavier:
@williamcroberts my opinion is that people that are installing from source (instead using their distro package manger) should already know what they are doing. I don't think that we should add this kind of backward compatibility in the autotools scripts.

About the path to install the tools, I think that makes sense to just have all the tools in bin.

williamcroberts:
[...]

PR #409, opened 2017-08-16
autotools: move tools from sbin to bin

williamcroberts:
sbin is for critical startup utilities. tpm2-tools should
be in regular bin.

Verified by setting DESTDIR=~/tmp and that make install sends
it to /tmp/usr/local/bin.

Fixes: #404


williamcroberts:
@martinezjavier Where code lives is not really a security solution, PATH can be manipulated, etc.
dbus has policy access controls and has LSM integration with SE Linux for those so inclined to set it up.

With that said, the security model around most usages of the tpm allow untrusted resources on the machine to access and use the tpm.

martinezjavier:
@williamcroberts exactly, that's what I tried to say...

PR #415, opened 2017-08-18
Improve log messages support

martinezjavier:
Currently the tools only have options to set two different log levels, -V to print all the log levels or not passing an option to only print error messages.

But users may want more fine grained control on what is printed by the tools, for example printing debug information but without the file and line numbers or to only print warning and error messages.

The latter is useful if the tools are used by scripts, which otherwise will have to redirect the tools standard output to /dev/null or other tricks to avoid cluttering the scripts output.

For example in the shell scripts where I'm using the tpm2-tools, I need to do things like this:

```
if ! tpm2_createprimary -A "$auth" -g "$hash" -G "$key" -C "$primary_context" > /dev/null; then
```
Because the tpm2_createprimary tool prints some information that is not relevant for the user of the script.

I thought that it would be nice to have a quiet option that prevents the tool to print any output, so instead I can have this:

```
if ! tpm2_createprimary -Q -A "$auth" -g "$hash" -G "$key" -C "$primary_context"; then
```

So this patch improves the log message support, by adding a debug and quiet options to provide this control for log levels. It also makes the default log level to be info, that way info, warning and error messages are printed if no option is used.

### DIFF
--- a/man/common-options.troff
+++ b/man/common-options.troff
@@ -11,4 +11,9 @@ Display version information for this tool.
 .TP
 \fB\-V,\ \-\-verbose\fR
 Increase the information that the tool prints to the console during its
-execution.
+execution. When using this option the file and line number are printed.
+.TP
+.TP
+\fB\-Q,\ \-\-quiet\fR
+Decrease the information that the tool prints to the console during its
+execution. With this option the tools output messages are not printed.
--- a/tools/main.c
+++ b/tools/main.c
@@ -35,6 +35,8 @@
 #include "main.h"
 #include "options.h"
 
+bool output_enabled = true;
+
 /*
  * This program is a template for TPM2 tools that use the SAPI. It does
  * nothing more than parsing command line options that allow the caller to
@@ -71,6 +73,10 @@ main (int   argc,
     }
     if (opts.verbose)
         log_set_level(log_level_verbose);
+
+    if (opts.quiet) {
+        disable_output();
+    }
 
     sapi_context = sapi_init_from_options (&opts);
     if (sapi_context == NULL)
--- a/tools/main.c
+++ b/tools/main.c
@@ -59,7 +59,7 @@ main (int   argc,
     switch (sanity_check_common (&opts)) {
     case 1:
         execute_man (argv[0], envp);
-        LOG_ERR ("failed to load manpage, check your environment / PATH\n");
+        LOG_ERR ("failed to load manpage, check your environment / PATH");
         exit (1);
     case 2:
         exit (1);
--- a/tools/main.h
+++ b/tools/main.h
@@ -32,7 +32,10 @@
 #define MAIN_H
 
 #include <sapi/tpm20.h>
+#include <stdbool.h>
 #include "options.h"
+
+extern bool output_enabled;
 
 int
 execute_tool (int              argc,
@@ -40,5 +43,26 @@ execute_tool (int              argc,
               char             *envp[],
               common_opts_t    *opts,
               TSS2_SYS_CONTEXT *sapi_context);
+
+/*
+ * Prints output messages if not disabled. Output messages are what tools prints
+ * to the standard output during normal execution.
+ */
+#define TOOL_OUTPUT(fmt, ...)                   \
+    do {                                        \
+        if (output_enabled) {                   \
+            printf(fmt, ##__VA_ARGS__);         \
+        }                                       \
+    } while (0)
+
+static inline void enable_output(void)
+{
+    output_enabled = true;
+}
+
+static inline void disable_output(void)
+{
+    output_enabled = false;
+}
 
 #endif /* MAIN_H */
--- a/lib/options.c
+++ b/lib/options.c
@@ -226,7 +226,7 @@ get_common_opts (int                    *argc_param,
     char **argv = *argv_param;
 
     int c = 0, option_index = 0;
-    char *arg_str = "-T:d:R:p:hvV";
+    char *arg_str = "-T:d:R:p:hvVQ";
     struct option long_options [] = {
         {
             .name    = "tcti",
@@ -270,6 +270,12 @@ get_common_opts (int                    *argc_param,
             .has_arg = no_argument,
             .flag    = NULL,
             .val     = 'V',
+        },
+        {
+            .name    = "quiet",
+            .has_arg = no_argument,
+            .flag    = NULL,
+            .val     = 'Q',
         },
         {
             .name    = "version",
@@ -350,6 +356,9 @@ get_common_opts (int                    *argc_param,
             break;
         case 'V':
             common_opts->verbose = true;
+            break;
+        case 'Q':
+            common_opts->quiet = true;
             break;
         case 'v':
             common_opts->version = true;
--- a/lib/options.c
+++ b/lib/options.c
@@ -196,7 +196,7 @@ append_arg_to_vector (int*  argc,
     if (new_argv != NULL) {
         new_argv[*argc - 1] = arg_string;
     } else {
-        LOG_ERR("Failed to realloc new_argv to append string %s: %s\n",
+        LOG_ERR("Failed to realloc new_argv to append string %s: %s",
                 arg_string,
                 strerror (errno));
     }
--- a/lib/options.h
+++ b/lib/options.h
@@ -76,6 +76,7 @@
     .socket_port    = TCTI_SOCKET_DEFAULT_PORT, \
     .help           = false, \
     .verbose        = false, \
+    .quiet          = false, \
     .version        = false, \
 }
 
@@ -100,6 +101,7 @@ typedef struct {
     uint16_t  socket_port;
     int       help;
     int       verbose;
+    int       quiet;
     int       version;
 } common_opts_t;
 
--- a/lib/tpm2_policy.c
+++ b/lib/tpm2_policy.c
@@ -56,7 +56,7 @@ static bool evaluate_populate_pcr_digests(TPML_PCR_SELECTION pcr_selections,
         unsigned long filesize = 0;
         bool result = files_get_file_size(raw_pcrs_file, &filesize);
         if (!result) {
-            LOG_ERR("Could not retrieve raw_pcrs_file size\n");
+            LOG_ERR("Could not retrieve raw_pcrs_file size");
             return false;
         }
         if (filesize != expected_pcr_input_file_size) {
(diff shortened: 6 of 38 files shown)


lib/log.c
```diff
@@ -65,8 +65,5 @@ _log (log_level level, const char *file, unsigned lineno, const char *fmt, ...)
     /* Print the user supplied message */
     vfprintf (stream, fmt, argptr);
 
-    /* always add a new line so the user doesn't have to */
-    fprintf (stream, "\n");
```
tstruk: Hi @martinezjavier,
Instead of removing the new line here and adding it to every invocation of the log macro
can we add new set macros that don't add the '\n'. The user can then choose if \n should be printed or not
bu using different macros. What about something like this below. Same effect, but much less churn:

diff --git a/lib/log.c b/lib/log.c
index 8078a00..1d4e690 100644
--- a/lib/log.c
+++ b/lib/log.c
@@ -42,7 +42,7 @@ get_level_msg (log_level level)
 }
 
 void
-_log (log_level level, const char *file, unsigned lineno, const char *fmt, ...)
+_log (log_level level, const char *file, unsigned lineno, bool nl, const char *fmt, ...)
 {
 
     /* Skip printing messages outside of the log level */
@@ -62,8 +62,8 @@ _log (log_level level, const char *file, unsigned lineno, const char *fmt, ...)
     /* Print the user supplied message */
     vfprintf (stderr, fmt, argptr);
 
-    /* always add a new line so the user doesn't have to */
-    fprintf (stderr, "\n");
+    if (nl)
+        fprintf (stderr, "\n");
 
     va_end(argptr);
 }
diff --git a/lib/log.h b/lib/log.h
index 32f43ec..5e477f3 100644
--- a/lib/log.h
+++ b/lib/log.h
@@ -11,7 +11,8 @@ typedef enum log_level log_level;
  *
  * Use this to log all error conditions.
  */
-#define LOG_ERR(fmt, ...) _log(log_level_error, __FILE__, __LINE__, fmt, ##__VA_ARGS__)
+#define LOG_ERR(fmt, ...) _log(log_level_error, __FILE__, __LINE__, 1, fmt, ##__VA_ARGS__)
+#define LOG_ERR_NONL(fmt, ...) _log(log_level_error, __FILE__, __LINE__, 0, fmt, ##__VA_ARGS__)
 
 /*
  * Prints an warning message. The fmt and variadic arguments mirror printf.
@@ -19,7 +20,8 @@ typedef enum log_level log_level;
  * Use this to log a warning. A warning is when something is wrong, but it is not a fatal
  * issue.
  */
-#define LOG_WARN(fmt, ...) _log(log_level_warning, __FILE__, __LINE__, fmt, ##__VA_ARGS__)
+#define LOG_WARN(fmt, ...) _log(log_level_warning, __FILE__, __LINE__, 1, fmt, ##__VA_ARGS__)
+#define LOG_WARN_NONL(fmt, ...) _log(log_level_warning, __FILE__, __LINE__, 0, fmt, ##__VA_ARGS__)
 
 /*
  * Prints an informational message. The fmt and variadic arguments mirror printf.
@@ -27,7 +29,8 @@ typedef enum log_level log_level;
  * Informational messages are only shown when verboseness is increased. Valid messages
  * would be debugging type messages where additional, extraneous information is printed.
  */
-#define LOG_INFO(fmt, ...) _log(log_level_verbose, __FILE__, __LINE__, fmt, ##__VA_ARGS__)
+#define LOG_INFO(fmt, ...) _log(log_level_verbose, __FILE__, __LINE__, 1, fmt, ##__VA_ARGS__)
+#define LOG_INFO_NONL(fmt, ...) _log(log_level_verbose, __FILE__, __LINE__, 0, fmt, ##__VA_ARGS__)
 
 enum log_level
 {
@@ -45,6 +48,6 @@ bool
 log_set_level (log_level level);
 
 void
-_log (log_level level, const char *file, unsigned lineno, const char *fmt, ...) __attribute__ ((format (printf, 4, 5)));
+_log (log_level level, const char *file, unsigned lineno, bool nl, const char *fmt, ...) __attribute__ ((format (printf, 5, 6)));
 
 #endif /* SRC_LOG_H_ */

martinezjavier: @tstruk thanks a lot for your feedback. In fact I thought about that option too, but at the end I decided not to do it because:

* I couldn't come with a good name for the macros that wouldn't add the new line.
* Other output printing functions that I'm familiar with (i.e: printk, pr_* and dev_* in Linux) don't add a new-line char to the string and let the developer to device. 
* There's a lot of places in tpm2-tools where LOG_* macros are called with a new-line char manually added to the string. So I wouldn't say that the interface is very obvious, it seems that many people expects a printf-like behavior.

So I thought that more macros would just add more confusion to the developers for little gain.

Having said that, I don't have a strong opinion on this so I'll be happy to do what you suggest if people decide that's the better approach.

tstruk:
On 08/18/2017 10:23 AM, Javier Martinez Canillas wrote:
> So I thought that more macros would just add more confusion to the developers for little gain.
> 
> Having said that, I don't have a strong opinion on this so I'll be happy to do what you suggest if people decide that's the better approach.
> 

So let's have @williamcroberts make the call.
Thanks,
-- 
Tadeusz

lib/log.h
```diff
@@ -22,16 +22,24 @@ typedef enum log_level log_level;
...
+ * for debugging purposes but just to verbose during normal execution.
+ */
+#define LOG_DEBUG(fmt, ...) _log(log_level_debug, __FILE__, __LINE__, fmt, ##__VA_ARGS__)
+
+/*
  * Prints an informational message. The fmt and variadic arguments mirror printf.
  *
- * Informational messages are only shown when verboseness is increased. Valid messages
- * would be debugging type messages where additional, extraneous information is printed.
+ * Use this to log an informational message. Informational message is what tools prints
```
williamcroberts: Informational messages are what tools print...

man/common-options.troff
```diff
@@ -11,4 +11,12 @@ Display version information for this tool.
...
-execution.
+execution. When using this option the file and line number are printed.
+.TP
+\fB\-B,\ \-\-debug\fR
+Increase the information that the tool prints to the console during its
+execution. When using this option debug information is printed.
+.TP
+\fB\-Q,\ \-\-quiet\fR
+Decrease the information that the tool prints to the console during its
+execution. Wit this option only warning and errors messages are printed.
```
williamcroberts: s/Wit/With

lib/context-util.c
```diff
@@ -139,18 +139,18 @@ tcti_tabrmd_init (void)
 
     rc = tss2_tcti_tabrmd_init(NULL, &size);
     if (rc != TSS2_RC_SUCCESS) {
-        LOG_ERR ("Failed to get size for TABRMD TCTI context: 0x%" PRIx32, rc);
+        LOG_ERR ("Failed to get size for TABRMD TCTI context: 0x%" PRIx32 "\n", rc);
```
williamcroberts: The only point I see for these is when printing hex dumps, or things with multiple invocations of printf. Perhaps we could format the data as a string and output it so we don't need to modify these.

lib/log.h
martinezjavier: Ok.

man/common-options.troff
martinezjavier: Thanks for pointing this typo.

lib/context-util.c
martinezjavier: @williamcroberts I'm not a big fan of manipulating strings in C if it can be avoided. There's always the risk to introduce subtle bugs that can later be exploited as security vulnerabilities.

I also fail to see what's so inconvenient about needing a new line character. I would guess that most C programmers' muscle memory is trained to add a '\n' since are used to printf() behavior.

But if you have a strong opinion about this, I can convert the multiple printous as a string and print it with a single LOG_*() call.

We would still need a similar patch than this one though, since if we go that route then we will need to fix all the places that are needlessly adding a '\n' (which kind of proves my point that people are used to add a new line char):

```
$ git grep "LOG_" tools/ | grep "\\\n\"" | wc -l                  
146
```

williamcroberts: The old code that the LOG_ macros replaced was often time missing new lines or adding superfluous formatting. The goal of the LOG_ macros was to provide consistent behavior of tool output that is not part of normal operation.

The more I think about this, we really don't need LOG_ macros for output, except to control quiet. The LOG_ macros were designed for optional/error output in a specific formatted version. I don't think general tool output is appropriate for them.

Maybe we could just place a macro in main.h TOOL_OUTPUT() that outputs to stdout and they control the formatting completely.

martinezjavier: Sounds good to me. I'll do that.

BTW, do you want the patch that splits on stderr and stdout or do you prefer to keep LOG_INFO() printing to stderr? (I prefer stdout but have no strong opinion)

williamcroberts: Anything warning or debug wise should go to stderr. Anything the tool *should* output, should go to stdout.

tools/main.h
```diff
@@ -32,13 +32,36 @@
...
               common_opts_t    *opts,
               TSS2_SYS_CONTEXT *sapi_context);
 
+/*
+ * Prints output messages if not disabled. Output messages are what tools prints
+ * to the standard output during normal execution.
+ */
+#define TOOL_OUTPUT(fmt, ...)                   \
+    do {                                        \
+        if (output_enabled)                     \
```
williamcroberts: brackets on single if

tools/main.h
```diff
@@ -32,13 +32,36 @@
 #define MAIN_H
 
 #include <sapi/tpm20.h>
+#include <stdbool.h>
 #include "options.h"
 
+bool output_enabled;
```
williamcroberts: You're getting lucky here that this is a tentative declaration and the linker isn't complaining. This is extern by default, but I would explicitly extern this and put the definition in main.c. This way we can avoid any linker issues if the code gets moved around later.

tools/main.c
```diff
@@ -69,8 +69,15 @@ main (int   argc,
         showVersion (argv[0]);
         exit (0);
     }
+
     if (opts.verbose)
         log_set_level(log_level_verbose);
```
williamcroberts: Does the dis-coupling of verbose and quiet do what you want when quiet is enabled? I'm ok with the behavior, as it matches grep:
```
echo foo | grep ",[" -q
grep: Invalid regular expression
```

williamcroberts:
I think if you rebase this on master your travis build should start up again

tools/main.h
martinezjavier: Right, agreed.

tools/main.h
martinezjavier: Ok.

tools/main.c
martinezjavier: I believe so. I decoupled it purposefully because they go to a different standard stream anyways. Everything that's printed by verbose goes to stderr and everything that's affected by quiet is what goes to stdout.

PR #416, opened 2017-08-21
tpm2_dump_capability: Display specification minor version correctly

webmeister:
value needs to be converted to float before doing the division, otherwise the minor version will be truncated to zero.


webmeister:
Is the Travis integration broken? It reports errors for something in tpm2-abrmd, but nothing that has been touched by my commit.

williamcroberts:
Hopefully #417 gets the builds running again, and I can restart your build.

williamcroberts:
on #418

PR #419, opened 2017-08-21
coveralls: fix uploads

williamcroberts:
Coveralls was not being updated as ENABLE_COVERAGE was exported in
a seperate environment and not preserved when invoking the coveralls
upload script. Move the coveralls upload invocation within that
environment.

### DIFF
--- a/.ci/coveralls-upload.sh
+++ b/.ci/coveralls-upload.sh
@@ -1,16 +1,17 @@
 #!/usr/bin/env bash
 
 if [ "$ENABLE_COVERAGE" != "true" ]; then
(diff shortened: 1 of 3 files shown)


williamcroberts:
I didn't understand that ALL encrypted env vars are not available to PRs. See https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml for information.

Thus to trigger scans, we would need to start a trusted CI job.

williamcroberts:
Actually we would need the first change to get the ENABLE_COVERAGE set and make it so when ENABLE_COVERAGE is set, COVERALLS_REPO_TOKEN can be not-set.

PR #420, opened 2017-08-21
test: test tpm2_rc_decode

williamcroberts:

### DIFF
--- a/test/system/test_tpm2_rc_decode.sh
+++ b/test/system/test_tpm2_rc_decode.sh
@@ -1,0 +1,173 @@
+#;**********************************************************************;
+#
+# Copyright (c) 2016, Intel Corporation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of Intel Corporation nor the names of its contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+#;**********************************************************************;
+#!/bin/bash
+
+onerror() {
+    echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
+    exit 1
+}
+trap onerror ERR
+
+#
+# codes was generated from the TPM_RC constants in:
+# https://github.com/01org/tpm2-tss/blob/master/include/sapi/tss2_tpm2_types.h#L68
+# Some of these may not be used correctly, which is OK, as tpm2_rc_decode never
+# fails and should attempt to decode it or print some unkown status. This gives
+# us coverage for both known and unkown/malformed inputs.
+#
+# Details on error code encoding can be found at:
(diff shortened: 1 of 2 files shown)


test/system/test_tpm2_rc_decode.sh
```diff
@@ -0,0 +1,165 @@
...
+  [TPM_RC_9]=0x900
+  [TPM_RC_A]=0xA00
+  [TPM_RC_B]=0xB00
+  [TPM_RC_C]=0xC00
+  [TPM_RC_D]=0xD00
+  [TPM_RC_E]=0xE00
+  [TPM_RC_F]=0xF00
+  [TPM_RC_N_MASK]=0xF00
+)
+
```
martinezjavier: This is not correct, the values in https://github.com/01org/tpm2-tss/blob/master/include/sapi/tss2_tpm2_types.h#L68 are not a list of error codes but instead a table of RC constants that is what error codes use it to encode information about the error.

The document "Trusted Platform Module Library Part 2: Structures Family “2.0” Level 00 Revision 01.38" explains how the information is decoded in section "6.6.2 Response Code Formats", and has an example for the error code when the TPM2_Startup() command was called with a wrong startupType.

In this case the command will consist of two components:

* TPM_RC_VALUE (value is out of range or is not correct for the context)
* RC_Startup_startupType

RC_Startup_startupType is encoded as TPM_RC_P (a parameter related error) + TPM_RC_1 (for the first parameter).

So the error code is:

error code = TPM_RC_VALUE + RC_Startup_startupType
error code = (RC_FMT1 + 0x004) + (TPM_RC_P + TPM_RC_1)
error code = (0x080 + 0x004) + (0x40 + 0x100)
error code = 0x1c4

The tpm2_rc_decode tool is used to decode the errors and get this information:

```
$ tpm2_rc_decode 0x1c4                                       
error layer                 
  hex: 0x0                  
  identifier: TSS2_TPM_ERROR_LEVEL                      
  description: Error produced by the TPM                
format 1 error code         
  hex: 0x04                 
  identifier: TPM_RC_VALUE  
  description: value is out of range or is not correct for the context                                           
parameter                   
  hex: 0x100                
  identifier:  TPM_RC_1     
  description:  (null)
```

So I think what we should test instead is if the tpm2_rc_decode tool is printing the correct information for a couple of well known error codes.

test/system/test_tpm2_rc_decode.sh
```diff
@@ -0,0 +1,165 @@
...
+onerror() {
+    echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
+    exit 1
+}
+trap onerror ERR
+
+declare -A codes
+codes=(
+  [TPM_RC_SUCCESS]=0x0
```
williamcroberts: It's still a return code, what does strerror(0) return. In theory we should be testing all sorts of codes, even unknown/invalid codes.

test/system/test_tpm2_rc_decode.sh
williamcroberts: I autogenerated the table, I can drop anything that was incorectly added.

test/system/test_tpm2_rc_decode.sh
martinezjavier: Fair enough, although currently the tpm2_rc_decode tool doesn't even support TPM_RC_SUCCESS

```
$ tpm2_rc_decode 0x0
error layer
  hex: 0x0
  identifier: TSS2_TPM_ERROR_LEVEL
  description: Error produced by the TPM
ERROR: version 1.2 TPM_RCs are not supported
```
Btw, this error is not propagated and so the tool exits with success:

```
$ echo $?                                                    
0
```
I think both issues should be fixed (not related to this PR of course).

test/system/test_tpm2_rc_decode.sh
martinezjavier: Ok, If you want to test passing not valid response codes to the tool, then I think that you should at least check the command exit code (ideally compare the output with the expected human readable response code name for the valid ones).

test/system/test_tpm2_rc_decode.sh
```diff
@@ -0,0 +1,177 @@
...
+# us coverage for both known and unkown/malformed inputs.
+#
+# Details on error code encoding can be found at:
+# Section 6.6.2 of t "Trusted Platform Module Library Part 2: Structures Family “2.0” Level 00 Revision 01.38"
+#  - https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf
+#
+declare -A codes
+codes=(
+  [TPM_RC_SUCCESS]=0x0
+  [TPM_RC_BAD_TAG]=0x1E
```
martinezjavier: The tool also doesn't support this error code, but IIUC the spec correctly it should since a TPM2 with 1.2 compatibility could return this response code:

```
$ tpm2_rc_decode 0x01E
error layer
  hex: 0x0
  identifier: TSS2_TPM_ERROR_LEVEL
  description: Error produced by the TPM
ERROR: version 1.2 TPM_RCs are not supported
```

test/system/test_tpm2_rc_decode.sh
```diff
@@ -0,0 +1,177 @@
...
+  [TPM_RC_REFERENCE_S2]=0x91A
+  [TPM_RC_REFERENCE_S3]=0x91B
+  [TPM_RC_REFERENCE_S4]=0x91C
+  [TPM_RC_REFERENCE_S5]=0x91D
+  [TPM_RC_REFERENCE_S6]=0x91E
+  [TPM_RC_NV_RATE]=0x920
+  [TPM_RC_LOCKOUT]=0x921
+  [TPM_RC_RETRY]=0x922
+  [TPM_RC_NV_UNAVAILABLE]=0x923
+  [TPM_RC_NOT_USED]=0x97F
```
martinezjavier: I would only include until here. As mentioned in my previous comment, the next ones are not response codes but defines used to report additional information about the error and are encoded in a response code.

In fact, if you look at Table 16 in TCG part 2 doc, it says "Additional Defines" before listing these.

martinezjavier: Probably worth adding a comment in https://github.com/01org/tpm2-tss/blob/master/include/sapi/tss2_tpm2_types.h#L68 to make more clear this distinction. What do you think @flihp ?

williamcroberts: That's fine.

flihp: Not sure. Adding clarifying comments to the headers / data types for stuff like this feels like a slippery slope to me.

williamcroberts:
@martinezjavier Thank you for participating in the review.

PR #423, opened 2017-08-22
travis: remove tpm2-abrmd patch that has already been merged

martinezjavier:
The tpm2-tools repository has a local hack to allow travis to apply a set
of patches that are needed for tpm2-abrmd to be build with tpm2-tss master
branch. But one of those patches have already been merged in tpm2-abrmd so
trying to apply it fails, which makes the other patches to not be applied.


williamcroberts:
It's quite a hack :-)

Issue #422, opened 2017-08-22
No tools are available for Hash/HMAC/Event Sequences


liuqun:
There isn't any tool that uses the following Hash, HMAC, or event sequences API:
1. Hash sequences commands:
 Tss2_Sys_HashSequenceStart() / Tss2_Sys_SequenceUpdate() / Tss2_Sys_SequenceComplete()
1. HMAC sequences commands:
 Tss2_Sys_HMAC_Start() / Tss2_Sys_SequenceUpdate() / Tss2_Sys_SequenceComplete()
1. Event sequences commands:
 Tss2_Sys_HashSequenceStart() / Tss2_Sys_SequenceUpdate() / Tss2_Sys_EventSequenceComplete()

In my opinion, we may improve the existing tools by introducing sequences commands.
- An enhanced version of "**tpm2_hash**" to support any input file larger than 1024 bytes.
- An enhanced version of "**tpm2_hmac**" to support any input file as well.
- A simple version of "**tpm2_pcrevent**" that uses single Tss2_Sys_PCR_Event() command to handle with data within 1024 bytes. ("**tpm2_pcrextend**" requires an existing hash digest, while "tpm2_pcrevent" doesn't.)
- An enhanced version of "tpm2_pcrevent" which uses the event sequences to handle larger files.

williamcroberts:
You could just make the existing tools smart enough to handle files bigger than 1024 bytes, like how nvread and nvwrite were just fixed. As far as tpm2_pcrevent, that should be a separate tool from extend.

williamcroberts:
I updated hash on PR #424. The only downside, is that when a file is < 1024 bytes, there is 3 call outs to the TPM. However, handling stdin when it was a fifo and unknown file size was important to me. We could optimize that later if need be and detect when seek() fails on the fifo (illegal seek).

williamcroberts:
For tpm2_pcrevent, I have that started, but the sequencing is throwing an error "TPM_RC_AUTH_MISSING" on the Tss2_Sys_EventSequenceComplete call.

See:
https://github.com/williamcroberts/tpm2.0-tools/tree/tpm2_pcrevent

williamcroberts:
Ahh I figured it out. TPM2_SequenceComplete has @ on both pcrHandle and sequenceHandle. Per the spec, @ is a decorator that "indicates that an authorization session is required for use of the entity associated with the handle.". Thus for the SequenceComplete, one needs to specify 2 entries to the command auths array.

[...]

PR #424, opened 2017-08-23
hmac/hash large files and clean up option handling

williamcroberts:
This is a partial fix to #422

### DIFF
--- a/lib/tpm_hash.c
+++ b/lib/tpm_hash.c
@@ -1,116 +1,38 @@
+#include <errno.h>
 #include <string.h>
 
 #include <sapi/tpm20.h>
 
+#include "log.h"
+#include "files.h"
+#include "tpm_hash.h"
 #include "tpm2_util.h"
 
-UINT32 tpm_hash(TSS2_SYS_CONTEXT *sapi_context, TPMI_ALG_HASH hashAlg,
-        UINT16 size, BYTE *data, TPM2B_DIGEST *result) {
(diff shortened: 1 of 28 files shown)


lib/tpm_hash.c
```diff
@@ -134,3 +136,75 @@ UINT32 tpm_hash_sequence(TSS2_SYS_CONTEXT *sapi_context, TPMI_ALG_HASH hash_alg,
...
+
+    TPMS_AUTH_COMMAND cmdAuth = {
+        .sessionHandle = TPM_RS_PW,
+        .nonce = TPM2B_EMPTY_INIT,
+        .hmac = TPM2B_EMPTY_INIT,
+        .sessionAttributes = SESSION_ATTRIBUTES_INIT(0),
+    };
+    TPMS_AUTH_COMMAND *cmdSessionArray[1] = { &cmdAuth };
+    TSS2_SYS_CMD_AUTHS cmdAuthArray = { 1, &cmdSessionArray[0] };
+
```
williamcroberts: So we really want to concatenate all the files and hash them all, ie:
```
cat a b c | sha256sum
```
or do we want sha256sum a b c behavior?

williamcroberts: I think we would want to default to, shells have a way to combine files, so we don't need it here. Considering that sha256sum gives us some kind of "standard" for expected behavior.

By changing this behavior, we can grab file size when creating the open file list, this was we can optimize TPM trips.

PR #428, opened 2017-08-27
CHANGELOG: Document that new --quiet option was added

martinezjavier:
A new option was added to supress the tools output, but a note was not
added to the CHANGELOG.


CHANGELOG
```diff
@@ -1,4 +1,5 @@
 next
+* Support for a --quiet option to suppress tools output. Many tools converted to use it.
```
williamcroberts: to suppress tool output to stdout...

martinezjavier: @williamcroberts right, I added more information to the changelog entry.

PR #431, opened 2017-08-29
Drop exec man

williamcroberts:


williamcroberts:
@martinezjavier found a really interesting bug with -I-.... the old option handling code ended up with a new_argv array one larger than passed in during tpm2_create in test_tpm2_unseal.sh.

I'm not 100% sure on how/why that worked, as it ended up duplicating the -O <optarg> optarg value...

martinezjavier:
@williamcroberts I see, I'll take a look to this tomorrow since I was busy today with other stuff.

lib/options.c
```diff
@@ -316,16 +291,12 @@ get_common_opts (int                    *argc_param,
...
-        switch (c) {
-        case 1: { /* positional arguments */
-            char **tmp = append_arg_to_vector (&new_argc, new_argv, optarg);
-            if (tmp == NULL) {
-                free(new_argv);
-                return 2;
-            }
-            new_argv = tmp;
+        if (new_argc >= argc) {
+            LOG_ERR("Error parsing common options");
```
williamcroberts: free(new_argv);

martinezjavier:
@williamcroberts the -I- issue you found is just a symptom of the common options logic changing the getopt semantics.

IOW, it should be valid to place the argument just after the command option since the optstring leads to no ambiguity about when an option ends and the argument starts. So for example this **should** work but doesn't:

```
$ tpm2_create -gsha1 -Gkeyedhash -oobj.pub -Oobj.priv -Isensitive.data -cprimary.context
```

test/system/test_tpm2_unseal.sh
```diff
@@ -106,7 +106,7 @@ if [ $? != 0 ];then
     exit 1
 fi
 
-tpm2_create -g $alg_create_obj -G $alg_create_key -o $file_unseal_key_pub -O $file_unseal_key_priv -I- -c $file_primary_key_ctx -L $file_policy -A $obj_attr <<< $secret
+tpm2_create -g $alg_create_obj -G $alg_create_key -o $file_unseal_key_pub -O $file_unseal_key_priv -I - -c $file_primary_key_ctx -L $file_policy -A $obj_attr <<< $secret
```
martinezjavier: As mentioned, I think that this change is not correct since is just papering over the real issue which is that the common options logic is changing the getopt semantics and requiring a space between the option and the argument which shouldn't be needed.

lib/options.c
```diff
@@ -178,31 +178,6 @@ get_common_opts_from_env (common_opts_t *common_opts)
     if (env_str != NULL)
```
martinezjavier: I comment here since it seems that Github doesn't allow to comment on the commit message:

s/reallocing/reallocating. Also I think it's more clear if you rephrase as:

"Rather than reallocting 1 new array item for the new argv array, just calloc up to argc count".

lib/options.c
```diff
@@ -316,16 +291,13 @@ get_common_opts (int                    *argc_param,
...
     {
-        switch (c) {
-        case 1: { /* positional arguments */
-            char **tmp = append_arg_to_vector (&new_argc, new_argv, optarg);
-            if (tmp == NULL) {
-                free(new_argv);
-                return 2;
-            }
-            new_argv = tmp;
+        if (new_argc >= argc) {
```
martinezjavier: if (new_argc > argc) I believe, since you could execute a command without any common options.

martinezjavier:
@williamcroberts so I had time to dig about this a little bit.

The problem is that when the user provided options are processed in get_common_opts(), the optstring only contains the common options.

So getopt_long() know nothing about the tool specific options and so it threats an option like "-I-" as two different (unknown) options instead of an -I option with - as an argument (as it will be processed by tpm2_tool due having "I:" in its opstring).

In the same way, passing a -gsha1 option (which is totally valid for tpm2_create due "g:" being in its opstring) it will processed as if 5 (unknown) options where passed.

So from the common options getopt_long() point of view, these two are equivalent:

``` 
$ tpm2_create -gsha1 

$ tpm2_create -g -s -h -a -1
```
I don't see an easy way to solve this. Do you have some ideas?

williamcroberts:
@martinezjavier I see two options:
1. Argpasre:
argparse supports child parsers which would fit this bill nicely. Granted C argparse is quite clunky, and kind of forces our hand to inline documentation for the help output. Which I guess would be nice for environments without manpages (maybe kill that exec on this approach completely). The downsides would be binary size and double maintenance on documentation.

We would have to add an entry point before execute tool that gets the child parser.

2. single getopt_long
We could add another entry point and get the options from tool, where they can return an options string and long options array and a callback, we could then have only one spot where getopt_long is called, and at that moment in time, getopt_long would have all the information available to it.

Either approach means that the tools will need a static global context for state, which isn't a problem per-say, considering it's a single threaded tool. However, if anything comes to light in tpm-tool, it shouldn't be assumed to be thread safe, we could lock in the lua calls or use the thread local attribute on the context. This is me just rambling, about possible potential future issues.

Ill be surprised if you read this far :-)

martinezjavier:
@williamcroberts I read it all :smiley: 

Yes, my thoughts were also similar to what you describe in (2).

lib/options.c
williamcroberts: Yes equality should should be allowed.

williamcroberts:
@martinezjavier, let's just do that. I think it's the simplest way to correct command line parsing.

martinezjavier:
@williamcroberts Ok, I'll attempt to fix this but probably next week.

I'll think that would also be good to have a test case for CLI options parsing trying different combinations.

williamcroberts:
@martinezjavier agreed.

PR #435, opened 2017-08-29
[WIP] tpm2_import: new tool that uses tpm2_import to import a key

williamcroberts:
TODO:
1. --parent-key-public should be optional, just read it in if not specified.
2. I don't understand the difference between the imported private, public and
   symetric. Is this limited to just importing an AES-128 key? If so why do I
   need a private and public key?
3. Should the mandatory inputs be options or arguments?

Limited to importing a symmetric AES-128 key.
^^^
Is this true?

Add tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,6 @@ addons:
   apt:
     packages:
     - autoconf-archive
-    - openssl
-    - libssl-dev
     - libcurl4-openssl-dev
     - libdbus-1-dev
     - libglib2.0-dev
@@ -65,6 +63,13 @@ install:
     - sha256sum libcmocka0_1.0.1-2_amd64.deb | grep -q 797155b45a8288a860c4ed9dd3f161420f09ebf362de30166d9f6b98bfc27dd0
     - sudo dpkg -i libcmocka0_1.0.1-2_amd64.deb
     - sudo dpkg -i libcmocka-dev_1.0.1-2_amd64.deb
+    # openssl 1.0.2g
+    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.0.0_1.0.2g-1ubuntu4.6_amd64.deb
+    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl-dev_1.0.2g-1ubuntu4.6_amd64.deb
+    - sha256sum libssl1.0.0_1.0.2g-1ubuntu4.6_amd64.deb | grep -q b858321c5858533399320fc2d849962c25288fe17f545dc4fb3ce99630cea5fe
+    - sha256sum libssl-dev_1.0.2g-1ubuntu4.6_amd64.deb | grep -q e4ad54c12aaf88d65c04eaccf1e07baf81ab3d14acf7e118137e8b1e10bdd06b
+    - sudo dpkg -i libssl1.0.0_1.0.2g-1ubuntu4.6_amd64.deb
+    - sudo dpkg -i libssl-dev_1.0.2g-1ubuntu4.6_amd64.deb
 
 script:
     - ./.ci/travis-build-and-run-tests.sh
--- a/test/system/test_tpm2_import.sh
+++ b/test/system/test_tpm2_import.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+#;**********************************************************************;
+#
+# Copyright (c) 2017, Intel Corporation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# 3. Neither the name of Intel Corporation nor the names of its contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+#;**********************************************************************;
+
+onerror() {
+    echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
+    exit 1
+}
+trap onerror ERR
+
+cleanup() {
+    tpm2_evictcontrol -A o -H 0x81010005 -S 0x81010005 2>/dev/null
+    rm -f import_key.ctx  import_key.name  import_key.priv  import_key.pub \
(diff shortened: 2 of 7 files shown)


williamcroberts:
@idesai can you help answer this?

I don't understand the difference between the imported private, public and
symetric. Is this limited to just importing an AES-128 key? If so why do I
need a private and public key?

What would we need to do to import private/pub key pairs if it doesn't support it now? Is this possible?

idesai:
@williamcroberts import_priv and import_pub are describing the sensitive and public information about the symmetric key. Another place you can see this is also tpm2_create tool when used to create a symmetric key produces private(encrypted sensitive) and public(key description). You can then use this in tpm2_load tool that needs both the encrypted sensitive and public data to produce a useable key handle.

idesai:
@williamcroberts a test is added part of the commit  test/system/test_tpm2_import.sh

williamcroberts:
@idesai Ahh when discussing private and public were talking about the tpm's notion of a public and private portions of a key object. This is easy to confuse with private and public portions of a key.

Ok I see now that the symmetric AES 128 key  is the one being imported, and the public/private are output that can be used in load later. I can make the manpage more clear about this.

So we would probably want the ability to load all supported symmetric keys? Can tpm2_import asymmetric keys?

test/system/test_tpm2_import.sh
```diff
@@ -0,0 +1,68 @@
...
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF 
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS 
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN 
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
+# THE POSSIBILITY OF SUCH DAMAGE.
+#;**********************************************************************;
+#!/bin/sh
+
```
williamcroberts: Update to use traps for handling file cleanup (to ensure that it is done). Also use ERR trap to print the line number so we don't have to explicitly check return codes.

test/system/test_tpm2_import.sh
```diff
@@ -0,0 +1,68 @@
...
+rm -f parent.ctx parent.pub  plain.dec.ssl  plain.enc  plain.txt  sym.key
+
+tpm2_createprimary -G 1 -g 0xb -A o -C parent.ctx 
+tpm2_evictcontrol -A o -c parent.ctx -S 0x81010005 
+
+dd if=/dev/urandom of=sym.key bs=1 count=16
+
+tpm2_readpublic -H 0x81010005 --opu parent.pub
+
+tpm2_import -k sym.key -H 0x81010005 -f parent.pub -q import_key.pub \
```
williamcroberts: make -f optional

williamcroberts: How would -q work with an asymmetric key?

test/system/test_tpm2_import.sh
```diff
@@ -0,0 +1,68 @@
...
+-r import_key.priv
+
+if [ $? != 0 ];then
+	    echo "tpm2_import Failed"
+		exit 1
+fi
+tpm2_load  -H 0x81010005 -u import_key.pub -r import_key.priv -n import_key.name \
+-C import_key.ctx
```
williamcroberts: I'm just noticing outputs unused, likely tpm2_load if not given a -n option can just not output the name, or just send it to stdout...

test/system/test_tpm2_import.sh
```diff
@@ -0,0 +1,68 @@
...
+openssl enc -in plain.enc -out plain.dec.ssl -d -K `xxd -p sym.key` -iv 0 \
+-aes-128-cfb
+
+diff plain.txt plain.dec.ssl
+if [ $? != 0 ];then
+	echo "tpm2_import test failed"
+	exit 1
+fi
+
+tpm2_evictcontrol -A o -H 0x81010005 -S 0x81010005
```
williamcroberts: I also see maybe a case to make tpm2_evictcontrol smart enough that if assume -H if -S is not provided, or maybe vice versa?

williamcroberts:
I don't understand why version 1.0.2h of openssl is missing this routine:
```
../tools/tpm2_import.c: In function ‘encrypt_seed_with_tpm2_rsa_public_key’:
../tools/tpm2_import.c:135:5: error: implicit declaration of function ‘RSA_padding_add_PKCS1_OAEP_mgf1’ [-Werror=implicit-function-declaration]
     int return_code = RSA_padding_add_PKCS1_OAEP_mgf1(encoded,
```

williamcroberts:
Darn, it's really version 1.0.1f according to the apt-get install output: 1.0.1f-1ubuntu2.22

Grabing the source for the package from: https://launchpad.net/ubuntu/+source/openssl/1.0.1f-1ubuntu2.22

untaring it and searching yields nothing:
```
grep -rn RSA_padding_add_PKCS1_OAEP_mgf1
```

idesai:
Dang! openssl hmm.... @williamcroberts when I wrote this app, the ssl version on my system was: OpenSSL 1.0.2g  1 Mar 2016.

test/system/test_tpm2_import.sh
idesai: @williamcroberts, agreed. If not given the -n option it should just print it to stdout.

test/system/test_tpm2_import.sh
idesai: @williamcroberts I believe the unique data should have the asymmetric public key.

idesai:
@williamcroberts yup that's the TPM's notion of sensitive and public data, yeah that can be confusing :)

test/system/test_tpm2_import.sh
idesai: @williamcroberts agreed. it is not very intuitive when trying to remove the persistent handle.

williamcroberts:
@idesai my system had the same openssl version. The configure script handles the version check, so I see no need to mention that in the manpage.

idesai:
@williamcroberts will this be merged?

williamcroberts:
@idesai yes, Ill try and do it soon.

williamcroberts:
oh thank god for restore branch... got a little overzealous with my house keeping.

williamcroberts:
@martinezjavier is the version bump of openssl going to affect you when you go to package? This change requires: 1.0.2g

martinezjavier:
@williamcroberts at least for Fedora it won't be an issue, since the last version of the compat-openssl10 package is 1.0.2j. I don't know for the other distros.

Did you check if these API change in OpenSSL 1.1.x? I fixed the issues we had on commit 5cee30cbc3da ("kdfa: allow to build with OpenSSL 1.1.x"), it would be nice if we make sure that no regressions are introduced (I didn't check BTW, just mentioning in case).

idesai:
@williamcroberts, cool this branch is back online, i was looking for it :) 
did I miss any questions? or is it ready to be merged?

idesai:
@williamcroberts made the changes per TSS commit: libmu: convert names to comply with TCG spec.

martinezjavier:
> Did you check if these API changed in OpenSSL 1.1.x? I fixed the issues we had on commit 5cee30c ("kdfa: allow to build with OpenSSL 1.1.x"), it would be nice if we make sure that no regressions are introduced (I didn't check BTW, just mentioning in case).

@williamcroberts so this branch added a build error with OpenSSL 1.1.0...

```
tools/tpm2_import.c:99:11: error: dereferencing pointer to incomplete type ‘RSA {aka struct rsa_st}’
     if ((r->n == NULL && n == NULL) || (r->e == NULL && e == NULL)) {
           ^~
make: *** [Makefile:1858: tools/tpm2_import.o] Error 1
make: *** Waiting for unfinished jobs....
```

martinezjavier:
I created PR #563 that fixes this issue among others I found while reading the code.

.travis.yml
```diff
@@ -65,6 +63,13 @@ install:
     - sha256sum libcmocka0_1.0.1-2_amd64.deb | grep -q 797155b45a8288a860c4ed9dd3f161420f09ebf362de30166d9f6b98bfc27dd0
     - sudo dpkg -i libcmocka0_1.0.1-2_amd64.deb
     - sudo dpkg -i libcmocka-dev_1.0.1-2_amd64.deb
+    # openssl 1.0.2g
+    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.0.0_1.0.2g-1ubuntu4.6_amd64.deb
+    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl-dev_1.0.2g-1ubuntu4.6_amd64.deb
+    - sha256sum libssl1.0.0_1.0.2g-1ubuntu4.6_amd64.deb | grep -q b858321c5858533399320fc2d849962c25288fe17f545dc4fb3ce99630cea5fe
+    - sha256sum libssl-dev_1.0.2g-1ubuntu4.6_amd64.deb | grep -q e4ad54c12aaf88d65c04eaccf1e07baf81ab3d14acf7e118137e8b1e10bdd06b
+    - sudo dpkg -i libssl1.0.0_1.0.2g-1ubuntu4.6_amd64.deb
+    - sudo dpkg -i libssl-dev_1.0.2g-1ubuntu4.6_amd64.deb
```
liuqun: It seemd to be a typo mistake that the commit message e63b52b said version 1.0.1g, which should be 1.0.2g

PR #437, opened 2017-08-30
add missing license header to files

williamcroberts:


liuqun:
- "// Copyright (c) 2017, Intel Corporation" for *.c *.h

but

- "# Copyright (c) 2016, Intel Corporation" for *.sh

test/system/test_algs.sh
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
martinezjavier: I usually prefer to use the absolute path to bash instead of using env (since with env someone can put a malicious bash binary in your $PATH). These are test scripts so I don't think that matters too much, but still I wonder if you have a strong opinion on this.

The script has a #!/bin/sh shebang after the license comment, which as you said makes no sense since isn't visible to the kernel's program loader. I see that your patch removes it in some scripts but miss in others.

williamcroberts: My script must have missed some.

PR #446, opened 2017-08-31
option fixing: move to single getopt

williamcroberts:
Parsing with multiple invocations of getopt
and strpping known options before invoking the
tool has resulted in some bugs. getopt is set
up to have all known options set up before the
call. A side affect of the old design was that
tool options could end up in a weird state.

For instance -I- for tpm2_create would not be parsed
correctly and an extra optarg would be added to
argv.

Redesign the option handling to use a single getopt
call, and develop new entry points for the tools.

New Entry Points:
1. tpm2_tool_onstart - The initial entry point to get
   the option arguments for getopt. This entry point
   also sets callbacks for:
    - option handling
    - argument handling

2. tpm2_tool_onrun - The tool executes the requested
   functionality.

3. tpm2_tool_exit - this is guarenteed to be called
   before exiting while the tcti context is still
   valid. This can be used for any cleanup done in a
   previous stage.

TODO:
 - clean ups
 - port other tools
 - error handling from main exits, tools need to
   clean up, do we have a tool_exit() call or
   just tell tools to set an at exit handler
   during 1.
   - maybe rename entry point 1 to tool_start?
     I personally like this flow:
     - tool_start
     - tool_execute
     - tool_exit


williamcroberts:
@martinezjavier let me know if your ok with this as the base, we can work on the tpm2-option-revamp branch

lib/tpm2_options.h
```diff
@@ -0,0 +1,176 @@
...
+ *  to handle arguments.
+ * @return
+ *  NULL on failure or an initialized tpm2_options object.
+ */
+tpm2_options *tpm2_options_new(const char *short_opts, size_t len,
+        const struct option *long_opts, tpm2_option_handler on_opt,
+        tpm2_arg_handler on_arg);
+
+/**
+ * Concatinates two tpm2_options objects, with src appended on
```
martinezjavier: s/Concatinates/Concatenates

tools/tpm2_tool.h
```diff
@@ -33,36 +33,54 @@
...
+/**
+ * prints output to stdout respecting the quiet option.
+ * Ie when quiet, don't print.
+ * @param fmt
+ *  The format specifier, ala printf.
+ * @param ...
+ *  The varargs, just like printf.
  */
-#define TOOL_OUTPUT(fmt, ...)                   \
+#define tpm2_tool_output(fmt, ...)                   \
```
martinezjavier: I would do this as a separate patch.

tools/tpm2_hash.c
```diff
@@ -57,6 +57,11 @@ struct tpm_hash_ctx {
     TSS2_SYS_CONTEXT *sapi_context;
 };
 
+static tpm_hash_ctx ctx = {
+    .hierarchyValue = TPM_RH_NULL,
+    .halg = TPM_ALG_SHA1,
+};
```
martinezjavier: You didn't answer my previous comment about passing a void * to the callbacks instead of using global variables.

lib/tpm2_options.c
```diff
@@ -0,0 +1,362 @@
...
+#ifndef VERSION
+  #warning "VERSION Not known at compile time, not embedding..."
+  #define VERSION "UNKNOWN"
+#endif
+
+#define TPM2TOOLS_ENV_TCTI_NAME      "TPM2TOOLS_TCTI_NAME"
+
+struct tpm2_options {
+    tpm2_option_handler on_opt;
+    tpm2_arg_handler on_arg;
```
martinezjavier: I suggested calling these on_callback or on_handler, I really think is much clear that on_opt.

tools/tpm2_tool.h
```diff
@@ -33,36 +33,54 @@
...
-static inline void enable_output(void)
-{
-    output_enabled = true;
-}
-
-static inline void disable_output(void)
-{
-    output_enabled = false;
-}
-
```
martinezjavier: Same for removing these. The patch will already be big to add unrelated changes.

lib/tpm2_options.c
williamcroberts: Sure, not a problem. I think I lost some of your comments in the noise.

tools/tpm2_hash.c
williamcroberts: We could definitely do that, but they would need to malloc there structure in the onstart() routine, wasn't sure if the complexity was worth it, given that everything is single threaded.

tools/tpm2_unseal.c
```diff
@@ -38,16 +38,16 @@
 #include <getopt.h>
 #include <sapi/tpm20.h>
 
+#include "../lib/tpm2_options.h"
```
martinezjavier: Do we really need these relative paths? $(top_srcdir)/lib/tcti is included in the list of directories to search for header files.

williamcroberts: no we don't, I have no idea how that hunk got added.

lib/tpm2_options.c
```diff
@@ -0,0 +1,362 @@
...
+
+    /* Get the options from the tool */
+    if (tool_opts) {
+        result = tpm2_options_cat(&opts, tool_opts);
+        if (!result) {
+            goto out;
+        }
+    }
+
+    /* Parse the options, calling the tool callback if unknown */
```
martinezjavier: You didn't answer my question about tools options being processed before -Q or -V common options.

tools/tpm2_tool.h
williamcroberts: I'm not worried about this one.

tools/tpm2_hash.c
martinezjavier: Right, let's just keep it as global vars for now to keep the patch simpler and we can later explore if the added complexity is worth it.

tools/tpm2_tool.h
martinezjavier: Ok, the patch will be huge anyways so I'm OK with sneaking these changes.

tools/tpm2_tool.h
williamcroberts: Normally I would be all about keeping patches as small as possible, with everything separated out. However, with this gigantic change, I am less worried about granularity.

If I just write our own option parser for main(), we can get away leaving everything else, I am very tempted to go in that direction. That would be the smallest, simplest change to fix our option issue.

lib/tpm2_options.c
martinezjavier: One solution I thought is to iterate over the options twice, in the first iteration only the common options are processed and the unknown options are not handled and in the second iteration is the opposite, common options are omitted and the unknown options are handled to the tools callbacks.

williamcroberts: Oh it's in the doxygen code comments in the new PR. LOG_INFO and tpm_tool_output() should not be considered valid until after on_opt is processed, on_arg and on is considered valid.

tools/tpm2_tool.h
martinezjavier: Do you mean not using getopt() ? I'm not sure about that, writing correct/safe C code is already hard and parsers are specially prone to subtle errors. I'm obviously not talking about you, just saying in general.

lib/tpm2_options.c
martinezjavier: Ah, sorry I missed that! That works for me.

tools/tpm2_hash.c
williamcroberts: If you really want to put the tool context structures on the heap, I can do it, I don't feel *that* strongly about it. I have the rest of your changes queued up locally, we can revisit tomorrow. Thanks.

martinezjavier: No strong opinion really, I trust your judgment since you did the change and have a better mental model of the code.

martinezjavier: @williamcroberts thanks a lot for addressing all the things I pointed out.

I'll work on porting the tools tomorrow based on the branch you left today.

williamcroberts: @martinezjavier ill upload the changes to your comments and merge into this branch, feel free to submit PRs for this to that branch.

lib/tpm2_options.c
williamcroberts: @martinezjavier with respect to iterating over them twice, getopt will still get confused on what things have optargs and what don't, we'd still have to create a sanitized argv for the tools. I tried duplicating argv and passing one to the tools and passing the other to main, but it had it's own issues.

getopt really needs to know all the options upfront AFAICT, even argparse has the same limitation, requiring children parsers present at the parse call.

martinezjavier: @williamcroberts yes, agreed. As mentioned I'm OK with having documented that the tpm2_tool_output() and LOG_* calls should not be considered valid by the tools until all the options have been processed.

They can just store any debug info in the callback and later print custom messages if they want in their tpm2_tool_onrun() handler.

tools/tpm2_hash.c
martinezjavier: @williamcroberts great, I'll work on top of that branch then and submit PR so you can squash them before pushing to master.

PR #447, opened 2017-09-01
configure: Check for md2man-roff.

flihp:
md2man-roff is required for the build to succeed. If it's missing
this commit will cause the configure script to fail with a helpful
error message.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,6 +6,11 @@ LT_INIT
 AM_INIT_AUTOMAKE([foreign
                   subdir-objects])
 AC_CONFIG_FILES([Makefile])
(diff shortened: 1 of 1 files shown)


flihp:
this resolves #447

martinezjavier:
@flihp isn't making configure to fail to harsh? I think is better to instead just don't attempt to generate the man pages for markdown (and so far only the tpm2_pcrevent tool uses it).

I'll propose an alternate patch for you to review.

williamcroberts:
I'm leaning towards #448

williamcroberts:
Going with #448, id rather attempt to move forward rather than fail completely.

PR #448, opened 2017-09-01
build: Don't generate man pages from markdown if md2man-roff isn't found

martinezjavier:
The md2man-roff tool is required to generate roff man pages from markdown
for some tools, but this may not be available on some systems. Instead of
making the build to fail, just don't attempt to generate these man pages.

Fixes: #447

Reported-by: Philip Tricca <philip.b.tricca@intel.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -261,8 +261,12 @@ man8_MANS = \
     man/man8/tpm2_rc_decode.8 \
     man/man8/tpm2_dictionarylockout.8 \
     man/man8/tpm2_createpolicy.8 \
-    man/man8/tpm2_pcrextend.8 \
+    man/man8/tpm2_pcrextend.8
+
+if HAVE_MD2MAN_ROFF
+    man8_MANS += \
     man/man8/tpm2_pcrevent.8
+endif
 
 MAN_DEPS := man/common-options.troff man/tcti-options.troff \
            man/tcti-environment.troff man/alg-common.troff \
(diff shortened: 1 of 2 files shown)


martinezjavier:
@flihp please let me know if this approach is suitable for you. I think that we should at least allow people to build the tools and the man pages that are already using troff, even when md2man-roff isn't available.

martinezjavier:
@webmeister yeah, it's the same situation in Fedora. I may have to package md2man in order to update the tpm2-tools package for the next upcoming release.

williamcroberts:
@webmeister  and @martinezjavier, You only need md2man if your building the source and packaging manpages, so you wouldn't be a dependency of binary packages. However, if your packaging source the host would need it.

Makefile.am
```diff
@@ -307,10 +311,12 @@ MARKDOWN_COMMON_DEPS = \
 man/man8/%.8 : man/%.8.md $(MARKDOWN_COMMON_DEPS)
 	rm -f $@
 	mkdir -p man/man8
+if HAVE_MD2MAN_ROFF
```
williamcroberts: I don't think you need this block, just remove the man page from man8_MANS should be sufficient.

martinezjavier: Yes, I believe you are right.

martinezjavier:
@williamcroberts sure, you don't need it as a Requires dependency but you need it as a BuildRequires dependency (speaking in SPEC / RPM terms).

Makefile.am
martinezjavier: @williamcroberts updated the branch removing the block. Thanks a lot for catching this.

PR #451, opened 2017-09-01
[WIP] port some tools to the new options parsing infra

martinezjavier:
work-in-progress pull request to port tools to the new options infrastructure. The tests in test/system are used to check that no regressions are introduced after the port.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -49,10 +49,10 @@ LDADD = \
 bin_PROGRAMS = \
     tools/tpm2_activatecredential \
     tools/tpm2_akparse \
-    tools/tpm2_hash
+    tools/tpm2_hash \
+    tools/tpm2_certify
 
 # NEED TO PORT
-#    tools/tpm2_certify \
 #    tools/tpm2_create \
 #    tools/tpm2_createpolicy \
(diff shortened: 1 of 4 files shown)


martinezjavier:
@williamcroberts sorry, I was quite busy today at the end with other stuff so I only had time to port one tool. But I can continue working on this next week.

BTW, I also tried to port tpm2_certify and wondered if you thought already what to do with the tools that have a lot of flags that are used after the getopt() call.

I guess we should put those flags in the tool context struct and process them on the tpm2_tool_onrun() handler or do you have a better idea?

martinezjavier:
@williamcroberts I'm doing a push -f each time that I find some time to port a new tool. Is this a reasonable process for you? I don't think I'll have more time to work on this today though, so feel free to port others tools on top and I can continue based on your branch tomorrow.

williamcroberts:
@martinezjavier just commit separate patches and ill squash down at the end. Also, so we don't conflict, if you start at the head of the tool list ill start at the tail and we can meet in the middle.

martinezjavier:
@williamcroberts perfect, that's what I've been doing. We are on the same page then.

PR #465, opened 2017-09-06
Another set of tools ported to the new option handling

martinezjavier:
This is yet another set of tools ported to the new option handling code. While porting this I noticed that my previous ports had some issues:

* I wrongly removed the LOG_ERR() usage in the .on_option handlers because I misunderstood that were not safe to use at some point (but only LOG_INFO and tpm2_tool_output shouldn't be use there).
* I forgot to remove the <getopt.h> header inclusion that's not needed anymore since tools now include "tpm2_options.h".
* Didn't check if tools really needed the sapi_context to be in the context struct. In most cases just the tpm2_tool_onrun() function should pass it to the functions that interact with the TPM.

I'll fix those and propose a patch on top of tpm2-option-revamp.

BTW, I notice a couple of bugs in tpm2_{option,tools}.c so I provide a fix up as a part of this PR.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,6 +65,7 @@ bin_PROGRAMS = \
     tools/tpm2_hmac \
     tools/tpm2_listpersistent \
     tools/tpm2_load \
+    tools/tpm2_loadexternal \
     tools/tpm2_send_command \
     tools/tpm2_sign \
     tools/tpm2_startup \
@@ -73,7 +74,6 @@ bin_PROGRAMS = \
     tools/tpm2_verifysignature
 
 # NEED TO PORT
-#    tools/tpm2_loadexternal \
 #    tools/tpm2_makecredential \
 #    tools/tpm2_nvdefine \
 #    tools/tpm2_nvlist \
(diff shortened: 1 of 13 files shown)


williamcroberts:
I like that the diffstat is more delete than add... I think it shows it heading in the right direction.

williamcroberts:
@martinezjavier I am doing rsadecrypt right now, were getting close and I don't want to clash with you, let me know what the last one you want to send up is, and ill work to that.

martinezjavier:
@williamcroberts I don't think I'll have more time to work on this today, so go ahead and I'll continue tomorrow if you don't finish today.

I'll also send you tomorrow the fixes for the previous tools that I mentioned.

williamcroberts:
Sounds good, I should be able to finish this out today

williamcroberts:
@martinezjavier pushing the last changes now, the only thing failing is:
```
Loaded key handle:  800000ff
ERROR: Expected options (H or c) and (k or C) and f and o
activatecredential test fail, please check the environment or parameters!
 test_tpm2_activecredential.sh Fail 
```

martinezjavier:
@williamcroberts Ok, great.

williamcroberts:
FYI I rebased and squashed that branch

PR #470, opened 2017-09-07
Fixups for tools ported to the new option handling

martinezjavier:
This pull request contains fixes for issues I found when double checking the ported tools to the new option handling code. After this I think that it should be ready to merge (unless I missed some issues).


marckleinebudde:
After merging https://github.com/martinezjavier/tpm2-tools/tree/tpm2-option-revamp-fixups it still fails the same way. See:
https://github.com/martinezjavier/tpm2-tools/commit/db1ab22c0fa85df6cef1838268889dc709019d32#commitcomment-24145112

martinezjavier:
@marckleinebudde yes, I didn't mean that it will be fixed by this PR. What I said is that you could wait for this PR to be merged and fix on top.

Having said that, probably makes sense to merge your fix since is a minimal conflict for @williamcroberts to fix on rebase before pushing this.

williamcroberts:
I just had to cherry-pick all the changes over since I rebased it, but they should all be on tpm2-option-revamp.

williamcroberts:
@martinezjavier These were the changes I was looking for. I won't rebase the tpm2-option-revamp branch until the final push to master Monday morning, so send anything my way you're dying to get in the merge, if not, don't worry about it.

martinezjavier:
@williamcroberts sounds good. As mentioned, I don't think I'll have anything else after these.

Hopefully we wouldn't introduce any regression on this rework.

williamcroberts:
@martinezjavier now that we know that session support is there, we need to get that worked into the next release and drop the pcr policy support added to tools like tpm2_unseal.

I need to spend some time in the ticket tracker next week and re-organize "next"

martinezjavier:
@williamcroberts does the support exist already in the resource managers? My understanding was that @flihp was going to add it to the tpm2-abrmd in the following weeks.

Also, it's not clear to me what's the current situation with the in-kenrel resource manager, Jarkko only pointed out to the patch-set that added the support in v4.12, but he didn't answer to the specific question about the Context{Save,Load} behavior in the kernel.

I have in my pending list to study the [TSS TAB and Resource Manager
Specification](https://trustedcomputinggroup.org/wp-content/uploads/TSS-TAB-and-Resource-Manager-00-91-PublicReview.pdf) document, which I believe is the one that @flihp mentioned and also the TPM resource manager implementation in the kernel.

It would be good to cut at some point a tpm2-tss and tpm2-tools release that has this support, even if in latter version we can improve due new features in the resource managers.

PR #471, opened 2017-09-07
tpm2_getmanufec: RetrieveEndorsementCredentials(): fix curl_easy_seto…

marckleinebudde:
…pt() call

curl_easy_setopt() expects the 3rd parameter to be a long, but it's a bool here and fails with:

| tools/tpm2_getmanufec.c: In function 'RetrieveEndorsementCredentials':                                                                                                                                         
| tools/tpm2_getmanufec.c:451:944: error: call to '_curl_easy_setopt_err_long' declared with attribute warning: curl_easy_setopt expects a long | argument for this option [-Werror]
| cc1: all warnings being treated as errors                                                                                                                                                                                                                                                 

This patch fixes the problem by casting the parameter to long.

### DIFF
--- a/tools/tpm2_getmanufec.c
+++ b/tools/tpm2_getmanufec.c
@@ -448,7 +448,7 @@ int RetrieveEndorsementCredentials(char *b64h)
      * If verbose is set, add in diagnostic information for debugging connections.
      * https://curl.haxx.se/libcurl/c/CURLOPT_VERBOSE.html
      */
-    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, verbose);
+    rc = curl_easy_setopt(curl, CURLOPT_VERBOSE, (long)verbose);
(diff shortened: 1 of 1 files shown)


martinezjavier:
@marckleinebudde thanks for your patch, it looks correct to me. But we are about to land a very big rework of the tools and this fix will conflict with this. Please take a look to PR #470. Could you please do it on top of that when it's merged? That should happen on Monday AFAIK.

BTW, you don't need to close a PR, delete and create your branch and open it again to change your commits. Just do git push -f and Github will refresh the changes in the opened PR.

marckleinebudde:
Still fails for me after merging - See pull request: https://github.com/martinezjavier/tpm2-tools/pull/1

williamcroberts:
This is on https://github.com/01org/tpm2-tools/tree/tpm2-option-revamp

PR #473, opened 2017-09-08
Tpm2 option revamp

williamcroberts:


williamcroberts:
@martinezjavier Here it is, the PR of doom. On the plus side, we removed 800 lines of code, I love that.

martinezjavier:
@williamcroberts great, negative diffstats are the best :smiley:

lib/tpm2_options.c
```diff
@@ -0,0 +1,372 @@
...
+
+    tpm2_option_code rc = tpm2_option_code_err;
+    bool result = false;
+
+    UNUSED(envp);
+
+    struct option long_options [] = {
+        { "tcti",    required_argument, NULL, 'T' },
+        { "help",    no_argument,       NULL, 'h' },
+        { "verbose", no_argument,       NULL, 'v' },
```
liuqun: Bug #967 Common option "-v" should print version, not verbose

lib/tpm2_options.c
```diff
@@ -0,0 +1,372 @@
...
+        { "quiet",   no_argument,       NULL, 'Q' },
+        { "version", no_argument,       NULL, 'V' },
```
liuqun: "version" should use lower case short option "-v"

PR #474, opened 2017-09-08
man: use links in manpages for common options

williamcroberts:
Rather than using the @@ decorators to mark where options
are, use relative links.

The @@ decarotors are not useable in the markdown rendering on
github, but the relative links, are. Thus, the non-built man
pages will render in a useable way on github.


williamcroberts:
@flihp this seems to work for me, and now the markdown rendered manpages on github work.

williamcroberts:
Holding off to submit a large batch of markdown converted man pages....

Issue #462, opened 2017-09-06
Add a test case for tpm2_nvreadlock tool


martinezjavier:
There isn't a test for tpm2_nvreadlock and the tools isn't used in other tests either.

martinezjavier:
The test should be something like this:

```
    $ tpm2_nvdefine -x 0x1500018 -a 0x40000001 -s 32 -t "ownerread|policywrite|ownerwrite|read_stclear"
    
    $ tpm2_nvwrite -x 0x1500018 -a 0x40000001 -f data.txt
    66 6f 6f 62 61 72 0a
    
    $ tpm2_nvread -x 0x1500018 -a 0x40000001 -s 14 -o 0
    000000: e8 0d cd a9 c0 7f 00 00 e8 0d cd a9 c0 7f 00 00  ................
    000010: 30 e0 50 f4 e7 55 00 00 05 00 00 00 00 00 00 00  0.P..U..........
    
    $ tpm2_nvreadlock -x 0x1500018 -a 0x40000001
    
    $ tpm2_nvread -x 0x1500018 -a 0x40000001 -s 14 -o 0
    ERROR: Failed to read NVRAM area at index 0x1500018 (22020120). Error:0x148
    
    $ tpm2_rc_decode 0x148
    error layer
      hex: 0x0
      identifier: TSS2_TPM_ERROR_LEVEL
      description: Error produced by the TPM
    format 0 error code
      hex: 0x48
      name: TPM_RC_NV_LOCKED
      description: NV access locked.
```

PR #475, opened 2017-09-10
tpm2_nvread: fix memcpy() arguments

cardoe:
The first argument is the destination which needs to be offset based on
how much has already been read. The third argument is the size that we
should be reading. Correct the spelling of the variable name.

Signed-off-by: Doug Goldstein <cardoe@cardoe.com>

### DIFF
--- a/tools/tpm2_nvread.c
+++ b/tools/tpm2_nvread.c
@@ -141,7 +141,7 @@ static bool nv_read(TSS2_SYS_CONTEXT *sapi_context) {
     }
 
     bool result = false;
-    UINT16 data_offest = 0;
+    UINT16 data_offset = 0;
     while (ctx.size_to_read) {
 
         UINT16 bytes_to_read = ctx.size_to_read > MAX_NV_BUFFER_SIZE ?
@@ -158,8 +158,8 @@ static bool nv_read(TSS2_SYS_CONTEXT *sapi_context) {
         ctx.size_to_read -= nv_data.t.size;
         ctx.offset += nv_data.t.size;
 
-        memcpy(data_buffer, nv_data.t.buffer, data_offest);
-        data_offest += nv_data.t.size;
+        memcpy(data_buffer + data_offset, nv_data.t.buffer, nv_data.t.size);
+        data_offset += nv_data.t.size;
     }
 
     hexdump(data_buffer, data_size);


cardoe:
You can see the bug in action in #462.

tools/tpm2_nvread.c
```diff
@@ -158,8 +158,8 @@ static bool nv_read(TSS2_SYS_CONTEXT *sapi_context) {
         ctx.size_to_read -= nv_data.t.size;
         ctx.offset += nv_data.t.size;
 
-        memcpy(data_buffer, nv_data.t.buffer, data_offest);
-        data_offest += nv_data.t.size;
+        memcpy(data_buffer + data_offset, nv_data.t.buffer, nv_data.t.size);
+        data_offset += nv_data.t.size;
     }
```
martinezjavier: @cardoe I also noticed this issue but didn't dig further. Thanks for the fix, it looks good to me.

tools/tpm2_nvread.c
```diff
@@ -158,8 +158,8 @@ static bool nv_read(TSS2_SYS_CONTEXT *sapi_context) {
...
 
     hexdump(data_buffer, data_size);
```
martinezjavier: I think there's a bug here too, shouldn't this be?

```
hexdump(data_buffer, data_offset);
```

Also, do you know why there's a data_offset and ctx.offset? It seems the state is duplicated or am I missing something?

williamcroberts: data_offset == data_size at the end of the loop.

martinezjavier: @williamcroberts well, that's not true on my test:

```
$ echo -n foobar > data.txt

$ hexdump -C -v data.txt 
00000000  66 6f 6f 62 61 72                                 |foobar|
00000006

$ tpm2_nvwrite -x 0x1500018 -a 0x40000001 -f data.txt
66 6f 6f 62 61 72 

$ tpm2_nvread -x 0x1500018 -a 0x40000001 -s 6 -o 0
data_size 32 ctx.size_to_read 0 data_offset 6
000000: 66 6f 6f 62 61 72 00 00 c8 ad 07 91 78 7f 00 00  foobar......x...
000010: 10 60 b2 e9 92 55 00 00 05 00 00 00 00 00 00 00  .`...U..........
```

When hexdump is called, data_size = 32 and data_offset = 6. That's why 32 bytes are printed instead of just 6.

The problem is that nv_public.t.nvPublic.dataSize > nv_data.t.size so that's why I suggested that change, that way you print the actual number of bytes read. After using data_offset instead of data_size:

```
$ tpm2_nvread -x 0x1500018 -a 0x40000001 -s 6 -o 0                    
000000: 66 6f 6f 62 61 72                                foobar
```

martinezjavier:
@williamcroberts I created PR #476, feel free to comment there.

PR #478, opened 2017-09-11
test: add a tpm2_readlock test case to tpm2_nv.sh

martinezjavier:
There's no test for tpm2_nvreadlock, add a test case for it to the NV
test. Also print a success message if test finish successfully like's
done in the other tests.

Fixes: #462


martinezjavier:
@williamcroberts indeed, I just removed the file at the end to make it consistent with the rest of the script.

I'll fix it as a separate patch.

martinezjavier:
Created PR #482. I think we should create an issue to make sure that all the test scripts do this.

williamcroberts:
SGTM

PR #481, opened 2017-09-11
Man page to md

williamcroberts:
Convert a few  more tools to use the markdown manpages.


man/tpm2_create.8.md
```diff
@@ -0,0 +1,105 @@
+tpm2_create 8 "AUGUST 2017" Linux "User Manuals"
+==================================================
```
martinezjavier: In general the change looks good to me. I just have one question about the header.

The man man page, list section 8 as:

"8 System administration commands (usually only for root)".

But since the tpm2-tools are now installed in bindir instead of sbindir, I wonder if we shouldn't use section 1 instead:

"1   Executable programs or shell commands"

That will also be in line with the "User Manuals" comment, otherwise it should be "System manager's manual" or something like that.

Last question is the "Linux" (which was wrongly Intel before). Many commands instead mention there the package name (i.e: tpm2-tools). I think that's more informative than just "Linux" (specially since AFAIU tpm2-tools is portable to other systems).

So in summary I would change it to:

```
tpm2_create 1 "SEPTEMBER 2017" tpm2-tools "User Manuals"
```

williamcroberts: I'm ok with that, I was questioning section 8 myself. I'll make these changes today and tack on another 6 or so man pages.

williamcroberts:
We are blocked on TSS issue: https://github.com/01org/tpm2-tss/issues/531

PR #482, opened 2017-09-11
test: make sure temp files are cleaned up on test_tpm2_nv.sh exit

martinezjavier:
If the test is interrupted, then temporary files may not be cleaned up
correctly. Ensure that the files are always removed on the script exit.


martinezjavier:
@williamcroberts Ok, I've added as a separate commit since I believe that are separate cleanups.

williamcroberts:
Another negative diffstat, I like it.

PR #485, opened 2017-09-13
OpenSSL output formats and interoperability draft

mgerstner:
Follow-up on PR #427. I don't expect this to be accepted yet, I just would like to know if this matches your expectations, before I start touching a dozen of other tools and writing documentation for it.

### DIFF
--- a/lib/conversion.c
+++ b/lib/conversion.c
@@ -1,0 +1,201 @@
+//**********************************************************************;
+// Copyright (c) 2017, SUSE Linux GmbH
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// 3. Neither the name of Intel Corporation nor the names of its contributors
+// may be used to endorse or promote products derived from this software without
+// specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+
+#include <string.h>
+#include <errno.h>
+
+#include <openssl/rsa.h>
+#include <openssl/pem.h>
+#include <openssl/bn.h>
+#include <openssl/err.h>
+
+#include "conversion.h"
+#include "files.h"
+#include "log.h"
+#include "tpm2_alg_util.h"
+#include "tpm2_util.h"
+
+static bool tpm2_convert_pubkey_ssl(TPMT_PUBLIC *public, pubkey_format format, const char *path);
+
+pubkey_format tpm2_parse_pubkey_format(const char *label) {
+    if (strcasecmp(label, "der") == 0) {
+        return pubkey_format_der;
+    }
+    else if (strcasecmp(label, "pem") == 0) {
+        return pubkey_format_pem;
+    }
+    else if (strcasecmp(label, "tss") == 0) {
+        return pubkey_format_tss;
+    }
+
+    LOG_ERR("Invalid public key output format '%s' specified", label);
+
+    return pubkey_format_err;
+}
+
+signature_format tpm2_parse_signature_format(const char *label) {
+    if (strcasecmp(label, "tss") == 0) {
+        return signature_format_tss;
+    }
+    else if (strcasecmp(label, "plain") == 0) {
+        return signature_format_plain;
+    }
+
+    LOG_ERR("Invalid signature output format '%s' specified", label);
+
+    return signature_format_err;
+}
+
+static void print_ssl_error(const char *failed_action) {
+    char errstr[256] = {0};
+    unsigned long errnum = ERR_get_error();
+
+    ERR_error_string_n(errnum, errstr, sizeof(errstr));
+    LOG_ERR("%s: %s", failed_action, errstr);
+}
+
+bool tpm2_convert_pubkey(TPM2B_PUBLIC *public, pubkey_format format, const char *path) {
+
+    if (format == pubkey_format_der || format == pubkey_format_pem) {
+        return tpm2_convert_pubkey_ssl(&public->t.publicArea, format, path);
+    }
+    else if (format == pubkey_format_tss) {
+        // save raw as is
+        return files_save_bytes_to_file(path, (UINT8 *)public, sizeof(*public));
+    }
+
+    LOG_ERR("Unsupported public key output format.");
+    return false;
+}
+
+static bool tpm2_convert_pubkey_ssl(TPMT_PUBLIC *public, pubkey_format format, const char *path) {
+    bool ret = false;
+    FILE *fp = NULL;
+    RSA *ssl_rsa_key = NULL;
+
+    // need this before the first SSL call for getting human readable error
+    // strings in print_ssl_error()
+    ERR_load_crypto_strings();
+
+    if (public->type != TPM_ALG_RSA) {
+        LOG_ERR("Unsupported key type for requested output format. Only RSA is supported.");
+        goto error;
+    }
+
+    UINT32 exponent = (public->parameters).rsaDetail.exponent;
+    if (exponent == 0) {
+        exponent = RSA_DEFAULT_PUBLIC_EXPONENT;
+    }
+
+    // OpenSSL expects this in network byte order
+    exponent = tpm2_util_hton_32(exponent);
+    ssl_rsa_key = RSA_new();
+    if (!ssl_rsa_key) {
+        print_ssl_error("Failed to allocate OpenSSL RSA structure");
+        goto error;
+    }
+
+    ssl_rsa_key->e = BN_bin2bn((void*)&exponent, sizeof(exponent), NULL);
+    ssl_rsa_key->n = BN_bin2bn(
+            (public->unique).rsa.t.buffer,
+            (public->unique).rsa.t.size,
+            NULL);
+
+    if (!ssl_rsa_key->n || !ssl_rsa_key->e) {
+        print_ssl_error("Failed to convert data to SSL internal format");
+        goto error;
+    }
+
+    fp = fopen(path, "wb");
+    if (!fp) {
+        LOG_ERR("Failed to open public key output file '%s': %s",
+            path, strerror(errno));
+        goto error;
+    }
+
+    int ssl_res = 0;
+
+    switch(format) {
+    case pubkey_format_pem:
+        ssl_res = PEM_write_RSA_PUBKEY(fp, ssl_rsa_key);
+        break;
+    case pubkey_format_der:
+        ssl_res = i2d_RSA_PUBKEY_fp(fp, ssl_rsa_key);
+        break;
+    default:
+        LOG_ERR("Invalid OpenSSL target format %d encountered", format);
+        goto error;
+    }
+
+    if (ssl_res <= 0) {
+        print_ssl_error("OpenSSL public key conversion failed");
+        goto error;
+    }
+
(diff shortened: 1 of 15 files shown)


lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+#include <openssl/rsa.h>
+#include <openssl/pem.h>
+#include <openssl/bn.h>
+#include <openssl/err.h>
+
+#include "log.h"
+#include "files.h"
+#include "tpm2_util.h"
+#include "tpm2_alg_util.h"
+#include "conversion.h"
```
williamcroberts: sort each include block alphabetically, but leave the blocks how you have them.

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+    // strings in print_ssl_error()
+    ERR_load_crypto_strings();
+
+    if (public->type != TPM_ALG_RSA) {
+        LOG_ERR("Unsupported key type for requested output format. Only RSA is supported.");
+        goto error;
+    }
+
+    UINT32 exponent = (public->parameters).rsaDetail.exponent;
+
```
williamcroberts: no newline here

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+    if (exponent == 0) {
+        exponent = RSA_DEFAULT_PUBLIC_EXPONENT;
+    }
+
+    // OpenSSL expects this in network byte order
+    exponent = tpm2_util_hton_32(exponent);
+
+    ssl_rsa_key = RSA_new();
+
```
williamcroberts: no newline here

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+            (public->unique).rsa.t.size,
+            NULL);
+
+    if (!ssl_rsa_key->n || !ssl_rsa_key->e) {
+        print_ssl_error("Failed to convert data to SSL internal format");
+        goto error;
+    }
+
+    fp = fopen(path, "wb");
+
```
williamcroberts: no newline here and throughout where you check for a return value

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+    int ssl_res = 0;
+
+    switch(format) {
+    case pubkey_format_pem:
+        ssl_res = PEM_write_RSA_PUBKEY(fp, ssl_rsa_key);
+        break;
+    case pubkey_format_der:
+        ssl_res = i2d_RSA_PUBKEY_fp(fp, ssl_rsa_key);
+        break;
+    default:
```
williamcroberts: Is this possible? Shouldn't this be a hard error as an unsupported format type?

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+}
+
+bool tpm2_convert_signature(TPMT_SIGNATURE *signature, signature_format format, const char *path) {
+
+    if (format == signature_format_tss) {
+        /* TODO fix serialization */
+        return files_save_bytes_to_file(path, (UINT8 *)signature,
+            sizeof(*signature));
+    }
+    else if (format == signature_format_plain) {
```
williamcroberts: You have switches in some places for this and an if/else here, I would be consistent. I like the switch myself if we need to add another output format.

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        bool ret = files_save_bytes_to_file(path, buffer, size);
+        free(buffer);
+
+        return ret;
+    }
+
+    LOG_ERR("Unsupported signature output format.");
+    return false;
```
williamcroberts: Yeah this is a hard error here, but not on line https://github.com/01org/tpm2-tools/pull/485/files#diff-98bdc2e9547136dd6f16812a7289b524R157

lib/conversion.h
```diff
@@ -0,0 +1,81 @@
...
+    pubkey_format_tss, pubkey_format_pem, pubkey_format_der, pubkey_format_err
+};
+
+typedef enum signature_format signature_format;
+
+enum signature_format {
+    signature_format_tss, signature_format_plain, signature_format_err
+};
+
+/*
```
williamcroberts: doxygen comments here and throughout the header

tools/tpm2_quote.c
```diff
@@ -205,6 +210,80 @@ static UINT16  calcSizeofTPMT_SIGNATURE( TPMT_SIGNATURE *sig )
...
+    size_t num_bytes;
+    FILE *fp = fopen(path,"wb");
+
+    if (!fp) {
+        printf("Output file %s cannot be created: %s\n",
+                path, strerror(errno));
+        goto error;
+    }
+
+    num_bytes = calcSizeofTPM2B_ATTEST(quoted);
```
williamcroberts: Where did this macro come from sapi headers?

Couldn't one use:
```
BUFFER_SIZE(typeof(quoted), field)
```

tools/tpm2_quote.c
```diff
@@ -205,6 +210,80 @@ static UINT16  calcSizeofTPMT_SIGNATURE( TPMT_SIGNATURE *sig )
...
 }
 
+static bool write_tss_file(TPM2B_ATTEST *quoted, TPMT_SIGNATURE *signature,
+        const char *path) {
+    bool res = false;
+    size_t num_bytes;
+    FILE *fp = fopen(path,"wb");
+
+    if (!fp) {
+        printf("Output file %s cannot be created: %s\n",
```
williamcroberts: LOG_ERR

williamcroberts: And throughout

tools/tpm2_quote.c
```diff
@@ -205,6 +210,80 @@ static UINT16  calcSizeofTPMT_SIGNATURE( TPMT_SIGNATURE *sig )
...
+    bool res = false;
+    FILE *fp = fopen(path, "wb");
+
+    if (!fp) {
+        printf("Message output file %s cannot be created: %s\n",
+                path, strerror(errno));
+        goto error;
+    }
+
+    res = files_write_bytes(fp, (UINT8*)(quoted->b).buffer, (quoted->b).size);
```
williamcroberts: Cant you replace this whole routine with ```files_save_bytes_to_file```

tools/tpm2_readpublic.c
```diff
@@ -92,8 +95,16 @@ static int read_public_and_save(TSS2_SYS_CONTEXT *sapi_context) {
     }
     tpm2_tool_output("\n");
 
-    return files_save_bytes_to_file(ctx.outFilePath, (UINT8 *) &public,
-            sizeof(public));
+    pubkey_format format = pubkey_format_tss;
```
williamcroberts: set this in the static global, then you can just update it in the option handler, no need for checking if the f flag is set.

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+    }
+    ERR_free_strings();
+
+    return ret;
+}
+
+bool tpm2_convert_signature(TPMT_SIGNATURE *signature, signature_format format, const char *path) {
+
+    if (format == signature_format_tss) {
+        /* TODO fix serialization */
```
williamcroberts: We actually have been discussing serialization issues on: https://github.com/01org/tpm2-tss/pull/533

The issue here is that this dumps all padding, so its output not portable across architectures.

lib/tpm2_alg_util.c
```diff
@@ -330,3 +330,57 @@ bool pcr_parse_digest_list(char **argv, int len,
...
 }
+
+UINT8* tpm2_extract_plain_signature(UINT16 *size, TPMT_SIGNATURE *signature) {
+
+    UINT8 *buffer = NULL;
+    *size = 0;
+
+    switch (signature->sigAlg) {
+    case TPM_ALG_RSASSA:
+        *size = sizeof((*signature).signature.rsassa.sig.t.buffer);
```
martinezjavier: I would argue that using  the arrow operator is a much common idiom than doing a pointer dereference here, i.e:

```
*size = sizeof(signature->signature.rsassa.sig.t.buffer);
```
This comment is also for all the places in the code that do the same.

lib/tpm2_alg_util.c
```diff
@@ -330,3 +330,57 @@ bool pcr_parse_digest_list(char **argv, int len,
...
+        memcpy(buffer, (*signature).signature.rsassa.sig.t.buffer, *size);
+        break;
+    case TPM_ALG_HMAC: {
+        TPMU_HA *hmac_sig = &((*signature).signature.hmac.digest);
+        *size = tpm2_alg_util_get_hash_size((*signature).signature.hmac.hashAlg);
+        buffer = malloc(*size);
+        if(! buffer) {
+            goto nomem;
+        }
+        memcpy(buffer, &(hmac_sig->na), *size);
```
martinezjavier: No need for the parenthesis here, the arrow operator has a higher precedence than the address-of operator.

lib/tpm2_alg_util.c
```diff
@@ -330,3 +330,57 @@ bool pcr_parse_digest_list(char **argv, int len,
...
+UINT8* tpm2_extract_plain_signature(UINT16 *size, TPMT_SIGNATURE *signature) {
+
+    UINT8 *buffer = NULL;
+    *size = 0;
+
+    switch (signature->sigAlg) {
+    case TPM_ALG_RSASSA:
+        *size = sizeof((*signature).signature.rsassa.sig.t.buffer);
+        buffer = malloc(*size);
+        if(! buffer) {
```
martinezjavier: Please remove this space.

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
+//**********************************************************************;
+// Copyright (c) 2017, Intel Corporation
```
martinezjavier: Shouldn't this be SUSE Linux GmbH instead?

lib/conversion.c
```diff
@@ -0,0 +1,204 @@
...
+    unsigned long errnum = ERR_get_error();
+
+    ERR_error_string_n(errnum, errstr, sizeof(errstr));
+    LOG_ERR("%s: %s", failed_action, errstr);
+}
+
+bool tpm2_convert_pubkey(TPM2B_PUBLIC *public, pubkey_format format, const char *path) {
+
+    if (format == pubkey_format_der || format == pubkey_format_pem) {
+        return tpm2_convert_pubkey_ssl(&((public->t).publicArea), format, path);
```
martinezjavier: Again, you can just write this as ```tpm2_convert_pubkey_ssl(&public->t.publicArea), format, path)``` since the ```->``` and ```.``` operators have more precedence than ```&```.

tools/tpm2_quote.c
mgerstner: I really don't want to touch this, because the `calcSizeofTPM2B_ATTEST()` uses the size field of the buffer while `BUFFER_SIZE()` uses `sizeof` the structure.

In my tests the data returned by Tss2_Sys_Quote() is much less than the size of the TPM2B_ATTEST structure for some reason. Either I don't understand some things here, or this might be worth opening an issue. However, changing to `BUFFER_SIZE()` would cause hundreds of zero bytes to be output into the quote file - at least on my machine.

The TPM chip actually only returns 121 bytes of data, while the structure is 1280 bytes large.

mgerstner:
I've adjusted according to your comments, what do you think now?

mgerstner:
I will rebase the changes for the final PR

tools/tpm2_quote.c
williamcroberts: The statement, "I really don't want to touch this, because the calcSizeofTPM2B_ATTEST() uses the size field of the buffer while BUFFER_SIZE() uses sizeof the structure." is 100% wrong,

`BUFFER_SIZE()` uses the field within the struct.

williamcroberts: I see your using the size stored in the structure, but why are you adding 2?

williamcroberts: Ahh I see, now it makes sense it's not in the diffstat. Thos calcSize routines were inherited and already apart of tpm2_quote. Those routines look very wrong. I would remove those, as that's the cruft we want to get rid of in the repository.

We should be using the size fields and sizeof() and other non-magic ways to determine the sizes of buffers, or just serialize the buffer to disk.

The current problem of casting a struct to a byte() array, is that any padding goes to disk with it, and the file becomes machine dependent.

williamcroberts:
@mgerstner it needs tests. We can deal with the sizeof calculation cleanups in a different PR, no need to do that here. The calc issue is #495

liuqun:
Hi @mgerstner 

There is a command line tool `pki --keyid` from project strongswan, which can be used to convert any tpm handle into a pubkey of OpenSSL PEM/DER format. Hope it‘ll be helpful for you.

- https://wiki.strongswan.org/projects/strongswan/wiki/TpmPlugin

mgerstner:
@williamcroberts what exactly needs tests? are there any changes for me left to do here then?

williamcroberts:
@mgerstner For testing, you could output to PEM format and make sure that openssl can use it.

mgerstner:
@williamcroberts Of course I've tested the changes and the output formats behave as (I) expected. I also ran the utilities through valgrind.

So are you requesting a unit test here?

williamcroberts:
@mgerstner no need for a unit test under test/unit. But integration tests under test/system would be great. We need something that exercises this code path in the test/system location.

liuqun:
There are some usefull tools to make OpenSSH work perfectly with TPM protected private keys

- libtpm2-pk11.so: PKCS11 provider for TPM 2.0 chips
https://github.com/irtimmer/tpm2-pk11

- libsimple-tpm-pk11.so: PKCS11 provider for TPM 1.2 chips
https://github.com/ThomasHabets/simple-tpm-pk11

mgerstner:
To get this PR into shape I've rebased the follow-up commits to clean up the history. I've sync-merged with your current master branch. I added man page documentation for the new switches.

Regarding the system test I considered just extending some of the existing tests like `test_tpm2_readpublic.sh`, however they mostly use the `TPM_ALG_KEYEDHASH` keys, so I can't test the OpenSSL logic with them.

So adding a dedicated test would be easier I guess. Do you run these tests against the tpm simulator?

I will be on vacation for the next two weeks so I can't work on this for a while.

williamcroberts:
@mgerstner we do run it against the simulator

mgerstner:
So here it is: a system test that covers the newly added functionality.

man/tpm2_quote.1.md
```diff
@@ -77,6 +84,12 @@ tpm2_quote -k 0x81010002 -P "hex:123abc" -L sha1:16,17,18+sha256:16,17,18 -o out
 
 [Github Issues](https://github.com/01org/tpm2-tools/issues)
 
+<<<<<<< HEAD
```
williamcroberts: bad merge, ill fix this up.

PR #490, opened 2017-09-14
tpm2_listpersistent: print algorithms names instead of their hex values

martinezjavier:
Now that most tools support to specify the algorithms using pretty names
instead of hexadecimal values, there is no need to keep printing these.

### DIFF
--- a/tools/tpm2_listpersistent.c
+++ b/tools/tpm2_listpersistent.c
@@ -40,6 +40,7 @@
 
 #include "tpm2_options.h"
 #include "files.h"
+#include "tpm2_alg_util.h"
 #include "tpm2_tool.h"
 #include "tpm2_util.h"
 
@@ -66,9 +67,13 @@ int readPublic(TSS2_SYS_CONTEXT *sapi_context,
         return -1;
     }
 
+
+    TPMI_ALG_PUBLIC type = outPublic.t.publicArea.type;
+    TPMI_ALG_HASH nameAlg = outPublic.t.publicArea.nameAlg;
+
     printf("  {\n");
-    printf("\tType: 0x%x\n ", outPublic.t.publicArea.type);
-    printf("\tHash algorithm(nameAlg): 0x%x\n ", outPublic.t.publicArea.nameAlg);
+    printf("\tKey algorithm: %s(0x%x)\n", tpm2_alg_util_algtostr(type), type);
+    printf("\tHash algorithm: %s(0x%x)\n", tpm2_alg_util_algtostr(nameAlg), nameAlg);
     printf("\tAttributes: 0x%x\n", outPublic.t.publicArea.objectAttributes.val);
     printf("  }\n");
 
--- a/tools/tpm2_listpersistent.c
+++ b/tools/tpm2_listpersistent.c
@@ -40,6 +40,7 @@
 
 #include "tpm2_options.h"
 #include "files.h"
+#include "log.h"
 #include "tpm2_alg_util.h"
 #include "tpm2_tool.h"
 #include "tpm2_util.h"
@@ -63,7 +64,7 @@ int readPublic(TSS2_SYS_CONTEXT *sapi_context,
     rval = Tss2_Sys_ReadPublic(sapi_context, objectHandle, 0, &outPublic, &name, &qualifiedName, &sessionsDataOut);
     if(rval != TPM_RC_SUCCESS)
     {
-        printf("\nTPM2_ReadPublic error: rval = 0x%0x\n\n",rval);
+        LOG_ERR("\nTPM2_ReadPublic error: rval = 0x%0x\n",rval);
         return -1;
     }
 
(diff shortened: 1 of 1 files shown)


tools/tpm2_listpersistent.c
```diff
@@ -67,8 +68,9 @@ int readPublic(TSS2_SYS_CONTEXT *sapi_context,
     }
 
     printf("  {\n");
-    printf("\tType: 0x%x\n ", outPublic.t.publicArea.type);
-    printf("\tHash algorithm(nameAlg): 0x%x\n ", outPublic.t.publicArea.nameAlg);
+    printf("\tType: %s\n", tpm2_alg_util_algtostr(outPublic.t.publicArea.type));
+    printf("\tHash algorithm(nameAlg): %s\n",
+           tpm2_alg_util_algtostr(outPublic.t.publicArea.nameAlg));
```
williamcroberts: These need to be converted to tool output at some point.

williamcroberts: Actually we should keep the hex values and add the name to the output?

Hash Alg: sha256(0xB)

martinezjavier: @williamcroberts sounds good to me.

w.r.t tool output, I wondered about the same but thought what use could have this tool without any output. But now thinking about it, I think that everything should be converted to tool output besides the persistent handle.

That way someone can do something like this:

```
$ handle=$(tpm2_listpersistent -Q)
$ tpm2_evictcontrol -A o -H $handle -S $handle
```
What do you think?

martinezjavier: @williamcroberts Ok, I just pushed a commit to explain what I meant before reading this comment.

I thought that the idea of -Q was to be less verbose, I wonder what could be the case for someone just wanting the exit code of a tool that list stuff. But I see your point of the option being consistent.

williamcroberts: @martinezjavier thinkg of -q with grep, no output just lets you know if there is a match. Here it's less obvious of a use case (I don't have one per-say, outside of testing listpersistent command). But by redirecting stdout, you can control the output you need if its formatted nicely.

martinezjavier: If you want I can re-spin having the handle also printed with tpm2_tool_output().

In any case if someone finds that just printing the handle could be useful (I guess it may be like the example script snippet I shared), a tool specific option (-s, --summary ?) can be added to only prints the persistent handle.

williamcroberts: If the output is simply parsed via grep + cut, then we don't need any more support in the tool itself.

martinezjavier: @williamcroberts Ok, updated the last commit as you suggested.

martinezjavier: @williamcroberts agreed about just have parsed by other tools. Now what I think that could be useful though is being able to filter depending on the persistent object associated key and hash algorithms. Similar to what's done in tpm2_pcrlist.

tools/tpm2_listpersistent.c
```diff
@@ -89,20 +95,20 @@ int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
...
     {
-        printf("\n......GetCapability: Get persistent object list Error."
-               " TPM Error:0x%x......\n", rval);
+        LOG_ERR("\n......GetCapability: Get persistent object list Error."
+               " TPM Error:0x%x......", rval);
         return 1;
     }
 
-    printf( "%d persistent objects defined.\n", capabilityData.data.handles.count);
+    tpm2_tool_output( "%d persistent objects defined.", capabilityData.data.handles.count);
```
williamcroberts: Maybe something less wordy:
count:<num>

williamcroberts: Do we even need this, as we have persistent handle[0]...n below?

tools/tpm2_listpersistent.c
```diff
@@ -89,20 +95,20 @@ int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
...
     UINT32 i;
     for( i=0; i < capabilityData.data.handles.count; i++ )
     {
-        printf("\n  %d. Persistent handle: 0x%x\n", i, capabilityData.data.handles.handle[i]);
+        tpm2_tool_output("\n  %d. Persistent handle: 0x%x\n", i, capabilityData.data.handles.handle[i]);
```
williamcroberts: no caps, just "persistent handle" maybe do array notation:

persistent handle[0]:0x12345678

Would spaces in the output add trailing/leading whitespace to variables in bash?

williamcroberts: We'd probably want readPublic to have easier syntax to parse...
```
persistent-handle[0]:0x12345678 key-alg:sha1 hash-alg:sha256 attrs:0x56
``
This way to get key-alg for a handle...
```
command | cut -d' ' -f 2-2 | cut -d\: -f 2-2
sha1
```

File: tools/tpm2_listpersistent.c [hunk 4]
```diff
@@ -62,15 +64,19 @@ int readPublic(TSS2_SYS_CONTEXT *sapi_context,
...
-    printf("  }\n");
+
+    TPMI_ALG_PUBLIC type = outPublic.t.publicArea.type;
+    TPMI_ALG_HASH nameAlg = outPublic.t.publicArea.nameAlg;
+
+    tpm2_tool_output("  {\n");
+    tpm2_tool_output("\tKey algorithm: %s(0x%x)\n", tpm2_alg_util_algtostr(type), type);
+    tpm2_tool_output("\tHash algorithm: %s(0x%x)\n", tpm2_alg_util_algtostr(nameAlg), nameAlg);
+    tpm2_tool_output("\tAttributes: 0x%x\n", outPublic.t.publicArea.objectAttributes.val);
+    tpm2_tool_output("  }\n");
```
williamcroberts: not thrilled about this format style, see comments below.

tools/tpm2_listpersistent.c
martinezjavier: @williamcroberts the intention of this patch was just to replace the printf() calls for the tool output and log err, but I do agree with you that the current output is far from ideal: not easier for machines to parse yet no really readable for humans either.

So I'll add another commit on top that takes your comments into account and changes the tool output. I likely do it tomorrow though since now I'm busy with other stuff.

williamcroberts: @martinezjavier that sounds perfect.

martinezjavier:
@williamcroberts I'm going to close this. I've the patches we discussed yesterday and will create a new PR but I'll wait for your markdown PR #491 to be merged first to avoid conflicts.

PR #496, opened 2017-09-15
Man page to md

williamcroberts:
25 down, 15 to go...


man/tpm2_loadexternal.1.md
```diff
@@ -0,0 +1,72 @@
...
+
+[common tcti options](common/tcti.md)
+
+EXAMPLES
+--------
+
+Load a public area generated by tpm2_create, tpm2_readpublic or manually
+
+```
+tpm2_loadexternal -H <e|o|p|n> -u <pubKeyFileName> -C object.context
```
martinezjavier: Here again the example is inconsistent with the other man pages. The options are listed above so it makes no sense to list here again, but yes I agree with you that we should fix as a separate issue.

PR #498, opened 2017-09-15
tpm2_listpersistent enhancements

martinezjavier:
This is a v2 of PR #490. As suggested by @williamcroberts, it instead improves the output format used by the tool and also adds a -g and -G options to filter by the hash and key algorithms like's done in other tools (i.e: tpm2_pcrlist).

### DIFF
--- a/man/tpm2_listpersistent.1.md
+++ b/man/tpm2_listpersistent.1.md
@@ -19,18 +19,35 @@ tpm2_listpersistent(1) - display all defined persistent objects.
 
 OPTIONS
 -----------
-This tool takes no tool specific options.
+These options for listing the persistent objects:
+
+  * `-g`, `--halg`=_ALGORITHM_:
+    Only display persistent objects using this hash algorithm. Algorithms should
+    follow the " formatting standards, see section "Algorithm Specifiers".
+    Also, see section "Supported Hash Algorithms" for a list of supported
+    hash algorithms.
+
+  * `-G`, `--kalg`=_KEY\_ALGORITHM_:
+    Only display persistent objects using this key algorithm. It accepts friendly
+    names just like `-g` option. See section "Supported Public Object Algorithms"
+    for a list of supported object algorithms.
 
 [common options](common/options.md)
 
 [common tcti options](common/tcti.md)
 
+[supported hash algorithms](common/hash.md)
+
+[supported public object algorithms](common/object-alg.md)
+
+[algorithm specifiers](common/alg.md)
 
 EXAMPLES
 --------
 
 ```
 tpm2_listpersistent
+tpm2_listpersistent -g sha256 -G ecc
 
 ```
 
--- a/tools/tpm2_listpersistent.c
+++ b/tools/tpm2_listpersistent.c
@@ -40,11 +40,14 @@
 
 #include "tpm2_options.h"
 #include "files.h"
+#include "log.h"
+#include "tpm2_alg_util.h"
 #include "tpm2_tool.h"
 #include "tpm2_util.h"
 
 int readPublic(TSS2_SYS_CONTEXT *sapi_context,
-               TPMI_DH_OBJECT    objectHandle)
+               TPMI_DH_OBJECT    objectHandle,
+               int               index)
 {
     UINT32 rval;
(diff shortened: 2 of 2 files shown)


tools/tpm2_listpersistent.c
```diff
@@ -62,19 +92,37 @@ int readPublic(TSS2_SYS_CONTEXT *sapi_context,
...
-    printf("\tType: 0x%x\n ", outPublic.t.publicArea.type);
-    printf("\tHash algorithm(nameAlg): 0x%x\n ", outPublic.t.publicArea.nameAlg);
-    printf("\tAttributes: 0x%x\n", outPublic.t.publicArea.objectAttributes.val);
-    printf("  }\n");
+    TPMI_ALG_PUBLIC type = outPublic.t.publicArea.type;
+    TPMI_ALG_HASH nameAlg = outPublic.t.publicArea.nameAlg;
+    UINT32 attrs = outPublic.t.publicArea.objectAttributes.val;
+
+    if ((ctx.type && (type != ctx.type)) ||
+        (ctx.nameAlg && (nameAlg != ctx.nameAlg)))
```
williamcroberts: So TPM_ALG_ERROR aka 0 is that what you want here, ie if it's not zero run the filter. Perhaps initialize these to TPM_ALG_NULL? Not sure what/if the benefit is, i just want to make sure its clear that your checking against TPM_ALG_ERROR here.

man/tpm2_listpersistent.1.md
```diff
@@ -19,18 +19,35 @@ tpm2_listpersistent(1) - display all defined persistent objects.
...
+  * `-g`, `--halg`=_ALGORITHM_:
+    Only display persistent objects using this hash algorithm. Algorithms should
+    follow the " formatting standards, see section "Algorithm Specifiers".
+    Also, see section "Supported Hash Algorithms" for a list of supported
+    hash algorithms.
+
+  * `-G`, `--kalg`=_KEY\_ALGORITHM_:
+    Only display persistent objects using this key algorithm. It accepts friendly
+    names just like `-g` option. See section "Supported Public Object Algorithms"
+    for a list of supported object algorithms.
```
williamcroberts: The problem with filters, is folks always want more... But this is a start...

tools/tpm2_listpersistent.c
martinezjavier: Yes, I noticed that TPM_ALG_ERROR is 0 but I thought it would be OK since either the members won't be initialized or the option would had been wrong and failed.

But now that you mention, probably I should init to TPM_ALG_NULL to check in the tpm2_tool_onstart() that the user didn't provide wrong arguments. And while looking at that I noticed a flaw in tpm2_alg_util_from_optarg().

If you provide a hash algorithm to the key option or viceversa, tpm2_alg_util_from_optarg() will happily convert the argument to a valid TPM_ALG value.

martinezjavier: @williamcroberts this is what I meant BTW:

```
$ tpm2_listpersistent 
persistent-handle[0]:0x81000000 key-alg:ecc hash-alg:sha256 object-attr:0x30072

$ tpm2_listpersistent -g foo
ERROR: Invalid hash algorithm, got "foo"

$ tpm2_listpersistent -g ecc
$
```

In any case this is independent to this change, so I'll just update whit your suggestion and we should fix tpm2_alg_util_from_optarg() as a separate issue.

williamcroberts: That's not an issue in ```tpm2_alg_util_from_optarg```, the caller should check if the algorithm is a hash algorithm via: ```tpm2_alg_util_is_hash_alg```

I thought about adding flags to the optarg conversion routine to specify what types of algorithms are allowed...

```
union flags {
  struct {
    UINT8 hash            : 1;
    UINT8 asymmetric : 1;
    UINT8 symmetric   : 1;
    UINT8 unused        : 5;
  };
UINT8 val;
};
```

And then based on the flags, check if its valid.

martinezjavier: @williamcroberts got it, I missed that function. I'll add a fixup then.

PR #499, opened 2017-09-15
man: add Public Object Algorithms section to tpm2_create page

martinezjavier:
The tpm2_create tool man page is missing the section about key algorithms.

While being there, also change the examples to list the pretty names.

### DIFF
--- a/man/tpm2_create.1.md
+++ b/man/tpm2_create.1.md
@@ -45,7 +45,8 @@ These options for creating the tpm entity:
 
   * `-G`, `--kalg`=_KEY\_ALGORITHM_:
     The algorithm associated with this object. It accepts friendly names just
-    like -g option.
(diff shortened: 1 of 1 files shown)


martinezjavier:
@williamcroberts I noticed that the commit in this PR was never merged. It was that I closed but I don't remember doing it...

In any case, I've rebased on current master and reopened the PR.

williamcroberts:
because the branch was deleted and restored, you need to re-open a new PR.

martinezjavier:
Ok, I found another bug in tpm2_load man page so I'll create a new PR with both fixes.

martinezjavier:
@williamcroberts I didn't know  thatdeleting a branch before being merged would automatically close the PR, but that makes sense. At least now I know how I closed it.

Issue #500, opened 2017-09-15
Hierarchy authorization input overwriting primary object authorization 


idesai:
Primary objects authentication password sensitive input is overwritten by the hierarchy authentication password data resulting in creation of primary objects with passwords same as hierarchy authentication password.

idesai:
@williamcroberts can close this, fixed in 0dd15d2

PR #501, opened 2017-09-15
Fix for hierarchy authentication password overwriting the primary object's authentication password

idesai:
Fix for issue #500

### DIFF
--- a/tools/tpm2_createprimary.c
+++ b/tools/tpm2_createprimary.c
@@ -54,6 +54,7 @@ TPM_HANDLE handle2048rsa;
 typedef struct tpm_createprimary_ctx tpm_createprimary_ctx;
 struct tpm_createprimary_ctx {
     TPMS_AUTH_COMMAND session_data;
+    TPM2B_SENSITIVE_CREATE inSensitive;
     TPM2B_PUBLIC in_public;
     TPMI_ALG_PUBLIC type;
     TPMI_ALG_HASH nameAlg;
@@ -75,6 +76,7 @@ static tpm_createprimary_ctx ctx = {
         .hmac = TPM2B_EMPTY_INIT,
         .sessionAttributes = SESSION_ATTRIBUTES_INIT(0),
     },
+    .inSensitive = TPM2B_SENSITIVE_CREATE_EMPTY_INIT,
     .in_public = TPM2B_EMPTY_INIT,
     .type = TPM_ALG_RSA,
     .nameAlg = TPM_ALG_SHA1,
@@ -154,7 +156,6 @@ int setup_alg(void) {
 
 int create_primary(TSS2_SYS_CONTEXT *sapi_context) {
     UINT32 rval;
-    TPM2B_SENSITIVE_CREATE inSensitive = TPM2B_EMPTY_INIT;
     TPMS_AUTH_RESPONSE sessionDataOut;
     TSS2_SYS_CMD_AUTHS sessionsData;
     TSS2_SYS_RSP_AUTHS sessionsDataOut;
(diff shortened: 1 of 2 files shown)


tools/tpm2_createprimary.c
```diff
@@ -178,16 +179,16 @@ int create_primary(TSS2_SYS_CONTEXT *sapi_context) {
     sessionsData.cmdAuthsCount = 1;
     sessionsDataOut.rspAuthsCount = 1;
 
-    inSensitive.t.sensitive.data.t.size = 0;
-    inSensitive.t.size = inSensitive.t.sensitive.userAuth.b.size + 2;
+    ctx.inSensitive.t.sensitive.data.t.size = 0;
```
williamcroberts: instead of using TPM2B_EMPTY_INIT, can you create a new macro for this that sets both size fields to 0?
```
TPM2B_SENSITIVE_CREATE_EMPTY_INIT
```

tools/tpm2_createprimary.c
```diff
@@ -178,16 +179,16 @@ int create_primary(TSS2_SYS_CONTEXT *sapi_context) {
...
+    ctx.inSensitive.t.size = ctx.inSensitive.t.sensitive.userAuth.b.size + 2;
```
williamcroberts:
```
ctx.inSensitive.t.size + sizeof(ctx.inSensitive.t.size) would be better than adding 2....
```

williamcroberts:
@flihp @tstruk Take a look at this PR. Even generating some of the nested structure size fields require packed structures. I think the attribute(packed) is going to have to live on all the public structure definitions..... This code is so subtle/error prone.

tools/tpm2_createprimary.c
```diff
@@ -178,16 +179,16 @@ int create_primary(TSS2_SYS_CONTEXT *sapi_context) {
     sessionsData.cmdAuthsCount = 1;
     sessionsDataOut.rspAuthsCount = 1;
 
-    inSensitive.t.sensitive.data.t.size = 0;
-    inSensitive.t.size = inSensitive.t.sensitive.userAuth.b.size + 2;
+    ctx.inSensitive.t.size = ctx.inSensitive.t.sensitive.userAuth.b.size + 
+        sizeof(ctx.inSensitive.t.size);
```
williamcroberts: This requires attribute(packed) on structures to work correctly....

I can't wait for .t and .b to go away....

williamcroberts:
@tstruk Is it because its a variable length struct? I wonder what happens on ARM  as ARM requires word alignment at 4 byte boundaries for LDR and STR instructions. Ie will there be 2 bytes of padding between buf and size? I would imagine not since buf is of type BYTE and it's a variable length array...

The most annoying thing is having to set size on that:
It's defined as:
```
typedef union {
    struct  {
        UINT16  size;
        TPMS_SENSITIVE_CREATE  sensitive;
    } t;
    TPM2B   b;
    } TPM2B_SENSITIVE_CREATE
```
I don't understand the point of wrapping a TPMS_SENSITIVE_CREATE with a size field when TPMS_SENSITIVE_CREATE  contains TPM2B fields that have size... its awkward.

tstruk:
On 09/20/2017 01:16 PM, William Roberts wrote:
> @tstruk <https://github.com/tstruk> Is it because its a variable length struct? I wonder what happens on ARM as ARM requires word alignment at 4 byte boundaries for LDR and STR instructions. Ie will there be 2 bytes of padding between buf and size? I would imagine not since buf is of type BYTE and it's a variable length array...

@williamcroberts That's is right. There is no point adding padding between UINT16 and UINT8[N]

> 
> The most annoying thing is having to set size on that:
> It's defined as:
> 
> |typedef union { struct { UINT16 size; TPMS_SENSITIVE_CREATE sensitive; } t; TPM2B b; } TPM2B_SENSITIVE_CREATE |
> 
> I don't understand the point of wrapping a TPMS_SENSITIVE_CREATE with a size field when TPMS_SENSITIVE_CREATE contains TPM2B fields that have size... its awkward.
> 

I think this is because TPMS_SENSITIVE_CREATE consists of two TPM2B members and each of them have its own size, and 
TPMS_SENSITIVE_CREATE.b.size is the total size of the whole thing.

PR #516, opened 2017-09-20
Fixes for tpm2_getmanufec

martinezjavier:
When testing the tool I noticed that it some bugs. These commits fixes them.

### DIFF
--- a/tools/tpm2_getmanufec.c
+++ b/tools/tpm2_getmanufec.c
@@ -133,7 +133,7 @@ int set_key_algorithm(TPM2B_PUBLIC *inPublic) {
         break;
     default:
         LOG_ERR("\nThe algorithm type input(%4.4x) is not supported!", ctx.algorithm_type);
-        return -1;
+        return 1;
     }
 
     return 0;
@@ -195,7 +195,7 @@ int createEKHandle(TSS2_SYS_CONTEXT *sapi_context)
                 if (tpm2_util_hex_to_byte_structure(ctx.endorse_passwd, &sessionData.hmac.t.size,
                                       sessionData.hmac.t.buffer) != 0) {
                         LOG_ERR("Failed to convert Hex format password for endorsePasswd.");
-                        return -1;
+                        return 1;
                 }
         }
     }
@@ -211,7 +211,7 @@ int createEKHandle(TSS2_SYS_CONTEXT *sapi_context)
              if (tpm2_util_hex_to_byte_structure(ctx.ek_passwd, &inSensitive.t.sensitive.userAuth.t.size,
                                    inSensitive.t.sensitive.userAuth.t.buffer) != 0) {
                   LOG_ERR("Failed to convert Hex format password for ekPasswd.");
-                  return -1;
+                  return 1;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


tools/tpm2_getmanufec.c
```diff
@@ -184,36 +184,35 @@ int createEKHandle(TSS2_SYS_CONTEXT *sapi_context)
...
-                        return 1;
-                }
+    if (ctx.endorse_passwd && strlen(ctx.endorse_passwd) > 0) {
+        if (!ctx.hex_passwd) {
+            memcpy(&sessionData.hmac.t.buffer[0], ctx.endorse_passwd,
+                   sessionData.hmac.t.size );
+        } else {
+            sessionData.hmac.t.size = sizeof(sessionData.hmac) - 2;
+
+            if (tpm2_util_hex_to_byte_structure(ctx.endorse_passwd,
```
williamcroberts: Oooh old -X option code...

williamcroberts: Update this to use the password util routines, that's my only request for this PR, the rest looks great.

martinezjavier:
@williamcroberts Ok, since is an independent change I've added it as a separate commit. Also while testing the password options, I noticed that the endorsement password wasn't working so I fixed it before doing the conversion.

And also added a test case to ensure that the password options won't regress in the future and cleaned the test a bit.

williamcroberts:
Fail Snippet:
```
WARN: TLS communication with the said TPM manufacturer server setup with SSL_NO_VERIFY!
persistent-handle[0]:0x81000000 key-alg:rsa hash-alg:sha256 object-attr:0x300b2
tpm2_evictcontrol -H $handle -A o $opass on line 80 failed: 1
[31m test_tpm2_getmanufec.sh Fail [0m
Success: setting up the number of allowed tries in the lockout parameters
```

It passed on CLANG, so I restarted the build.

tools/tpm2_getmanufec.c
```diff
@@ -582,13 +526,19 @@ static bool on_option(char key, char *value) {
...
     case 'i':
-        return_val = tpm2_util_string_to_uint32(value, &ctx.auth_session_handle);
+        return_val = tpm2_util_string_to_uint32(value, &ctx.endorse_session_data.sessionHandle);
+        if (!return_val) {
+            LOG_ERR("Could not convert session handle to number, got: \"%s\"",
+                    value);
+            return false;
+        }
+
+        return_val = tpm2_util_string_to_uint32(value, &ctx.owner_session_data.sessionHandle);
```
williamcroberts: This could be a lot more compact:
```
UINT32 tmp;
return_val = tpm2_util_string_to_uint32(value, &tmp);
if (error)...
fi
ctx.owner_session_data.sessionHandle =
ctx.endorse_session_data.sessionHandle = tmp;
```
or
```
return_val = tpm2_util_string_to_uint32(value, &ctx.endorse_session_data.sessionHandle)
                    && tpm2_util_string_to_uint32(value, &ctx.owner_session_data.sessionHandle);
if (error)....
fi
```

williamcroberts:
@martinezjavier looks broken with GCC

tools/tpm2_getmanufec.c
martinezjavier: indeed, I wasn't thinking clearly.

martinezjavier:
@williamcroberts yes, not sure why though. I'll investigate.

martinezjavier:
@williamcroberts the CI test passes now.

I just pushed a new version since found an error in the updated man page, but should be ready to be merged.

Issue #335, opened 2017-07-12
Make sure that all tests work on a real TPM hardware


martinezjavier:
The tpm2.0-tools tests have been mostly used to avoid regressions by running them on the Travis continuous integration environment each time that a change is proposed.

The tests are ran using the tpm2-abrmd TCTI and the ibmswtpm2 TPM2 simulator.

But it would be good to also make sure that the tests run correctly on real TPM hardware. Most of the tests already do but some were not working and were fixed (e.g: commit 99391407a44e) and some tests are still failing like [test_algs_tpm2_quote.sh](https://github.com/01org/tpm2.0-tools/blob/master/test/system/test_algs_tpm2_quote.sh) and [test_tpm2_createpolicy.sh](https://github.com/01org/tpm2.0-tools/blob/master/test/system/test_tpm2_createpolicy.sh)

```
$ ./test_all.sh
 ...
 Tests passed: 28 
 Tests Failed: 2
```

test_tpm2_createpolicy.sh fails because the expected digests that are used for comparison are hardcoded to the values expected by the TPM simulator:

```
declare -A expected_policy_digest (["sha1"]="f28230c080bbe417141199e36d18978228d8948fc10a6a24921b9eba6bb1d988"
 ["sha256"]="33e36e786c878632494217c3f490e74ca0a3a122a8a4f3c5302500df3b32b3b8")
```
Similarly, test_tpm2_quote.sh fails because the passed qualifying data is hardcoded to a value expected by the TPM simulator:

```
nonce=12345abcde12345abcde12345abcde12345abcde12345abcde12345abcde12345abcde12345abcde12345abcde12345abcde
```

jiazhang0:
Currently, all test cases with some fixes (already submitted or being reviewed) are passed on Nationz Z32H320TC, except test_tpm2_createpolicy which needs a generic design (See https://github.com/01org/tpm2-tools/issues/551).

$./test.sh 
test_tpm2_activecredential.sh ... PASSED
test_tpm2_akparse.sh ... PASSED
test_tpm2_certify.sh ... PASSED
\Expected PCR 0 "sha1" bank to be 0x3, found: "9872964b9b40cdd0363fcd6af8c267c9cb34200b"
test_tpm2_createpolicy.sh ... FAILED
test_tpm2_createprimary.sh ... PASSED
test_tpm2_create.sh ... PASSED
test_tpm2_dictionarylockout.sh ... PASSED
[...]

PR #517, opened 2017-09-20
test: use ECC algorithm instead of RSA for primary key in tpm2_unseal

martinezjavier:
Creating a primary key may take a lot of time, specially when using the
RSA algorithm. Since for test we can use any algorithm, use ECC instead
to speed up the test on most systems.

While being there, do other cleanups like using pretty names instead of
hexadecimal values and use the -E option in the tpm2_create command so
the needed object attributes are set by the tool.

### DIFF
--- a/test/system/test_tpm2_unseal.sh
+++ b/test/system/test_tpm2_unseal.sh
@@ -30,15 +30,13 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF 
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
-alg_primary_obj=0x0004
-alg_primary_key=0x0001
-alg_create_obj=0x000B
-alg_create_key=0x0008
-alg_pcr_policy=0x0004
+alg_primary_obj=sha256
+alg_primary_key=ecc
+alg_create_obj=sha256
+alg_create_key=keyedhash
+alg_pcr_policy=sha1
 
 pcr_ids="0,1,2,3"
-
-obj_attr=0x492 # fixedTPM, fixedParent, adminWithPolicy, noDA
 
 file_pcr_value=pcr.bin
 file_input_data=secret.data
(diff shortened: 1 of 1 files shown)


test/system/test_tpm2_unseal.sh
```diff
@@ -106,7 +104,7 @@ if [ $? != 0 ];then
     exit 1
 fi
 
-tpm2_create -g $alg_create_obj -G $alg_create_key -u $file_unseal_key_pub -r $file_unseal_key_priv -I- -c $file_primary_key_ctx -L $file_policy -A $obj_attr <<< $secret
+tpm2_create -g $alg_create_obj -G $alg_create_key -u $file_unseal_key_pub -r $file_unseal_key_priv -I- -c $file_primary_key_ctx -L $file_policy -E <<< $secret
```
williamcroberts: I can't recall the spec offhand, but is ECC a mandatory or optional requirement of tpm2.0. How does this work on a real tpm2.0 device.

How does this affect #335 ?

martinezjavier: @williamcroberts it's mandatory according to [TCG PC Client Platform TPM Profile (PTP) Specification (Family “2.0” Level 00 Revision 01.03 v22 May 22, 2017)](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PC_Client_Platform_TPM_Profile_PTP_Specification_Family_2.0_Revision_1.3v22.pdf).

If you look at "Table 2 - PC Client TPM Algorithms", both the TPM_ALG_ECDH (for {en,de}cryption) and TPM_ALG_ECDSA (for signing) are listed as mandatory.

This affects real TPM2.0 devices in the way that generating ECC keys is much faster than generating RSA keys. How much it depends on the chip, but on some I found that creating a RSA primary key can take up to 40 seconds!

You can measure this easily on your system with:

```
$ time tpm2_createprimary -A o -g sha256 -G rsa -C primary.context

CreatePrimary Succeed ! Handle: 0x80ffffff


real    0m7.245s
user    0m0.002s
sys     0m0.058

$ time tpm2_createprimary -A o -g sha256 -G ecc -C primary.context

CreatePrimary Succeed ! Handle: 0x80ffffff


real    0m0.469s
user    0m0.003s
sys     0m0.058s
```

So how this affect issue #355, well is much faster to execute this test on a real TPM2.0. Or put it on another way, using RSA just makes the test to take more time for no good reasons IMHO.

williamcroberts: Perfect Mandatory was what I needed to know.

Issue #486, opened 2017-09-13
Ensure that scripts are properly cleaning up their temporary files on exit


martinezjavier:
Some tests create temporary files but don't remove it. These should be cleaned up.

Also, some tests are just removing the temporary files at the end of their execution. So if the script finishes early due some errors (or a user cancels it with a signal), the files are never removed. Instead scripts should define a cleanup function that's set as a trap handler for the EXIT signal.

For an example, see commit 69bd17f53b48 ("test: make sure temp files are cleaned up on test_tpm2_nv.sh exit").

williamcroberts:
This has been opened before, I hacked test_wrapper in test_all.sh to clean things up when test scripts don't, but it's quite a hack. You can likely blame that for the ticket information.

martinezjavier:
@williamcroberts I see, you mean commit f569e46ad20f ("tests: cleanup test artifacts") that fixed issue #125.

martinezjavier:
So we should remove that once all the tests are properly cleaning their temporary files.

williamcroberts:
@martinezjavier yeah that nasty hack should be removed when the tests are doing things correctly.

PR #519, opened 2017-09-21
Test cleanup

williamcroberts:
This makes progress towards #486


test/system/test_tpm2_activecredential.sh
```diff
@@ -4,81 +4,62 @@
...
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
-handle_ek=0x81010009
-handle_ak=0x8101000a
```
martinezjavier: Any reason you are removing these variables? My rule of thumps is to have variables in the tests if the values are used at least twice. That way if for some reason needs to be changed in the future, will be less error prone.

test/system/test_tpm2_akparse.sh
```diff
@@ -4,52 +4,55 @@
...
+}
+trap onerror ERR
+
+cleanup() {
+    rm -f ak.out akparse.out ek.pub ak.pub ak.name &>/dev/null
 
-rm $output_ekparse -rf 
+    # Evict persistent handles, we want them to always succeed and never trip
+    # the onerror trap.
+	tpm2_evictcontrol -Q -A o -H 0x8101000b 2>/dev/null || true
```
martinezjavier: You could instead do:

```
if tpm2_listpersistent | grep -q 0x8101000b; then
    tpm2_evictcontrol -Q -A o -H 0x8101000b
fi
```

test/system/test_tpm2_akparse.sh
```diff
@@ -4,52 +4,55 @@
...
+cleanup
 
- tpm2_akparse -f $file_input_data  -k $output_akparse
-if [ $? != 0 ];then
-	echo "akparse fail, please check the environment or parameters!"
-	exit 1
-fi
+tpm2_getpubek -Q -H 0x8101000b -g rsa -f ek.pub
 
-echo "akparse successfully!"
```
martinezjavier: I always find these messages to be useful. Yes, you could check $? but having instant feedback when running a test is a good thing IMHO.

martinezjavier: Having said that, I don't think "akparse successfully!" is even a correct english phrase so probably needs to be fixed. But making all the tests to output a consistent succeed message would be nice.

test/system/test_tpm2_create_all.sh
```diff
@@ -4,83 +4,69 @@
...
-gAlg=
-GAlg=
+onerror() {
+    echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
+    exit 1
+}
+trap onerror ERR
+
+cleanup() {
+  rm -f key.pub key.priv policy.bin &>/dev/null
```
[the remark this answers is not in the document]
williamcroberts: I guess `-f` is sufficient then

test/system/test_tpm2_activecredential.sh
williamcroberts: When the error is logged, the lines values are not expanded. Maybe it was some side affact of having the on error handler before the variable declarations.

test/system/test_tpm2_akparse.sh
williamcroberts: Thought about it, but didn't as I wanted to avoid 2 tpm calls for a flush

test/system/test_tpm2_akparse.sh
williamcroberts: test_all.sh handles that

martinezjavier: Yes, but no when run it as an standalone test.

williamcroberts: True, but I have some fancy output changes to test_all.sh...
I'd like the tests to be executable stand-alone scripts, but one can check the return for 0 for success.

Run these locally with test_all.sh and let me know what you think of the output.

williamcroberts:
Interesting failure for tpm2_getpubak and tpm2_getpubek:
```
[1;33m\[0m
** (tpm2-abrmd:24487): WARNING **: tpm2_response_get_handle: insufficient buffer to get handle
ERROR: TPM2_CreatePrimary Error. TPM Error:0x9a2
getpubek fail, please check the environment or parameters!

[1;31mtest_tpm2_getpubak.sh ... PASSED[0m

[1;33m\[0m
** (tpm2-abrmd:24487): WARNING **: tpm2_response_get_handle: insufficient buffer to get handle
ERROR: TPM2_CreatePrimary Error. TPM Error:0x9a2
getpubek fail, please check the environment or parameters!

[1;31mtest_tpm2_getpubek.sh ... PASSED[0m
```

It says passed but it's getting added to the fail log... weird.

williamcroberts:
Ahh its a typo, I never changed PASSED to FAILED.. oops

williamcroberts:
Im running abrmd at:  150356fb let my try updating locally and see what happens.

https://github.com/01org/tpm2-abrmd/commit/b4031c3b83eaca9ef9387aa0b5233cee14fca819

idesai:
@williamcroberts I have seen this error before with older resource manager and newer tools that triggered the structure packing discussion.

williamcroberts:
@martinezjavier maybe we can make it so test.sh runs them all but allow arguments to run a specific test:
``
./test.sh
./test tpm2_getmanufec
./test tpm2_getmanufec tpm2_getrandom
```

test/system/test_tpm2_akparse.sh
martinezjavier: Ok for me.

test/system/test_tpm2_akparse.sh
martinezjavier: @williamcroberts cool, I didn't have time to finish reviewing all the patches in the PR today so I didn't notice those changes but the spinning thing is nice indeed.

But I don't think those are mutually exclusive, after all the wrapper doesn't care about the test output, only its exit code.

Or the wrapper could pass an argument to the tests so only print their exit status when executed as standalone. I also mentioned that one could check $? but that's not user friendly in my opinion.

martinezjavier:
@williamcroberts running the tests through a test.sh script sounds good to me. As long as this is well documented then I don't think it will be a problem for users.

williamcroberts:
@idesai weird though, travis is grabbing from master

test/system/test_tpm2_akparse.sh
williamcroberts: We can do that, that's simple enough

williamcroberts:
I cannot get this to fail locally, i'll have to break this up into hunks and submit them in smaller batches

williamcroberts:
Darn, SourceForge is down, ie where we get the tpm simulator.

PR #522, opened 2017-09-26
test: fix test errors with test_tpm2_pcrevent.sh

jiazhang0:
The returned identify of sha1 algorithm may be "sha" and thus calling
shasum rather than sha1sum will fail for sure. Fix the alg if returning
"sha".

In addition, there is no equivalent like sha256sum for sm3_256.
Currently we simply ignore testing this sort of hash algorithm.

### DIFF
--- a/test/system/test_tpm2_pcrevent.sh
+++ b/test/system/test_tpm2_pcrevent.sh
@@ -62,6 +62,12 @@ cat $hash_in_file | tpm2_pcrevent > $hash_out_file
 # Verify output as expected.
 for l in `cat $hash_out_file`; do
   alg=`echo -n $l | cut -d\: -f 1-1`
+
+  if ! which "$alg"sum >/dev/null 2>&1; then
+      echo "Ignore checking $alg algorithm due to unavailable \"${alg}sum\" program"
+      continue
+  fi
+
   hash=`echo -n $l | cut -d\: -f 2-2`
   check=`"$alg"sum $hash_in_file | cut -d' ' -f 1-1`
   if [ "$check" != "$hash" ]; then


test/system/test_tpm2_pcrevent.sh
```diff
@@ -58,10 +58,18 @@ tpm2_pcrevent -Q < $hash_in_file
 
 # Test that fifo stdin works
 cat $hash_in_file | tpm2_pcrevent > $hash_out_file
+cat $hash_out_file
```
williamcroberts: Don't add any superfluous output to the testing code, this cat doesn't add any value, delete it.

test/system/test_tpm2_pcrevent.sh
```diff
@@ -58,10 +58,18 @@ tpm2_pcrevent -Q < $hash_in_file
...
 
 # Verify output as expected.
 for l in `cat $hash_out_file`; do
   alg=`echo -n $l | cut -d\: -f 1-1`
+  if [ x"$alg" == x"sha" ]; then
+    alg="sha1"
+  elif [ x"$alg" == x"sm3_256" ]; then
+    echo "Ignore checking the hash algorithm $alg"
+    continue
+  fi
```
williamcroberts: the elif clause is pointless, delete it.

tools/tpm2_createprimary.c
```diff
@@ -184,6 +184,7 @@ int create_primary(TSS2_SYS_CONTEXT *sapi_context) {
 
     if(setup_alg())
         return -1;
+    tpm2_tool_output("ObjectAttribute: 0x%08X\n", ctx.in_public.t.publicArea.objectAttributes.val);
```
williamcroberts: this patch looks fine

test/system/test_tpm2_pcrevent.sh
```diff
@@ -58,10 +58,18 @@ tpm2_pcrevent -Q < $hash_in_file
 
 # Test that fifo stdin works
 cat $hash_in_file | tpm2_pcrevent > $hash_out_file
+cat $hash_out_file
 
 # Verify output as expected.
 for l in `cat $hash_out_file`; do
   alg=`echo -n $l | cut -d\: -f 1-1`
+  if [ x"$alg" == x"sha" ]; then
+    alg="sha1"
```
martinezjavier: @williamcroberts this is in fact a bug in the tpm2_alg_util_algtostr() function.

The spec makes clear that both TPM_ALG_SHA and TPM_ALG_SHA1 refer to the SHA-1 algorithm, but the function table has entries for both even when map to the same hex value (0x0004). So instead I believe that TPM_ALG_SHA shouldn't be defined and always map 0x0004 to "sha1".

Otherwise you can't really convert TPM_ALG_SHA1 to its mapped string since TPM_ALG_SHA will always match first:

```
tpm2_tool_output("hash algorithm is %s\n", tpm2_alg_util_algtostr(TPM_ALG_SHA1));

hash algorithm is sha
```

test/system/test_tpm2_pcrevent.sh
martinezjavier: Instead of hardcoding the algorithms that don't have a sum command, couldn't you check if "$alg"sum exists?

test/system/test_tpm2_pcrevent.sh
williamcroberts: This crossed my mind yesterday. Yeah I could see the argument both ways, since they both map to the same numerical value, should sha and sha1 exist as friendly identifiers? I could see a case for both arguments, but today I am leaning towards killing sha. Let's take PR https://github.com/01org/tpm2-tools/pull/526 for the algutil changes and then you can update this PR with just the tpm2_tool_output() changes.

test/system/test_tpm2_pcrevent.sh
jiazhang0: OK thanks I will send a new PR.

PR #524, opened 2017-09-26
Doc improve

williamcroberts:


williamcroberts:
@martinezjavier How much do you hate the idea of jamming an xxd dump of the simulator into the .ci folder to avoid having to wget it from Source Forge, who is still having issues...?

martinezjavier:
@williamcroberts I don't really like it. It pollutes the repository for everyone just to keep the CI happy. I also wonder the licensing implications of distributing the tarball in the project...

If source force is not reliable, I would much prefer to keep the tarball hosted in another service and the CI script to download it from there (probably as a fallback only if source forge fails).

If you don't want to depend on another service, I guess you could just create an empty github project and commit it there? That way you only depend on a single service (since if github is down then the project is down anyways).

Lastly, what's the advantage of having it as an hex dump and not just in the tarball binary form?

williamcroberts:
As far as hex form Git doesn't do well with large binary files....
I'd like to just use source forge, but 2 days of downtime seems unruly. I think a separate github for CI would be nice anyways, so people don't have to get our ci stuff with the code.

PR #530, opened 2017-09-28
tpm2_dump_capability: add dumping ECC curves

OYTIS:
Add option to dump elliptic curves supported by the device.
Corresponding capability selector is TPM_CAM_ECC_CURVES.

Signed-off-by: Anton Gerasimov <anton@advancedtelematic.com>

### DIFF
--- a/tools/tpm2_dump_capability.c
+++ b/tools/tpm2_dump_capability.c
@@ -56,6 +56,18 @@ typedef struct capability_map_entry {
  */
 capability_map_entry_t capability_map[] = {
     {
+        .capability_string = "algorithms",
+        .capability        = TPM_CAP_ALGS,
+        .property          = TPM_ALG_FIRST,
+        .count             = MAX_ALG_LIST_SIZE,
+    },
+    {
+        .capability_string = "commands",
+        .capability        = TPM_CAP_COMMANDS,
+        .property          = TPM_CC_FIRST,
+        .count             = MAX_CAP_CC,
+    },
+    {
         .capability_string = "properties-fixed",
         .capability        = TPM_CAP_TPM_PROPERTIES,
         .property          = PT_FIXED,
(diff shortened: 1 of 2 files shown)


tools/tpm2_dump_capability.c
```diff
@@ -508,6 +557,7 @@ dump_command_attr_array (TPMA_CC     command_attributes[],
     for (i = 0; i < count; ++i)
         dump_command_attrs (command_attributes [i]);
 }
+
 /*
```
martinezjavier: Unrelated new line added.

tools/tpm2_dump_capability.c
```diff
@@ -562,6 +612,10 @@ static int dump_tpm_capability (TPMU_CAPABILITIES *capabilities) {
         dump_algorithms (capabilities->algorithms.algProperties,
                          capabilities->algorithms.count);
         break;
+    case TPM_CAP_ECC_CURVES:
+	dump_ecc_curves (capabilities->eccCurves.eccCurves,
+                         capabilities->eccCurves.count);
+	break;
```
martinezjavier: I think it would be a good idea to keep the capabilities sorted according to 6.12 TPM_CAP (Capabilities) table in Trusted Platform Module Library part 2.

Rest of the patch looks good, thank a lot for proposing it.

OYTIS: Sorry I don't quite get it. The previous sorting [PROPERTIES, ALGS, COMMANDS] didn't correspond 6.12 as well.

martinezjavier: Indeed and that's not your fault, but I meant that we should try to keep it sorted from now on (since is quite likely that more properties will be added the future).

We can later move TPM_CAP_TPM_PROPERTIES to the end to match the table order.

OYTIS: OK, this makes sense.

PR #537, opened 2017-10-03
man: use pandoc instead of md2man for man pages generation

martinezjavier:
The tools man pages used to be static roff files, and were converted to be
generated by md2man-roff from markdown documents.

But md2man is a rather obscure project, no major distros have the ruby gem
packaged, is basically a one man project and its last release was almost 2
years ago. Finally, it depends on the binman and opener projects from the
same author that have even older releases and less development.

While packaging these ruby gems is trivial, the fact that are no packages
is an indication that no many projects use it.

The pandoc project on the other hand has hundreds of contributors, it's in
active development and all major Linux distributions have packages for it.

So let's use pandoc to generate the man pages instead of md2man. This will
make easier not only for distros to update the tpm2-tools package but also
for developers and users to build tpm2-tools (with man pages) from source.


martinezjavier:
@williamcroberts yes, it does. What doesn't do is to take the literal `` as a bold text in the generated roff file.

IOW, md2man-roff converts both **foo** and `foo` in the markdown to **foo** in the man page (actually \fBfoo\fR but you know what I mean).

That's why I modified `tpm2_foo` [OPTIONS] in the synopsis to **tpm2_foo** [OPTIONS] so the generated roff file remains the same.

But if you prefer to have it as `tpm2_foo` so it can be rendered as a literal in Github and no as bold text in the man page, I can change it.

PR #540, opened 2017-10-05
Add -G option to tpm2_quote

harisokanovic:


man/tpm2_quote.1.md
```diff
@@ -41,7 +41,12 @@
 
   * **-o**, **--outFile**:
 
-    Output file path, recording the two structures output by tpm2_quote function.
+    Output file path, recording the two structures (quote and signature) output by tpm2_quote function.
+    Contains only one structure (quote) when -s is also specified
+
+  * **-s**, **--outSigFile**:
```
martinezjavier: Please don't use camel case for neither options nor variables. I know that you are just being consistent with the current implementation but that's something we want to change project wide.

tools/tpm2_quote.c
```diff
@@ -49,12 +49,12 @@ typedef struct {
...
 
 static TPMS_AUTH_COMMAND sessionData;
-static char *outFilePath;
+static char *outFilePath, *outSigFilePath;
 static TPM2B_DATA qualifyingData = TPM2B_EMPTY_INIT;
 static TPML_PCR_SELECTION  pcrSelections;
 static bool is_auth_session;
 static TPMI_SH_AUTH_SESSION auth_session_handle;
-static int k_flag, c_flag, l_flag, g_flag, L_flag, o_flag;
+static int k_flag, c_flag, l_flag, g_flag, L_flag, o_flag, s_flag;
```
martinezjavier: You don't really need the s_flag, just check for outSigFilePath (but rename it to not use camel case convention).

tools/tpm2_quote.c
```diff
@@ -261,19 +261,39 @@ static int quote(TSS2_SYS_CONTEXT *sapi_context, TPM_HANDLE akHandle, TPML_PCR_S
         printf("OutFile: %s Can Not Be Created !\n",outFilePath);
         return -2;
     }
+
+    FILE *sigfp = NULL;
+    if(s_flag)
+    {
```
martinezjavier: Please check the coding style at misc/coding_standard_c.txt and follow that.

man/tpm2_quote.1.md
harisokanovic: No problem. I changed the log opt name.
Also note that someone already added a -f option equivalent to my proposed -s, so I dropped the first patch in this series. I'll resubmit the -G part.

tools/tpm2_quote.c
harisokanovic: Fixed

PR #543, opened 2017-10-07
tpm2_createprimary: support to print the object attributes

jiazhang0:
This does the same as tpm2_create.


tools/tpm2_createprimary.c
```diff
@@ -184,6 +184,7 @@ int create_primary(TSS2_SYS_CONTEXT *sapi_context) {
 
     if(setup_alg())
         return -1;
+    tpm2_tool_output("ObjectAttribute: 0x%08X\n", ctx.in_public.t.publicArea.objectAttributes.val);
```
williamcroberts: We should probably be printing these out nicer, see #557

PR #546, opened 2017-10-09
[V4] tpm2_create: fix RC_ATTRIBUTES error when running test_tpm2_encryptde…

jiazhang0:
Changelog:

- V4
  * Rebase.

- V3
  * Explicitly mentioning TPM 2.0 spec rev 1.16.
  * Place the tpm2_create wrapper to the common piece test_helpers.sh.

Note that this patch is based on the one submitted to https://github.com/01org/tpm2-tools/pull/552
 
### DIFF
--- a/test/system/test_helpers.sh
+++ b/test/system/test_helpers.sh
@@ -99,3 +99,18 @@ hash_alg_supported() {
         fi
     done
 }
+
+# Certain TPM 2.0 chip, e.g, Nationz Z32H320TC, and tpm2simulator referring
+# to TPM 2.0 spec rev 1.16 may have an errata, disallowing both decrypt and
+# sign set for a symcipher object, and in response to RC_ATTRIBUTES. In
+# order to work around it, we attempt to run tpm2_create again with sign bit
+# clear.
+create_object() {
+    local alg_create_obj=0x20072
+
+    if tpm2_create $@ 2>&1 | grep -q 'Create Object Failed ! ErrorCode: 0x2c2'; then
+        tpm2_create -Q -A $alg_create_obj $@
+    else
+        true
+    fi
+}
--- a/test/system/test_tpm2_create.sh
+++ b/test/system/test_tpm2_create.sh
@@ -57,7 +57,7 @@ tpm2_createprimary -Q -A p -g sha1 -G rsa -C context.out
 # values.
 for gAlg in `populate_hash_algs mixed`; do
     for GAlg in rsa 0x08 ecc 0x25; do
-        tpm2_create -Q -c context.out -g $gAlg -G $GAlg -u key.pub -r key.priv
+        create_object -c context.out -g $gAlg -G $GAlg -u key.pub -r key.priv
         cleanup keep_context
     done
 done
@@ -67,7 +67,7 @@ cleanup keep_context
 echo "f28230c080bbe417141199e36d18978228d8948fc10a6a24921b9eba6bb1d988" \
 | xxd -r -p > policy.bin
 
-tpm2_create -Q -c context.out -g sha256 -G 0x1 -L policy.bin -u key.pub -r key.priv -E
+create_object -c context.out -g sha256 -G 0x1 -L policy.bin -u key.pub -r key.priv -E
 
 cmp -i 14:0 -n 32 key.pub policy.bin -s
 
--- a/test/system/test_tpm2_encryptdecrypt.sh
+++ b/test/system/test_tpm2_encryptdecrypt.sh
@@ -31,6 +31,8 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
 
+source test_helpers.sh
+
 onerror() {
     echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
     exit 1
@@ -60,7 +62,7 @@ tpm2_takeownership -Q -c
 
 tpm2_createprimary -Q -A e -g sha1 -G rsa -C primary.ctx
 
-tpm2_create -Q -g sha256 -G symcipher -u key.pub -r key.priv -c primary.ctx
+create_object -g sha256 -G symcipher -u key.pub -r key.priv -c primary.ctx
 
 tpm2_load -Q -c primary.ctx -u key.pub -r key.priv -n key.name -C decrypt.ctx
 
--- a/test/system/test_tpm2_quote.sh
+++ b/test/system/test_tpm2_quote.sh
@@ -30,6 +30,9 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
+
+source test_helpers.sh
+
 alg_primary_obj=sha256
 alg_primary_key=rsa
 alg_create_obj=0x000B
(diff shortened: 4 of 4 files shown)


jiazhang0:
@martinezjavier This failure just hits the reference code documented in the page 116 (L414-421), TPM 2.0 spec rev 1.16.

martinezjavier:
@jiazhang0 1.16 is an old revision of the spec. Did you check what rev 01.38 says?

Also, can you please mention the section and specific document (i.e: what Part of the spec, since just saying TPM2.0 spec is quite general due being many documents).

AFAICS, it's absolutely valid for symmetric cipher objects to have the TPMA_OBJECT sign attribute SET. 
This is what says in **Table 31 - Definition of (UINT32) TPMA_OBJECT Bits** in spec doc **Trusted Platform Module Library Part 2: Structures Level 00 Revision 01.38 September 29, 2016**:

```
sign / encrypt

SET (1): For a symmetric cipher object, the private portion of the key may be
used to encrypt. For other objects, the private portion of the key may be used to
sign.
CLEAR (0): The private portion of the key may not be used to sign or encrypt.
```

Finally, this change breaks test_tpm2_encryptdecrypt.sh.

jiazhang0:
@martinezjavier You are right. Both decrypt and sign set is allowed. Here is the errata for spec 1.16:

```
2.7 Sign/decrypt attribute encoding
According to the current definition in Part 1, Table 24, a SymCipher object can be used to encrypt and decrypt data if the decrypt attribute is SET. This is incorrect. The current Table 24 is only valid for KeyedHash and asymmetric objects.

For a SymCipher object the sign attribute encodes encryption and the decrypt attribute encodes decryption. Thus, a SymCipher object can only be used for encryption if the sign attribute is SET. To allow encryption and decryption, both the sign and decrypt attribute need to be SET. This allows SymCipher objects to differentiate between encryption and decryption.

This issue affects the command TPM2_EncryptDecrypt() and the table in Part 2 that defines the allowed scheme values for SymCipher object .
```

So it is correct for a symcipher object with both sign and decrypt set. And 1.38 doesn't have the reference codes thinking of wrong combination if both sign and decrypt are set.

Therefore, it is possible that certain TPM 2.0 chips may refer to the wrong words and reference code to design the symcipher object inside TPM firmware. Also, the tpm2simulator has the same errata if both sign and decrypt set. 

I will make a workaround in test case to work out this errata.

jiazhang0:
@martinezjavier I created a new commit "https://github.com/01org/tpm2-tools/pull/546/commits/e3b46c82f1430e9502278b09705b4c7cbbe42cbd" to work around this sort of errata issue. Note that this commit is based on the commit which was approved at https://github.com/01org/tpm2-tools/pull/552

test/system/test_tpm2_create.sh
```diff
@@ -47,6 +47,17 @@ cleanup() {
...
 trap cleanup EXIT
 
+# Certain TPM 2.0 chip, e.g, Nationz Z32H320TC, and tpm2simulator referring
+# to spec 1.16 may have an errata, disallowing both decrypt and sign set for
+# a symcipher object, and in response to RC_ATTRIBUTES. In order to work
+# around it, we attempt to run tpm2_create again with sign bit clear.
+create() {
+    tpm2_create $@ 2>&1 |
+        grep -q 'Create Object Failed ! ErrorCode: 0x2c2' && {
+            tpm2_create -Q -A 0x20072 $@
```
williamcroberts: Why not use the friendly name attributes here?

test/system/test_tpm2_encryptdecrypt.sh
```diff
@@ -60,7 +60,14 @@ tpm2_takeownership -Q -c
 
 tpm2_createprimary -Q -A e -g sha1 -G rsa -C primary.ctx
 
-tpm2_create -Q -g sha256 -G symcipher -u key.pub -r key.priv -c primary.ctx
+# Certain TPM 2.0 chip, e.g, Nationz Z32H320TC, and tpm2simulator referring
+# to spec 1.16 may have an errata, disallowing both decrypt and sign set for
```
williamcroberts: Can you elaborate this reference, 1.16 is vague.

test/system/test_tpm2_quote.sh
```diff
@@ -77,7 +77,14 @@ tpm2_takeownership -c
...
 
-tpm2_create -Q -g $alg_create_obj -G $alg_create_key -u $file_quote_key_pub -r $file_quote_key_priv  -c $file_primary_key_ctx
+# Certain TPM 2.0 chip, e.g, Nationz Z32H320TC, and tpm2simulator referring
+# to spec 1.16 may have an errata, disallowing both decrypt and sign set for
+# a symcipher object, and in response to RC_ATTRIBUTES. In order to work
+# around it, we attempt to run tpm2_create again with sign bit clear.
+tpm2_create -g $alg_create_obj -G $alg_create_key -u $file_quote_key_pub -r $file_quote_key_priv -c $file_primary_key_ctx 2>&1 |
+    grep -q 'Create Object Failed ! ErrorCode: 0x2c2' && {
+        tpm2_create -g $alg_create_obj -G $alg_create_key -A 0x20072 -u $file_quote_key_pub -r $file_quote_key_priv -c $file_primary_key_ctx
+    }
```
williamcroberts: if your copying and pasting this issue arround, it looks like maybe a test_helpers.sh that can be sourced is prudent, and then you just call that common piece. This way, when error codes get cleaned up, were not changing it in N > 1 spots.

test/system/test_tpm2_create.sh
jiazhang0: @williamcroberts It is unsupported to use friendly naming scheme for the attributes option. The option 'A' for tpm2_createprimary supports.

jiazhang0:
@williamcroberts @martinezjavier Plz help to review this new PR. Thanks.

test/system/test_tpm2_createprimary.sh
```diff
@@ -53,6 +55,11 @@ cleanup
 # Keep the algorithm specifiers mixed to test friendly and raw
 # values.
 for gAlg in 0x04 sha256 0x0C; do
+    if [ x"`hash_alg_supported $gAlg`" == x"" ]; then
```
williamcroberts: you don't need to put the x in there when you use quoted strings.

test/system/test_helpers.sh
```diff
@@ -0,0 +1,63 @@
...
+
+    if [ "$alg" = "${alg//[^0-9a-fA-FxX]/}" ]; then
+        alg=${hash_alg_map_table["$alg"]}
+        [ -z "$alg" ] && return
+    fi
+
+    local t_alg
+    for t_alg in $algs_supported; do
+        if [ "$t_alg" = "$alg" ]; then
+            echo "$alg"
```
williamcroberts: Get rid of all the added echos. We only want output on error.

test/system/test_tpm2_createprimary.sh
```diff
@@ -53,6 +55,11 @@ cleanup
 # Keep the algorithm specifiers mixed to test friendly and raw
 # values.
 for gAlg in 0x04 sha256 0x0C; do
```
williamcroberts: Why not just change the function to give you all supported hash algorithms.... then you don't need this filter approach. You just loop over all supported hash algorithms

test/system/test_tpm2_createprimary.sh
```diff
@@ -53,6 +55,11 @@ cleanup
...
+    if [ x"`hash_alg_supported $gAlg`" == x"" ]; then
+        echo "Ignore testing the unsupported hash algorithm \"$gAlg\""
```
williamcroberts: get rid of superfluous non-error output, like this.

jiazhang0:
Hold this PR till https://github.com/01org/tpm2-tools/pull/552 is approved.

PR #549, opened 2017-10-10
test_tpm2_create.sh,test_tpm2_createprimary.sh: test sha256 instead o…

jiazhang0:
…f sha384

sha384 is not so common on a real TPM 2.0 chip so it is better to test
sha256 instead.

Otherwise, the following failure would happen.

ObjectAttribute: 0x00060072
ERROR:
Create Object Failed ! ErrorCode: 0x2c3

tpm2_create -c context.out -g $gAlg -G $GAlg -u key.pub -r key.priv on
line 58 failed: 1

### DIFF
--- a/test/system/test_tpm2_create.sh
+++ b/test/system/test_tpm2_create.sh
@@ -53,7 +53,7 @@ tpm2_createprimary -Q -A p -g sha1 -G rsa -C context.out
 
 # Keep the algorithm specifiers mixed to test friendly and raw
 # values.
-for gAlg in sha1 0x0B sha384; do
+for gAlg in sha1 0x0B sha256; do
     for GAlg in rsa 0x08 ecc 0x25; do
         tpm2_create -Q -c context.out -g $gAlg -G $GAlg -u key.pub -r key.priv
         cleanup keep_context
--- a/test/system/test_tpm2_createprimary.sh
+++ b/test/system/test_tpm2_createprimary.sh
@@ -52,7 +52,7 @@ cleanup
 
 # Keep the algorithm specifiers mixed to test friendly and raw
 # values.
-for gAlg in 0x04 sha256 0x0C; do
+for gAlg in sha1 0x04 sha256; do
     for GAlg in 0x01 keyedhash ecc 0x25; do
         for Atype in o e p n; do
             tpm2_createprimary -Q -A $Atype -g $gAlg -G $GAlg -C context.out


test/system/test_tpm2_create.sh
```diff
@@ -53,7 +53,7 @@ tpm2_createprimary -Q -A p -g sha1 -G rsa -C context.out
 
 # Keep the algorithm specifiers mixed to test friendly and raw
 # values.
-for gAlg in sha1 0x0B sha384; do
+for gAlg in sha1 0x0B sha256; do
```
martinezjavier: The test is already testing for sha256, is just using the hex value (0x0B) since both pretty names and hex values are supported. So your change is testing sha256 twice. So just remove sha384 and keep sha1 and 0x0B (we are using both pretty names and hex values on purpose to make sure that the parse logic isn't regressed).

Your 0x2c3 error code is indeed TPM_RC_HASH (hash algorithm not supported or not appropriate). So if this hash algorithm is not common on real TPMs, then I agree with you that the test could omit it.

In your commit message please mention issue #335 (Make sure that all tests work on a real TPM hardware) since this change also help on that regard.

test/system/test_tpm2_createprimary.sh
```diff
@@ -52,7 +52,7 @@ cleanup
 
 # Keep the algorithm specifiers mixed to test friendly and raw
 # values.
-for gAlg in 0x04 sha256 0x0C; do
+for gAlg in sha1 0x04 sha256; do
```
martinezjavier: Same comment than before, sha1 is already tested (0x04). So just remove 0x0C (sha384).

test/system/test_tpm2_create.sh
martinezjavier: A long term solution for this should be to query the TPM about the supported algorithms and only test those, i.e:

for gAlg in $(tpm2_dump_capability -c algorithms | grep sha | awk '/TPMA_ALGORITHM for ALG_ID:/{print $4}'); do

jiazhang0:
@martinezjavier Thanks for your comments. I have sent V2 PR. Plz review it.

martinezjavier:
@jiazhang0 you don't need to close the PR and open a new one, just change your commits and force a push to the same branch, i.e:

```
$ git commit -a --fixup=$SHA1
$ git rebase -i  $SHA1~1
$ git push -f $BRANCH
```

jiazhang0:
@martinezjavier Next time I will do it that way.

PR #550, opened 2017-10-11
nvread: fix missing 'f' in optstring

Fabo:
nvread is not understanding the "-f" parameter because the
option is missing from the optstring. Add it to the optstring.

Signed-off-by: Fabien Parent <fparent@baylibre.com>

### DIFF
--- a/tools/tpm2_nvread.c
+++ b/tools/tpm2_nvread.c
@@ -261,7 +261,7 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
         { "input-session-handle",1,          NULL, 'S' },
     };
 
-    *opts = tpm2_options_new("x:a:s:o:P:S:", ARRAY_LEN(topts), topts, on_option, NULL);
+    *opts = tpm2_options_new("x:a:f:s:o:P:S:", ARRAY_LEN(topts), topts, on_option, NULL);
 
     return *opts != NULL;
 }


tools/tpm2_nvread.c
```diff
@@ -261,7 +261,7 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
         { "input-session-handle",1,          NULL, 'S' },
```
martinezjavier: I think is valuable to always include a reference in the message to the commit that introduced the issue, i.e:

Commit 797d90f1e216 ("tpm2_nvread: Add "output" option (-f)") added an
option to define a file path to store the data read from the NV area.

But it didn't add the 'f' option character to the opstring and so isn't
usable. Add it to the optstring.

tools/tpm2_nvread.c
```diff
@@ -261,7 +261,7 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
...
     };
 
-    *opts = tpm2_options_new("x:a:s:o:P:S:", ARRAY_LEN(topts), topts, on_option, NULL);
+    *opts = tpm2_options_new("x:a:f:s:o:P:S:", ARRAY_LEN(topts), topts, on_option, NULL);
```
martinezjavier: Change looks good to me.

PR #553, opened 2017-10-12
test_tpm2_createpolicy.sh: fix the compare between PCR 0 and hex value

jiazhang0:
This fix resolves the following errors:

./test_tpm2_createpolicy.sh: line 57: [[:
9872964b9b40cdd0363fcd6af8c267c9cb34200b: value too great for base
(error token is "9872964b9b40cdd0363fcd6af8c267c9cb34200b")
Failure: Creating Policy Digest with PCR policy for index 0 and sha1
pcr index hash

### DIFF
--- a/test/system/test_tpm2_createpolicy.sh
+++ b/test/system/test_tpm2_createpolicy.sh
@@ -54,10 +54,10 @@ do
     # This script expects PCR index 0 to be equal to 0x03
     #
     value=`tpm2_pcrlist -L $halg:0 | grep PCR_00 | cut -d\: -f 2-`
-    if [[ $value -ne 0x03 ]]; then
+    if [[ 16#$value -ne 0x03 ]]; then
         echo "Expected PCR 0 \"$halg\" bank to be 0x3, found: \"$value\"" 1>&2
         exit 1
-	fi
+    fi
 
     tpm2_createpolicy -P -L $halg:0 -f policy.out
 


test/system/test_tpm2_createpolicy.sh
```diff
@@ -54,10 +54,10 @@ do
     # This script expects PCR index 0 to be equal to 0x03
     #
     value=`tpm2_pcrlist -L $halg:0 | grep PCR_00 | cut -d\: -f 2-`
-    if [[ $value -ne 0x03 ]]; then
+    if [[ 16#$value -ne 0x03 ]]; then
```
martinezjavier: According to the [bash manual](https://www.gnu.org/software/bash/manual/bashref.html#Shell-Arithmetic), constants with a leading 0 are interpreted as octal numbers.

So I guess this test only worked because the PCR0 for the SHA-1 bank is always "0000000000000000000000000000000000000003" and the octal 03 is equal to the hexadecimal 0x03 but it would fail for other values were this isn't true.

I agree with your change, although I would reword the commit message to something like:

```
If the base is omitted in an arithmetic expression bash assumes a base 10,
unless the constant starts with a 0 or 0x. In these cases, base 8 and 16
are used respectively.

So the test was only working because assumes a PCR0 value of 3, which has
the same value in both bases, but this isn't true for all values and will
lead to the following issue:

./test_tpm2_createpolicy.sh: line 57: [[:
9872964b9b40cdd0363fcd6af8c267c9cb34200b: value too great for base
(error token is "9872964b9b40cdd0363fcd6af8c267c9cb34200b")
Failure: Creating Policy Digest with PCR policy for index 0 and sha1
pcr index hash

Fix this by explicitly define a base 16 so bash know that the constant
value is an hexadecimal number.
```

Now, I think this is really a bug in the tpm2_pcrlist tool. All other tools prints hexadecimal values with a `0x` prefix, so we should to the same for tpm2_pcrlist. That would fix your particular issue, although I'm not against being more strict and also defining the base in this check as your patch does.

jiazhang0:
@martinezjavier I have updated the commit as you suggested. Thanks for your comments.

PR #555, opened 2017-10-13
test_tpm2_createpolicy.sh: use file to supply expected PCR value

### LINKED ISSUE #551: The tpm2_createpolicy test excepts a PCR0 value that's only true in the TPM simulator
It would be good if the test could be make more generic, this is also needed to solve issue #335.

webmeister:
For the simulator, PCR0 contains the value 0x03, but this is not necessarily the case for other TPM implementations. Therefore, use a file to supply the expected PCR value to tpm2_createpolicy, so that the expected policy digest remains unchanged and the test passes for all TPMs.

We tried a similar approach some time ago. I've now reimplemented it, but had no chance to test it, so I hope that this works :)

Fixes #551, #335.

### DIFF
--- a/test/system/test_tpm2_createpolicy.sh
+++ b/test/system/test_tpm2_createpolicy.sh
@@ -39,27 +39,23 @@ onerror() {
 trap onerror ERR
 
 cleanup() {
-    rm -f policy.out
+    rm -f pcr.in policy.out
 }
 trap cleanup EXIT
 
+declare -A digestlengths=(["sha1"]=20 ["sha256"]=32)
 declare -A expected_policy_digest=(["sha1"]="f28230c080bbe417141199e36d18978228d8948fc10a6a24921b9eba6bb1d988"
                                    ["sha256"]="33e36e786c878632494217c3f490e74ca0a3a122a8a4f3c5302500df3b32b3b8")
 
-for halg in sha1 sha256
+for halg in ${!digestlengths[@]}
 do
     cleanup
 
-    #
-    # This script expects PCR index 0 to be equal to 0x03
-    #
-    value=`tpm2_pcrlist -L $halg:0 | grep PCR_00 | cut -d\: -f 2-`
-    if [[ 16#$value -ne 0x03 ]]; then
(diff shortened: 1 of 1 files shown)


webmeister:
I've made the changes, but I'm not sure whether GitHub handles it correctly that I've opened a separate pull request (#556) for the /dev/null stuff (since more than this one file was affected). The commit seems now to be part of both pull requests.

martinezjavier:
Probably git is able to handle since the sha-1 is the same in both cases so merging with the rebase option will only merge the second commit.

Having said that, I think it's better to just include it in this PR and close #556.

test/system/test_tpm2_createpolicy.sh
```diff
@@ -50,16 +50,16 @@ for halg in sha1 sha256
...
-    if [[ $value -ne 0x03 ]]; then
-        echo "Expected PCR 0 \"$halg\" bank to be 0x3, found: \"$value\"" 1>&2
-        exit 1
-	fi
+    # Create file containing expected PCR value
+    if [ $halg = sha1 ]; then
+        digestlength=20
+    else
+        digestlength=32
+    fi
```
williamcroberts: this is brittle. I would look it up in the hash table and use wc -c to get the length.

webmeister: Not that brittle, since the possible values for $halg are defined just four lines above, but if you can tell me where to find this hash table I can change the algorithm.

martinezjavier: @webmeister maybe something like this would be nicer?

```
declare -A digestlengths=(["sha1"]=20 ["sha256"]=32)
...

for halg in ${!digestlengths[@]}
do
   ...
    head -c $((${digestlengths[$halg]} - 1)) /dev/zero > pcr.in
    ...
done
```

webmeister: Thanks, that keeps the values nicely together. I'll update the patch accordingly.

williamcroberts:
You changed the mode from 0755 --> 0644, which causes a permission denied. You'll need to fix that, otherwise this looks good.

webmeister:
Argh, sorry, this is what you get for editing files on Windows...

PR #559, opened 2017-10-14
conversion: fix build error with OpenSSL 1.1.x

martinezjavier:
Since OpenSSL version 1.1.0, the RSA structure (struct rsa_st) is opaque
to programs (https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes).

So the structure members can't be accessed directly and instead helpers
functions should be used.

Because the API changes, building tpm2-tools with OpenSSL 1.1.x leads to
the following compile error:

lib/conversion.c: In function ‘tpm2_convert_pubkey_ssl’:
lib/conversion.c:126:16: error: dereferencing pointer to incomplete type ‘RSA {aka struct rsa_st}’
     ssl_rsa_key->e = BN_bin2bn((void*)&exponent, sizeof(exponent), NULL);
                ^~
make: *** [Makefile:1842: lib/conversion.o] Error 1

### DIFF
--- a/lib/conversion.c
+++ b/lib/conversion.c
@@ -100,6 +100,7 @@ static bool tpm2_convert_pubkey_ssl(TPMT_PUBLIC *public, pubkey_format format, c
     bool ret = false;
     FILE *fp = NULL;
     RSA *ssl_rsa_key = NULL;
+    BIGNUM *e = NULL, *n = NULL;
 
     // need this before the first SSL call for getting human readable error
     // strings in print_ssl_error()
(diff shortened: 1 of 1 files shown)


martinezjavier:
@mgerstner could you please review this change?

lib/conversion.c
```diff
@@ -123,17 +124,22 @@ static bool tpm2_convert_pubkey_ssl(TPMT_PUBLIC *public, pubkey_format format, c
...
+    if (!n || !e) {
         print_ssl_error("Failed to convert data to SSL internal format");
         goto error;
     }
 
+#if OPENSSL_VERSION_NUMBER < 0x1010000fL /* OpenSSL 1.1.0 */
+    ssl_rsa_key->e = e;
+    ssl_rsa_key->n = n;
+#else
+    RSA_set0_key(ssl_rsa_key, n, e, NULL);
```
mgerstner: `RSA_set0_key` returns 0 on failure. That should be handled.

In that error case it would also be needed to call `BN_free` on `e` and `n`, because ownership has not been transferred to the `ssl_rsa_key` structure.

lib/conversion.c
```diff
@@ -123,17 +124,22 @@ static bool tpm2_convert_pubkey_ssl(TPMT_PUBLIC *public, pubkey_format format, c
...
-    ssl_rsa_key->n = BN_bin2bn(
-            (public->unique).rsa.t.buffer,
-            (public->unique).rsa.t.size,
-            NULL);
+    e = BN_bin2bn((void*)&exponent, sizeof(exponent), NULL);
+    n = BN_bin2bn(public->unique.rsa.t.buffer, public->unique.rsa.t.size,
+                  NULL);
 
-    if (!ssl_rsa_key->n || !ssl_rsa_key->e) {
+    if (!n || !e) {
```
mgerstner: When only one of these conversions fails then we'll leak the other one, because no `BN_free` is called on it.

This is tricky, because once the BIGNUMs are successfully stored in the `ssl_rsa_key` structure, `RSA_free` will take care of freeing that memory.

Maybe tracking the ownership of `e` and `n` can be done with an extra `bool` variable that triggers the appropriate `BN_free` in the error handling code at the end of the function.

williamcroberts: Maybe just NULL out n and e when the ownership transfer occurs.

martinezjavier: @mgerstner thanks a lot for pointing out these issues. I'm not really familiar with the OpenSSL lAPI so I missed these details.

@williamcroberts agreed that the cleaner option is to NULL these when are owned by the RSA struct. I'll re-spin the commit.

lib/conversion.c
martinezjavier: @mgerstner done in the next version.

lib/conversion.c
martinezjavier: @mgerstner @williamcroberts does it look Ok for you now?

mgerstner: It looks good to me, thank you.

PR #563, opened 2017-10-17
tpm2_import: fix build error with OpenSSL 1.1.0 and other issues

martinezjavier:
This pull request contains fixes for the tpm2_import tool. I noticed these while fixing a build error with OpenSSL 1.1.0.

@idesai could you please review these since you are the author of the tpm2_import tool.

@mgerstner I would really appreciate your review too, specially for the OpenSSL API usage.

### DIFF
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -117,6 +117,8 @@ static void ssl_RSA_set0_key(RSA *r, BIGNUM *n, BIGNUM *e, BIGNUM *d) {
 }
 
 static bool encrypt_seed_with_tpm2_rsa_public_key(void) {
+    bool rval = false;
+
     //Public Modulus
     FILE *fp = fopen(ctx.parent_key_public_file, "rb");
     if (fp == NULL) {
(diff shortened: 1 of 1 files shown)


tools/tpm2_import.c
```diff
@@ -176,7 +178,15 @@ static bool encrypt_seed_with_tpm2_rsa_public_key(void) {
         LOG_ERR("BN_bin2bn failed\n");
         goto error;
     }
+#if OPENSSL_VERSION_NUMBER < 0x1010000fL /* OpenSSL 1.1.0 */
```
mgerstner: The `ssl_RSA_set0_key` does currently not return an error code, but it does have an error return path. Should this path be taken for some reason then we'll end up without a valid `rsa` structure and we'll leak `n`.

You could make it a real drop-in replacement (i.e. rename `ssl_RSA_set0_key` to `RSA_set0_key` and add the return code handling). This would also get rid of the additional #if/#else preprocessor branch.

martinezjavier: @mgerstner indeed, I've created PR #564 with your suggested cleanup.

PR #567, opened 2017-10-19
lib: clean up potential memory leak

snits:
In case where sapi_ctx_init fails in sapi_init_from_options, free tcti_ctx.

This is specific to the 2.X branch.


snits:
Patch for 2.X cleaning up a problem our coverity scan detected. In the error path of sapi_init_from_options tcti_ctx was not being freed. Master no longer has this code from what I see, so doesn't apply there.

williamcroberts:
@snits thanks, the free coverity scan did not detect this.

PR #568, opened 2017-10-19
[V2] test_tpm2_import: work around TPM 2.0 spec 1.16 errata on decrypt and…

jiazhang0:
Changelog:

- V2:
  - Add space between if and (
  - Use if (ctx.flags.A) instead of if (ctx.flags.A == 1)
  - Add { } around ctx.import_key_public.t.publicArea.objectAttributes.val = ctx.objectAttibutes;
  - Add space between option character and }


jiazhang0:
@williamcroberts Here is a quick diff against V1:
index 0055131..742570c 100644
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -293,8 +293,9 @@ static bool create_import_key_public_data_and_name(void) {
 
-    if(ctx.flags.A == 1)
+    if (ctx.flags.A) {
         ctx.import_key_public.t.publicArea.objectAttributes.val = ctx.objectAttr
+    }
 
     tpm2_tool_output("ObjectAttribute: 0x%08X\n",
                      ctx.import_key_public.t.publicArea.objectAttributes.val);
@@ -582,12 +583,12 @@ static bool on_option(char key, char *value) {
 bool tpm2_tool_onstart(tpm2_options **opts) {
 
     const struct option topts[] = {
-      { "input-key-file",     required_argument, NULL, 'k'},
-      { "parent-key-handle",  required_argument, NULL, 'H'},
-      { "parent-key-public",  required_argument, NULL, 'f'},
-      { "import-key-private", required_argument, NULL, 'r'},
-      { "object-attributes", required_argument, NULL, 'A'},
+      { "input-key-file",     required_argument, NULL, 'k' },
+      { "parent-key-handle",  required_argument, NULL, 'H' },
+      { "parent-key-public",  required_argument, NULL, 'f' },
+      { "import-key-private", required_argument, NULL, 'r' },
+      { "import-key-public",  required_argument, NULL, 'q' },
+      { "object-attributes",  required_argument, NULL, 'A' },
     };
 
     setbuf(stdout, NULL);

williamcroberts:
Moved to: https://github.com/01org/tpm2-tools/pull/570

PR #570, opened 2017-10-21
test_tpm2_import: work around spec errata

williamcroberts:
TPM 2.0 spec 1.16 errata on decrypt and sign set for a symcipher.

Certain TPM 2.0 chip, e.g, Nationz Z32H320TC, and tpm2simulator
referring to TPM 2.0 spec 1.16 may have an errata, disallowing both
decrypt and sign set for a symcipher object, and in response to
RC_ATTRIBUTES.

In order to work around it, we attempt to run tpm2_importe again with
sign bit clear.

This change partially resolves the issue #335 (Make sure that all
tests work on a real TPM hardware) for helping on it regard.

### DIFF
--- a/test/system/test_helpers.sh
+++ b/test/system/test_helpers.sh
@@ -114,3 +114,13 @@ create_object() {
         true
     fi
 }
+
+import_object() {
+    local alg_import_obj=0x20040
+
+    if tpm2_import $@ 2>&1 | grep -q 'ERROR: Failed Key Import 000002C2'; then
+        tpm2_import -A $alg_import_obj $@
+    else
+        true
+    fi
+}
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -51,6 +51,7 @@
 #include "tpm_kdfa.h"
 #include "tpm2_options.h"
 #include "tpm2_util.h"
+#include "tpm2_tool.h"
 
 #define SYM_KEY_SIZE 16
 #define max_buffer_size  1024
@@ -83,6 +84,10 @@ struct tpm_import_ctx {
     TPM2B_DATA enc_sensitive_key;
     TPM2B_MAX_BUFFER encrypted_inner_integrity;
     TPM2B_MAX_BUFFER encrypted_duplicate_sensitive;
+    UINT32 objectAttributes;
+    struct {
+        UINT16 A : 1;
+    } flags;
 };
 
 static tpm_import_ctx ctx = { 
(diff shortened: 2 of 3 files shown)


test/system/test_helpers.sh
```diff
@@ -114,3 +114,13 @@ create_object() {
         true
     fi
 }
+
+import_object() {
+    local alg_import_obj=0x20040
+
+    if tpm2_import $@ 2>&1 | grep -q 'ERROR: Failed Key Import 000002C2'; then
```
martinezjavier: If this is an errata of 1.16, can we instead of assuming the TPM2_Import is failing due that, check if the TPM2 revision is 1.16 and CLEAR the sign and decrypt bit? This can be checked with something like following:
```
tpm2_dump_capability -c properties-fixed | grep TPM_PT_REVISION | cut -d ':' -f2 | tr -d '[:space:]'
```

Or extend tpm2_dump_capability to only print the revision.

tools/tpm2_import.c
```diff
@@ -564,17 +583,18 @@ static bool on_option(char key, char *value) {
...
-      { "input-key-file",     required_argument, NULL, 'k'},
-      { "parent-key-handle",  required_argument, NULL, 'H'},
-      { "parent-key-public",  required_argument, NULL, 'f'},
-      { "import-key-private", required_argument, NULL, 'r'},
-      { "import-key-public",  required_argument, NULL, 'q'},
+      { "input-key-file",     required_argument, NULL, 'k' },
+      { "parent-key-handle",  required_argument, NULL, 'H' },
+      { "parent-key-public",  required_argument, NULL, 'f' },
+      { "import-key-private", required_argument, NULL, 'r' },
+      { "import-key-public",  required_argument, NULL, 'q' },
```
martinezjavier: Please don't change the these entries, if you want to do some style cleanups do it as a separate commit.

tools/tpm2_import.c
```diff
@@ -564,17 +583,18 @@ static bool on_option(char key, char *value) {
...
+      { "object-attributes",  required_argument, NULL, 'A' },
```
martinezjavier: This new option has to be added to the man page.

williamcroberts:
FYI The check failing is likely due to tss. Im trouble shooting that now:

test/system/test_helpers.sh
williamcroberts: This tool needs a yaml output, that would make this way easier to parse in a programmatic way.

Something that's been building in my head is that each tool outputs a tree of output to tool_main and then tool main can format and output....

williamcroberts:
Ok the bug is in the tss between 3bc96032d47...3f1828ccb31

Ill set up a git bisection and find it.

jiazhang0:
@martinezjavier @williamcroberts Thanks for your comments and review.

I will implement a generic framework to handle errata for tpm2.0-tools these days, implying this patch and previous tpm2_create will be reworked under the new framework. If you agree, plz close this PR. I confirm the new framework would be better than current implementation.

tools/tpm2_import.c
jiazhang0: The above issues are addressed by https://github.com/01org/tpm2-tools/pull/577.

test/system/test_helpers.sh
jiazhang0: I implemented a generic framework (https://github.com/01org/tpm2-tools/pull/578) to handle TPM errata. Plz help to review it.

PR #573, opened 2017-10-23
Revert "[REVERTME] travis: get build working"

williamcroberts:
This reverts commit 03f51f5fd262de6b7deb459d5f1791a39efc8568.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,6 @@ install:
     - popd && popd
     - git clone https://github.com/01org/TPM2.0-TSS.git
     - pushd TPM2.0-TSS
-    - git checkout 65a40b4c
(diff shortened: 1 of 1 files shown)


williamcroberts:
Fixed in: https://github.com/01org/tpm2-tss/pull/590

martinezjavier:
@williamcroberts I would prefer to only pollute the repo with these kind of things of hacks if the issue on tpm2-tss isn't fixed in a sensible time frame, but I don't think that's worth it to have a hack + tpm2-tss + hack revert all in a single day.

williamcroberts:
@martinezjavier I don't either, but I gotta keep things moving. Eventually id like the travis stuff in its own repo

PR #577, opened 2017-10-24
[V2] tpm2_import: add -A option to specify the object attributes

jiazhang0:

### DIFF
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -50,6 +50,7 @@
 #include "files.h"
 #include "tpm_kdfa.h"
 #include "tpm2_options.h"
+#include "tpm2_tool.h"
 #include "tpm2_util.h"
 
 #define SYM_KEY_SIZE 16
@@ -83,6 +84,7 @@ struct tpm_import_ctx {
     TPM2B_DATA enc_sensitive_key;
     TPM2B_MAX_BUFFER encrypted_inner_integrity;
     TPM2B_MAX_BUFFER encrypted_duplicate_sensitive;
+    UINT32 objectAttributes;
 };
 
 static tpm_import_ctx ctx = { 
@@ -287,6 +289,13 @@ static bool calc_sensitive_unique_data(void) {
 static bool create_import_key_public_data_and_name(void) {
 
     IMPORT_KEY_SYM_PUBLIC_AREA(ctx.import_key_public)
+
+    if (ctx.objectAttributes) {
+        ctx.import_key_public.t.publicArea.objectAttributes.val = ctx.objectAttributes;
+    }
+
+    tpm2_tool_output("ObjectAttribute: 0x%08X\n",
+                     ctx.import_key_public.t.publicArea.objectAttributes.val);
 
     memcpy(ctx.import_key_public.t.publicArea.unique.sym.t.buffer,
             ctx.import_key_public_unique_data, SHA256_DIGEST_SIZE);
@@ -553,6 +562,12 @@ static bool on_option(char key, char *value) {
     case 'r':
         ctx.import_key_private_file = value;
         break;
+    case 'A':
+        if(!tpm2_util_string_to_uint32(value, &ctx.objectAttributes)) {
+            LOG_ERR("Invalid object attribute, got\"%s\"", value);
+            return false;
+        }
+        break;
     default:
         LOG_ERR("Invalid option");
         return false;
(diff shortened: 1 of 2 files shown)


tools/tpm2_import.c
```diff
@@ -51,6 +51,7 @@
 #include "tpm_kdfa.h"
 #include "tpm2_options.h"
 #include "tpm2_util.h"
+#include "tpm2_tool.h"
```
martinezjavier: Can you please keep the headers sorted alphabetically? So this should be after tpm2_options.h.

tools/tpm2_import.c
```diff
@@ -83,6 +84,10 @@ struct tpm_import_ctx {
     TPM2B_DATA enc_sensitive_key;
     TPM2B_MAX_BUFFER encrypted_inner_integrity;
     TPM2B_MAX_BUFFER encrypted_duplicate_sensitive;
+    UINT32 objectAttributes;
```
martinezjavier: Why do you need this variable? It isn't used AFAICT.

tools/tpm2_import.c
```diff
@@ -83,6 +84,10 @@ struct tpm_import_ctx {
...
+    struct {
+        UINT16 A : 1;
+    } flags;
```
martinezjavier: I don't think this flag is needed either, you can just check if ctx.objectAttributes != 0

tools/tpm2_import.c
```diff
@@ -288,6 +293,13 @@ static bool create_import_key_public_data_and_name(void) {
 
     IMPORT_KEY_SYM_PUBLIC_AREA(ctx.import_key_public)
 
+    if (ctx.flags.A) {
+        ctx.import_key_public.t.publicArea.objectAttributes.val = ctx.objectAttributes;
+    }
+
+    tpm2_tool_output("ObjectAttribute: 0x%08X\n",
+                     ctx.import_key_public.t.publicArea.objectAttributes.val);
+
```
martinezjavier: I think this should only be printed if ctx.objectAttributes was assigned a value.

tools/tpm2_import.c
jiazhang0: It is used to retrieve the value of -A option from commandline:

```
 +    case 'A':
 +        if(!tpm2_util_string_to_uint32(value, &ctx.objectAttributes)) {
 +            LOG_ERR("Invalid object attribute, got\"%s\"", value);
 +            return false;
 +        }
 +        ctx.flags.A = 1;
 +        break;
```

tools/tpm2_import.c
jiazhang0: This design refers to tpm2_create.c. Yes, if we have -A option only, it is not necessary to use  flags.

tools/tpm2_import.c
jiazhang0: This also refers to tpm2_create and tpm2_createprimary.

jiazhang0:
V2 Changelog:
- minor changes according to martinezjavier's comments.
- here is a quick diff against V1:
```
index ae83aa7..dfdd4ae 100644
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -50,8 +50,8 @@
 #include "files.h"
 #include "tpm_kdfa.h"
 #include "tpm2_options.h"
-#include "tpm2_util.h"
 #include "tpm2_tool.h"
+#include "tpm2_util.h"
 
 #define SYM_KEY_SIZE 16
 #define max_buffer_size  1024
@@ -85,9 +85,6 @@ struct tpm_import_ctx {
     TPM2B_MAX_BUFFER encrypted_inner_integrity;
     TPM2B_MAX_BUFFER encrypted_duplicate_sensitive;
     UINT32 objectAttributes;
-    struct {
-        UINT16 A : 1;
-    } flags;
 };
 
 static tpm_import_ctx ctx = { 
@@ -293,7 +290,7 @@ static bool create_import_key_public_data_and_name(void) {
 
-    if (ctx.flags.A) {
+    if (ctx.objectAttributes) {
         ctx.import_key_public.t.publicArea.objectAttributes.val = ctx.objectAttributes;
     }
 
@@ -570,7 +567,6 @@ static bool on_option(char key, char *value) {
             LOG_ERR("Invalid object attribute, got\"%s\"", value);
             return false;
         }
-        ctx.flags.A = 1;
         break;
     default:
         LOG_ERR("Invalid option");
```

tools/tpm2_import.c
martinezjavier: ah, sorry. I misread and thought you were adding both a struct tpm_import_ctx member and a global variable.

PR #578, opened 2017-10-24
[V4] tpm2_errata: implement a generic handling for errata

jiazhang0:
The TPM errata should be addressed and handled transparently by design.

In order to handle an errata, the errata info should be recorded in
tpm2_errata.[ch], and the affected tools need to query the errata
handler with a pre-defined errata number and to-be-fixed contents.

During the initialization, the TPM spec and errata info will be
remembered and used to check the errata query issued by the tools.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -122,6 +122,8 @@ lib_libcommon_a_SOURCES = \
     lib/rc-decode.h \
     lib/tpm2_alg_util.c \
     lib/tpm2_alg_util.h \
+    lib/tpm2_errata.c \
+    lib/tpm2_errata.h \
     lib/tpm2_header.h \
     lib/tpm2_nv_util.c \
     lib/tpm2_nv_util.h \
--- a/lib/tpm2_errata.c
+++ b/lib/tpm2_errata.c
@@ -1,0 +1,269 @@
+//**********************************************************************;
+// Copyright (c) 2017, Alibaba Group
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// 3. Neither the name of Intel Corporation nor the names of its contributors
+// may be used to endorse or promote products derived from this software without
+// specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+#include <stdarg.h>
+#include <stdbool.h>
+
+#include "log.h"
+#include "tpm2_errata.h"
+#include "tpm2_util.h"
+
+struct tpm2_errata_desc {
+    UINT32 spec_level;          /* spec level */
+    UINT32 spec_rev;            /* spec revision */
+    UINT32 errata_ver;          /* errata version */
+    void (*fixup)(va_list *ap); /* errata correction handler */
+    const char *name;           /* full section name in errata doc */
+};
+
+/*
+ * Published spec and errata information.
+ *
+ * Note that TPM_PT_YEAR and TPM_PT_DAY_OF_YEAR retrieved
+ * from capability query only have the values of the
+ * release date of the specification if the TPM does not
+ * implement an errata. So the spec info are also given.
+ */
+static struct tpm2_errata_info {
+    UINT32 spec_level;
+    UINT32 spec_rev;
+    UINT32 errata_ver;
+    UINT32 year;
+    UINT32 day_of_year;
+} known_errata_info[] = {
+    /* Specification Revision 1.16 October 30, 2014 */
+    { 00, 116, 000, 2014, 303 },
+    /* Errata Version 1.2 February 16, 2015 */
+    { 00, 116, 120, 2015,   4 },
+    /* Errata Version 1.3 June 16, 2015 */
+    { 00, 116, 130, 2015, 167 },
+    /* Errata Version 1.4 January 15, 2016 */
+    { 00, 116, 140, 2016,  15 },
+    /* Errata Version 1.5 September 21, 2016 */
+    { 00, 116, 150, 2016, 265 },
+    /* Specification Revision 1.38 September 29, 2016 */
+    { 00, 138, 000, 2016, 273 },
+    /* Errata Version 1.0 January 16, 2017 */
+    { 00, 138, 100, 2017,  16 },
+    /* Errata Version 1.1 March 2, 2017 */
+    { 00, 138, 110, 2017,  61 },
+};
+
+static struct tpm2_errata_info *this_errata_info;
+
+static void fixup_sign_decrypt_attribute_encoding(va_list *ap);
+
+/*
+ * Beware of that each record contains the first errata
+ * version with the corresponding correction. This rule
+ * allows errata_match() to function properly.
+ */
+static struct tpm2_errata_desc errata_desc_list[] = {
+    [SPEC_116_ERRATA_2_7] = {
+        .spec_level = 00,
+        .spec_rev = 116,
+        .errata_ver = 120,
+        .fixup = fixup_sign_decrypt_attribute_encoding,
+        .name = "Sign/decrypt attribute encoding",
+    },
+    /*
+     * Append the new errata descriptor here.
+     */
+};
+
+static bool errata_match(struct tpm2_errata_desc *errata);
+static struct tpm2_errata_desc *errata_query(tpm2_errata_index_t index);
+
+/*
+ * Request an errata correction.
+ * @index: the errata to be queried.
+ *
+ * This function requests an errata correction to work
+ * around a known issue well documented in errata doc.
+ * If the request is valid and known, the queried errata
+ * will be applied by the corresponding pre-defined errata
+ * correction handler. The fixup process is transparent to
+ * the callers so there is no return values. Any tools can
+ * call this function to apply an errata if necessary.
+ *
+ * Return value:
+ * N/A
+ */
+void tpm2_errata_fixup(tpm2_errata_index_t index, ...) {
+
+    struct tpm2_errata_desc *errata;
+
+    errata = errata_query(index);
+    if (!errata) {
+        return;
+    }
+
+    bool res = errata_match(errata);
+    if (res == false) {
+        return;
+    }
+
+    va_list ap;
+
+    va_start(ap, index);
+    errata->fixup(&ap);
+    va_end(ap);
+
+    LOG_INFO("Errata %s applied\n", errata->name);
+}
+
+/*
+ * Query the full name of an errata.
+ * @index: the errata to be queried.
+ *
+ * Return value:
+ * the full name of an errata, or NULL if the query fails.
+ */
+const char *tpm2_errata_name(tpm2_errata_index_t index) {
+
+    struct tpm2_errata_desc *errata;
+
+    errata = errata_query(index);
+    if (!errata) {
+        return NULL;
+    }
+
+    return errata->name;
+}
+
+/*
+ * Initialize errata subsystem.
+ * @sapi_ctx: SAPI context to be queried.
+ *
+ * Return value:
+ * return true if TPM_SPEC is known and valid, or false if opposite.
+ */
+bool tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx) {
+
+    TPMS_CAPABILITY_DATA capability_data;
+    TPMI_YES_NO more_data;
+    TSS2_RC rc;
+
+    rc = Tss2_Sys_GetCapability(sapi_ctx, NULL, TPM_CAP_TPM_PROPERTIES,
+                                PT_FIXED, MAX_TPM_PROPERTIES, &more_data,
+                                &capability_data, NULL);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOG_ERR("Failed to GetCapability: capability: 0x%x, property: 0x%x, "
+                "TSS2_RC: 0x%x\n", TPM_CAP_TPM_PROPERTIES, PT_FIXED, rc);
+        return false;
+    } else if (more_data == YES) {
+        LOG_WARN("More data to be queried: capability: 0x%x, property: "
+                 "0x%x\n", TPM_CAP_TPM_PROPERTIES, PT_FIXED);
+    }
+
+    /* Distinguish current spec level 0 */
+    UINT32 spec_level = -1;
+    UINT32 spec_rev = 0;
+    UINT32 day_of_year = 0;
+    UINT32 year = 0;
+    TPML_TAGGED_TPM_PROPERTY *properties = &capability_data.data.tpmProperties;
+    size_t i;
+    for (i = 0; i < properties->count; ++i) {
+        TPMS_TAGGED_PROPERTY *property = properties->tpmProperty + i;
+
+        if (property->property == TPM_PT_LEVEL) {
+            spec_level = property->value;
+        } else if (property->property == TPM_PT_REVISION) {
+            spec_rev = property->value;
+        } else if (property->property == TPM_PT_DAY_OF_YEAR) {
+            day_of_year = property->value;
+        } else if (property->property == TPM_PT_YEAR) {
+            year = property->value;
+            /* Short circuit because this is the last item we need */
+            break;
+        } else if (property->property > TPM_PT_YEAR) {
+            break;
+        }
+    }
+
+    if (!spec_rev || !day_of_year || !year) {
+        LOG_ERR("Invalid TPM_SPEC parameter\n");
+        return false;
+    }
+
+    /* Determine the TPM spec and errata */
+    for (i = 0; i < ARRAY_LEN(known_errata_info); ++i) {
+         if (known_errata_info[i].day_of_year == day_of_year &&
+             known_errata_info[i].year == year &&
+             known_errata_info[i].spec_rev == spec_rev &&
+             known_errata_info[i].spec_level == spec_level) {
+             this_errata_info = known_errata_info + i;
+             break;
+         }
+    }
+
+    if (!this_errata_info) {
+        LOG_ERR("Unknow TPM_SPEC. spec_level: %d, spec_rev: 0x%x, "
+                "year: %d, day_of_year: %d\n", spec_level, spec_rev,
+                year, day_of_year);
+        return false;
+    }
+
+    LOG_INFO("TPM_SPEC: spec level %d, spec rev %f, errata ver %f\n",
+             this_errata_info->spec_level,
+             (float)this_errata_info->spec_rev / 100,
+             (float)this_errata_info->errata_ver / 100);
+
+    return true;
+}
+
+static void fixup_sign_decrypt_attribute_encoding(va_list *ap) {
+
+    TPMA_OBJECT *attrs = va_arg(*ap, TPMA_OBJECT *);
+
+    attrs->sign = 0;
+}
+
+static bool errata_match(struct tpm2_errata_desc *errata) {
+
+    if (!this_errata_info) {
+        LOG_ERR("Unrecognized TPM_SPEC for errata check\n");
+        return false;
+    }
+
+    return errata->errata_ver > this_errata_info->errata_ver &&
+           errata->spec_rev >= this_errata_info->spec_rev &&
+           errata->spec_level == this_errata_info->spec_level;
+}
+
+static struct tpm2_errata_desc *errata_query(tpm2_errata_index_t index) {
+
+    if ((size_t)index >= ARRAY_LEN(errata_desc_list)) {
+        LOG_ERR("Invalid errata index queried: %u\n", (unsigned int)index);
+        return NULL;
+    }
+
+    return &errata_desc_list[index];
+}
--- a/lib/tpm2_errata.h
+++ b/lib/tpm2_errata.h
@@ -1,0 +1,84 @@
+//**********************************************************************;
+// Copyright (c) 2017, Alibaba Group
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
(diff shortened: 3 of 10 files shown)


jiazhang0:
Here is a quick diff showing how tpm2_import queries the errata handler:

index ae83aa7..744c6e4 100644
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -52,6 +52,7 @@
 #include "tpm2_options.h"
 #include "tpm2_util.h"
 #include "tpm2_tool.h"
+#include "tpm2_errata.h"
 
 #define SYM_KEY_SIZE 16
 #define max_buffer_size  1024
@@ -293,6 +294,11 @@ static bool create_import_key_public_data_and_name(void) {
 
+    if (tpm2_errata_fixup(SPEC_116_ERRATA_2_7,
+                          &ctx.import_key_public.t.publicArea.objectAttributes))
+        tpm2_tool_output("Errata %s applied\n",
+                         tpm2_errata_name(SPEC_116_ERRATA_2_7));
+
     if (ctx.flags.A) {
         ctx.import_key_public.t.publicArea.objectAttributes.val = ctx.objectAttr
     }

Makefile.am
```diff
@@ -140,7 +140,9 @@ lib_libcommon_a_SOURCES = \
     lib/tpm2_options.c \
     lib/tpm2_options.h \
     lib/tpm_session.c \
-    lib/tpm_session.h
+    lib/tpm_session.h \
+    lib/tpm2_errata.c \
+    lib/tpm2_errata.h
```
martinezjavier: Can you please keep this sorted alphabetically?

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+#include <stdbool.h>
+#include <stdarg.h>
+
+#include "tpm2_errata.h"
+#include "log.h"
+#include "tpm2_util.h"
```
martinezjavier: And also sort these.

jiazhang0:
Here is the quick diff against V1:
```
diff --git a/Makefile.am b/Makefile.am
index 484318e..3585e71 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -122,6 +122,8 @@ lib_libcommon_a_SOURCES = \
     lib/rc-decode.h \
     lib/tpm2_alg_util.c \
     lib/tpm2_alg_util.h \
+    lib/tpm2_errata.c \
+    lib/tpm2_errata.h \
     lib/tpm2_header.h \
     lib/tpm2_nv_util.c \
     lib/tpm2_nv_util.h \
@@ -140,9 +142,7 @@ lib_libcommon_a_SOURCES = \
     lib/tpm2_options.c \
     lib/tpm2_options.h \
     lib/tpm_session.c \
-    lib/tpm_session.h \
-    lib/tpm2_errata.c \
-    lib/tpm2_errata.h
+    lib/tpm_session.h
 
 TOOL_SRC := tools/tpm2_tool.c tools/tpm2_tool.h
 
diff --git a/lib/tpm2_errata.c b/lib/tpm2_errata.c
index f6bc8c2..ce1de39 100644
--- a/lib/tpm2_errata.c
+++ b/lib/tpm2_errata.c
@@ -28,11 +28,11 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 // THE POSSIBILITY OF SUCH DAMAGE.
 //**********************************************************************;
-#include <stdbool.h>
 #include <stdarg.h>
+#include <stdbool.h>
 
-#include "tpm2_errata.h"
 #include "log.h"
+#include "tpm2_errata.h"
 #include "tpm2_util.h"
 
 struct tpm2_errata_desc {

```

martinezjavier:
@jiazhang0 thanks for the re-spin, I didn't finish reviewing the code yet though :smiley:

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+//**********************************************************************;
+#include <stdarg.h>
+#include <stdbool.h>
+
+#include "log.h"
+#include "tpm2_errata.h"
+#include "tpm2_util.h"
+
+struct tpm2_errata_desc {
+    UINT32 spec_rev;
```
martinezjavier: Do you think that makes sense to also have a member for the spec "Level" (TPM_PT_LEVEL) ?

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+
+static bool errata_match(struct tpm2_errata_desc *errata) {
+
+    return errata->spec_rev >= this_errata_info->spec_rev &&
+           errata->errata_ver > this_errata_info->errata_ver;
+}
+
+static bool errata_query(int nr, struct tpm2_errata_desc **errata) {
+
+    if (nr < 0 || (size_t)nr >= sizeof(errata_desc_list) / sizeof(*errata)) {
```
martinezjavier: There's an ARRAY_LEN() macro in lib/tpm2_util.h

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+        TPMS_TAGGED_PROPERTY *property = properties->tpmProperty + i;
+
+        if (property->property == TPM_PT_REVISION)
+            spec_rev = property->value;
+        else if (property->property == TPM_PT_DAY_OF_YEAR)
+            day_of_year = property->value;
+        else if (property->property == TPM_PT_YEAR)
+            year = property->value;
+        else if (property->property > TPM_PT_YEAR)
+            break;
```
martinezjavier: Probably is worth adding a comment that TPM_PT_YEAR and TPM_PT_DAY_OF_YEAR only have the values of the release date of the specification if the TPM does not implement an errata, but that the values indicate the release date of the errata document if the TPM implements the errata.

This wasn't evident to me and I had to read the TPM library part 2 to figure out.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+    /* Errata Version 1.1 March 2, 2017 */
+    { 138, 110, 2017,  61 },
+}, *this_errata_info;
+
+static void
+fixup_sign_decrypt_attribute_encoding(struct tpm2_errata_desc *errata,
+                                      va_list *ap) {
+
+    TPMA_OBJECT *attrs = va_arg(*ap, TPMA_OBJECT *);
+printf("applied\n");
```
williamcroberts: get rid of this, if you want any verbose information their is macros in log.h like the one you used below.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+    va_end(ap);
+
+    return true;
+}
+
+const char *tpm2_errata_name(int nr) {
+
+    struct tpm2_errata_desc *errata;
+
+    if (errata_query(nr, &errata) == false)
```
williamcroberts: coding standard don't hide routines in conditionals.
```
bool res = errata_query()
if (!res) {
  //error
}
```

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+}
+
+void tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx) {
+
+    TPMS_CAPABILITY_DATA      capability_data;
+    TPML_TAGGED_TPM_PROPERTY *properties;
+    TSS2_RC                   rc;
+    TPMI_YES_NO               more_data;
+    size_t                    i;
+    UINT32                    spec_rev, day_of_year, year;
```
williamcroberts: Coding standard: declare variables at use/nearest to use, unless its something used throughout the whole routine like something that can be referenced in a goto block.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+
+    rc = Tss2_Sys_GetCapability(sapi_ctx,
```
williamcroberts: ```TSS2_RC rc = Tss2_Sys_GetCapability()```

lib/tpm2_errata.c
williamcroberts: Also, no multiple type declarations of variables, ie one line for one variable.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+                                &capability_data,
+                                NULL);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOG_ERR("Failed to GetCapability: capability: 0x%x, property: 0x%x, "
+                "TSS2_RC: 0x%x\n", TPM_CAP_TPM_PROPERTIES, PT_FIXED, rc);
+        return;
+    } else if (more_data == YES)
+        LOG_WARN("More data to be queried: capability: 0x%x, property: "
+                 "0x%x\n", TPM_CAP_TPM_PROPERTIES, PT_FIXED);
+
```
williamcroberts: brackets around everything

tools/tpm2_tool.c
```diff
@@ -134,6 +135,8 @@ int main(int argc, char *argv[], char *envp[]) {
     /* TODO SAPI INIT */
     TSS2_SYS_CONTEXT *sapi_context = sapi_ctx_init(tcti);
 
+    tpm2_errata_init(sapi_context);
```
martinezjavier: Maybe this function should be called by sapi_ctx_init()?

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+    if (errata_query(nr, &errata) == false)
+        return false;
+
+    va_start(ap, nr);
+    errata->fixup(errata, &ap);
+    va_end(ap);
+
+    return true;
+}
```
martinezjavier: Can you please add the commits that adds the errata queries for the tpm2 tools? I think that would make more easier to understand how these functions will be used.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+    .spec_rev = 116,
+    .errata_ver = 120,
+    .fixup = fixup_sign_decrypt_attribute_encoding,
+    .name = "Sign/decrypt attribute encoding",
+};
+
+/* Add the new errata descriptor here */
+
+static struct tpm2_errata_desc *errata_desc_list[] = {
+    [SPEC_116_ERRATA_2_7] = &sign_decrypt_attribute_encoding,
```
williamcroberts: Does this really need the pointer indirection... cant you just declare everything inline.

[SPEC_116_ERRATA] = { 116, 120, fixup_sign_decrypt_attribute_encoding, "Sign/decrypt attribute encoding" }

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+    { 116, 140, 2016,  15 },
+    /* Errata Version 1.5 September 21, 2016 */
+    { 116, 150, 2016, 265 },
+    /* Specification Revision 1.38 September 29, 2016 */
+    { 138, 000, 2016, 273 },
+    /* Errata Version 1.0 January 16, 2017 */
+    { 138, 100, 2017,  16 },
+    /* Errata Version 1.1 March 2, 2017 */
+    { 138, 110, 2017,  61 },
+}, *this_errata_info;
```
williamcroberts: Declare this separately. 
Also, wouldn't we want this all to be static, I don't see any point in having it extern.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,212 @@
...
+    if (!this_errata_info) {
+        LOG_ERR("Unknow TPM_SPEC. spec_rev: 0x%x, year: %d, day_of_year: %d\n",
+                spec_rev, year, day_of_year);
+        return;
+    }
+
+    LOG_INFO("TPM_SPEC: spec rev %f, errata ver %f\n",
+             (float)this_errata_info->spec_rev / 100,
+             (float)this_errata_info->errata_ver / 100);
+}
```
williamcroberts: Failures in this routine are fatal at errata lookup, which fails if initialization is not performed... Should looking up the tpm version fail in the tools?

lib/tpm2_errata.h
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Note that it is unnecessary to describe errata version because the
+ * section number should be kept consistent accoss all errata versions
+ * for a specific spec version.
+ */
+enum {
+    SPEC_116_ERRATA_2_7,
+};
+
+void tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx);
```
williamcroberts: doxygen style comments for all extern'd routines.

tools/tpm2_tool.c
williamcroberts: A failure here should minimally print a warning message that errata application cannot be applied. The fix routine on this un-initialized state should warn what errata fixup was not applied and return success. Ie try and continue in the face of errors until the tpm gives us a hard stop.

lib/tpm2_errata.h
```diff
@@ -0,0 +1,53 @@
...
+bool tpm2_errata_fixup(int nr, ...);
```
williamcroberts: No one uses this?

williamcroberts: Also, you define enum's for this but pass int. Use the actual enum type where callers should pass enums, it makes it much simpler to understand your interface.

williamcroberts:
Overall I like the direction, but please review the coding style document: https://github.com/01org/tpm2-tools/blob/master/misc/coding_standard_c.txt

There is even an eclipse CDT format definition file: https://github.com/01org/tpm2-tools/blob/master/misc/formatters/tpm2.0-eclipse-cdt-formatter.xml

1. You need unit tests on this, that it generally performs as your expecting
2. You need an in-tree user of the fixup. If it's being called, I missed that, I didn't
    see any callers of tpm2_errata_fixup
3. Follow the coding standard. Yes I know tools is a mixed bag at the moment, but I am trying to 
    minimize adding code that deviates from it.

tools/tpm2_tool.c
jiazhang0: The semantics of sapi init is simple and atomic, so this is why I place errata_int outside it. But I don't mind moving it to sapi init.

lib/tpm2_errata.c
jiazhang0: No problem. Actually I have planed to do it soon. That tool would give the full errata list for a running TPM.

lib/tpm2_errata.h
jiazhang0: I will submit a patch which shows how to use tpm2_errata_fixup(). The top comment gives a preview of that patch.

tools/tpm2_tool.c
jiazhang0: If tpm2_errata_init fails, this should be a critical issue, because one of items in TPM_PT_LEVEL, TPM_PT_REVISION, TPM_PT_DAY_OF_YEAR or TPM_PT_YEAR is incorrect in hardware level. But maybe such a tpm can still work as expected. So I would prefer to prompt a warning for this failure.

For tpm2_errata_name and tpm2_errata_fixup, returning false means the requesting errata cannot be applied due to some reasons, one of which includes the same reason where tpm2_errata_init fails.

Therefore, I prefer tpm2_errata_fixup without return value, and let tpm2_errata_fixup print the "Errata %s applied".

jiazhang0:
@martinezjavier @williamcroberts 

V3 changelog:
- Employ ARRAY_LEN() for walking through errata_desc_list and known_errata_info
- Add and use TPM_PT_LEVEL for errata search
- Comment improvement
- Separately declare this_errata_info 
- Declare errata record inside errata_desc_list[]
- Coding style corrections
- Take out errata_match from errata_query
- Standardize the errata index as tpm2_errata_index_t
- https://github.com/01org/tpm2-tools/pull/579 uses this subsystem to handle errata fixup

lib/tpm2_errata.c
```diff
@@ -0,0 +1,269 @@
...
+    /* Errata Version 1.1 March 2, 2017 */
+    { 00, 138, 110, 2017,  61 },
+};
+
+static struct tpm2_errata_info *this_errata_info;
+
+static void fixup_sign_decrypt_attribute_encoding(va_list *ap);
+
+/*
+ * Beaware of that each record contains the first errata
```
williamcroberts: spelling: Beware

lib/tpm2_errata.c
```diff
@@ -0,0 +1,269 @@
...
+ *
+ * Return values:
+ * N/A
+ */
+void tpm2_errata_fixup(tpm2_errata_index_t index, ...) {
+
+    struct tpm2_errata_desc *errata;
+
+    errata = errata_query(index);
+    if (!errata)
```
williamcroberts: brackets on everything, even single line if statements. Here and throughout.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,269 @@
...
+
+    LOG_INFO("Errata %s applied\n", errata->name);
+}
+
+/*
+ * Query the full name of an errata.
+ * @index: the errata to be queried.
+ *
+ * Return the full name of an errata, or NULL if the query fails.
+ */
```
williamcroberts: These doxygen comments should be in the header, here and throughout.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,269 @@
...
+    UINT32 spec_level;
+    UINT32 spec_rev;
+    UINT32 day_of_year;
+    UINT32 year;
+    TPML_TAGGED_TPM_PROPERTY *properties;
+    size_t i;
+
+    /* distinguish current spec level 0 */
+    spec_level = -1;
+    spec_rev = day_of_year = year = 0;
```
williamcroberts: assign at declaration.

lib/tpm2_errata.c
```diff
@@ -0,0 +1,269 @@
...
+        LOG_WARN("More data to be queried: capability: 0x%x, property: "
+                 "0x%x\n", TPM_CAP_TPM_PROPERTIES, PT_FIXED);
+    }
+
+    UINT32 spec_level;
+    UINT32 spec_rev;
+    UINT32 day_of_year;
+    UINT32 year;
+    TPML_TAGGED_TPM_PROPERTY *properties;
+    size_t i;
```
williamcroberts: Should be just above the for loop. Remember: "Closest Usage"

lib/tpm2_errata.c
```diff
@@ -0,0 +1,269 @@
...
+
+    /* distinguish current spec level 0 */
+    spec_level = -1;
+    spec_rev = day_of_year = year = 0;
+    properties = &capability_data.data.tpmProperties;
```
williamcroberts: ```TPML_TAGGED_TPM_PROPERTY *properties =  &capability_data.data.tpmProperties;```

lib/tpm2_errata.h
```diff
@@ -0,0 +1,53 @@
...
+ * because the section number should be kept consistent
+ * accoss all errata versions for a specific spec revision.
+ */
+typedef enum {
+    SPEC_116_ERRATA_2_7,
+} tpm2_errata_index_t;
+
+bool tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx);
+void tpm2_errata_fixup(tpm2_errata_index_t index, ...);
+const char *tpm2_errata_name(tpm2_errata_index_t index);
```
williamcroberts: These should have the doxygen comments for the routines.

jiazhang0:
Appreciate your patient review! If this is the last PR rev, I will then submit a new tool called tpm2_flushcontext and then the errata tool tpm2_errata. The former was written and tested for a while.

Here is a quick diff against V4:
```
diff --git a/lib/tpm2_errata.c b/lib/tpm2_errata.c
index 7c85968..0a2acff 100644
--- a/lib/tpm2_errata.c
+++ b/lib/tpm2_errata.c
@@ -81,7 +81,7 @@ static struct tpm2_errata_info *this_errata_info;
 static void fixup_sign_decrypt_attribute_encoding(va_list *ap);
 
 /*
- * Beaware of that each record contains the first errata
+ * Beware of that each record contains the first errata
  * version with the corresponding correction. This rule
  * allows errata_match() to function properly.
  */
@@ -113,7 +113,7 @@ static struct tpm2_errata_desc *errata_query(tpm2_errata_index_t index);
  * the callers so there is no return values. Any tools can
  * call this function to apply an errata if necessary.
  *
- * Return values:
+ * Return value:
  * N/A
  */
 void tpm2_errata_fixup(tpm2_errata_index_t index, ...) {
@@ -121,12 +121,14 @@ void tpm2_errata_fixup(tpm2_errata_index_t index, ...) {
     struct tpm2_errata_desc *errata;
 
     errata = errata_query(index);
-    if (!errata)
+    if (!errata) {
         return;
+    }
 
     bool res = errata_match(errata);
-    if (res == false)
+    if (res == false) {
         return;
+    }
 
     va_list ap;
 
@@ -141,15 +143,17 @@ void tpm2_errata_fixup(tpm2_errata_index_t index, ...) {
  * Query the full name of an errata.
  * @index: the errata to be queried.
  *
- * Return the full name of an errata, or NULL if the query fails.
+ * Return value:
+ * the full name of an errata, or NULL if the query fails.
  */
 const char *tpm2_errata_name(tpm2_errata_index_t index) {
 
     struct tpm2_errata_desc *errata;
 
     errata = errata_query(index);
-    if (!errata)
+    if (!errata) {
         return NULL;
+    }
 
     return errata->name;
 }
@@ -158,7 +162,8 @@ const char *tpm2_errata_name(tpm2_errata_index_t index) {
  * Initialize errata subsystem.
  * @sapi_ctx: SAPI context to be queried.
  *
- * Return true if TPM_SPEC is known and valid, or false if opposite.
+ * Return value:
+ * return true if TPM_SPEC is known and valid, or false if opposite.
  */
 bool tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx) {
 
@@ -178,18 +183,13 @@ bool tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx) {
                  "0x%x\n", TPM_CAP_TPM_PROPERTIES, PT_FIXED);
     }
 
-    UINT32 spec_level;
-    UINT32 spec_rev;
-    UINT32 day_of_year;
-    UINT32 year;
-    TPML_TAGGED_TPM_PROPERTY *properties;
+    /* Distinguish current spec level 0 */
+    UINT32 spec_level = -1;
+    UINT32 spec_rev = 0;
+    UINT32 day_of_year = 0;
+    UINT32 year = 0;
+    TPML_TAGGED_TPM_PROPERTY *properties = &capability_data.data.tpmProperties;
     size_t i;
-
-    /* distinguish current spec level 0 */
-    spec_level = -1;
-    spec_rev = day_of_year = year = 0;
-    properties = &capability_data.data.tpmProperties;
-
     for (i = 0; i < properties->count; ++i) {
         TPMS_TAGGED_PROPERTY *property = properties->tpmProperty + i;
 
@@ -201,7 +201,7 @@ bool tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx) {
             day_of_year = property->value;
         } else if (property->property == TPM_PT_YEAR) {
             year = property->value;
-            /* short circuit because this is the last item we need */
+            /* Short circuit because this is the last item we need */
             break;
         } else if (property->property > TPM_PT_YEAR) {
             break;
diff --git a/lib/tpm2_errata.h b/lib/tpm2_errata.h
index 57946d8..3b0af2f 100644
--- a/lib/tpm2_errata.h
+++ b/lib/tpm2_errata.h
@@ -46,8 +46,39 @@ typedef enum {
     SPEC_116_ERRATA_2_7,
 } tpm2_errata_index_t;
 
+/*
+ * Initialize errata subsystem.
+ * @sapi_ctx: SAPI context to be queried.
+ *
+ * Return value:
+ * return true if TPM_SPEC is known and valid, or false if opposite.
+ */
 bool tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx);
+
+/*
+ * Request an errata correction.
+ * @index: the errata to be queried.
+ *
+ * This function requests an errata correction to work
+ * around a known issue well documented in errata doc.
+ * If the request is valid and known, the queried errata
+ * will be applied by the corresponding pre-defined errata
+ * correction handler. The fixup process is transparent to
+ * the callers so there is no return values. Any tools can
+ * call this function to apply an errata if necessary.
+ *
+ * Return value:
+ * N/A
+ */
 void tpm2_errata_fixup(tpm2_errata_index_t index, ...);
+
+/*
+ * Query the full name of an errata.
+ * @index: the errata to be queried.
+ *
+ * Return value:
+ * the full name of an errata, or NULL if the query fails.
+ */
 const char *tpm2_errata_name(tpm2_errata_index_t index);
 
 #endif /* TPM2_ERRATA_H */

```

martinezjavier:
@jiazhang0 cool news about tpm2_flushcontext, I had writing one in my TODO for a while but never found the time.

lib/tpm2_errata.h
```diff
@@ -0,0 +1,84 @@
...
+/*
+ * Initialize errata subsystem.
+ * @sapi_ctx: SAPI context to be queried.
+ *
+ * Return value:
+ * return true if TPM_SPEC is known and valid, or false if opposite.
+ */
+bool tpm2_errata_init(TSS2_SYS_CONTEXT *sapi_ctx);
+
+/*
```
williamcroberts: doxygen has /** for the starter. This is close enough. Thanks.

PR #584, opened 2017-10-27
test: don't extend a PCR with more digest values than supported banks

martinezjavier:
The list of digest values used to extend a PCR are defined by passing a
TPML_DIGEST_VALUES to TPM2_PCR_Extend. By passing a list of values, the
PCR can be extended in all the supported banks on a single TPM2 command.

Since there's no check if the algorithm ID is different for each digest
used, different digest values could be passed for the same PCR bank so a
PCR will be extended multiple times for the same hash algorithm bank.

But this isn't the preferred way to use it, as explained in the TCG spec:

"While the semantics of this command allow multiple extends to a single
PCR bank, this is not the preferred use and the limit on the number of
entries in the list make this use somewhat impractical."

Also, the number of banks will vary depending of the hash algorithms that
are supported by each TPM2 chip, so the test shouldn't make assumption on
the maximum number of digests that could be used. Instead, it should only
attempt to extend the PCR for the banks that are known to be supported.

The test though, assumes that 3 hash digests could be passed since that's
what supported by the TPM2 simulator (sha1, sha256 and sha384). So change
the test to make sure that it will work on all TPM2 chips.

And also change the test to not attempt to extend the same PCR multiple
times for the same bank since the spec says that's not the preferred way
to use the TPM2_PCR_Extend command.

Finally, the test is checking that attempting to extend a PCR with more
digest values than supported banks will fail but there's no check for the
command exit status to make sure that it failed.

### DIFF
--- a/test/system/test_tpm2_pcrextend.sh
+++ b/test/system/test_tpm2_pcrextend.sh
@@ -45,6 +45,7 @@ onerror() {
 }
 trap onerror ERR
 
+digests=""
 # test a single algorithm based on what is supported
 for alg in `tpm2_pcrlist -s | cut -d\  -f 3-`; do
   alg=`echo $alg | cut -d\( -f 1-1`;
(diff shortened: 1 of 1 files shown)


test/system/test_tpm2_pcrextend.sh
```diff
@@ -45,6 +45,7 @@ onerror() {
 }
 trap onerror ERR
 
+digests=""
 # test a single algorithm based on what is supported
 for alg in `tpm2_pcrlist -s | cut -d\  -f 3-`; do
```
williamcroberts: Maybe in a different patch at some date, we can use test_helpers.sh that can be sourced. Looking at this maybe using hash_alg_supported function would be better.

test/system/test_tpm2_pcrextend.sh
```diff
@@ -55,26 +56,33 @@ for alg in `tpm2_pcrlist -s | cut -d\  -f 3-`; do
 
   hash=${alg_hashes[$alg]}
 
+  if [ ! -z  $digests ]; then
+      digests="$digests,"
+  fi
+
+  digests="$digests$alg=$hash"
```
williamcroberts: I think this test was missing testing individual hash specs, I think we would probably want to test that a single hash spec works, maybe test them individually here while building a larger digest spec list.

martinezjavier: @williamcroberts individual hash specs are tested, since the next line in the shell script does `tpm2_pcrextend 9:$alg=$hash`, or what did you mean?

williamcroberts: Wow that was right in front of me, I missed it. Nothing to do here then.

PR #588, opened 2017-10-30
[V0] tpm2_getcap: correct option names

jiazhang0:
handles-hmac-session and handles-policy-session should be replaced by
handles-loaded-session and handles-saved-session. According to the
spec, TPM_HT_LOADED_SESSION and TPM_HT_HMAC_SESSION have the same
value, as do TPM_HT_SAVED_SESSION and TPM_HT_POLICY_SESSION, but they
are used in different context.

### DIFF
--- a/man/tpm2_getcap.1.md
+++ b/man/tpm2_getcap.1.md
@@ -52,11 +52,11 @@ command.
     * handles-nv-index:
       Display handles about NV Indices.
 
-    * handles-hmac-session:
-      Display handles about HMAC sessions.
+    * handles-loaded-session:
+      Display handles about both loaded HMAC and policy sessions.
 
-    * handles-policy-session:
-      Display handles about policy sessions.
+    * handles-saved-session:
+      Display handles about saved sessions.
 
 [common options](common/options.md)
 
--- a/tools/tpm2_getcap.c
+++ b/tools/tpm2_getcap.c
@@ -116,15 +116,15 @@ capability_map_entry_t capability_map[] = {
         .count             = MAX_CAP_HANDLES,
     },
     {
-        .capability_string = "handles-hmac-session",
+        .capability_string = "handles-loaded-session",
         .capability        = TPM_CAP_HANDLES,
-        .property          = HMAC_SESSION_FIRST,
+        .property          = LOADED_SESSION_FIRST,
         .count             = MAX_CAP_HANDLES,
     },
     {
-        .capability_string = "handles-policy-session",
+        .capability_string = "handles-saved-session",
         .capability        = TPM_CAP_HANDLES,
-        .property          = POLICY_SESSION_FIRST,
+        .property          = ACTIVE_SESSION_FIRST,
         .count             = MAX_CAP_HANDLES,
     },
 };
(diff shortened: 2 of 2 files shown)


jiazhang0:
I have the questions:
- Why the spec doesn't define HR_LOADED_SESSION and TPM_SAVED_SESSION?
- Why there is no SAVED_SESSION_FIRST for the property.

I see ACTIVE_SESSION_FIRST and ACTIVE_SESSION_LAST are mentioned that they are used in TPM2_GetCapability command but there is no definition HR_ACTIVE_SESSION in spec. I'm confused here. I think we should have a consistent name scheme for *_FIRST|LAST, HR_* and TPM_HT_* in spec. Currently, the * may be LOADED_SESSION, SAVED_SESSION and ACTIVE_SESSION.

williamcroberts:
When it comes to spec questions, @flihp or @idesai might know.

williamcroberts:
@jiazhang0 Can you comment on which part of the spec, everything else looks good.

jiazhang0:
@williamcroberts I'm considering which names are best for option names.

According to the spec rev 1.38 part 1 (PDF page 100), the requesting properties could be either TPM_HT_LOADED_SESSION (0x02) or TPM_HT_SAVED_SESSION (0x03).

```
When TPM2_GetCapability() is used to obtain a list of sessions that are active but not on the TPM, the caller would use a handle with an MSO of 0316 which normally would reference a policy session. The TPM will respond with a list of session handles that are in use, but not on the TPM. Since the TPM does not keep a record of whether the saved session context was an HMAC or policy session, all of the handles in the list will have an MSO of 0216.

The TPM is required to maintain a list of all, currently assigned session handles as well as the correct "version number" for any saved session contexts.
```

The returned handles should be called the active sessions (in used but not on the TPM).

And the spec rev 1.38 part 3 (code version, PDF page 396 and 399) has the following description:

```
For this capability, use of TPM_HT_LOADED_SESSION and TPM_HT_SAVED_SESSION is allowed. Requesting handles with a handle type of TPM_HT_LOADED_SESSION will return handles for loaded sessions. The returned handle values will have a handle type of either TPM_HT_HMAC_SESSION or TPM_HT_POLICY_SESSION. If saved sessions are requested, all returned values will have the TPM_HT_HMAC_SESSION handle type because the TPM does not track the session type of saved sessions.

NOTE: TPM_HT_LOADED_SESSION and TPM_HT_HMAC_SESSION have the same value, as do TPM_HT_SAVED_SESSION and TPM_HT_POLICY_SESSION. It is not possible to request that the TPM return a list of loaded HMAC sessions without including the policy sessions.
``` 

TPM_HT_LOADED_SESSION is documented well and I have no concern about it. But TPM_HT_SAVED_SESSION seems to not conform to its name. So if the user specifies --handles-saved-session, the results are not expected. Maybe --handles-active-session is better. If so, why the spec doesn't define TPM_HT_ACTIVE_SESSION?

williamcroberts:
@jiazhang0 I see your dilemma, but I would do ```--handles-saved-session``` As for the spec, try and get an email to @flihp, maybe there is a bit of a misnomer here that needs to be corrected. The spec isn't perfect.

jiazhang0:
@williamcroberts So I created a fix in tpm2-abrmd (https://github.com/01org/tpm2-abrmd/commit/bc1b31e7dab08f5d6151b0d72816fcce3d6eb022) to see whether the new names are fine. There is no object to it so this fix is also right.

PR #592, opened 2017-10-30
tpm2_print_quote: Tool to view TPM generated quote objects

harisokanovic:
Takes a list of file paths as input containing TPMS_ATTEST objects.
Prints enclosed elements to stdout, most significant of which are the
PCR selection list and digest. Does NOT authenticate quotes, which can
be done with other tools -- E.g. see edc3e022 for openssl example.

Signed-off-by: Haris Okanovic <haris.okanovic@ni.com>


tools/tpm2_print_quote.c
```diff
@@ -0,0 +1,220 @@
...
+    UINT32 x;
+    if(fread(&x, sizeof(UINT32), 1, fd) != 1) {
+        x = 0;
+        *read_error = true;
+    }
+    x = tpm2_util_ntoh_32(x);
+    return x;
+}
+
+static UINT64 fdread_uint64(FILE* fd, bool* read_error) {
```
williamcroberts: these pretty much all exist in files.h

tools/tpm2_print_quote.c
```diff
@@ -0,0 +1,220 @@
...
+    UINT64 x;
+    if(fread(&x, sizeof(UINT64), 1, fd) != 1) {
+        x = 0;
+        *read_error = true;
+    }
+    x = tpm2_util_ntoh_64(x);
+    return x;
+}
+
+static bool print_hex(FILE* fd, long unsigned int size, bool* read_error) {
```
williamcroberts: hex_dump exists in tpm2_nvread IIRC, move it to a common utility routine

Makefile.am
```diff
@@ -189,6 +190,7 @@ tools_tpm2_pcrextend_SOURCES = tools/tpm2_pcrextend.c $(TOOL_SRC)
 tools_tpm2_pcrevent_SOURCES = tools/tpm2_pcrevent.c $(TOOL_SRC)
 tools_tpm2_rc_decode_SOURCES = tools/tpm2_rc_decode.c $(TOOL_SRC)
 tools_tpm2_import_SOURCES = tools/tpm2_import.c $(TOOL_SRC)
+tools_tpm2_print_quote_SOURCES = tools/tpm2_print_quote.c
```
williamcroberts: I think this would be a good time to start tools/aux where we can place this tool and move tpm2_rc_decode and tpm2_akparse. As well as make it a configure option.

williamcroberts: Also this needs a manpage

tools/tpm2_print_quote.c
```diff
@@ -0,0 +1,220 @@
...
+    if(fdread_uint32(fd, &read_error) != TPM_GENERATED_VALUE || read_error) {
+        LOG_ERR("%s: Bad magic", filename);
+        goto end;
+    }
+
+    // check type (must be a quote)
+    if(fdread_uint16(fd, &read_error) != TPM_ST_ATTEST_QUOTE || read_error) {
+        LOG_ERR("%s: Not a quote object", filename);
+        goto end;
+    }
```
williamcroberts: Considering that we have both of these magic values, and this tool could be extended to dump other things, maybe call it tpm2_print <objects> and design a way to make it easy to identify what object you have and look up a parser for it.

tools/tpm2_print_quote.c
```diff
@@ -0,0 +1,220 @@
...
+
+    printf("filename=%s\n", filename);
+    FILE* fd = fopen(filename, "rb");
+    if(!fd) {
+        LOG_ERR("%s: Could not open file", filename);
+        return result;
+    }
+
+    // check magic
+    if(fdread_uint32(fd, &read_error) != TPM_GENERATED_VALUE || read_error) {
```
williamcroberts: Part of the coding standard (albeit not always followed) is to not hide functions in conditional blocks (with the exception of when it makes the code simpler). This doesn't fall into the exception.

Makefile.am
harisokanovic: I like the idea of auxiliary tools, which don't directly interact with TPM. I'll move some things around.

I'm not sure what you mean by "configure option". Are you saying aux tools should optionally build?

williamcroberts: yes aux tools id see as being optionally built. Maybe opt in be default and allow opt out. @martinezjavier thoughts?

martinezjavier: @williamcroberts I wonder what would be the value of not building them. Are these auxiliary tools (tpm2_akparse, tpm2_print_quote) meant to be needed for TPM2 use cases or are mostly for debugging problems like tpm2_rc_decode? If the latter, I wonder if tools/debug shouldn't be a better name.

Also, this sub-directory will only have some meaning when executing the built tools in the source code directory, right? Since regardless of their path, all will be part of bin_PROGRAMS and will end in the bindir when are installed with make install.

So maybe it will be confusing for users to only have a sub-set of the existing tpm2 tools in bindir if the configure option is disabled.

williamcroberts: @martinezjavier I see aux tools as tools that don't invoke tpm commands. They wouldn't be needed, but mostly for conversion routines, debugging, etc. I don't want to bloat the core tool set with things that are not needed all the time.

I also don't package things for distro use, so I wasn't sure if there is any convention/wisdom you might have in this area.

I could see how some packages having extra tools could be confusing.

So is our consensus, since we don't have a very strong opinion just adding them directly into the build for now?

martinezjavier: @williamcroberts I see. What many packages do to have a core package and another package that has additional stuff (usually the convention is to use a -extras suffix, i.e: tpm2-tools-extras).

But I guess it will depend whether people will really be interested in doing this split or it's most likely that you will either want all the tools (the ones using TPM2 commands and the auxiliary ones) or none.

So yeah, I think for now we can just keep all of them in the build. Although I don't really have a strong opinion.

Somehow related to this, we should allow tools that don't use the SAPI context to not fail if the context can't be created (in fact I think it shouldn't even be attempted to create a context for these). For example, this shouldn't really fail:

```
$ tpm2_rc_decode 0x18a
ERROR: Failed to initialize tcti context: 0x1
```

tools/tpm2_print_quote.c
harisokanovic: tpm2_nvread's hexdump() prints a multi-line output similar to the hexdump utility. I prefer a single-line output in tpm2_print so that it's easily parseable by grep/cut (in shell scripts). For example, I have a use case where we need to authenticate a quote and display only PCR digest and qualifier. This is trivial to implement in bash in the current form.

tools/tpm2_print_quote.c
harisokanovic: Fixed, thanks!

tools/tpm2_print_quote.c
harisokanovic: I agree. Renamed tool to tpm2_print under tools/aux dir with a slightly different API.

tools/tpm2_print_quote.c
harisokanovic: Let me know what you think of the current implementation

harisokanovic:
@williamcroberts  I see the following in build log: command "wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.0.0_1.0.2g-1ubuntu4.6_amd64.deb" failed and exited with 8 during .
This look unrelated to my change. Could you restart the build if that's the case?

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+
+#include <stdio.h>
+#include "log.h"
+#include "files.h"
+#include "tpm2_util.h"
+#include "tpm2_alg_util.h"
+
+#define FILE_2_SHORT_ERROR(filename) LOG_ERR("%s: File too short", (filename))
```
williamcroberts: Prefer static inline over macro function when possible. Is this really needed?

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+
+static void print_hex_buffer(const UINT8* buff, UINT32 size) {
```
williamcroberts: use the hexdump routine in tpm2_util.h

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+    while(size-- > 0) {
+        printf("%2.2x", (unsigned int)*buff);
+        ++buff;
+    }
+    printf("\n");
+}
+
+static bool print_hex(FILE* fd, const char* const filename, UINT32 size) {
```
williamcroberts: Update the internals of the tpm2_util routine for hex dump to use FILE streams and then keep the old interfaces but use fmemopen() under the hood.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+            printf("\n");
+            return false;
+        }
+        printf("%2.2x", (unsigned int)byte);
+    }
+    printf("\n");
+    return true;
+}
+
+static bool print_tpm2b_hex(FILE* fd, const char* const filename) {
```
williamcroberts: same lets get this all collapsed as far as implementation in tpm2_util.c using fmemopen

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+    union {
+        UINT8  u8;
+        UINT32 u32;
+        UINT64 u64;
+    } numb;
+
+    if(!files_read_64(fd, &numb.u64)) {
+        goto read_error;
+    }
+    printf("%s.clockInfo.clock=%llu\n", prefix, (long long unsigned int)numb.u64);
```
williamcroberts: Everything should be output in yaml format, for example, see #618 
Also use tpm2_tool_output() so it respects the -Q option.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+
+    if(!files_read_32(fd, &numb.u32)) {
```
williamcroberts: again, per the coding standard no function in conditionals except for rare circumstances.
```
bool res = files_read_32(...);
if (!res) {
  LOG_ERR(...);
  goto read_error;
}
...
```

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+
+    if(fread(&numb.u8, 1, 1, fd) != 1) {
+        goto read_error;
+    }
+    printf("%s.clockInfo.safe=%u\n", prefix, (unsigned int)numb.u8);
+
+    // success
+    return true;
+
+    read_error:
```
williamcroberts: no indent here on the goto label

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+        goto read_error;
+    }
+    printf("%s.clockInfo.resetCount=%lu\n", prefix, (long unsigned int)numb.u32);
+
+    if(!files_read_32(fd, &numb.u32)) {
+        goto read_error;
+    }
+    printf("%s.clockInfo.restartCount=%lu\n", prefix, (long unsigned int)numb.u32);
+
+    if(fread(&numb.u8, 1, 1, fd) != 1) {
```
williamcroberts: files_read_bytes with size of 1

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+    const char* object_type = NULL;
+    const char* filename = NULL;
+    FILE* fd = NULL;
+
+    if(argc == 3) {
+        object_type = argv[1];
+        filename = argv[2];
+    }
+    else {
+        LOG_ERR("Usage: tpm2_print <object_type> <filename>");
```
williamcroberts: no default to stdin so you can do pipes and file redirection?

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+        goto error;
+    }
+
+    printf("filename=%s\n", filename);
```
williamcroberts: LOG_INFO, these seems worthless as far as normal tool output goes.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+    fd = fopen(filename, "rb");
+    if(!fd) {
+        LOG_ERR("%s: Could not open file", filename);
+        goto error;
```
williamcroberts: no need for the goto.
return 1

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+    int fread_result = fread(&garbage, 1, 1, fd);
+    UNUSED(fread_result);
+    printf("feof=%d\n", (feof(fd) != 0) ? 1 : 0);
+
+    // success
+    fclose(fd);
+    return 0;
+
+    error:
+    if(fd) {
```
williamcroberts: if you get rid of line 248 and make it a return, you don't need this check. Also `if (fd) {` you are missing the space.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    // print number of bytes read
+    printf("ftell=%ld\n", ftell(fd));
+
+    // check EOF by reading one more byte
+    UINT8* garbage;
+    int fread_result = fread(&garbage, 1, 1, fd);
+    UNUSED(fread_result);
+    printf("feof=%d\n", (feof(fd) != 0) ? 1 : 0);
```
williamcroberts: Is this really needed output?

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,279 @@
...
+        LOG_ERR("Unsupported object_type %s; only TPMS_ATTEST is presently supported", object_type);
+        goto error;
+    }
+
+    // print number of bytes read
+    printf("ftell=%ld\n", ftell(fd));
+
+    // check EOF by reading one more byte
+    UINT8* garbage;
+    int fread_result = fread(&garbage, 1, 1, fd);
```
williamcroberts: Could you re-work this logic with files_load_bytes_from_file_or_stdin? It would require you to know how much data you need upfront.

williamcroberts:
@harisokanovic are you going to update this to the comments for change, or did you and I miss something?

harisokanovic:
Hi William,

I got caught up in other work again. Will post an update shortly with changes you suggested.

Thanks,
Haris

tools/aux/tpm2_print.c
harisokanovic: LOG_ERR() makes use of __FILE__ and __LINE__ macros which don't expand past a function definition, but do expand though macro definition. In other words, the logged line numbers would be wrong if FILE_2_SHORT_ERROR() were a function (inline or otherwise).

tools/aux/tpm2_print.c
harisokanovic: I didn't see the `plain` option last I looked. That works for me! Will add tpm2_util_hexdump_file() to accept FILE* and refactor tpm2_util_hexdump().

tools/aux/tpm2_print.c
harisokanovic: Done

tools/aux/tpm2_print.c
harisokanovic: Done

tools/aux/tpm2_print.c
harisokanovic: Fixed

tools/aux/tpm2_print.c
harisokanovic: Fixed

tools/aux/tpm2_print.c
harisokanovic: Fixed

tools/aux/tpm2_print.c
harisokanovic: I didn't have a need for it, but I don't see why not. Added.

tools/aux/tpm2_print.c
harisokanovic: Done

tools/aux/tpm2_print.c
harisokanovic: Done

tools/aux/tpm2_print.c
harisokanovic: I though it might be useful to know if you've read the whole file or not, but not strictly necessary. I can drop it.

tools/aux/tpm2_print.c
harisokanovic: Fixed

tools/aux/tpm2_print.c
harisokanovic: Dropped per comments below.

tools/aux/tpm2_print.c
harisokanovic: Done

harisokanovic:
I opened https://github.com/intel/tpm2-tools/pull/734 with my updates since this one was closed.

PR #593, opened 2017-10-31
Flihp encrypt decrypt 2

williamcroberts:


lib/tpm2_util.h
```diff
@@ -93,6 +93,19 @@
...
+/*
+ * This macro is useful as a wrapper around SAPI functions to automatically
+ * retry function calls when the RC is TPM_RC_RETRY.
+ */
+#define TSS2_RETRY_EXP(expression)                         \
+    ({                                                     \
+        TSS2_RC __result = 0;                              \
+        do {                                               \
+            __result = (expression);                       \
+        } while ((__result & 0x0000ffff) == TPM_RC_RETRY); \
```
williamcroberts: @flihp I'm surprised there isn't a mask for this in the TSS

williamcroberts:
This is on the 1.x branch as well as master (good).

Branch 1.X: https://github.com/01org/tpm2-tss/commit/ce374d153cde9a44a09200ad801908bb5991da2c
Master: https://github.com/01org/tpm2-tss/commit/bbcc5c03ded19de78cfdede81b8d8b542d33b5e1

lib/tpm2_util.h
flihp: I've had a few things that I've wanted to work on and cleaning up the response code decoding is one of them. I don't even think the mask is necessary given the way TPM_RC_RETRY and the like are defined.

PR #594, opened 2017-11-01
Nv cleanups

williamcroberts:
Adds PCR policy support for nv read and write operations.

Fixes: #348

### DIFF
--- a/man/tpm2_nvread.1.md
+++ b/man/tpm2_nvread.1.md
@@ -42,6 +42,17 @@
   * **-S**, **--input-session-handle**=_SIZE_:
     Optional Input session handle from a policy session for authorization.
 
+  * **-L**, **--set-list**==_PCR\_SELECTION\_LIST_:
+
+    The list of pcr banks and selected PCRs' ids.
+    _PCR\_SELECTION\_LIST_ values should follow the
+    pcr bank specifiers standards, see section "PCR Bank Specfiers".
+
+  * **-F**,**--pcr-input-file=_PCR\_INPUT\_FILE_
+
+    Optional Path or Name of the file containing expected pcr values for the specified index.
+    Default is to read the current PCRs per the set list.
+
 [common options](common/options.md)
 
 [common tcti options](common/tcti.md)
--- a/tools/tpm2_nvread.c
+++ b/tools/tpm2_nvread.c
@@ -36,9 +36,12 @@
 
 #include <sapi/tpm20.h>
 
+#include "pcr.h"
+#include "tpm_session.h"
 #include "tpm2_options.h"
 #include "tpm2_password_util.h"
 #include "log.h"
+#include "tpm2_policy.h"
 #include "tpm2_nv_util.h"
 #include "tpm2_tool.h"
 #include "tpm2_util.h"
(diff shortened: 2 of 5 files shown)


williamcroberts:
Fixes: #348

man/tpm2_nvread.1.md
```diff
@@ -42,6 +42,17 @@
...
+  * **-L**, **--set-list**==_PCR\_SELECTION\_LIST_:
+
+    The list of pcr banks and selected PCRs' ids.
+    _PCR\_SELECTION\_LIST_ values should follow the
+    pcr bank specifiers standards, see section "PCR Bank Specfiers".
+
+  * **-F**,**--pcr-input-file=_PCR\_INPUT\_FILE_
+
+    Optional Path or Name of the file containing expected pcr values for the specified index.
+    Default is to read the current PCRs per the set list.
```
martinezjavier: Here it says that -F is optional...

tools/tpm2_nvread.c
```diff
@@ -259,14 +277,46 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
...
+    *opts = tpm2_options_new("x:a:f:s:o:P:S:L:F:", ARRAY_LEN(topts),
+            topts, on_option, NULL);
 
     return *opts != NULL;
 }
 
 int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
 
-    return nv_read(sapi_context, flags) != true;
+    if (ctx.flags.L && ctx.flags.F) {
```
martinezjavier: ...but here is checking for both, so it makes -F to be a required option. Instead, it should only check for ctx.flags.L since tpm2_policy_build() can take a NULL argument for the raw_pcrs_file parameter. If is NULL, then  Tss2_Sys_PCR_Read() will be called for the PCR selection.

tools/tpm2_nvread.c
```diff
@@ -52,12 +55,18 @@ struct tpm_nvread_ctx {
     UINT32 offset;
     TPMS_AUTH_COMMAND session_data;
     char *output_file;
+    char *raw_pcrs_file;
+    SESSION *policy_session;
+    TPML_PCR_SELECTION pcr_selection;
+    struct {
+        UINT8 L : 1;
+        UINT8 F : 1;
```
martinezjavier: I don't think the F flag is needed, you can just check for raw_pcrs_file being NULL (although the check is not needed as explained in my other comment).

I also wonder if the L flag is needed, maybe you can check for pcr_selection.count != 0 ?

tools/tpm2_nvread.c
martinezjavier: @williamcroberts I noticed that the tpm2_unseal tool had the same bug, so I fixed on PR #595.

williamcroberts: Ok sounds good, I literally copied you!

tools/tpm2_nvread.c
williamcroberts: I thought about that, but I remember you saying you liked seeing if (flag.x) { do something } vs if (!file) { do something }

martinezjavier: @williamcroberts I don't remember mentioning that, but I don't have a strong opinion either.

Issue #479, opened 2017-09-11
tpm2_create attribute and policy options


williamcroberts:
Could use the nv utils for handling attribute specification rather than the current mask and -E options.

Issue #489, opened 2017-09-14
Persistent objects can't be persistent by using a loaded transient object handle


martinezjavier:
This does work:

```
$ tpm2_createprimary -A o -g sha256 -G ecc -C primary.context

CreatePrimary Succeed ! Handle: 0x80ffffff

$ tpm2_evictcontrol -A o -c primary.context -S 0x81010000
persistentHandle: 0x81010000
```

But this doesn't:

```
$ tpm2_createprimary -A o -g sha256 -G ecc -C primary.context

CreatePrimary Succeed ! Handle: 0x80ffffff

$ tpm2_evictcontrol -A o -H 0x80ffffff -S 0x81010000
persistentHandle: 0x81010000
ERROR: EvictControl failed, error code: 0xa000a

$ tpm2_rc_decode 0xa000a
error layer
  hex: 0xa0000
  identifier: TSS2_TCTI_ERROR_LEVEL
  description: Error from the TCTI
base error code
  identifier: TSS2_BASE_RC_IO_ERROR
  description: IO failure
```

The reason why the former works is that [files_load_tpm_context_from_file()](https://github.com/01org/tpm2-tools/blob/master/lib/files.c#L236) has a side effect, since it also does a call to Tss2_Sys_ContextLoad().

IIUC this is just another case of the context load/save/flush issue that @williamcroberts and @idesai have been [discussing](https://github.com/01org/tpm2-tools/issues/484#issuecomment-328962870) in issue #484. But I still wanted it to have documented since is currently not working.

electrickite:
@martinezjavier Were these commands run against a resource manager? Or against the TPM device directly?

martinezjavier:
@electrickite the reported issue was when using the in-kernel resource manager (although the same error occurs with the tpm2-abrmd). Sending commands directly to the TPM will of course work because the session is alive in the TPM and never flushed.

The problem as I mentioned is that the session is flushed when the SAPI context is released. The corrct thing to do is to allow the tools to save and load sessions now that the tpm2-abrmd has support for it. With the in-kernel resource manager this may never work since is a design decision to flush all the sessions after the file descriptor is closed. 

* With tpm2-abrmd:
```
$ tpm2_createprimary -H o -g sha256 -G ecc -C primary.context
ObjectAttribute: 0x00030072

CreatePrimary Succeed ! Handle: 0x800000ff

$ tpm2_evictcontrol -A o -H 0x800000ff -S 0x81010000
[...]

PR #597, opened 2017-11-06
Friendly Object Attributes

williamcroberts:
This provides public object attribute friendly name resolution to a few tools:
tpm2_create
tpm2_createprimary
tpm2_listpersistent

It is not meant to be exhaustive of all tools, but instead lay a significant framework for moving
forward and scoped to fix issue #479.

Fixes: #479


lib/tpm2_attr_util.c
```diff
@@ -242,39 +239,151 @@ static bool nt(TPMA_NV *nv, char *arg) {
...
+    dispatch_no_arg_add(clear_stclear),   // 27
+    dispatch_no_arg_add(readlocked),      // 28
+    dispatch_no_arg_add(written),         // 29
+    dispatch_no_arg_add(platformcreate),  // 30
+    dispatch_no_arg_add(read_stclear),    // 31
+};
+
+static bool fixedtpm(TPMA_OBJECT *obj, char *arg) {
+
+    (void) arg;
```
martinezjavier: Wasn't the convention in the project to use the UNUSED() macro for this?

Now, it seems we are using this macro in a lot of places so I wonder if we shouldn't just add a `-Wno-unused-parameter` to the `EXTRA_CFLAGS` in `configure.ac` or something along those lines since we are silent the warning anyways by casting to void.

tools/tpm2_listpersistent.c
```diff
@@ -103,16 +105,30 @@ int readPublic(TSS2_SYS_CONTEXT *sapi_context, TPMI_DH_OBJECT objectHandle) {
 
     TPMI_ALG_PUBLIC type = outPublic.t.publicArea.type;
     TPMI_ALG_HASH nameAlg = outPublic.t.publicArea.nameAlg;
-    UINT32 attrs = outPublic.t.publicArea.objectAttributes.val;
+    char *attrs = tpm2_attr_util_obj_attrtostr(
+            outPublic.t.publicArea.objectAttributes);
+    bool should_free = !!attrs;
+    if (!attrs) {
+        LOG_WARN("Could not convert objectAttributes, converting to hex output");
+        char tmp[11]; /* UINT32 in hex (8) + "0x" + '\0' */
```
martinezjavier: Maybe it will be easier to read if you just allocate the array in attrs and copy there directly. That way you can get rid of the should_free and tmp variables and just check if attrs != NULL to free.

test/system/test_tpm2_getmanufec.sh
```diff
@@ -76,6 +76,7 @@ tpm2_getmanufec -H $handle -U -E ECcert2.bin -f test_ek.pub -o $opass -e $epass
                 https://ekop.intel.com/ekcertservice/
 
 tpm2_listpersistent | grep -q $handle
+tpm2_listpersistent
```
martinezjavier: Why you are adding this to the test? At least it should be mentioned in the commit message.

If is just to see if the friendly attribute names are printed, I don't think is the correct way to do it. Instead we need a tpm2_listpersistent integration test as mentioned in issue #489.

man/common/obj-attrs.md
```diff
@@ -0,0 +1,19 @@
...
+which can be found here:
+
+<https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf>
+
+Nice names are calculated by taking the name field of table 31 and removing the
+prefix **TPMA_OBJECT_** and lowercasing the result. Thus, **TPMA_OBJECT_FIXEDTPM** becomes
+fixedtpm. Nice names can be joined using the bitwise or "|" symbol.
+
+For instance, to set The fields **TPMA_OBJECT_FIXEDTPM**,
+**TPMA_OBJECT_NODA**, and **TPMA_OBJECT_sign**, the argument
```
martinezjavier: s/TPMA_OBJECT_sign/TPMA_OBJECT_SIGN

lib/tpm2_attr_util.c
williamcroberts: Yes that is correct. I'm not a big fan of adding compiler flags to undo warnings. I'd rather the code be explicit about an unused variable.

tools/tpm2_listpersistent.c
williamcroberts: I'm not sure what your saying but I have another approach to this that's simpler.

test/system/test_tpm2_getmanufec.sh
williamcroberts: This is a change I had locally that I never meant to submit just for getting tpm2_listpersistent output with an object defined.

williamcroberts:
Yes this will be cherry-picked to 3.X as it was slated for 3.X. I branched 3.X early so when the major API change in TSS hits, 3.X will already be on a stable tss version. I plan on taking everything on the next milestone to 3.X and master.

williamcroberts:
3.X here: https://github.com/01org/tpm2-tools/pull/598

Issue #508, opened 2017-09-19
tpm2_rsadecrypt/encrypt, make -i an argument and -o optional


williamcroberts:
Without -o output to stdout in hex format via tool_output. With -o, output to a file as a binary format.

Drop -I as it should be an argument or default to stdin.

PR #608, opened 2017-11-09
Rsa decrypt options

williamcroberts:


tools/tpm2_rsaencrypt.c
```diff
@@ -46,14 +47,13 @@ struct tpm_rsaencrypt_ctx {
     struct {
         UINT8 k : 1;
         UINT8 I : 1;
-        UINT8 o : 1;
         UINT8 c : 1;
-        UINT8 unused : 4;
+        UINT8 unused : 5;
```
martinezjavier: Is the unused field really needed here?

The bit-fields will be packed in an unsigned char anyways so I don't see the value in explicitly defining it (specially since you need to update it each time you add/remove a bit-field).

tools/tpm2_rsaencrypt.c
```diff
@@ -112,16 +118,11 @@ static bool on_option(char key, char *value) {
     }
         break;
     case 'o': {
-        bool result = files_does_file_exist(optarg);
-        if (result) {
-            return false;
-        }
-        ctx.output_file_path = optarg;
-        ctx.flags.o = 1;
```
martinezjavier: This is changing some semantics here, I agree with your change but I think it should be commented in the commit message.

man/tpm2_rsaencrypt.1.md
```diff
@@ -41,7 +41,9 @@ The key referenced by keyHandle is **required** to be:
```
martinezjavier: I really think Github should have a way to comment on commit messages, so I'll comment here:

You need to add a Fixes: #608

martinezjavier: Ah, nevermind. I see that you are adding the tag in the next commit so feel free to ignore this comment.

tools/tpm2_rsaencrypt.c
```diff
@@ -46,14 +46,14 @@ typedef struct tpm_rsaencrypt_ctx tpm_rsaencrypt_ctx;
 struct tpm_rsaencrypt_ctx {
     struct {
         UINT8 k : 1;
-        UINT8 I : 1;
         UINT8 c : 1;
-        UINT8 unused : 5;
+        UINT8 unused : 6;
```
martinezjavier: Same comment than in previous commit, I think we can just drop this.

tools/tpm2_rsaencrypt.c
```diff
@@ -131,24 +121,35 @@ static bool on_option(char key, char *value) {
...
+            on_option, on_args);
 
     return *opts != NULL;
 }
 
 static bool init(TSS2_SYS_CONTEXT *sapi_context) {
 
-    if (!((ctx.flags.k || ctx.flags.c) && ctx.flags.I)) {
+    if (!(ctx.flags.k || ctx.flags.c)) {
         LOG_ERR("Expected options I and (k or c)");
```
martinezjavier: You forgot to remove the `I` option from the error log printout.

williamcroberts: Done

tools/tpm2_rsaencrypt.c
williamcroberts: Done

man/tpm2_rsaencrypt.1.md
williamcroberts: Yes becuase #508 had both -I and -o modifications, so I don't close it until all the work is done.

tools/tpm2_rsaencrypt.c
williamcroberts: Doesn't this clarify it, "Make -o option optional and default to stdout. When
printing to stdout, use an xxd compatible hexdump
format."

That's the commit message body.

martinezjavier: @williamcroberts I didn't mean that part but the fact that now the output file would be overwritten if already existed. But before your change it would fail if the file already existed.

tools/tpm2_rsaencrypt.c
williamcroberts: Sorry this is from me working with Borland back in the day that would complain if your bitfield was not equal to the width of the type. It's just become habit to be explicit here, but it's not needed and ill stop now :-)

PR #610, opened 2017-11-09
[WIP] Allow tpm2-tools 3.X branch to build against the tpm2-tss 1.x branch

martinezjavier:
This pull request contains patches that allows the tpm2-tools to be built against the tpm2-tss 1.x branch.

There's one issue remaining though so this PR isn't in a merge-able state: the tpm2_import tool uses some TPM marshalling helpers that are only present in the tpm2-tss master branch, so this isn't something trivial to solve and some understanding from the TSS marshalling code is needed.

I wanted to share the patches anyways so you could review the other ones while the last issue is solved.


williamcroberts:
Just drop tpm2_import then

Makefile.am
```diff
@@ -63,7 +63,6 @@ bin_PROGRAMS = \
     tools/tpm2_getrandom \
     tools/tpm2_hash \
     tools/tpm2_hmac \
-    tools/tpm2_import \
```
williamcroberts: This is fine, just nuke this tool.

martinezjavier: @williamcroberts Ok, I'll remove the this tool then. I guess that's OK since the implementation is tightly coupled with the SAPI 2.0 version.

martinezjavier:
I forgot to test locally with `--enable-unit` and so didn't realize that the test_tpm2_alg_util unit test is failing. Didn't have time to look at this but I'll do it tomorrow.

williamcroberts:
moved: https://github.com/intel/tpm2-tools/pull/613

Looks like that activatecredential issue is on 3.x as well. Ill have to bisect this and find where I introduced the error.

PR #614, opened 2017-11-09
Correct tpm2_activecredential test

williamcroberts:
* Revert the HACK to make the test pass
* Fix the issue
* misc fixes

These all need to go to the 3.X branch tomorrow.


tools/tpm2_pcrlist.c
```diff
@@ -281,11 +264,8 @@ static bool show_pcr_values(void) {
                 return false;
             }
 
-            if (ctx.format.yaml) {
-                tpm2_tool_output("  %-2d : ", pcr_id);
-            } else {
-                tpm2_tool_output("PCR_%02d:", pcr_id);
-            }
```
martinezjavier: I'm not sure about this change. But I understand your goal of having all the tools to print the output in YAML format for consistency, so I guess it's OK to deprecate the old printout.

PR #618, opened 2017-11-14
tpm2_nvlist: yaml output support

williamcroberts:
This PR has commits for:
 * Fixing a BE to host endian conversion
 * tpm2_nvlist: output in yaml format
 * changelog update

### DIFF
--- a/tools/tpm2_nvlist.c
+++ b/tools/tpm2_nvlist.c
@@ -68,14 +68,15 @@ static void print_nv_public(TPM2B_NV_PUBLIC *nv_public) {
     tpm2_tool_output("  size: %d\n",
                nv_public->nvPublic.dataSize);
 
-    tpm2_tool_output("  authorization policy: ");
+    if (nv_public->nvPublic.authPolicy.size) {
+        tpm2_tool_output("  authorization policy: ");
 
-    UINT16 i;
-    for(i=0; i<nv_public->nvPublic.authPolicy.size; i++) {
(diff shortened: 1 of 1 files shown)


tools/tpm2_nvlist.c
```diff
@@ -49,18 +50,32 @@ static void print_nv_public(TPM2B_NV_PUBLIC *nv_public) {
...
             tpm2_util_ntoh_32(nv_public->nvPublic.attributes.val));
-    printf("\tThe size of the data area(dataSize):%d\n ",
-            nv_public->nvPublic.dataSize);
-    printf("\tAuthorization Policy for R/W/D: ");
-    int i;
+
+    tpm2_tool_output("  size: %d\n",
+               nv_public->nvPublic.dataSize);
+
+    tpm2_tool_output("  authorization policy: ");
```
martinezjavier: Can we only print this if `nv_public->nvPublic.authPolicy.size > 0` ? Or print `None` or something if `nv_public->nvPublic.authPolicy.size == 0`. Otherwise it feels as if the output got truncated.

williamcroberts: That's fine. My reasoning for the way it is is that one can always get the key in the yaml file and the value is null. However, it's just as easy to check if key exists. So ill do it that way and then it's less for a human to read as well.

williamcroberts:
Ughhh... This is why I hate having direct push access. I just accidentally pushed my changes into master. Ill edit this to be code review fixups.

williamcroberts:
I just enabled branch protections, hopefully that will help prevent dumb mistakes.

PR #624, opened 2017-11-15
master-errata enable option

williamcroberts:


williamcroberts:
Im going to put  f400ceb on a separate PR to 3.X and master, as we need it for bug fix.

lib/tpm2_errata.c
```diff
@@ -143,6 +143,9 @@ void tpm2_errata_fixup(tpm2_errata_index_t index, ...) {
         return;
     }
 
+    LOG_WARN("Applying errata: \"%s\", use --disable-errata to disable!",
+            errata->name);
```
martinezjavier: If we flip the logic to instead have an option to enable erratas, I think we should change this to LOG_INFO().

tools/tpm2_tool.c
```diff
@@ -139,7 +139,11 @@ int main(int argc, char *argv[], char *envp[]) {
     /* TODO SAPI INIT */
     TSS2_SYS_CONTEXT *sapi_context = sapi_ctx_init(tcti);
 
-    tpm2_errata_init(sapi_context);
+    if (!flags.disable_errata) {
+        tpm2_errata_init(sapi_context);
+    } else {
+       LOG_INFO("Errata disabled");
```
martinezjavier: And this could also be removed if we flip the logic.

lib/tpm2_errata.c
williamcroberts: Agreed.

tools/tpm2_tool.c
williamcroberts: Agreed.

williamcroberts:
@martinezjavier I am inclined to disable by default as well, lets let this soak for a day for @jiazhang0 to respond. If no argument contrary will flip the logic tomorrow and submit on master and 3.X

lib/tpm2_errata.c
williamcroberts: There is an Errata applied statement at the end...

lib/tpm2_options.c
```diff
@@ -252,12 +252,17 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv, char **envp,
...
-        { "help",    no_argument,       NULL, 'h' },
-        { "verbose", no_argument,       NULL, 'v' },
-        { "quiet",   no_argument,       NULL, 'Q' },
-        { "version", no_argument,       NULL, 'V' },
+        { "tcti",           required_argument, NULL, 'T' },
+        { "help",           no_argument,       NULL, 'h' },
+        { "verbose",        no_argument,       NULL, 'v' },
+        { "quiet",          no_argument,       NULL, 'Q' },
+        { "version",        no_argument,       NULL, 'V' },
+        { "enable-errata", no_argument,       NULL, 'Z' },
```
williamcroberts: nit: space

jiazhang0:
@williamcroberts @martinezjavier I totally agree with disabling errata by default because there is no way to enforce manufacturer strictly following the spec. Since we already see a broken TPM firmware, it means we will see more in future.

PR #625, opened 2017-11-15
log: fix verbose option

williamcroberts:
log_set_level() was never being called when verbose
was enabled. So set it when verbose is enabled.

This fixes LOG_INFO not showing up.


williamcroberts:
@martinezjavier it did work at some point. Likely during the option revamping, but I didn't go looking.

martinezjavier:
yeah, likely. Anyways, it's fixed now :smiley:

PR #627, opened 2017-11-15
master-add unit tests for tpm2_errata

williamcroberts:
TODO:
Add negative tests

### DIFF
--- a/test/unit/test_tpm2_errata.c
+++ b/test/unit/test_tpm2_errata.c
@@ -1,0 +1,185 @@
+//**********************************************************************;
+// Copyright (c) 2017, Intel Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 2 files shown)


williamcroberts:
@martinezjavier @jiazhang0 any early feedback on testing the errata code? Id like to get the test coverage up and considering we've been hitting issues in this, I figured this was as good as any place to start.

Plus I figured out how to get cmocka mock stuff working which is pretty nice for emulating tpm responses.

martinezjavier:
@williamcroberts yes, I think it would be nice to have such tests. I'll let @jiazhang0 to review it though, since he is the one who wrote the errata code.

test/unit/test_tpm2_errata.c
```diff
@@ -0,0 +1,201 @@
...
+            .objectAttributes = {
+                .val = TPMA_OBJECT_SIGN
+            },
+        },
+    };
+    tpm2_errata_fixup(SPEC_116_ERRATA_2_7,
+                      &in_public.publicArea.objectAttributes);
+
+    assert_int_equal(in_public.publicArea.objectAttributes.sign, 1);
+
```
jiazhang0: Inside errata code, the errata_match() will be true if this_errata_info is older than the known errata SPEC_116_ERRATA_2_7, where this_errata_info is retrieved from the tpm hardware. So I suspect setcaps() cannot affect this_errata_info.

So a tpm manufactured according to the original 1.16 spec will trigger the assertion because sign is fixed up to be 0.

williamcroberts:
Submitted on 3.X as #632

PR #630, opened 2017-11-17
Account for type and API changes in the tpm2-tss libraries and headers.

flihp:
These are all simple renames. There are no functional changes in this patch.


flihp:
Still two failing test cases.

williamcroberts:
@flihp I'll look at this today

flihp:
Thanks. I'll be doing the same in a bit.

flihp:
`test_tpm2_import.sh` failing on `tpm2_import`. Failing function is `RSA_public_encrypt`. Adding a bit of code to decode the OpenSSL error code reveals:
OpenSSL error code is: 50770023
This decodes to the string: `error:0306B067:bignum routines:BN_div:div by zero`

williamcroberts:
@flihp cherry-pick these two changes, they fix the tests with respect to quoting issues on empty strings, etc. So now more tests fail...
5f907f992d3e035fc288189e5e0ab10714bc53b3
c86b5cb5ca6ba08a7ef6f25777d10dea30716286

flihp:
thaaaaaaanks

flihp:
are you sure you meant 5f907f9?

williamcroberts:
@flihp sorry, wrong sha, i meant: a459b03c01431b59b974e7dae63aaea59953449e

flihp:
NP. Just picked these into this branch / PR. Still just one test failing (import), same as before.

williamcroberts:
@flihp what did you have to fix up?

flihp:
stuff like: https://github.com/intel/tpm2-tools/pull/630/commits/654c5f0700011d5bbcce276991d356a4494ff69a#diff-487a18693efe653813e8fbff6bf590adL156

williamcroberts:
Ahh this all funnels back to my goal of fixing up the poor output of the tools! I've actually decreed that YAML is going to be our default output.

I thought you meant you had to fixup my commits to the test scripts as far as those two commits go.

flihp:
getting closer to a solution: `tpm2_readpublic` is producing public keys with an additional 256 bytes of 0's tacked on to the end of the file

martinezjavier:
@flihp it sounds something similar to the issue fixed by commit df751ae5bea0 ("tpm2_unseal: fix unsealed data object serialization").

flihp:
that is exactly what's happening here, using sizeof (struct) not size field from structure.

williamcroberts:
@flihp how close are you to having this sorted? You sound close. Im trying not to stomp on your toes/duplicate work. Ping me if you need anything.

williamcroberts:
Submitted on 3.X as #631

williamcroberts:
@flihp #634 will make your happy when it's done

flihp:
@williamcroberts : #634 isn't addressing my current problem though I'm sure it's an improvement. I think https://github.com/intel/tpm2-tools/blob/master/lib/conversion.c#L85 is a bug similar to https://github.com/intel/tpm2-tools/commit/df751ae5bea0bb057c9ee4cb0c1176c48ff68492. Unfortunately "fixing" this bug may be uncovering a new one. I'll make a separate issue for this.

williamcroberts:
@flihp So what's the current issue, I see tpm2_import is still failing.

flihp:
Yes but it looks like the issue isn't in the import function. The difference I'm seeing is in the public key dumped by `tpm2_readpublic`. The sizeof this file changed from 360 bytes to 616 with the update to the TSS libraries. It looks like the conversion function that dumps the TPM2B_PUBLIC structure to a file is using sizeof (struct) when it should be consulting the `size` field from the structure: https://github.com/intel/tpm2-tools/blob/master/lib/conversion.c#L85.

Changing this code to use `public->size + sizeof (public->size)` seems to truncate the public structure though, producing a file 284 bytes.

flihp:
sorry link in the previous post should have been https://github.com/intel/tpm2-tools/blob/master/lib/conversion.c#L92

martinezjavier:
I thought we had fixed all the serialization issues, didn't we have an issue filed for that?

flihp:
not that I found, not open at least.

williamcroberts:
No open ones, I thought we got them too. I guess not.

flihp:
Also worth noting that I don't know for sure that this is actually part of the problem. Just a weird thing I ran across trying to debug this.

flihp:
I'll make an issue for it though.

martinezjavier:
@flihp this is the one that I was remembering https://github.com/intel/tpm2-tools/issues/229. But it seems that not all the tools got fixed and some of them remained as TODO.

So there's at least 7 more tools with the same issue...

```
$ git grep "TODO" | grep serialization | wc -l                         
7
```

williamcroberts:
@martinezjavier I see 7 bugs

williamcroberts:
@flihp I had the same issue with the last header changes. I noticed the file size change, but it appeared to still be working under most conditions.

flihp:
#635 is for the issue dumping the TPM2B_PUBLIC structure to disk.

How accurate do we think this 'TODO' count of 7 is? This thing with the TPM2B_PUBLIC structure isn't marked with a TODO. So it's not included in that count.

martinezjavier:
@flihp indeed, that's why I said *at least* 7 issues.

williamcroberts:
Yeah there is likely quite a few of them still and I bet theres still some tools that hand output all the files rather than using the files.h interfaces.

flihp:
I've updated #635 with a few notes. I don't think this is the root cause here anyway, just something I tripped over while debugging.

williamcroberts:
Man the code when loading everyhing up is based on a  lot of magic, like the modulus at bytes 102 and that it's TPM2_MAX_RSA_KEY_BYTES (512) long. This takes into a bunch of trailing 0s in the file.
```
    FILE *fp = fopen(ctx.parent_key_public_file, "rb");
    if (fp == NULL) {
        LOG_ERR("Failed accessing parent key public file.");
        return false;
    }
    if (fseek(fp, 102, SEEK_SET) != 0) {
        LOG_ERR("Expected parent key public data file size failure");
        fclose(fp);
        return false;
    }
    unsigned char pub_modulus[TPM2_MAX_RSA_KEY_BYTES] = { 0 };
    int ret = fread(pub_modulus, 1, TPM2_MAX_RSA_KEY_BYTES, fp);
    if (ret != TPM2_MAX_RSA_KEY_BYTES) {
        LOG_ERR("Failed reading public modulus from parent key public file");
        fclose(fp);
        return false;
    }
    fclose(fp);
```

williamcroberts:
Im getting this error: ERROR: Failed RSA_public_encrypt: "error:0406B06E:rsa routines:RSA_padding_add_none:data too large for key size"
At the RSA_public_encrypt(), IIRC only 256 can be encrypted with a 2048 bit key.

flihp:
Yeah that `fseek` is ugly. Should be marshal / unmarshal IMHO but this goes against your yaml goals.

williamcroberts:
@flihp yaml is only to stdout...files are binary. I have a patch that gets us throught that part

williamcroberts:
So if I don't hear anything by 10am pacific time, (Ill give folks time to wake up and comment), I am going to take this PR + a patch to disable tpm2_import tests. And then we can file a bug to fixup and cleanup the tool. We might need Imrans help.

@martinezjavier The second issue this raises is how brittle all the serialization issues to disk are. Should I just pause the 3.0-rc1 and we can pull all the fixes into that as well as master?

flihp:
Definitely the right thing to do w/r to saving the pubkey to a file and back. It breaks a number of other tests though. Looks like it may just be that they need to be fixed up to use this method as well. Looking into that now.

williamcroberts:
@flihp don't worry about the other tests. Lets focus on tpm2_import and see if we can fix this nightmare:
```
test.sh test_tpm2_import.sh
```
We might just want to take your current PR + a patch to disable tpm2_import tests, file a bug and get @idesai involved.

flihp:
@williamcroberts IMHO disabling the `tpm2_import` test is the right way to go about this. Per my last comment, switching over the method for dumping and reading public keys back isn't a piecemeal operation. Best to minimize the moving parts and disable this one test and then fix all of the public key to / from disk stuff in one go. I'll update this PR accordingly.

williamcroberts:
@flihp ahh I misunderstood.

idesai:
@williamcroberts ok I will await your serializations fixes

flihp:
@williamcroberts : There's now a commit disabling the `tpm2_import` test. This can be reverted once the key blob save / restore to / from disk stuff is reworked. Tagged you in for a review too.

martinezjavier:
> @martinezjavier The second issue this raises is how brittle all the serialization issues to disk are. Should I just pause the 3.0-rc1 and we can pull all the fixes into that as well as master?

yes, I was thinking the same. Since otherwise a serialized data written to disk may not be read anymore  after the fixes. It feels as if we will never have this release...

PR #640, opened 2017-11-21
master-serialization-fixes

williamcroberts:
Looking through the source for bad casts on files_save_bytes yields no hits after these patches.
```grep -n files_save_bytes *.c | grep UINT8```

Hopefully this puts all serialization issues to bed #639 .

I deleted tpm2_akparse and will update the CHANGELOG with it when it's merged. There was no reason to keep this tool after readpublic's output was updated to dump that in a yaml format.

Ill will port these to 3.X and cut rc1 soon, I need to get a fix for #637 on master to backport to 3.X.

Fixes: #639 
Fixes: #637

### CODE REVIEW

File: lib/conversion.c [hunk 1]
```diff
@@ -88,8 +88,7 @@ bool tpm2_convert_pubkey(TPM2B_PUBLIC *public, pubkey_format format, const char
         return tpm2_convert_pubkey_ssl(&public->publicArea, format, path);
     }
     else if (format == pubkey_format_tss) {
-        // save raw as is
-        return files_save_bytes_to_file(path, (UINT8 *)public, sizeof(*public));
+        return files_save_pubkey(public, path);
```
williamcroberts: files_save_public seems like a better name...

lib/files.c
```diff
@@ -568,3 +568,119 @@ bool files_load_bytes_from_file_or_stdin(const char *path, UINT16 *size, BYTE *b
...
+
+    size_t offset = 0;
+    TSS2_RC rc = Tss2_MU_TPM2B_SENSITIVE_Unmarshal(buffer, size, &offset, sensitive);
+    if (rc != TSS2_RC_SUCCESS) {
+        LOG_ERR("Error serializing signature structure: 0x%x", rc);
+        return false;
+    }
+
+    return rc == TPM2_RC_SUCCESS;
+}
```
williamcroberts: These could all be generated via macros....

tools/tpm2_akparse.c
```diff
@@ -94,7 +94,9 @@ static bool save_alg_and_key_to_file(const char *key_file, UINT16 alg_type,
             goto out;
         }
 
+        tpm2_tool_output("- ");
         tpm2_util_print_tpm2b(tmp);
+        tpm2_tool_output("\n");
```
williamcroberts: This output is flaky. Have it map to something like:
ECC:
```
algorithm: ecc
  x : fdjhdfkhj
  y: djhjfhg
```
RSA
```
algorithm: rsa
   foo: bar
```
Not sure what the rsa sub-name is offhand. Writing this so I don't forget TODO tomorrow.

lib/conversion.c
martinezjavier: Indeed.

martinezjavier: And probably makes sense to rename all the *_pubkey to *_public

lib/files.c
martinezjavier: Indeed, all the files_{save,load}_* functions seems to be the same. The only thing that changes is the second parameter type and marshal/unmarshal function used (which are also generated via macros).

martinezjavier:
@williamcroberts if the libmarshal helpers are backported to 3.X, then the tpm2_import tool could be enabled again. Since the reason why it was disabled was that it didn't build against tpm2-tss 1.3.0 due these functions not being there.

flihp:
@martinezjavier the `tpm2_import` tool had issues beyond the serialization functions. As I recall even after we switched to using the proper marshalling function from libmu the `tpm2_import` tool was still failing. These were blockers though so once they're in we can get back to fixing `tpm2_import`.

martinezjavier:
@flihp right, I didn't follow that discussion too closely.

flihp:
I guess my point is that there's still work to be done there after the marshalling functions are in place. Not sure how long we want to hold up the release on account of this especially with the upcoming U.S. holiday. I remember some discussion on time-boxing the release nearly 2 months ago ...

martinezjavier:
@flihp Absolutely agree. I just mentioning because I (wrongly) assumed that there wasn't more work besides the missing marshaling functions.

But if that's the case I'm in favour of just keeping the tpm2_import tool out of the 3.X release.

I'm very interested on having this release ASAP since I've some changes in another project waiting for it release for months.

williamcroberts:
@flihp @martinezjavier 

* tpm2_import is not in the 3.X release.
* This patch should take care of most of the serialization issues to disk. Hopefully it's all of them, I didn't 
  spot anymore.
* I plan on porting these to master on 11-23/24 and having rc1 out the following week. So only one week  
  delay.

williamcroberts:
Looking at the tpmu-types serialization code, with all the "na" member buisness, this looks insane for me to port over in a reasonable amount of effort given my familiarity with the code.

Offhand was "na" needed. Wouldn't the pointer to the struct be the top of struct? Ie &foo->na == foo

martinezjavier:
@williamcroberts I don't think that na field makes any sense. As you said it's guaranteed that a pointer to a structure points to its initial member.

We were only using for 3.X in tpm2_alg_util.c before and I just got rid of it with commit 8e7c52bf7cec ("lib: just copy the TPMU_HA union instead of copying its member").

williamcroberts:
@martinezjavier its a matter of taking apart those macros. I might actually just use C protobuf to create pack/unpack routines. It might be the most elegant solution here in that we won't need to deal with libmarshal and protobuf generate the code for us AFAIC: https://github.com/protobuf-c/protobuf-c

I'd like to see the justification for "na" by @tstruk and/or @flihp..

test/system/test_tpm2_unseal.sh
```diff
@@ -105,7 +127,9 @@ tpm2_create -Q -g $alg_create_obj -G $alg_create_key -u $file_unseal_key_pub -r
 
 tpm2_load -Q -c $file_primary_key_ctx  -u $file_unseal_key_pub  -r $file_unseal_key_priv -n $file_unseal_key_name -C $file_unseal_key_ctx
 
-unsealed=`tpm2_unseal -c $file_unseal_key_ctx -L ${alg_pcr_policy}:${pcr_ids} -F $file_pcr_value`
+tpm2_unseal -c $file_unseal_key_ctx -L ${alg_pcr_policy}:${pcr_ids} -F $file_pcr_value > "$file_unsealed_data_out"
+
+unsealed=`yaml_get "sensitive" "$file_unsealed_data_out" | xxd -r`
```
martinezjavier: I like the yaml output for tools list tpm2_pcrlist or tpm2_nvlist but I think this is taking it too far.

It's quite likely that people will want tpm2_unseal to just return the same sensitive data that was provided to tpm2_create when the object was created. Having it print as a yaml output with a single `sensitive:` field just makes user's life more dificult IMHO.

Same for tpm2_read, it should just read whatever was written in the NVRAM with tpm2_write. From a user POV she just wrote some bytes and wants to read again the same.

tools/tpm2_unseal.c
```diff
@@ -93,13 +93,14 @@ bool unseal_and_save(TSS2_SYS_CONTEXT *sapi_context) {
         return false;
     }
 
+    tpm2_tool_output("sensitive: |\n");
+    tpm2_util_hexdump(outData.buffer, outData.size, false, true);
```
martinezjavier: Same for using hexdump here, this just makes users to need a tool like `xxd` to get the original sealed data for no good reasons IMHO. And the same comment applies to tpm2_nv{read,write}.

flihp:
libmu questions are for @tstruk

test/system/test_tpm2_unseal.sh
williamcroberts: @martinezjavier I don't want raw bytes dumped to the terminal. Consider a user inspecting the contents and forgets to | to xxd or something, their going to end up with unescaped sequences hitting their terminal.

tools/tpm2_unseal.c
williamcroberts: Same response applies.

williamcroberts:
@martinezjavier I am less confident that these changes will hit 3.X as libmu seems much more work to port than anticipated.

test/system/test_tpm2_unseal.sh
martinezjavier: @williamcroberts well, the user should know what's doing. It will be the same if tries to read any raw binary data.

The reason why I added tpm2_unseal to print the unsealed data to the stdout in the first place in commit 8aa7f644386d ("tpm2_unseal: allow to write unsealed data to the standard output") was to avoid the need to store sensitive data to disk.

But printing the data as a hexdump inside a yaml makes basically that information unusable. Since you now need to parse the data and transform, which means that you need tools like xxd and a yaml parser (or at least grep) to get what I think is very basic information (the data that you sealed).

I think that at the very least we need to support both, it's OK if you don't want to make printing the raw data as the default (although I would) but we need a command line option to allow users IMHO.

martinezjavier: I think my point is that tpm2_unseal, obtains the sensitive part of the sealed object and tpm2_nvread reads a NVRAM index area. The tool shouldn't try to infer what kind of data that is.

f that data are bytes with ASCII values, then it will be friendly to print in a terminal and if is binary data, then it won't. 

It's absolutely valid to seal (or nvwrite) ASCII bytes for example and that will be printed correctly on a terminal. So it would be weird for a user to get a hexdump of the string "foobar".

Just like any tools reading data and printing to stdout (i.e: cat) does. The user should know what kind of information is reading and redirecting the stdout to another file if doesn't want to end in the terminal.

martinezjavier:
@williamcroberts yeah, I looked briefly when you said that it would need backporting and I thought that it wouldn't be trivial. Although I'm not familiar at all with libmu so I thought that was the reason.

test/system/test_tpm2_unseal.sh
williamcroberts: @martinezjavier let me explain further why I made this choice:
Consistent parsing scheme. As we  move forward, outputs will become backwards compatible and YAML makes this easy. We can add, arbitrarily to the outputs and the YAML code written to parse it will always work. I'm hoping the same for options, id like to get less churn in the releases as the code starts to become more robust, sane and well tested. Theirs still a long way to go, these tools were in quite awful state when I took over, and still are :-1: 

However, the tools you mention, it would be easy to add a -o <file> option that supports the "-" so you could send binary to stdout. This seems sane to me.

martinezjavier: @williamcroberts and one last argument against to output in yaml + hexdump is that it makes the operations non-reciprocal. As a user I would think why read doesn't work the same as write?

For example I wold be very confused with something like the following:

```
$ echo foobar > file.txt
$ cat file.txt
0000000 6f66 626f 7261 000a                    
0000007
```

As you can see from my lengthy responses, I've a very strong opinion on this :smiley:

martinezjavier:
> However, the tools you mention, it would be easy to add a -o option that supports the "-" so you could send binary to stdout. This seems sane to me.

I do agree to have yaml output on anything that's data not provided by the user. My point is that we should output the data provided by the user verbatim and not mangle it.

martinezjavier:
But I guess that's OK to have a add the `-` convention to the `-o` option so you can write to the stdout with `-o-`. I would just leave it as is but it's your call of course.

test/system/test_tpm2_unseal.sh
williamcroberts: Now echo some non-ascii, this is where it falls apart at the shell.
```
$ echo -e "\x02\xc5\xd8" > foo
$ cat foo
�
```

williamcroberts:
@martinezjavier I want you on board with changes, or at least can understand why I want to do something a certain way.

test/system/test_tpm2_unseal.sh
martinezjavier: Yes, I know that. My point was that `cat` doesn't try to be smart and just prints to stdout whatever reads from the file. If the user wrote binary data there and tries to read it, it's their problem.

martinezjavier:
@williamcroberts thanks, I really appreciate. I do understand your motivation and I agree with most of it. I just think that an exception should be made for data written by the user.

But as said, I'm also OK with having an option to control this behaviour . Most likely users will always use that option instead of the default yaml + hexdump though.

test/system/test_tpm2_unseal.sh
williamcroberts: I'm coming around to your argument. I'm gonna back out nvread to hexdump then as well as these yaml parts to these changes.

Ill change my stance to:
Any tool that has a compound output, is YAML. Other tools, data as is

martinezjavier:
@williamcroberts thanks a lot for taking my comments into account.

PR #644, opened 2017-11-23
Switch to new TPMA layout of tpm2-tss

AndreasFuchsSIT:
tpm2-tss and spec changed the definition of TPMAs from bitfield
structs to a signel value and defines.
This patch adapts this change.

P.S. This will of course not test on travis, since the version of tpm2-tss is too old over there...


williamcroberts:
Nice I picked this onto #640

Issue #646, opened 2017-11-27
Only check the lower 12 bits in TPM_RC for TPM response codes


martinezjavier:
The TPM only uses the lower 12 bits from the 32 bits of TPM_RC. The other layers uses some of the higher unused 20 bits to indicate from which layer of the stack the error comes from.

So add a macro that masks TPM_RC to only the bits used by the TPM for response codes.

PR #652, opened 2017-11-28
lib: mask TPM_RC to only get the bits set by the TPM

martinezjavier:
The TPM only uses the lower 12 bits from the 32 bits of TPM_RC, the other
layers uses some of the higher unused 20 bits to indicate from which layer
of the stack the error comes from.

So mask the low-order 12 bits of the TPM_RC variable instead comparing TPM
response codes directly. There's no need to mask when the check is against
TPM_RC_SUCCESS, since in this case the upper 20 bits are going to be zero.

Fixes: #646


williamcroberts:
Moved:
https://github.com/intel/tpm2-tools/pull/662

PR #659, opened 2017-11-28
3.x serialization fixes

williamcroberts:


williamcroberts:
@martinezjavier anything else you want on 3.X?

martinezjavier:
@williamcroberts good to see this PR!

> @martinezjavier anything else you want on 3.X? I see you have a ton of PRs, taking those in-tree.

I don't have anything else left, I pushed today everything that I had on my TODO list.

martinezjavier:
> I don't have anything else left, I pushed today everything that I had on my TODO list.

Actually, I lied. There's one more change I want to propose, I think the tools also need this change:

https://github.com/intel/tpm2-tss/pull/645

martinezjavier:
> Actually, I lied. There's one more change I want to propose, I think the tools also need this change:

I was correct at the end that I don't have any more changes, since the mentioned change wasn't  needed.

PR #663, opened 2017-11-28
Flushcon

williamcroberts:


williamcroberts:
I see this has a dependency on https://github.com/intel/tpm2-tools/pull/667

In the future, anything with dependencies put them in the same PR

Issue #664, opened 2017-11-28
tpm2_rsaencrypt is stuck


arvin4181:
tpm2_rsaencrypt is stuck while doing a read, take a look at the strace below. other general commands like tpm2_nvlist works fine.

$ strace tpm2_rsaencrypt keyFile.bak -c ctx.out1 -o encKeyFile.bak
execve("/usr/local/bin/tpm2_rsaencrypt", ["tpm2_rsaencrypt", "keyFile.bak", "-c", "ctx.out1", "-o", "encKeyFile.bak"], [/* 18 vars */]) = 0
brk(NULL)                               = 0x5590b82d0000
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
mmap(NULL, 12288, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f717ec67000
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
open("/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=24321, ...}) = 0
mmap(NULL, 24321, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f717ec61000
close(3)                                = 0
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
open("/usr/local/lib/libtcti-socket.so.0", O_RDONLY|O_CLOEXEC) = 3
read(3, "\177ELF\2\1\1\0\0\0\0\0\0\0\0\0\3\0>\0\1\0\0\0\340\24\0\0\0\0\0\0"..., 832) = 832
fstat(3, {st_mode=S_IFREG|0755, st_size=65296, ...}) = 0
mmap(NULL, 2114656, PROT_READ|PROT_EXEC, MAP_PRIVATE|MAP_DENYWRITE, 3, 0) = 0x7f717e842000
mprotect(0x7f717e846000, 2093056, PROT_NONE) = 0
mmap(0x7f717ea45000, 8192, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_DENYWRITE, 3, 0x3000) = 0x7f717ea45000
close(3)                                = 0
access("/etc/ld.so.nohwcap", F_OK)      = -1 ENOENT (No such file or directory)
[...]

PR #665, opened 2017-11-28
tpm2_rsaencrypt: fix file argument

williamcroberts:
Based on argument ordering and getopt handling, the code was getting lucky and
actually properly grabing the input file at argv[1]. However, the first
item in the array is really argv[0].

A result could be that the input file is NULL, which means read data
from stdin and the tool will hang.

Fixes: #664


williamcroberts:
Thankfully this looks like the only instance of this bug in the code.

PR #688, opened 2017-12-08
Makefile.am: Add LICENSE in tarball generated by the make dist target

martinezjavier:
The LICENSE file that's present in the source code is not distributed in the
tarball generated by the make dist target. Including full license text makes
it easier for people to comply with the desired license terms.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -249,7 +249,8 @@ test_unit_test_tpm2_errata_SOURCES  = test/unit/test_tpm2_errata.c
 
 endif
 
-EXTRA_DIST = $(top_srcdir)/man
+EXTRA_DIST = $(top_srcdir)/man \
+	     LICENSE
 
 if HAVE_PANDOC
     man1_MANS := \


martinezjavier:
@williamcroberts this should also go to 3.0... sorry that I just found this after the release, but it's because we didn't have a dist tarball in the past in https://github.com/intel/tpm2-tools and just the source code tarball.

For now I just left the package getting the tarball from:

https://github.com/intel/tpm2-tools/archive/%{version}/%{name}-%{version}.tar.gz

But it would be nice to get this fix for the next releases so we instead use:

https://github.com/intel/tpm2-tools/releases/download/%{version}/%{name}-%{version}.tar.gz

williamcroberts:
@martinezjavier it's too late now. We'll have to add it and cut 3.0.1 on Monday.

williamcroberts:
Can you send a cherry-pick to 3.X branch?

martinezjavier:
@williamcroberts done, PR #689.

Yes, no worries. As mentioned this doesn't even block packagers to update to the final release. It's just nicer to use the generated tarball, but it's hard to do it without a LICENSE file.

Thanks!

flihp:
I missed a few of the *.md files in the tss2 release as well :(

PR #694, opened 2017-12-12
configure: get version from a file

emmanuel-deloget:
Use the idea in tpm2-abrmd to generate a VERSION file and use this file
to get the software version instead of relying on the fact we're in an
active git tree (this is not the case when we fetch a tarball).

The VERSION file generated by bootstrap is then included in the
distribution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ missing
 Makefile
 Makefile.in
 test-driver
+VERSION
--- a/Makefile.am
+++ b/Makefile.am
@@ -230,7 +230,8 @@ test_unit_test_tpm2_password_util_SOURCES  = test/unit/test_tpm2_password_util.c
 
 endif
 
-EXTRA_DIST = $(top_srcdir)/man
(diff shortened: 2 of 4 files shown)


emmanuel-deloget:
Please ignore - I'll send a rebased one.

williamcroberts:
I'm not still convinced it's needed. I downloaded the tarball from `make dist` and ran `./configure`, and looking at the logs:
```
config.log:585:PACKAGE_STRING='tpm2-tools 3.0.1'
config.log:588:PACKAGE_VERSION='3.0.1'
config.log:610:VERSION='3.0.1'
```
It's pulling the version from somewhere. Im assuming the bootstrap command, since that's where it runs the version macro, is setting this up. It appears to be embedded in configure script.

With that said, I don't want to support this without reason. If you need to build from source, use git to clone it. If you need a source release, you have two options:
1. Use a dist tarball provided in the release section as a download
2. Use git and checkout a tag

Am I missing something here?

PR #695, opened 2017-12-12
configure: get version from a file

emmanuel-deloget:
Use the idea in tpm2-abrmd to generate a VERSION file and use this file
to get the software version instead of relying on the fact we're in an
active git tree (this is not the case when we fetch a tarball).

The VERSION file generated by bootstrap is then included in the
distribution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ missing
 Makefile
 Makefile.in
 test-driver
+VERSION
--- a/Makefile.am
+++ b/Makefile.am
@@ -257,7 +257,8 @@ EXTRA_DIST = $(top_srcdir)/man \
 	     LICENSE \
 	     MAINTAINERS.md \
 	     README.md \
-	     RELEASE.md
+	     RELEASE.md \
+	     VERSION
 
 if HAVE_PANDOC
     man1_MANS := \
--- a/bootstrap
+++ b/bootstrap
@@ -2,4 +2,5 @@
 
 AUTORECONF=${AUTORECONF:-autoreconf}
 
+git describe --tags --always --dirty > VERSION
 ${AUTORECONF} --install --sym
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_INIT([tpm2-tools],
-    [m4_esyscmd_s([git describe --tags --always --dirty])])
+    [m4_esyscmd_s([cat ./VERSION])])
 AC_CONFIG_MACRO_DIR([m4])
 AC_PROG_CC
 LT_INIT


martinezjavier:
@emmanuel-deloget I've also been bitten by this in the past (download a source tarball, build outside a git tree and get an empty PACKAGE_VERSION).

Now, with the tarball generated with the `make dist` target this is not an issue. Since it will be generated from an active git tree and the generated configure script will have a proper PACKAGE_VERSION.

When fetching a source code tarball as you said it's currently not working, since `git describe --tags --always --dirty` will fail due not being in a git repository. In this case the PACKAGE_VERSION in the configure script will just be empty.

I fail to understand though what's the difference with your approach than the current one. Since in both cases the user will have to execute `bootstrap` when downloading the source code tarball. So even if a VERSION file is distributed, it will be destroyed when the user executes `bootstrap`.

Sorry if I'm missing something.

emmanuel-deloget:
@martinezjavier you're not missing much. I was bitten by this a few days ago while trying to integrate the various tpm2-* projects from Intel in LEDE (where I do not need to execute bootstrap and I can provide my own VERSION file if needed). 

Now that you say it, I assumed that bootstrap was not needed in the packaged versions (i.e. it's executed before make dist and the archive contains both the VERSION file and the configure script). If this is wrong then yes, my change do not have any impact.

martinezjavier:
@emmanuel-deloget I'm not sure how the source code tarballs (i.e: https://github.com/intel/tpm2-tools/archive/3.0.1.tar.gz) are generated.

But if you download it, you can see that there's only a configure.ac and Makefile.am files. Neither a configure script nor Makefile.in. And the INSTALL.md fails says that you have to execute `./bootstrap`.

Assuming that the VERSION file is there and was generated, then it may work but only if users execute something like `autoreconf --install --sym` directly. If they use the `bootstrap` script, then will overwrite VERSION.

For the generated tarball with `make dist` (i.e: https://github.com/intel/tpm2-tools/releases/download/3.0.1/tpm2-tools-3.0.1.tar.gz), there's no bootstrap and already contain a generated configure script with the proper PACKAGE_VERSION.

So I would say that the proper solution for you is to use the latter instead of the former when building your packages.

williamcroberts:
I'm not still convinced it's needed. I downloaded the tarball from `make dist` and ran `./configure`, and looking at the logs:
```
config.log:585:PACKAGE_STRING='tpm2-tools 3.0.1'
config.log:588:PACKAGE_VERSION='3.0.1'
config.log:610:VERSION='3.0.1'
```
It's pulling the version from somewhere. Im assuming the bootstrap command, since that's where it runs the version macro, is setting this up. It appears to be embedded in configure script.

With that said, I don't want to support this without reason. If you need to build from source, use git to clone it. If you need a source release, you have two options:
1. Use a dist tarball provided in the release section as a download
2. Use git and checkout a tag

Am I missing something here?

williamcroberts:
@martinezjavier I generate those tarballs I put on the release tags with a simple `make dist`.

martinezjavier:
> Am I missing something here?

I don't think you are missing anything, that's basically what I also said.

martinezjavier:
> @martinezjavier I generate those tarballs I put on the release tags with a simple make dist.

I didn't mean for the https://github.com/intel/tpm2-tools/releases/download/3.0.1/tpm2-tools-3.0.1.tar.gz

I meant for https://github.com/intel/tpm2-tools/archive/3.0.1.tar.gz

Those tarballs definitely have different files so I don't think both are generated with `make dist`.

williamcroberts:
Oh I thought I had that link right. The archive link is just the same as the "download zip" link but for a tarball. This link to Stack Overflow is pertinent: https://stackoverflow.com/questions/8377081/github-api-download-zip-or-tarball-link

martinezjavier:
@williamcroberts ah, I see. So the source code tarball is basically snapshot of the git repo with the .git directory removed.

Anyways, my argument was that this change isn't needed nor a different over the current behaviour. So if @emmanuel-deloget agrees we should just close this PR.

martinezjavier:
@williamcroberts unrelated to this PR but something I noticed when comparing the files for both tarballs, the integration tests are not distributed by `make dist`. Is that on purpose or just missing in the `EXTRA_DIST` ?

williamcroberts:
@martinezjavier like everything, its missing EXTRA_DIST :+1: 

Oh and I want to comment on why @flihp added the VERSION file. His is a hardcoded version number, as get describe was providing weird numbers on master, for instance the tools master:
```
 $ git describe 
2.0.0-738-g459a42b
```
I'm not too concerned with master being off, as the commit number gets us some details.

martinezjavier:
Ok, I'll add the tests latter today unless someone beats me.

emmanuel-deloget:
After carefull thinking, this change is not necessary (I'll fix it within my LEDE dev tree ince I don't execute the bootstrap script as part of the build process ; but I admit I'm sometimes a bit weird).

PR #696, opened 2017-12-12
tpm2_clear: add tool

emmanuel-deloget:
Add a (crude) tpm2_clear tool which is used to clear the TPM2 and
restore it to a nearly blank state.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -49,6 +49,7 @@ LDADD = \
 bin_PROGRAMS = \
     tools/tpm2_activatecredential \
     tools/tpm2_certify \
+    tools/tpm2_clear \
     tools/tpm2_create \
     tools/tpm2_createpolicy \
     tools/tpm2_createprimary \
@@ -147,6 +148,7 @@ lib_libcommon_a_SOURCES = \
 
 TOOL_SRC := tools/tpm2_tool.c tools/tpm2_tool.h
 
+tools_tpm2_clear_SOURCES = tools/tpm2_clear.c $(TOOL_SRC)
 tools_tpm2_create_SOURCES = tools/tpm2_create.c $(TOOL_SRC)
 tools_tpm2_createprimary_SOURCES = tools/tpm2_createprimary.c $(TOOL_SRC)
 tools_tpm2_getcap_SOURCES = tools/tpm2_getcap.c $(TOOL_SRC)
@@ -263,6 +265,7 @@ if HAVE_PANDOC
     man1_MANS := \
     man/man1/tpm2_activatecredential.1 \
     man/man1/tpm2_certify.1 \
+    man/man1/tpm2_clear.1 \
     man/man1/tpm2_create.1 \
     man/man1/tpm2_createpolicy.1 \
     man/man1/tpm2_createprimary.1 \
--- a/man/tpm2_clear.1.md
+++ b/man/tpm2_clear.1.md
@@ -1,0 +1,43 @@
+tpm2_clear 1 "DECEMBER 2017" tpm2-tools
+==================================================
+
+NAME
+----
+
+tpm2_clear(1) - Send a clear command to the TPM.
+
+SYNOPSIS
+--------
+
+`tpm2_clear` [OPTIONS]
+
+DESCRIPTION
+-----------
+
+tpm2_clear(1) Send a `TPM2_Clear` command with `TPM2_RH_PLATFORM`.
+
+OPTIONS
+-------
+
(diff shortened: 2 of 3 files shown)


martinezjavier:
@emmanuel-deloget the tpm2_takeownership already has a `-c` option which I thought was used to clear the TPM, what's the difference between that option and this new tool?

emmanuel-deloget:
@martinezjavier tpm2_takeownership -c does a TPM2_Clear with TPM2_RH_LOCKOUT, this one does a TPM2_RH_PLATFORM which does not fail when you're DA-locked.

martinezjavier:
@emmanuel-deloget ah, I see. Wonder if we should either add this option to tpm2_takeownership or make tpm2_clear allow to either clear using either TPM2_RH_PLATFORM or TPM2_RH_LOCKOUT.

Also, IIUC your tool assumes that there is an empty password for the platform hierarchy but AFAICT this is usually not the case. I see that the platform hierarchy authorization is set to a random value at startup by the firmware.

Probably that's something that should be documented in the tool man page.

emmanuel-deloget:
@martinezjavier both -p and -P are already taken on tpm2_takeownership :) More seriously, thet fact that we work on the platform hierarchy here would be at odd with the goal of tpm2_takeowership (that' s how I conseive it but I may be wrong). 

About the empty password: as I said in my PR, this is a crude version of the program (I needed it, so I built it). Adding a lockoutAuth password should not be too difficult. In my tests, such a password was not needed (I used it to bypass the lockoutAuth in order to restore the tpm2 in a working state after I made a few errors). 

Also, I find weird that this password may be needed -- I cannot find this in the spec, so if you have a pointer, I'll happily read it. If the firmware really set it to a random value at startup, that would make this command entirely non-fonctionnal (as I have no way to retrieve this password). And of course, clearing the platform hierarchy is already PP-protected.

martinezjavier:
@emmanuel-deloget I'm not saying your tool is not needed, in fact I like it since the convention is to have the tools named after the TPM2 command code. Son for consistency with the other tools `tpm2_clear` is much better than `tpm2_takeownership -c` IMHO.

I just don't want to have two ways to do similar things in the tools, so users may be confused (as I was) about the difference between `tpm2_takeownsership -c` and `tpm2_clear`.

Now, going back to the password issue. I get the following when trying your tool on my Intel PTT TPM2 implemented in the ME:

```
$ tpm2_clear                                                 
ERROR: Tss2_Sys_Clear failed: 0x9a2                    
$ tpm2_rc_decode 0x9a2                                               
error layer                
  hex: 0x0                 
  identifier: TSS2_TPM_RC_LAYER                        
  description: Error produced by the TPM               
format 1 error code        
  hex: 0x22                
  identifier: TPM2_RC_BAD_AUTH                         
  description: authorization failure without DA implications                                                   
session                    
  hex: 0x100               
  identifier: TPM2_RC_1    
  description:  (null)
```

Which is caused by the fact that the platform hierarchy has an authorization password set. My understanding is that's true for most TPM2 out there.

This is another example: https://github.com/intel/tpm2-tss/issues/354

williamcroberts:
I would be OK with splitting out tpm2_takeownership as it's using two tpm functions that actually change this (i'm ok with reading public data in a tool to get a default size or something).
tpm2_takeownership could be tpm2_clear and tpm2_changeauth (or something like that).

emmanuel-deloget:
@williamcroberts @martinezjavier So to get this straight : 

1) remove option -c on tpm2_takeownership
2) rename tmp2_takeowership to tpm2_changeauth
3) make a tpm2_clear with the following characteristics
   - allow to work on the platform or lockout hierarchy
   - allow a (lockout) password to be specified

Am I correct?

While I'm at it, the tpm2_clear tool might also be a good place to implement TPM2_ClearControl() (that would disallow TPM2_Clear()). How does that sound?

martinezjavier:
@emmanuel-deloget sounds good to me. My only doubt now is if we should have a separate tpm2_clearcontrol tool... we have for example tpm2_nvreadlock and tpm2_nvread as separate tools.

But I don't really have a strong opinion on that, so whatever @williamcroberts and you decide.

williamcroberts:
@martinezjavier @emmanuel-deloget let's keep them seperate tools for clearcontrol and clear, since it does fall in line with nvread and nvreadlock.

williamcroberts:
Issue created for adding a clearlock tool: https://github.com/intel/tpm2-tools/issues/710

PR #697, opened 2017-12-12
merge 3.X into master

williamcroberts:
Lots of empty commits, but this forces a sync between the two branches where it matters and gets git describe version on master reporting 3.0 as the nearest tag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,8 +43,8 @@ install:
     - mkdir ibmtpm974 && pushd ibmtpm974 && tar axf ../ibmtpm974.tar.gz && pushd ./src && make
     - ./tpm_server &
     - popd && popd
-    - git clone https://github.com/01org/TPM2.0-TSS.git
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -43,8 +43,8 @@ install:
     - mkdir ibmtpm974 && pushd ibmtpm974 && tar axf ../ibmtpm974.tar.gz && pushd ./src && make
     - ./tpm_server &
     - popd && popd
-    - git clone https://github.com/01org/TPM2.0-TSS.git
-    - pushd TPM2.0-TSS
```
martinezjavier: I did the same change when fixing the Travis CI build on PR #701.

So probably you want to drop this commit and merge on top? Or I can re-spin my branch without that change, but it would create conflicts anyways.

williamcroberts: Merges make everything harder, ill fix up your branch after this.

PR #701, opened 2017-12-13
travis.yml: build and install autoconf-archive from source

martinezjavier:
Commit 590ed9d2ed21 ("travis: fix autconf dependency on tss") fixed the
tpm2-tss dependency with the AX_CODE_COVERAGE macro, by fetching latest
autoconf-archive and copying the macro into the tpm2-tss source code m4
directory.

But now tpm2-abrmd also uses the macro for code coverage and so instead
of duplicating the same logic, install autoconf-archive in the machine.

While being there, clone the tpm2-tss repository from the latest URL.


williamcroberts:
Moved: https://github.com/intel/tpm2-tools/pull/704

PR #704, opened 2017-12-14
Martinez fix travis

williamcroberts:


williamcroberts:
@martinezjavier I wonder if we should fall in line to tss/abrmd and have enable code coverage and update the travis coveralls to use it.

martinezjavier:
@williamcroberts I think it would good, yes.

PR #720, opened 2017-12-15
tests: move test scripts to seperate directory

williamcroberts:
Organize the test scripts into a supordinate directory.

This makes finding tests easier as it doesn't really rely on
magic name prefixes.


martinezjavier:
I don't think this is suitable for the 3.X branch. After a release, we should only do changes that are not invasive to users. In fact, I would say that only bug fixes are suitable for a released branch.

williamcroberts:
@martinezjavier Ill drop this from 3.X, as I consider this not-backwards-compatible. But as far as release branches go, new features will get a minor version bump. Ie if we ever add tpm2-import to 3.X the release would go to 3.1. Bug fixes increase the patch level, so 3.0.1 becomes 3.0.2. I might be abusing terms, but it's laid out in semantic versioning.

PR #721, opened 2017-12-15
tests: move test scripts to separate directory

williamcroberts:
Organize the test scripts into a supordinate directory.

This makes finding tests easier as it doesn't really rely on
magic name prefixes.


tools/tpm2_getmanufec.c
```diff
@@ -410,15 +410,19 @@ int RetrieveEndorsementCredentials(char *b64h)
...
+        goto out;
     }
 
     LOG_INFO("%s", b64);
 
-    int rc = RetrieveEndorsementCredentials(b64);
+    rc = RetrieveEndorsementCredentials(b64);
+out:
+    free(hash);
     free(b64);
```
martinezjavier: This is an unrelated change and should be in a separate patch.

williamcroberts: It is, it's on 0ec3a5a where as the move is on ff44794

martinezjavier: @williamcroberts ah, sorry. For some reasons I thought it was part of the same commit. Sorry for the noise.

PR #722, opened 2017-12-16
nv: don't exceed max nv buffer size for TSS

snits:
Currently if a tpm device supports a max nv buffer size
greater than tss supports we can get buffer overflows. Check
the size returned by the tpm device, and use the min of
that and the max supported by tss.


snits:
resending with style change

PR #734, opened 2017-12-18
tpm2_print: A tool to print TPM data structures

harisokanovic:
Takes an object type + file path as inputs and prints enclosed elements
to stdout in YAML.

Presently only supports TPMS_ATTEST objects for the purpose of printing
TPM generated quotes. Does NOT authenticate input file, which can be
done using other tools -- E.g. see edc3e022 for openssl example.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -48,6 +48,7 @@ LDADD = \
 
 # keep me sorted
 bin_PROGRAMS = \
+    tools/aux/tpm2_print \
     tools/tpm2_activatecredential \
     tools/tpm2_certify \
     tools/tpm2_changeauth \
@@ -154,6 +155,8 @@ lib_libcommon_a_SOURCES = \
     lib/tpm2_session.h
 
 TOOL_SRC := tools/tpm2_tool.c tools/tpm2_tool.h
+
+tools_aux_tpm2_print_SOURCES = tools/aux/tpm2_print.c $(TOOL_SRC)
 
 tools_tpm2_clear_SOURCES = tools/tpm2_clear.c $(TOOL_SRC)
 tools_tpm2_clearlock_SOURCES = tools/tpm2_clearlock.c $(TOOL_SRC)
--- a/lib/tpm2_options.h
+++ b/lib/tpm2_options.h
@@ -100,8 +100,11 @@ typedef bool (*tpm2_arg_handler)(int argc, char **argv);
  *
  * TPM2_OPTIONS_SHOW_USAGE:
  *  Enable printing a short usage summary (I.e. help)
+ * TPM2_OPTIONS_NO_SAPI:
+ *  Skip SAPI initialization. Removes the "-T" common option.
  */
 #define TPM2_OPTIONS_SHOW_USAGE 0x1
+#define TPM2_OPTIONS_NO_SAPI 0x2
 
 struct tpm2_options {
     struct {
--- a/lib/tpm2_options.h
+++ b/lib/tpm2_options.h
@@ -95,6 +95,14 @@ typedef bool (*tpm2_option_handler)(char key, char *value);
  */
 typedef bool (*tpm2_arg_handler)(int argc, char **argv);
 
+/**
+ * TPM2_OPTIONS_* flags change default behavior of the argument parser
+ *
+ * TPM2_OPTIONS_SHOW_USAGE:
+ *  Enable printing a short usage summary (I.e. help)
+ */
+#define TPM2_OPTIONS_SHOW_USAGE 0x1
+
 struct tpm2_options {
     struct {
         tpm2_option_handler on_opt;
@@ -102,7 +110,7 @@ struct tpm2_options {
     } callbacks;
     char *short_opts;
     size_t len;
-    bool show_usage;
+    UINT32 flags;
     struct option long_opts[];
 };
 
@@ -123,14 +131,14 @@ typedef struct tpm2_options tpm2_options;
  * @param on_arg
  *  An argument handling callback, which may be null if you don't wish
  *  to handle arguments.
- * @param show_usage
- *  Whether the tool wants a usage short summary text to be printed.
+ * @param flags
+ *  TPM2_OPTIONS_* bit flags
  * @return
  *  NULL on failure or an initialized tpm2_options object.
  */
 tpm2_options *tpm2_options_new(const char *short_opts, size_t len,
         const struct option *long_opts, tpm2_option_handler on_opt,
-        tpm2_arg_handler on_arg, bool show_usage);
+        tpm2_arg_handler on_arg, UINT32 flags);
 
 /**
  * Concatenates two tpm2_options objects, with src appended on
--- a/lib/tpm2_util.c
+++ b/lib/tpm2_util.c
@@ -346,6 +346,12 @@ static void tpm2_util_public_to_keydata(TPM2B_PUBLIC *public, tpm2_util_keydata 
     return;
 }
 
+void print_yaml_indent(size_t indent_count) {
+    while (indent_count--) {
+        tpm2_tool_output("  ");
+    }
+}
+
 void tpm2_util_public_to_yaml(TPM2B_PUBLIC *public) {
 
     tpm2_tool_output("algorithm:\n");
--- a/lib/tpm2_util.c
+++ b/lib/tpm2_util.c
@@ -31,9 +31,9 @@
 #include <ctype.h>
 #include <errno.h>
 #include <stdbool.h>
-#include <stdio.h>
 
 #include "log.h"
+#include "files.h"
 #include "tpm2_alg_util.h"
 #include "tpm2_attr_util.h"
 #include "tpm2_tool.h"
@@ -132,7 +132,7 @@ int tpm2_util_hex_to_byte_structure(const char *inStr, UINT16 *byteLength,
     return 0;
 }
 
-void tpm2_util_hexdump(BYTE *data, size_t len, bool plain) {
+void tpm2_util_hexdump(const BYTE *data, size_t len, bool plain) {
 
     if (!output_enabled) {
         return;
@@ -168,6 +168,26 @@ void tpm2_util_hexdump(BYTE *data, size_t len, bool plain) {
         }
         printf("\n");
     }
+}
+
+bool tpm2_util_hexdump_file(FILE *fd, size_t len, bool plain) {
+    BYTE* buff = (BYTE*)malloc(len);
+    if (!buff) {
+        LOG_ERR("malloc() failed");
+        return false;
+    }
+
+    bool res = files_read_bytes(fd, buff, len);
+    if (!res) {
+        LOG_ERR("Failed to read file");
+        free(buff);
+        return false;
+    }
+
+    tpm2_util_hexdump(buff, len, plain);
+
+    free(buff);
+    return true;
 }
 
 /* TODO OPTIMIZE ME */
--- a/lib/tpm2_util.c
+++ b/lib/tpm2_util.c
@@ -190,6 +190,17 @@ bool tpm2_util_hexdump_file(FILE *fd, size_t len, bool plain) {
     return true;
 }
 
+bool tpm2_util_print_tpm2b_file(FILE *fd)
+{
+    UINT16 len;
+    bool res = files_read_16(fd, &len);
+    if(!res) {
+        LOG_ERR("File read failed");
+        return false;
+    }
+    return tpm2_util_hexdump_file(fd, len, true);
+}
+
 /* TODO OPTIMIZE ME */
 UINT16 tpm2_util_copy_tpm2b(TPM2B *dest, TPM2B *src) {
     int i;
--- a/lib/tpm2_util.h
+++ b/lib/tpm2_util.h
@@ -33,6 +33,7 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
 
 #include <sapi/tpm20.h>
 
@@ -159,7 +160,24 @@ bool tpm2_util_string_to_uint16(const char *str, uint16_t *value);
  *  true for a plain hex string false for an xxd compatable
  *  dump.
  */
-void tpm2_util_hexdump(BYTE *data, size_t len, bool plain);
+void tpm2_util_hexdump(const BYTE *data, size_t len, bool plain);
+
+/**
+ * Prints an xxd compatible hexdump to stdout if output is enabled,
+ * ie no -Q option.
+ *
+ * @param fd
+ *  A readable open file.
+ * @param len
+ *  The length of the data to read and print.
+ * @param plain
+ *  true for a plain hex string false for an xxd compatable
+ *  dump.
+ * @return
+ *  true if len bytes were successfully read and printed,
+ *  false otherwise
+ */
+bool tpm2_util_hexdump_file(FILE *fd, size_t len, bool plain);
 
 /**
  * Prints a TPM2B as a hex dump.
--- a/lib/tpm2_util.h
+++ b/lib/tpm2_util.h
@@ -281,6 +281,13 @@ UINT64 tpm2_util_ntoh_64(UINT64 data);
 UINT32 tpm2_util_pop_count(UINT32 data);
 
 /**
+ * Prints whitespace indention for yaml output.
+ * @param indent_count
+ *  Number of times to indent
+ */
+void print_yaml_indent(size_t indent_count);
+
+/**
  * Convert a TPM2B_PUBLIC into a yaml format and output if not quiet.
  * @param public
  *  The TPM2B_PUBLIC to output in YAML format.
--- a/lib/tpm2_util.h
+++ b/lib/tpm2_util.h
@@ -189,6 +189,13 @@ static inline void tpm2_util_print_tpm2b(TPM2B *buffer) {
 }
 
 /**
+ * Reads a TPM2B object from FILE* and prints data in hex.
+ * @param fd
+ *  A readable open file.
+ */
+bool tpm2_util_print_tpm2b_file(FILE *fd);
+
+/**
  * Copies a tpm2b from dest to src and clears dest if src is NULL.
  * If src is NULL, it is a NOP.
  * @param dest
--- a/man/tpm2_print.1.md
+++ b/man/tpm2_print.1.md
@@ -1,0 +1,51 @@
+% tpm2_print(1) tpm2-tools | General Commands Manual
+%
+% DECEMBER 2017
+
+# NAME
+
+**tpm2_print**(1) - Prints TPM data structures
+
+# SYNOPSIS
+
+**tpm2_print** [*OPTIONS*]
+
+# DESCRIPTION
+
+**tpm2_print**(1) decodes a TPM data structure and prints enclosed
+elements to stdout as YAML.
+
+# OPTIONS
+
+  * **-t**, **--type**:
+
+    Required. Type of data structure. Only TPMS_ATTEST is presently
+    supported.
+
+  * **-f**, **--file**:
+
+    Optional. File containing TPM object. Reads from stdin if unspecified.
+
+[common options](common/options.md)
+
+[common tcti options](common/tcti.md)
+
+# EXAMPLES
+
+```
+tpm2_print -t TPMS_ATTEST -f /path/to/tpm/quote
+tpm2_print --type=TPMS_ATTEST --file=/path/to/tpm/quote
+cat /path/to/tpm/quote | tpm2_print --type=TPMS_ATTEST
+```
+
+# RETURNS
+
+0 on success. Non-zero otherwise.
+
+# BUGS
+
+[Github Issues](https://github.com/01org/tpm2-tools/issues)
+
+# HELP
+
+See the [Mailing List](https://lists.01.org/mailman/listinfo/tpm2)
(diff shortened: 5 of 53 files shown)


martinezjavier:
@harisokanovic can you please fix the compile error and resend?

```
 CC       tools/aux/tpm2_print.o
../tools/aux/tpm2_print.c: In function ‘print_TPMS_QUOTE_INFO’:
../tools/aux/tpm2_print.c:119:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for(long unsigned int i = 0; i < pcr_selection_count; ++i) {
     ^
../tools/aux/tpm2_print.c:119:5: note: use option -std=c99 or -std=gnu99 to compile your code
make[1]: *** [tools/aux/tpm2_print.o] Error 1
make[1]: Leaving directory `/home/travis/build/intel/tpm2-tools/tpm2-tools-3.0.1-54-gb286bf3/_build'
make: *** [distcheck] Error 1
```

lib/tpm2_util.c
```diff
@@ -132,42 +132,71 @@ int tpm2_util_hex_to_byte_structure(const char *inStr, UINT16 *byteLength,
     return 0;
 }
 
-void tpm2_util_hexdump(BYTE *data, size_t len, bool plain) {
+void tpm2_util_hexdump(const BYTE *data, size_t len, bool plain) {
+    FILE *fd = fmemopen((void*)data, len, "r");
+    if (!fd) {
+        /* TODO: should we change the API to allow error? */
```
martinezjavier: AFAICT, the only caller of this function currently is tpm2_util_public_to_yaml(), so it should be quite easy to change the function signature to propagate the error.

That said, I wonder if instead of using fmemopen() and calling tpm2_util_hexdump_file(), can we do the opposite and make tpm2_util_hexdump_file() read the file contents and call tpm2_util_hexdump()?

Since once the data is in memory, tpm2_util_hexdump() shouldn't really have an error path.

lib/tpm2_util.h
```diff
@@ -155,7 +156,24 @@ bool tpm2_util_string_to_uint16(const char *str, uint16_t *value);
...
+/**
+ * Prints an xxd compatible hexdump to stdout if output is enabled,
+ * ie no -Q option.
+ *
+ * @param fd
+ *  A readable open file.
+ * @param len
+ *  The length of the data to read and print.
+ * @param plain
+ *  true for a plain hex string false for an xxd compatable
```
martinezjavier: s/compatable/compatible

Makefile.am
```diff
@@ -48,6 +48,7 @@ LDADD = \
 
 # keep me sorted
 bin_PROGRAMS = \
+    tools/aux/tpm2_print \
```
martinezjavier: Did we decide that there would be an aux sub-directory? I remember the discussion but I had that at the end we decided to just have all the tools under tools.

man/tpm2_print.1.md
```diff
@@ -0,0 +1,51 @@
+% tpm2_print(1) tpm2-tools | General Commands Manual
+%
+% DECEMBER 2018
```
martinezjavier: I don't think this patch will take a year to be merged :smiley:

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+
+#include <stdio.h>
```
martinezjavier: Please insert a blank line here to separate system headers from project headers.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+#include "log.h"
+#include "files.h"
+#include "tpm2_util.h"
+#include "tpm2_alg_util.h"
+#include "tpm2_tool.h"
```
martinezjavier: Please sort these alphabetically.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+const char* file_path = "";
+
+#define FILE_2_SHORT_ERROR() LOG_ERR("%s: File too short", (file_path))
+
+static inline void print_yaml_indent(size_t indent_count)
+{
+    while (indent_count-- > 0) {
+        tpm2_tool_output("  ");
+    }
+}
```
martinezjavier: This probably should be in lib/tpm2_util.{c,h}

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+
+typedef enum {
+    file_type_unknown = 0,
+    file_type_TPMS_ATTEST
+} file_type_id;
+
+file_type_id file_type = file_type_unknown;
+const char* file_path = "";
+
+#define FILE_2_SHORT_ERROR() LOG_ERR("%s: File too short", (file_path))
```
martinezjavier: This is only used in two places, so I wonder the value of the macro. But it's probably just my personal taste, since I think that is an unnecessary level of indirection and makes the code harder to read.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+    tpm2_tool_output("restartCount: %lu\n", (long unsigned int)numb.u32);
+
+    res = files_read_bytes(fd, &numb.u8, 1);
+    if (!res) {
+        goto read_error;
+    }
+    print_yaml_indent(indent_count);
+    tpm2_tool_output("safe: %u\n", (unsigned int)numb.u8);
+
+    // success
```
martinezjavier: I think this comment is superfluous.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+    tpm2_tool_output("pcrSelect:\n");
+
+    print_yaml_indent(indent_count + 1);
+    tpm2_tool_output("count: %lu\n", (long unsigned int)pcr_selection_count);
+
+    print_yaml_indent(indent_count + 1);
+    tpm2_tool_output("pcrSelections:\n");
+
+    // read TPML_PCR_SELECTION array (of size count)
+    for(long unsigned int i = 0; i < pcr_selection_count; ++i) {
```
martinezjavier: This `i` declaration is what should be fixed.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+    };
+
+    *opts = tpm2_options_new("t:f:", ARRAY_LEN(topts), topts,
+        on_option, NULL);
+
+    return *opts != NULL;
+}
+
+int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
+    UNUSED(sapi_context);
```
martinezjavier: Not related with your patch, but this tool could be used even if there's no TPM available. But currently it will fail even when the SAPI context is not used. I've filled issue #737 for that.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+    UNUSED(flags);
+
+    bool res;
+    FILE* fd = stdin;
+
+    if (file_path && *file_path != 0) {
```
martinezjavier: Maybe change to `if (file_path && *file_path)` for consistency?

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+            goto error;
+        }
+        break;
+    default:
+        LOG_ERR("Invalid file_type");
+        goto error;
+        break;
+    }
+
+    // log number of bytes read
```
martinezjavier: Again, this and the following `// success` comments could be removed IMHO.

Makefile.am
williamcroberts: I don't think we ever concluded, I'm OK with an aux tool directory, what about you @martinezjavier ?

lib/tpm2_util.c
williamcroberts: I think I prefer it this way, as potentially big buffers won't need to be fully read into memory. Granted, most tpm data structures are quite small. Either way is fine by me in the end.

tools/aux/tpm2_print.c
williamcroberts: Yeah perhaps we need a way in onstart to indicate that they don't need a SAPI. tpm2_rc_decode has the same issue, and could go to aux as well at some point.

tools/aux/tpm2_print.c
williamcroberts: fopen("", "rb) will fail, don't even need that check.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+read_error:
+    FILE_2_SHORT_ERROR();
+error:
+    return false;
+}
+
+static bool on_option(char key, char *value) {
+    switch (key) {
+    case 't':
+        if (strcmp(value, "TPMS_ATTEST") == 0) {
```
williamcroberts: Flipping the logic makes this simpler. If not TPMS_ATTEST, log an error and return, no need for an else clause, just assign to type.
```
case 't':
    if (strcmp(value, "TPMS_ATTEST"))  {
        LOG_ERR("msg"):
        return 1;
    }
    file_type = file_type_TPMS_ATTEST;
    break;
```

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+    // success
+    return true;
+
+read_error:
+    FILE_2_SHORT_ERROR();
+error:
+    return false;
+}
+
+static bool print_TPMS_ATTEST_yaml(FILE* fd, size_t indent_count) {
```
williamcroberts: Whats the point if indent_count, all I see is it being set to 0, subordinate calls are settiing it to other values, so I see no point in setting it here.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+    bool res;
```
williamcroberts: declare and assign at first use on 196 per the coding standard.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,336 @@
...
+} file_type_id;
+
+file_type_id file_type = file_type_unknown;
+const char* file_path = "";
+
+#define FILE_2_SHORT_ERROR() LOG_ERR("%s: File too short", (file_path))
+
+static inline void print_yaml_indent(size_t indent_count)
+{
+    while (indent_count-- > 0) {
```
williamcroberts: no need for ```> 0``` when indent_count is 0, it's false and will stop.

tools/aux/tpm2_print.c
williamcroberts: Yes keep types consistent, I know their are some sketchy spots in this codebase, but Id like to not add more.

man/tpm2_print.1.md
```diff
@@ -0,0 +1,51 @@
...
+**tpm2_print** [*OPTIONS*]
+
+# DESCRIPTION
+
+**tpm2_print**(1) decodes a TPM data structure and prints enclosed
+elements to stdout as YAML.
+
+# OPTIONS
+
+  * **-t**, **--type**:
```
williamcroberts: Is there anyway we could figure this out for the user? I didn't see anything as a dead giveaway...

man/tpm2_print.1.md
harisokanovic: :) fixed

tools/aux/tpm2_print.c
harisokanovic: Added

tools/aux/tpm2_print.c
harisokanovic: done

tools/aux/tpm2_print.c
harisokanovic: Moved

tools/aux/tpm2_print.c
harisokanovic: Alright, I removed the macro since both you and William don't like it :)

tools/aux/tpm2_print.c
harisokanovic: Removed

tools/aux/tpm2_print.c
harisokanovic: Fixed. I guess my compiler defaults to C99 and missed this.

harisokanovic: I used `long unsigned int` to avoid extra casting on prints below. Longs are guaranteed to be at least 32 bits wide, but I can change it if you prefer.

tools/aux/tpm2_print.c
harisokanovic: I noticed that as well, but it hasn't been an issue for us yet. I'd prefer to compile out the SAPI initialization instead of a runtime flag, so that we can remove the library dependency as well.

tools/aux/tpm2_print.c
harisokanovic: `file_path` defaults to empty string, so we need `*file_path`.

tools/aux/tpm2_print.c
harisokanovic: Removed

tools/aux/tpm2_print.c
harisokanovic: Removed

tools/aux/tpm2_print.c
harisokanovic: I was just being consistent with the other print_*() functions. I can remove it.

tools/aux/tpm2_print.c
harisokanovic: Fixed

tools/aux/tpm2_print.c
harisokanovic: Changed

lib/tpm2_util.h
harisokanovic: Fixed

Makefile.am
martinezjavier: @williamcroberts I'm also OK with it. I just asked since it was not clear to me that there was an agreement on that.

tools/aux/tpm2_print.c
williamcroberts: @harisokanovic you'd still need the headers for data types, so it would only remove the link dependency.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,314 @@
...
+}
+
+int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
+    UNUSED(sapi_context);
+    UNUSED(flags);
+
+    bool res;
+    FILE* fd = stdin;
+
+    if (file_path && *file_path) {
```
williamcroberts: NULL file_path, no need for it to be "".
If NULL set file_path to something like "<stdin>" in here.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,314 @@
...
+    default:
+        LOG_ERR("Invalid file_type");
+        goto error;
+        break;
+    }
+
+    // log number of bytes read
+    LOG_INFO("Read %ld bytes from file_path=%s", ftell(fd), file_path);
+
+    fclose(fd);
```
williamcroberts: looks like your closing stdin, you need to check if (fd == stdin) don't fclose it.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,314 @@
...
+    case file_type_TPMS_ATTEST:
+        res = print_TPMS_ATTEST_yaml(fd);
+        if(!res) {
+            goto error;
+        }
+        break;
+    default:
+        LOG_ERR("Invalid file_type");
+        goto error;
+        break;
```
williamcroberts: no need for break after goto, unless it silences a silly compiler warning.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    // log number of bytes read
+    LOG_INFO("Read %ld bytes from file_path=%s", ftell(fd), file_path);
+
+    fclose(fd);
+    return 0;
+
+error:
```
williamcroberts: you can just put an out label and return an rc variable. Less duplication of cleanup code.
```
function() {
    int rc = 1;
    do_stuff()
     if (fail) {
          goto out;
     }
    rc = 0;

out:
     cleanup();
     return rc;
```

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,314 @@
...
+            LOG_ERR("%s: Could not open file", file_path);
+            return 1;
+        }
+    }
+    else {
+        LOG_INFO("Reading from stdin");
+    }
+
+    switch (file_type) {
+    case file_type_TPMS_ATTEST:
```
williamcroberts: You could check this before even attempting to read the file.... this way you can fail early before you even have to worry about closing the file.

tools/aux/tpm2_print.c
```diff
@@ -0,0 +1,314 @@
...
+#include "tpm2_tool.h"
+#include "tpm2_util.h"
+
+typedef enum {
+    file_type_unknown = 0,
+    file_type_TPMS_ATTEST
+} file_type_id;
+
+file_type_id file_type = file_type_unknown;
+const char* file_path = "";
```
williamcroberts: NULL makes more sense here, also static for both vars.

Also, follow convention and create a context structure for these...
```
typedef struct tpm2_print_ctx tpm2_print_ctx;
struct tpm2_print_ctx {
  struct {
    const char *path;
    file_type_id type;
  } file;
};

static tpm2_print_ctx ctx;

foo() {
  ctx.file.path = "foo";
  ctx.file.type = 43;
}
```

tools/aux/tpm2_print.c
williamcroberts: If you do it that way, you can also use https://github.com/intel/tpm2-tools/blob/master/lib/files.h#L67:
```
/**
 * Loads data from a file path or stdin enforcing an upper bound on size.
 * @param path
 *  The path to load data from, NULL means stdin.
 * @param size
 *  The maximum size.
 * @param buf
 *  The buffer to write the data into.
 * @return
 *  True on success or false otherwise.
 */
bool files_load_bytes_from_file_or_stdin(const char *path, UINT16 *size, BYTE *buf);
```

man/tpm2_print.1.md
harisokanovic: We could do something like assume type=TPMS_ATTEST when the first 4 bytes (magic) are ff544347, but that's a bit flaky. I'm also not sure that every TPM generated object uses the same magic. We can always add this later without breaking the API.

harisokanovic:
@williamcroberts , @martinezjavier , I posted an update with changes you requested + a first attempt at splitting TCTI and SAPI initialization from the arg parser, so that we can avoid it in aux tools. Looks like travis-ci is hitting a build error on an unrelated file. Any ideas?

williamcroberts:
@harisokanovic Thank you for your patience and update to the PR. I will look to make sure that master is still building, offhand, looks like somethings changed in the TSS so master may be broken on travis. In the meantime, if you can rebase your change on the latest and fix the conflicts, that would be great.

harisokanovic:
@williamcroberts, looks like the travis-ci build succeeded

lib/tpm2_options.h
```diff
@@ -45,8 +45,12 @@ union tpm2_option_flags {
         UINT8 verbose : 1;
         UINT8 quiet   : 1;
         UINT8 enable_errata  : 1;
+        struct {
+            char *name;
+            char *opts;
+        } tcti;
     };
-    UINT8 all;
+    UINT8 all : 3;
```
williamcroberts: Why did you change this?

harisokanovic: I moved sapi initialization to avoid a run-time dependency on libsapi. However, it doesn't matter in my use case (yet anyway), so I could go either way on this point.

I added an explicit 3-bit deceleration for `all`, since the rest of the bits in this data structure are char*'s. I agree it's a bit weird to add strings to the flags struct. I can move it if there's a better/preferred place.

williamcroberts: The three bit bitfield on all, kill it off. It will likely generate more code as the compiler cant just write 0 to a byte address, it would have to mask off and clear only those bits... ie read a byte, and them with 0x7 and write it back.

For the sapi split, I am not worried about the link dependency, so you'll have to change that up to the tpm2_options_new flag approach.

Otherwise, once those are done, ill be happy to pull this in.

williamcroberts: FYI, you can use: https://github.com/intel/tpm2-tools/pull/767 as a sample to port to master. This may or may not be done by the time you get back to this.

harisokanovic:
@williamcroberts , I replaced the SAPI split with a "flags" bitmask in tpm2_options_new() as you suggested.

harisokanovic:
Great, thanks. I forgot about the man pages. We should also move `tpm2_rc_decode` into the aux/ directory and apply the NO_SAPI flag.

williamcroberts:
See #769 for existing work items

PR #751, opened 2017-12-22
configure.ac: add --with-manpages-help option

martinezjavier:
When the -h / --help option is passwed to the tools, their man pages are
shown. But the tools should be allowed to be used even on machines where
the man pages were not installed.

Currently the tools print a usage text when are executed with no options
but the -h option should be available for those setups and also some of
the tools allow them to be executed with no options (i.e: when using the
standard input to read data).

So add a configure flag to allow the tools -h option to not show the man
pages and install fallback to print the summary text usage instead.

Suggested-by: Emmanuel Deloget <logout@free.fr>

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -37,6 +37,9 @@
 #include <getopt.h>
 #include <unistd.h>
 
+#include <sys/types.h>
+#include <sys/wait.h>
+
 #include "log.h"
 #include "tpm2_options.h"
 #include "tpm2_util.h"
@@ -199,12 +202,31 @@ static char *tcti_get_opts(char *optstr) {
     return &split[1];
 }
 
-static void execute_man(char *prog_name) {
+static bool execute_man(char *prog_name) {
 
-    char *manpage = basename(prog_name);
-    execlp("man", "man", manpage, NULL);
-    LOG_ERR("Could not execute \"man %s\" error: %s", manpage,
-            strerror(errno));
+    pid_t  pid;
+    int status;
+
+    if ((pid = fork()) < 0) {
+        LOG_ERR("Could not fork process to execute man, error: %s",
+                strerror(errno));
+        return false;
+    }
+
+    if (pid == 0) {
+        char *manpage = basename(prog_name);
+        execlp("man", "man", manpage, NULL);
+    } else {
+        if ((pid = waitpid(pid, &status, 0)) == -1) {
+            LOG_ERR("Waiting for child process that executes man failed, error: %s",
+                    strerror(errno));
+            return false;
+        }
+
+        return WEXITSTATUS(status) == 0;
+    }
+
+    return true;
 }
 
 static void show_version (const char *name) {
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -199,15 +199,10 @@ static char *tcti_get_opts(char *optstr) {
     return &split[1];
 }
 
-static void execute_man (char *prog_name, char *envp[]) {
+static void execute_man(char *prog_name) {
 
     char *manpage = basename(prog_name);
-    char *argv[] = {
-        "/man", // ARGv[0] needs to be something.
-        manpage,
-        NULL
-    };
-    execvpe ("man", argv, envp);
+    execlp("man", "man", manpage, NULL);
     LOG_ERR("Could not execute \"man %s\" error: %s", manpage,
             strerror(errno));
 }
@@ -291,7 +286,7 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv, char **envp,
             tcti_opts = tcti_get_opts(optarg);
             break;
         case 'h':
-            execute_man(argv[0], envp);
+            execute_man(argv[0]);
             show_help = false;
             goto out;
             break;
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -253,6 +253,30 @@ static void show_version (const char *name) {
             tcti_conf);
 }
 
+void tpm2_print_usage(const char *command, struct tpm2_options *tool_opts) {
+    unsigned int i;
+
+    printf("usage: %s%s%s\n", command,
+           tool_opts->callbacks.on_opt ? " [OPTIONS]" : "",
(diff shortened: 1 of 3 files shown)


martinezjavier:
@emmanuel-deloget I forgot this bit when proposing PR #747.

martinezjavier:
@williamcroberts I'm not completely happy with the name of the configure option... but I couldn't come wit ha better name.

williamcroberts:
@martinezjavier rather than making this a configure time option, couldn't we just fall back when exec(man) fails and print the non-manpage output?

This way we can avoid yet another configure time option.

martinezjavier:
> @martinezjavier rather than making this a configure time option, couldn't we just fall back when exec(man) fails and print the non-manpage output?

There's no way for the tool to know that exec failed currently since the process address space was already replaced by the man binary.

I could do that, but the patch will be less simple since we will need to fork a process, exec man, call waitpid to know the process exist status, etc. But I can rework it if you prefer that way.

williamcroberts:
@martinezjavier yes, do it that way.

martinezjavier:
@williamcroberts perfect, I'm on holidays now but I'll re-spin the branch as soon as I've time to do it.

lib/tpm2_options.c
```diff
@@ -199,17 +202,36 @@ static char *tcti_get_opts(char *optstr) {
...
+        if ((pid = waitpid(pid, &status, 0)) == -1) {
+            LOG_ERR("Waiting for child process that executes man failed, error: %s",
+                    strerror(errno));
+            return false;
+        }
+
+        return WEXITSTATUS(status) == 0;
+    }
+
+    return true;
```
williamcroberts: I hate to do this to you, and I forgot to mention it, can you use popen() here pretty please.

lib/tpm2_options.c
```diff
@@ -291,7 +337,8 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv, char **envp,
             tcti_opts = tcti_get_opts(optarg);
             break;
         case 'h':
-            execute_man(argv[0], envp);
+            if (!execute_man(argv[0], envp))
+                tpm2_print_usage(argv[0], tool_opts);
```
williamcroberts: brackets.

lib/tpm2_options.h
```diff
@@ -186,4 +186,6 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv, char **envp,
         tpm2_options *tool_opts, tpm2_option_flags *flags,
         TSS2_TCTI_CONTEXT **tcti);
 
+void tpm2_print_usage(const char *command, struct tpm2_options *tool_opts);
```
williamcroberts: doxygen style comment block description

lib/tpm2_options.c
martinezjavier: @williamcroberts can I know what's your rationale to prefer `popen()` here? I really don't want to use it, since it invokes the shell to spawn a new process (which introduces a whole set of possible attack vectors) and we don't really need to pipe anything here. We only care about the child process exit status and not its output.

But now that we are discussing this, I wonder if we are using the correct exec* variant here. For instance `execvpe()` is a GNU extension and I can't see a reason to use it.

Also we don't really need a `v` or `e` variant since the command-line arguments array only contains the command and the calling program's environment is used anyways, so I would just use the `execlp()` wrapper here:

```
        char *manpage = basename(prog_name);
        execlp("man", "man", manpage, NULL);
```

Another question is how much we care about the user being able to manipulate the `$PATH` variable and trick the tools to execute an arbitrary binary as `man`. Probably that's not an issue, unless the tpm2_* tools have the setuid bit set but that would be a bad idea anyways.

I'll re-spin the branch changing this and addressing the other issues you pointed out, and we can then continue the discussion about using popen() or not.

lib/tpm2_options.c
martinezjavier: done.

lib/tpm2_options.h
martinezjavier: done.

lib/tpm2_options.c
williamcroberts: If they could do this, they already hijacked your shell, which is the context the tools are most likely going to be executed out of. In the case of someone executing the tools programatically (ie not in shell context), the man pager feature shouldn't be getting invoked, as it makes no sense your tool arguments would be wrong enough.

It boils down to popen is more portable, and we should be taking more portable interfaces over GNU extensions/linux specific. And yes execlp() would work just fine. Even if someone  is manipulating PATH or the shell to exploit this, as even if it did get exploited, they should only be able to do things granted by the permission context the tools are running in, but they already had your shell hijacked, so it was game over anyways.

Now with that said, if the tools get erroneous input, that input shouldn't result in gaining execution control over the tool itself. Ie, giving a bad file as input shouldn't cause the user to be able to steal your ssh keys.

martinezjavier: Regardless the possibly security issues, I prefer the fork / exec / waitpid combination over popen, since as mentioned the parent process doesn't really care about the child process output and only the exit status.

The latest version shouldn't have GNU extensions, so it should be portable on POSIX compliant systems. Having said that, I could change to use popen if you have a strong opinion on this.

williamcroberts: The TSS is heading to Windows as well. So I would rather start considering being windows ready where portable interfaces exist, even in the face of Windows 4 Linux subsystem. I always prefer a C std interface over an OS specific interface, and maybe that should be in the coding standard.

Plus it makes your code a 2 liner:
```
FILE *f = popen("man ls", "we");
return pclose(f);
```

martinezjavier: Hmm, does Windows even have a man pager to start with? So I guess this part will have to be built out on Windows anyways?

I also wonder how realistic is to build the tools on Windows at this point.

It's less code but underneath it invokes a shell to exec the man binary for no good reasons (yes, it's more portable in theory but in practice is debatable as mentioned).

williamcroberts: But it would compile and run on Windows, just the manpager would be broken.
It's not very realistic to build the tools on Windows at this point AFAIK the TSS isn't ported yet.

I'll just take this as is. It's easy enough to change in the future if we need to and it won't be a user facing change if we do.

PR #758, opened 2018-01-09
tpm2_sign : provide -D option to sign a pre-computed hash

### LINKED ISSUE #753: Signing without hashing?
Would it be possible to add an option to tpm2_sign for providing just a message digest instead of the original message? When signing with a non-restricted key, it's not mandatory to have TPM to hash the message.

emmanuel-deloget:
Intruct tpm2_sign to allow it to handle signing pre-computed hashes. This fixes #753. 

When one wants to use option -D he needs to respect a few conditions: 

* the digest file size should match the selected halg
* the selected signing key shall not be restricted

When option -D is used, options -m and -t are (silently) ignored. 

The patch is bundled with a cleaning patch that remove optarg from the tools code (optarg is a global set by getopt_long() and passed to the tools on_option() as argument value ; thus, it's always available using variable value, and one may consider that optarg is, at this point, an implementation details that should not be shown to users). 

Best regards, 

-- Emmanuel Deloget

### DIFF
--- a/man/tpm2_sign.1.md
+++ b/man/tpm2_sign.1.md
@@ -46,6 +46,15 @@ data and validation shall indicate that hashed data did not start with
 
     The message file, containing the content to be  digested.
 
+  * **-D**, **--digest**=_DIGEST\_FILE_:
+
+    The digest file that shall be computed using the correct hash
+    algorithm. When this option is specified, a warning is generated and
+    **both the message file (-m) and the validation ticket (-t) are
+    ignored**.
+    You cannot use this option to sign a digest against a restricted
+    signing key.
+
   * **-t**, **--ticket**=_TICKET\_FILE_:
 
     The ticket file, containning the validation structure, optional.
--- a/tools/tpm2_loadexternal.c
+++ b/tools/tpm2_loadexternal.c
@@ -121,7 +121,7 @@ static bool on_option(char key, char *value) {
         ctx.flags.H = 1;
     break;
     case 'u':
-        if(!files_load_public(optarg, &ctx.public_key)) {
+        if(!files_load_public(value, &ctx.public_key)) {
             return false;;
         }
         ctx.flags.u = 1;
--- a/tools/tpm2_sign.c
+++ b/tools/tpm2_sign.c
@@ -177,7 +177,7 @@ static bool on_option(char key, char *value) {
         bool result = tpm2_util_string_to_uint32(value, &ctx.keyHandle);
         if (!result) {
             LOG_ERR("Could not format key handle to number, got: \"%s\"",
-                    optarg);
+                    value);
             return false;
         }
         ctx.flags.k = 1;
@@ -186,24 +186,24 @@ static bool on_option(char key, char *value) {
     case 'P': {
         bool result = tpm2_password_util_from_optarg(value, &ctx.sessionData.hmac);
         if (!result) {
-            LOG_ERR("Invalid key password, got\"%s\"", optarg);
+            LOG_ERR("Invalid key password, got\"%s\"", value);
             return false;
         }
         ctx.flags.P = 1;
     }
         break;
     case 'g': {
-        ctx.halg = tpm2_alg_util_from_optarg(optarg);
+        ctx.halg = tpm2_alg_util_from_optarg(value);
         if (ctx.halg == TPM2_ALG_ERROR) {
             LOG_ERR("Could not convert to number or lookup algorithm, got: \"%s\"",
-                    optarg);
+                    value);
             return false;
         }
         ctx.flags.g = 1;
     }
         break;
     case 'm':
-        ctx.inMsgFileName = optarg;
+        ctx.inMsgFileName = value;
         ctx.flags.m = 1;
         break;
     case 't': {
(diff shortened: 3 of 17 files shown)


emmanuel-deloget:
Ouch. Test failed because a library was not found :(

martinezjavier:
@emmanuel-deloget I've created PR #759 that should fix this issue.

tools/tpm2_loadexternal.c
```diff
@@ -121,7 +121,7 @@ static bool on_option(char key, char *value) {
         ctx.flags.H = 1;
     break;
     case 'u':
-        if(!files_load_public(optarg, &ctx.public_key)) {
```
martinezjavier: Indeed, these are left overs from when the tools were ported to the new option handling infrastructure.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 ### next
+  * tpm2_sign: add option -D  to provide an already computed digest to sign
```
martinezjavier: Usually @williamcroberts updates the Changelog after things are merged. I don't know if he prefers to do it or is just that people don't usually update it.

I'm not against your change of course, I just don't know what is expected and if he has a preference (we should probably document it in CONTRIBUTING.md so people if they should update it or not).

emmanuel-deloget: I was not sure if this change was a good idea. I've remove it. 

Thanks for your review!

man/tpm2_sign.1.md
```diff
@@ -46,6 +46,14 @@ data and validation shall indicate that hashed data did not start with
 
     The message file, containing the content to be  digested.
 
+  * **-D**, **--digest**=_DIGEST\_FILE_:
+
+    The digest file that shall be computed using the correct hash
+    algorithm. When this option is specified, both the message file and
+    the validation ticket are ignored.
```
williamcroberts: Make this complain loudly please.

man/tpm2_sign.1.md
```diff
@@ -46,6 +46,14 @@ data and validation shall indicate that hashed data did not start with
...
+    You cannot use this option to sign a digest against a restricted
+    signed key.
```
williamcroberts: "signing key" right, not "signed key" here?

tools/tpm2_sign.c
```diff
@@ -135,36 +139,38 @@ static bool init(TSS2_SYS_CONTEXT *sapi_context) {
...
+      }
+      if (file_size == 0) {
+          LOG_ERR("The message file \"%s\" is empty!", ctx.inMsgFileName);
+          return false;
+      }
+
+      if (file_size > 0xffff) {
+          LOG_ERR(
+                  "The message file was longer than a 16 bit length, got: %lu, expected less than: %d!",
+                  file_size, 0x10000);
```
williamcroberts: Use UINT16_MAX here from stdint.h for both the reported max (UINT16_MAX + 1) on line 158 and the check on line 155,

man/tpm2_sign.1.md
emmanuel-deloget: Yes, sorry :)

tools/tpm2_sign.c
emmanuel-deloget: Ok

man/tpm2_sign.1.md
emmanuel-deloget: I'll make it bold in the man page, and I'll write an info message.

emmanuel-deloget:
I pushed a new version, it hopefully resolve all the noted issues. It also add a few more tests in sign.sh: 

* test tpm2_sign with a validation token (generated with tpm2_hash)
* test tpm2_sign with options -D, -t and -m (and check that it generates a warning). 

Best regards, 

-- Emmanuel Deloget

man/tpm2_sign.1.md
williamcroberts: SGTM

tools/tpm2_sign.c
```diff
@@ -135,36 +143,38 @@ static bool init(TSS2_SYS_CONTEXT *sapi_context) {
...
+      }
+      if (file_size == 0) {
+          LOG_ERR("The message file \"%s\" is empty!", ctx.inMsgFileName);
+          return false;
+      }
+
+      if (file_size > UINT16_MAX) {
+          LOG_ERR(
+                  "The message file was longer than a 16 bit length, got: %lu, expected less than: %d!",
+                  file_size, 0x10000);
```
williamcroberts: Close (UINT16_MAX + 1) here or change the wording to "less than or equal to" and use UINT16_MAX, or just drop it and leave it less than 16 bits in length, and show what size you read, folks can figure that out from there.

williamcroberts:
minor change left on the 0x10000

tools/tpm2_sign.c
emmanuel-deloget: Can I change the message to "The message file is too big, got: %lu, expected less than: %u bytes" ?

"Longer than 16 bit length" seems weird to me :)

emmanuel-deloget:
I _think_ I got everyting right this time :) Of course, me being myself, that assertion might as well be wrong.

tools/tpm2_sign.c
```diff
@@ -135,36 +143,38 @@ static bool init(TSS2_SYS_CONTEXT *sapi_context) {
...
+      }
+      if (file_size == 0) {
+          LOG_ERR("The message file \"%s\" is empty!", ctx.inMsgFileName);
+          return false;
+      }
+
+      if (file_size > UINT16_MAX) {
+          LOG_ERR(
+                  "The message file \"%s\" is too large, got: %lu bytes, expected less than: %u bytes!",
+                  ctx.inMsgFileName, file_size, UINT16_MAX + 1);
```
williamcroberts: This looks good. Thanks for fixing this while you where there.

Issue #760, opened 2018-01-09
tpm2_nvlist crashes on systems without man(1)


emmanuel-deloget:
It does so because it tries to dereference tool_opts (which is NULL) in tpm2_print_usage(). 

I'll send a PR in a few minutes to fix that. 

Best regards, 

-- Emmanuel Deloget

PR #761, opened 2018-01-09
tools: fix print usage when the program takes no option, no args

emmanuel-deloget:
The symptom is: tpm2_nvlist -h crashes on a system that is missing
man(1).

The reason is: tpm2_nvlist takes no argument at all, and thus tool_opts
is NULL.

In oder to avoid the crash, don't dereference tool_opts when it's not a
valid pointer.

Fixes: #760

Signed-off-by: Emmanuel Deloget <logout@free.fr>

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -257,10 +257,10 @@ void tpm2_print_usage(const char *command, struct tpm2_options *tool_opts) {
     unsigned int i;
 
     printf("usage: %s%s%s\n", command,
-           tool_opts->callbacks.on_opt ? " [OPTIONS]" : "",
-           tool_opts->callbacks.on_arg ? " ARGUMENTS" : "");
+           tool_opts && tool_opts->callbacks.on_opt ? " [OPTIONS]" : "",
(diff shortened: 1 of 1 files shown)


martinezjavier:
@emmanuel-deloget I'm so sorry for introducing this issue. In the previous usage of the function I did check for tool_opts being NULL and also for  tool_opts->show_usage being set (which isn't for tools with no options), from https://github.com/intel/tpm2-tools/blob/master/tools/tpm2_tool.c#L114:

```
    if (argc == 1 && tool_opts && tool_opts->show_usage) {
        tpm2_print_usage(argv[0], tool_opts);
        return ret;
    }
```

So I think the correct fix is to move that check inside `tpm2_print_usage()` and just return `if (!tool_opts || tool_opts->show_usage)`, since I don't think that makes sense to print a usage text if no options or arguments are needed (it will also be consistent when the tools are executed with no options).

martinezjavier:
@emmanuel-deloget I've created PR #762 with an alternative fix as mentioned in my previous comment. Please let me know what you think about it.

And thanks a lot for pointing out this regression!

emmanuel-deloget:
Indeed, yours is better than mine :)

I'm closing this PR.

PR #762, opened 2018-01-09
lib: fix tpm2_print_usage() when tools have no options/args

martinezjavier:
For tools that takes no options and arguments, its struct tpm2_options is
NULL so trying to access it will lead to a NULL pointer dereference error.

This is checked in the main() function before calling tpm2_print_usage(),
but it was checked when tpm2_print_usage() was added as a fallback to the
--help option when executing the man pager fails.

Instead of relying of the caller to do the check, make tpm2_print_usage()
robust enough to cope with being called with a NULL struct tpm2_options.

After all, it makes no sense to print a usage text for a tool that takes
neither options nor arguments.

Reported-by: Emmanuel Deloget <logout@free.fr>

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -256,6 +256,10 @@ static void show_version (const char *name) {
 void tpm2_print_usage(const char *command, struct tpm2_options *tool_opts) {
     unsigned int i;
 
+    if (!tool_opts || !tool_opts->show_usage) {
+        return;
+    }
+
     printf("usage: %s%s%s\n", command,
            tool_opts->callbacks.on_opt ? " [OPTIONS]" : "",
            tool_opts->callbacks.on_arg ? " ARGUMENTS" : "");


lib/tpm2_options.c
```diff
@@ -256,6 +256,9 @@ static void show_version (const char *name) {
 void tpm2_print_usage(const char *command, struct tpm2_options *tool_opts) {
     unsigned int i;
 
+    if (!tool_opts || !tool_opts->show_usage)
+        return;
```
williamcroberts: brackets

martinezjavier: @williamcroberts sorry... I've been repeatedly getting wrong this. Lately did some work on a different project whose coding style has the "single statements not using brackets" rule so I get confused :)

williamcroberts: Yeah that'll happen, been there. Also switch between C and python a lot, that can be brutal sometimes.

emmanuel-deloget:
Can you also add a Fixes tag? I created issue #760 to keep track of the bug, so you'll probably want to mark it as fixed by this commit.

martinezjavier:
@emmanuel-deloget right, added now.

Issue #752, opened 2017-12-27
GCC 5.4.0 complains "-Werror=int-to-pointer-cast" on "test_tpm2_session.c"


liuqun:
tpm2-tools branch master commit 3ab33f7
When configured with cmocka unit test enabled, GCC 5.4.0 complains about casting from integer to pointer (intergers and pointers might have different sizes in arm-linux-gnueabihf-gcc) :
```
  CC       test/unit/test_unit_test_tpm2_session-test_tpm2_session.o
test/unit/test_tpm2_session.c: In function ‘__wrap_Tss2_Sys_StartAuthSession’:
test/unit/test_tpm2_session.c:116:24: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
     expected_data *e = (expected_data *) mock();
                        ^
cc1: all warnings being treated as errors
```

GCC on Raspberry Pi 3 (target: arm-linux-gnueabihf)
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/5/lto-wrapper
Target: arm-linux-gnueabihf
[...]

PR #763, opened 2018-01-10
Session tools

williamcroberts:
This patchset leaves all the old session handling stuff pretty much in place, with the exception of the session option -S using a file to pass the data versus just a hex string argument. I found tools needing more information about a session, so rather than adding options, we just encode it in the file along with the handle.

Outside of that, it adds tpm2_startauthsession and tpm2_policypcr tools, which essentially breaks out the tpm2_createpolicy functionality into discrete tools.

I'm looking for early feedback before I go continuing down this path of adding more policy event tools and updating all the "-S" options currently. I am mostly interested in large architectural changes to the code they may need to occur versus a pedantic code review, but both are welcome/appreciated.
@martinezjavier 

I tested this against abrmd, with a known issue of: https://github.com/intel/tpm2-abrmd/issues/285

### DIFF
--- a/lib/files.c
+++ b/lib/files.c
@@ -154,7 +154,7 @@ bool files_save_bytes_to_file(const char *path, UINT8 *buf, UINT16 size) {
 #define CONTEXT_VERSION 1
 
 bool files_save_tpm_context_to_file(TSS2_SYS_CONTEXT *sysContext, TPM2_HANDLE handle,
-        const char *path) {
+        FILE *stream) {
 
     TPMS_CONTEXT context;
 
@@ -163,13 +163,6 @@ bool files_save_tpm_context_to_file(TSS2_SYS_CONTEXT *sysContext, TPM2_HANDLE ha
         LOG_ERR(
                 "Tss2_Sys_ContextSave: Saving handle 0x%x context failed. TPM Error:0x%x",
                 handle, rval);
-        return false;
-    }
-
-    FILE *f = fopen(path, "w+b");
-    if (!f) {
-        LOG_ERR("Error opening file \"%s\" due to error: %s", path,
-                strerror(errno));
         return false;
     }
 
@@ -182,63 +175,71 @@ bool files_save_tpm_context_to_file(TSS2_SYS_CONTEXT *sysContext, TPM2_HANDLE ha
      * U16 contextBlobLength
      * BYTE[] contextBlob
      */
-    bool result = files_write_header(f, CONTEXT_VERSION);
+    bool result = files_write_header(stream, CONTEXT_VERSION);
     if (!result) {
-        LOG_ERR("Could not write header for file: \"%s\"", path);
+        LOG_ERR("Could not write context file header");
         goto out;
     }
 
     // UINT32
-    result = files_write_32(f, context.hierarchy);
+    result = files_write_32(stream, context.hierarchy);
     if (!result) {
-        LOG_ERR("Could not write hierarchy for file: \"%s\"", path);
+        LOG_ERR("Could not write hierarchy");
         goto out;
     }
 
-    result = files_write_32(f, context.savedHandle);
+    result = files_write_32(stream, context.savedHandle);
     if (!result) {
-        LOG_ERR("Could not write savedHandle for file: \"%s\"", path);
+        LOG_ERR("Could not write savedHandle");
         goto out;
     }
 
     // UINT64
-    result = files_write_64(f, context.sequence);
+    result = files_write_64(stream, context.sequence);
     if (!result) {
-        LOG_ERR("Could not write sequence for file: \"%s\"", path);
+        LOG_ERR("Could not write sequence");
         goto out;
     }
 
     // U16 LENGTH
-    result = files_write_16(f, context.contextBlob.size);
+    result = files_write_16(stream, context.contextBlob.size);
     if (!result) {
-        LOG_ERR("Could not write contextBob size file: \"%s\"", path);
+        LOG_ERR("Could not write contextBob size");
         goto out;
     }
 
     // BYTE[] contextBlob
-    result = files_write_bytes(f, context.contextBlob.buffer,
+    result = files_write_bytes(stream, context.contextBlob.buffer,
             context.contextBlob.size);
     if (!result) {
-        LOG_ERR("Could not write contextBlob buffer for file: \"%s\"", path);
+        LOG_ERR("Could not write contextBlob buffer");
     }
     /* result is set by file_write_bytes() */
 
 out:
-    fclose(f);
     return result;
 }
 
-bool files_load_tpm_context_from_file(TSS2_SYS_CONTEXT *sapi_context,
-        TPM2_HANDLE *handle, const char *path) {
+bool files_save_tpm_context_to_path(TSS2_SYS_CONTEXT *sysContext, TPM2_HANDLE handle,
+        const char *path) {
 
-    TSS2_RC rval;
-
-    FILE *f = fopen(path, "rb");
+    FILE *f = fopen(path, "w+b");
     if (!f) {
         LOG_ERR("Error opening file \"%s\" due to error: %s", path,
                 strerror(errno));
         return false;
     }
+
+    bool result = files_save_tpm_context_to_file(sysContext, handle, f);
+    fclose(f);
+    return result;
+}
+
+
+bool files_load_tpm_context_from_file(TSS2_SYS_CONTEXT *sapi_context,
+        TPM2_HANDLE *handle, FILE *fstream) {
+
+    TSS2_RC rval;
 
     /*
      * Reading the TPMS_CONTEXT structure to disk, format:
@@ -251,15 +252,15 @@ bool files_load_tpm_context_from_file(TSS2_SYS_CONTEXT *sapi_context,
      */
     UINT32 version;
     TPMS_CONTEXT context;
-    bool result = files_read_header(f, &version);
+    bool result = files_read_header(fstream, &version);
     if (!result) {
         LOG_WARN(
-                "The tpm context file \"%s\" does not appear in the proper format, assuming old format, this will be converted on the next save.",
-                path);
-        rewind(f);
-        result = files_read_bytes(f, (UINT8 *) &context, sizeof(context));
+            "The loaded tpm context does not appear to be in the proper format,"
+            "assuming old format, this will be converted on the next save.");
+        rewind(fstream);
+        result = files_read_bytes(fstream, (UINT8 *) &context, sizeof(context));
         if (!result) {
-            LOG_ERR("Could not load file \"%s\" into tpm context", path);
+            LOG_ERR("Could not load tpm context file");
             goto out;
         }
         /* Success load the context into the TPM */
@@ -273,25 +274,25 @@ bool files_load_tpm_context_from_file(TSS2_SYS_CONTEXT *sapi_context,
         goto out;
     }
 
-    result = files_read_32(f, &context.hierarchy);
+    result = files_read_32(fstream, &context.hierarchy);
     if (!result) {
         LOG_ERR("Error reading hierarchy!");
         goto out;
     }
 
-    result = files_read_32(f, &context.savedHandle);
+    result = files_read_32(fstream, &context.savedHandle);
     if (!result) {
         LOG_ERR("Error reading savedHandle!");
         goto out;
     }
 
-    result = files_read_64(f, &context.sequence);
+    result = files_read_64(fstream, &context.sequence);
     if (!result) {
         LOG_ERR("Error reading sequence!");
         goto out;
     }
 
-    result = files_read_16(f, &context.contextBlob.size);
+    result = files_read_16(fstream, &context.contextBlob.size);
     if (!result) {
         LOG_ERR("Error reading contextBlob.size!");
         goto out;
@@ -306,7 +307,7 @@ bool files_load_tpm_context_from_file(TSS2_SYS_CONTEXT *sapi_context,
         goto out;
     }
 
-    result = files_read_bytes(f, context.contextBlob.buffer,
+    result = files_read_bytes(fstream, context.contextBlob.buffer,
             context.contextBlob.size);
     if (!result) {
         LOG_ERR("Error reading contextBlob.size!");
@@ -324,6 +325,21 @@ load_to_tpm:
     result = true;
 
 out:
+    return result;
+}
+
+bool files_load_tpm_context_from_path(TSS2_SYS_CONTEXT *sapi_context,
+        TPM2_HANDLE *handle, const char *path) {
+
+    FILE *f = fopen(path, "rb");
+    if (!f) {
+        LOG_ERR("Error opening file \"%s\" due to error: %s", path,
+                strerror(errno));
+        return false;
+    }
+
+    bool result = files_load_tpm_context_from_file(sapi_context, handle, f);
+
     fclose(f);
     return result;
 }
--- a/lib/tpm2_session.c
+++ b/lib/tpm2_session.c
@@ -110,7 +110,7 @@ TPMI_ALG_HASH tpm2_session_get_authhash(tpm2_session *session) {
     return session->input->authHash;
 }
 
-TPMI_SH_AUTH_SESSION tpm2_session_get_session_handle(tpm2_session *session) {
+TPMI_SH_AUTH_SESSION tpm2_session_get_handle(tpm2_session *session) {
     return session->output.session_handle;
 }
 
@@ -271,7 +271,7 @@ bool tpm2_session_save(TSS2_SYS_CONTEXT *sapi_context, tpm2_session *session,
         return false;
     }
 
-    TPM2_HANDLE handle = tpm2_session_get_session_handle(session);
+    TPM2_HANDLE handle = tpm2_session_get_handle(session);
     result = files_save_tpm_context_to_file(sapi_context, handle, mem);
     if (!result) {
         goto out;
@@ -335,7 +335,7 @@ out:
 
 bool tpm2_session_restart(TSS2_SYS_CONTEXT *sapi_context, tpm2_session *s) {
 
-    TPMI_SH_AUTH_SESSION handle = tpm2_session_get_session_handle(s);
+    TPMI_SH_AUTH_SESSION handle = tpm2_session_get_handle(s);
 
     TSS2_RC rval = Tss2_Sys_PolicyRestart(sapi_context, handle, NULL, NULL);
     if (rval != TPM2_RC_SUCCESS) {
--- a/lib/tpm2_session.c
+++ b/lib/tpm2_session.c
@@ -317,7 +317,6 @@ bool tpm2_session_save(TSS2_SYS_CONTEXT *sapi_context, tpm2_session *session,
      }
 
      /* result is set by files_write_32() */
-
 out:
     if (mem) {
         fclose(mem);
--- a/lib/tpm2_session.c
+++ b/lib/tpm2_session.c
@@ -114,6 +114,10 @@ TPMI_SH_AUTH_SESSION tpm2_session_get_session_handle(tpm2_session *session) {
     return session->output.session_handle;
 }
 
+TPM2_SE tpm2_session_get_type(tpm2_session *session) {
+    return session->input->session_type;
+}
+
 //
 // This is a wrapper function around the TPM2_StartAuthSession command.
 // It performs the command, calculates the session key, and updates a
--- a/lib/tpm2_session.c
+++ b/lib/tpm2_session.c
@@ -25,16 +25,18 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //**********************************************************************;
 
-#include "tpm2_session.h"
-
+#include <errno.h>
 #include <stdbool.h>
+#include <stdio.h>
 #include <stdlib.h>
 
 #include <sapi/tpm20.h>
 
+#include "files.h"
 #include "log.h"
 #include "tpm_kdfa.h"
 #include "tpm2_alg_util.h"
+#include "tpm2_session.h"
 #include "tpm2_util.h"
 
 struct tpm2_session_data {
@@ -137,9 +139,11 @@ static bool start_auth_session(TSS2_SYS_CONTEXT *sapi_context,
 void tpm2_session_free(tpm2_session **session) {
 
     tpm2_session *s = *session;
-    free(s->input);
-    free(s);
-    *session = NULL;
+    if (s) {
+        free(s->input);
+        free(s);
+        *session = NULL;
+    }
 }
 
 tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
@@ -156,6 +160,10 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
 
     session->internal.nonceNewer.size = session->input->nonce_caller.size;
 
+    if (!sapi_context) {
+        return session;
+    }
+
     bool result = start_auth_session(sapi_context, session);
     if (!result) {
         tpm2_session_free(&session);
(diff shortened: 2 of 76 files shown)


lib/files.c
```diff
@@ -251,15 +252,15 @@ bool files_load_tpm_context_from_file(TSS2_SYS_CONTEXT *sapi_context,
...
     TPMS_CONTEXT context;
-    bool result = files_read_header(f, &version);
+    bool result = files_read_header(fstream, &version);
     if (!result) {
         LOG_WARN(
-                "The tpm context file \"%s\" does not appear in the proper format, assuming old format, this will be converted on the next save.",
-                path);
-        rewind(f);
-        result = files_read_bytes(f, (UINT8 *) &context, sizeof(context));
+            "The loaded tpm context does not appear in the proper format,"
```
martinezjavier: Is this sentence correct, shouldn't be "does not appear to be in the proper format" ?

lib/tpm2_session.c
```diff
@@ -156,6 +160,10 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
 
     session->internal.nonceNewer.size = session->input->nonce_caller.size;
 
+    if (!sapi_context) {
```
martinezjavier: Isn't this a bug in the caller? Why would a tool want to start a session without a SAPI context? 

I think this check should be at the beginning of the function and just return NULL. Or maybe I'm misunderstanding something?

lib/tpm2_session.c
```diff
@@ -164,3 +172,150 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
...
+ * It works by leveraging the fact that C does not allow negative array sizes.
+ * If the sizes are equal, the boolean equality operator will return 1, thus
+ * a subtraction of 1 yields 0, which is a legal array size in C. In the false
+ * case (ie sizes not equal), 0 - 1 is -1, which will cause the compiler to
+ * complain.
+ */
+#define COMPILE_ASSERT_SIZE(a, b) \
+    typedef char WRONG_SIZE_##a[(sizeof(a) == sizeof(b)) - 1]
+
+// We check that the TSS library does not change sizes unbeknownst to us.
```
martinezjavier: I just learned a new word today :smiley:

lib/tpm2_session.c
```diff
@@ -164,3 +172,150 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
...
+COMPILE_ASSERT_SIZE(TPM2_HANDLE, UINT32);
+COMPILE_ASSERT_SIZE(TPMI_ALG_HASH, UINT16);
+COMPILE_ASSERT_SIZE(TPM2_SE, UINT8);
```
martinezjavier: Hmm, I wonder if were aren't being too paranoid here. At the end of the day we depend on the TSS library to not break API and ABI compatibility, otherwise shouldn't we check the size of every possible type?

lib/tpm2_session.c
```diff
@@ -164,3 +172,150 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
...
+
+tpm2_session *tpm2_session_restore(const char *path) {
+
+    tpm2_session *s = NULL;
+
+    FILE *f = fopen(path, "rb");
+    if (!f) {
+        LOG_ERR("Could not open path \"%s\", due to error: \"%s\"",
+                path, strerror(errno));
+        goto out;
```
martinezjavier: You should just return NULL here, otherwise fclose(f) will be called when jumping to the out label.

lib/tpm2_session.c
```diff
@@ -164,3 +172,150 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
...
+
+    tpm2_session_data *d = tpm2_session_data_new(type);
+    if (!d) {
+        LOG_ERR("oom");
+        goto out;
+    }
+
+    tpm2_session_set_authhash(d, auth_hash);
+
+    s = tpm2_session_new(NULL, d);
```
martinezjavier: You have to check if tpm2_session_new() didn't return NULL, and free tpm2_session_data *d if that's the case.

lib/tpm2_session.c
```diff
@@ -164,3 +172,150 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
...
+        goto out;
+    }
+
+    /*
+     * Now write the session_type, handle and auth hash data to disk
+     */
+    result = files_write_header(session_file, SESSION_VERSION);
+    if (!result) {
+         LOG_ERR("Could not write context file header");
+         goto out;
```
martinezjavier: You need a new error label here that also closes session_file before returning.

lib/tpm2_session.c
```diff
@@ -164,3 +172,150 @@ tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
...
+     }
+
+     // UINT8 session type:
+     TPM2_SE session_type = session->input->session_type;
+     result = files_write_bytes(session_file, &session_type, sizeof(session_type));
+     if (!result) {
+         LOG_ERR("Could not write session type");
+         goto out;
```
martinezjavier: Same here, a label that closes session_file should be used.

tools/tpm2_createprimary.c
```diff
@@ -325,7 +325,7 @@ int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
         returnVal = create_primary(sapi_context);
 
         if (returnVal == 0 && ctx.flags.C) {
-            returnVal = files_save_tpm_context_to_file(sapi_context, ctx.handle2048rsa,
+            returnVal = files_save_tpm_context_to_path(sapi_context, ctx.handle2048rsa,
```
martinezjavier: Not related with your changes, but why the handles variables are called `handle2048rsa` in some tools? Shouldn't that just be called `handle`? I guess is a copy & paste error or am I missing something?

lib/tpm2_session.h
```diff
@@ -146,6 +148,37 @@ TPMI_SH_AUTH_SESSION tpm2_session_get_session_handle(tpm2_session *session);
 tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
         tpm2_session_data *data);
 
+/**
+ * Saves session data to disk allowing tpm2_session_from_file() to
+ * restore the session.
+ *
+ * @Note
+ * This is accomplished by calling:
+ *   - Tss2_Sys_ContextSave - marks to some RMs like abrmd not to flush this session
```
martinezjavier: I would say tpm2-abrmd instead of just abrmd.

man/tpm2_startauthsession.1.md
```diff
@@ -0,0 +1,71 @@
...
+the session handle in a yaml format to stdout as key value "session-handle"
+and the handle in hex (0x1234) format.
+
+# OPTIONS
+
+  * **-a**, **--auth-policy-session**:
+
+    Change the policy session from a *trial* session to a *policy* session.
+    **NOTE**: A *trial* session is used when building a policy and a *policy*
+    session is used when authenticating with a policy.
```
martinezjavier: Nice note.

man/tpm2_startauthsession.1.md
```diff
@@ -0,0 +1,71 @@
...
+    The hash algorithm used in computation of the policy digest. Algorithms
+    should follow the "formatting standards", see section "Algorithm Specifiers".
+    Also, see section "Supported Hash Algorithms" for a list of supported hash
+    algorithms.
+
+  * **-S**, **--session**=_SESSION_FILE_:
+
+    Saves the policy session data to a file. This file can then be used in subsequent
+    tools that can use a policy file for authorization or policy events. **NOTE**: That
+    this will not work with resource managers (RMs) outside of abrmd, as most RMs will
```
martinezjavier: s/abrmd/tpm2-abrmd

man/tpm2_startauthsession.1.md
```diff
@@ -0,0 +1,71 @@
...
+    flush session handles when a client disconects on the IPC channel. This will work
```
martinezjavier: shouldn't this be "from" the IPC channel?

man/tpm2_startauthsession.1.md
```diff
@@ -0,0 +1,71 @@
...
+    with direct TPM access, but note that internally this calls a *ContextSave* and a
+    *ContextLoad* on the session handle, thus the session **cannot** be saved/loaded
+    again.
```
martinezjavier: I would also add that the saved context can be loaded only once to prevent replay attacks. So subsequent tools need to save the context again if more than one tool are chained.

tools/tpm2_startauthsession.c
```diff
@@ -0,0 +1,138 @@
...
+    if (!session_data) {
+        LOG_ERR("oom");
+        return rc;
+    }
+
+    tpm2_session_set_authhash(session_data, ctx.session.halg);
+
+    tpm2_session *s = tpm2_session_new(sapi_context,
+            session_data);
+    if (!s) {
```
martinezjavier: You have to free tpm2_session_data *session_data before returning.

tools/tpm2_startauthsession.c
```diff
@@ -0,0 +1,138 @@
...
+        bool result = tpm2_session_save(sapi_context, s, ctx.output.path);
+        if (!result) {
+            goto out;
+        }
+    }
+
+    rc = 0;
+
+out:
+    tpm2_session_free(&s);
```
martinezjavier: Same here, session_data is not freed.

man/tpm2_startauthsession.1.md
```diff
@@ -0,0 +1,71 @@
...
+Start a *trial* session and save the session data to a file.
+```
+tpm2_startauthsession -S session.dat
+session-handle: 0x3000000
+```
+
+Start a *policy* session and save the session data to a file.
+```
+tpm2_startauthsession -S session.dat -a
+session-handle: 0x3000000
```
martinezjavier: It would be good if you also add an integration test for this new tool.

man/tpm2_policypcr.1.md
```diff
@@ -0,0 +1,93 @@
...
+
+[common tcti options](common/tcti.md)
+
+[supported hash algorithms](common/hash.md)
+
+[algorithm specifiers](common/alg.md)
+
+# EXAMPLES
+
+Start a *trial* session and build a PCR policy and use that policy in an object.
```
martinezjavier: I think you can remove one "and", i.e: Start a *trial* session, build a PCR policy and use that policy in an object.

man/tpm2_policypcr.1.md
```diff
@@ -0,0 +1,93 @@
...
+Then use a *policy* session to unseal some data.
+```
+# Step 1: Create a trial session and build a pcr policy via a policyPCR event to generate
```
martinezjavier: s/pcr/PCR for consistency.

man/tpm2_policypcr.1.md
```diff
@@ -0,0 +1,93 @@
...
+
+tpm2_policypcr -Q -S session.dat -L "sha1:0,1,2,3" -F pcr.dat -f policy.dat
+
+unsealed=`tpm2_unseal -S session.dat -c unseal.key.ctx`
+
+echo "$unsealed"
+
+tpm2_flushcontext -H "$handle"
+```
+
```
martinezjavier: I think it's very cool to have these long examples. How it's rendered this in the generated troff man page?

man/tpm2_policypcr.1.md
```diff
@@ -0,0 +1,93 @@
...
+#  policy hash.
+#
+# Step 4: Using the actual policy session from step 3 in tpm2_unseal to unseal the object.
+#
+
+tpm2_createprimary -H e -g sha256 -G ecc -C primary.ctx
+
+tpm2_pcrlist -Q -L "sha1:0,1,2,3 -o pcr.dat
+
+handle=`tpm2_startauthsession -S session.dat | cut -d' ' -f 2-2`
```
martinezjavier: This is an example why I would prefer to just print the session instead of having it inside a yaml output, the pipe and cut won't be needed in that case.

man/tpm2_policypcr.1.md
martinezjavier: BTW, this example will make a good integration test for the tool to include in this commit.

tools/tpm2_policypcr.c
```diff
@@ -0,0 +1,165 @@
...
+        LOG_ERR("Must specify -S session file");
+        fail = true;
+    }
+
+    if (!ctx.pcr_selection.count) {
+        LOG_ERR("Must specify -L pcr selection list");
+        fail = true;
+    }
+
+    if (fail) {
```
martinezjavier: Not sure I understand the reason of this fail variable, why not just return 1 in the error path of the two checks above?

tools/tpm2_policypcr.c
```diff
@@ -0,0 +1,165 @@
...
+        return 1;
+    }
+
+    s = tpm2_session_restore(ctx.session_path);
+    if (!s) {
+        return false;
```
martinezjavier: return 1

tools/tpm2_policypcr.c
```diff
@@ -0,0 +1,165 @@
...
+
+    TPM2B_DIGEST policy_digest = TPM2B_EMPTY_INIT;
+    result = tpm2_policy_get_digest(sapi_context, s,
+            &policy_digest);
+    if (!result) {
+        LOG_ERR("Could not build tpm policy");
+        goto out;
+    }
+
+    tpm2_tool_output("policy_digest: 0x");
```
martinezjavier: Again I wonder if we shouldn't just print the policy digest and in any case print additional info with -V.

I think most users that care about this information will prefer to just do:

```
digest=`tpm2_policypcr -S session.dat -L "sha1:0,1,2,3" -F pcr.dat -f policy.dat`
```

tools/tpm2_policypcr.c
```diff
@@ -0,0 +1,165 @@
...
+    tpm2_tool_output("\n");
+
+    if (ctx.policy_out_path) {
+        result = files_save_bytes_to_file(ctx.policy_out_path,
+                    (UINT8 *) &policy_digest.buffer,
+                    policy_digest.size);
+        if (!result) {
+            LOG_ERR("Failed to save policy digest into file \"%s\"",
+                    ctx.policy_out_path);
+            return false;
```
martinezjavier: shouldn't this be `goto out` too?

man/tpm2_unseal.1.md
```diff
@@ -38,15 +38,22 @@ alive and pass that session using the **--input-session-handle** option.
...
-  * **-S**, **--input-session-handle**=_SESSION\_HANDLE_:
+## Session Options
 
-    Optional Input session handle from a policy session for authorization.
+  Options used for controlling sessions and policy events.
+
+  * **-S**, **--session**=_SESSION\_FILE_:
+
+    Optional, A session file from **tpm2_startauthsession**(1)'s **-S** option. This session
+    is used in leui of starting a session a session and using the PCR policy options. **-L** is
```
martinezjavier: s/a session a session/a session.

man/tpm2_unseal.1.md
```diff
@@ -38,15 +38,22 @@ alive and pass that session using the **--input-session-handle** option.
...
+    mutually exclusive of this option.
 
   * **-L**, **--set-list**==_PCR\_SELECTION\_LIST_:
 
     The list of pcr banks and selected PCRs' ids.
     _PCR\_SELECTION\_LIST_ values should follow the
     pcr bank specifiers standards, see section "PCR Bank Specfiers".
+    **-S** is mutually exclusive of this option.
```
martinezjavier: I think it should be: mutually exclusive *with* this option

martinezjavier: Can you please add an example to the man page using this new option and also extend the unseal integration test to test for regressions?

test/system/tests/tcti/abrmd/extended-sessions.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
```
martinezjavier: The other tests scripts uses a different shebang. I don't really mind which one to use but I prefer if we keep it consistent.

Also the other tests have a header with Copyright, etc. I would also like it to have consistency with that.

test/system/tests/tcti/abrmd/extended-sessions.sh
```diff
@@ -0,0 +1,91 @@
...
+trap cleanup EXIT
+
+cleanup
+
+echo $secret > $file_input_data
+
+tpm2_clear
+
+#
+# Test an extended policy session beyond client connections. This is ONLY supported by abrmd
```
martinezjavier: s/abrmd/tpm2-abrmd

martinezjavier: Or maybe tabrmd since that's how we call the tpm2-abrmd TCTI.

lib/tpm2_session.c
williamcroberts: Because it will get silently truncated and debugging it will be a PITA.
Ie this generates no warnings on GCC with no other options enabled:
```
    uint32_t a = 2^10;
    uint8_t x = a;
```

lib/tpm2_session.c
williamcroberts: I think this was just some dumbness on my part.

lib/files.c
williamcroberts: Yes it should

lib/tpm2_session.c
williamcroberts: yes

lib/tpm2_session.c
williamcroberts: yep

lib/tpm2_session.c
williamcroberts: Agreed.

lib/tpm2_session.h
williamcroberts: roger

man/tpm2_startauthsession.1.md
```diff
@@ -0,0 +1,71 @@
...
+# SYNOPSIS
+
+**tpm2_startauthsession** [*OPTIONS*]
+
+# DESCRIPTION
+
+**tpm2_startauthsession**(1) Starts a session with the TPM. The default is
+to start a *trial* session unless the **-a** option is specified. It outputs
+the session handle in a yaml format to stdout as key value "session-handle"
+and the handle in hex (0x1234) format.
```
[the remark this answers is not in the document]
williamcroberts: Because I don't know if tpm2_startauthsession will ever need to output something else. So I can add to a YAML output and keep backwards compatibility. In this case it's literally session-handle: 0x1234 which is super easy to parse without a full blown YAML parser. In fact the test scripts just use cut.

man/tpm2_startauthsession.1.md
williamcroberts: That and disconects should be spelled correctly, "disconnects".

man/tpm2_startauthsession.1.md
williamcroberts: Not true. We already do a save/load in startauthsession so abrmd marks it as a non-flush-able handle on IPC disconnection.

Tools that take a -S will set the continue bits in the session field, and then to kill it, tpm2_flushcontext can take the handle. For convenience, I would like to add the -S option to flushcontext so one never really needs to parse out the handle.

man/tpm2_startauthsession.1.md
williamcroberts: It does: 104c26b11442075c332922100204fa2663ec2d7c

test/system/tests/tcti/abrmd/extended-sessions.sh
williamcroberts: Yeah their is a lot-of-cleanups to do in these patches, and it will match the others we have already established.

tools/tpm2_createprimary.c
williamcroberts: All legacy stuff handed to me.

tools/tpm2_startauthsession.c
```diff
@@ -0,0 +1,138 @@
+//**********************************************************************;
+// Copyright (c) 2017, Intel Corporation
```
williamcroberts: 2018

tools/tpm2_startauthsession.c
williamcroberts: Nope. tpm2_session_new() takes ownership. So when it fails it should free session_data. I did it specifically to make this code easier to write. Nothing should fail from tpm2_session_data_new() to when you call this, and you can avoid dealing with the session_data pointer once you call this.

```
tpm2_session *tpm2_session_new(TSS2_SYS_CONTEXT *sapi_context,
        tpm2_session_data *data) {

    tpm2_session *session = calloc(1, sizeof(tpm2_session));
    if (!session) {
        free(data);
        LOG_ERR("oom");
        return NULL;
    }

    session->input = data;

    session->internal.nonceNewer.size = session->input->nonce_caller.size;

    if (!sapi_context) {
        return session;
    }

    bool result = start_auth_session(sapi_context, session);
    if (!result) {
        tpm2_session_free(&session);
        return NULL;
    }

    return session;
}
```

tools/tpm2_startauthsession.c
williamcroberts: See comment above, the session  takes ownership of the data.
```
tpm2_session_data *d = tpm2_session_data_new(TRIAL);
if (!d) { bail };
tpm2_session_data_set_xxx(d, foo);
tpm2_session_data_set_xxx(d, foo);

tpm2_session *s = tpm2_session_new(sapi, d);
if (!s) { bail(); } // no need to worry about d
```

man/tpm2_policypcr.1.md
williamcroberts: tpmNonce output will need to be added at some point, so eventually it will need multiple outputs. Id like to get some things to stdout versus always dumping to a file.

With that said tpm2_flushcontext could take -S so no need for handle to be on stdout. But I think it would be informative for users to be able to see the outputs from a command on both stdout (human readable/parseable) and via binary files (if needed as an option).

man/tpm2_policypcr.1.md
williamcroberts: It looks OK in the manpage. And I agree, this is how I would like to get a lot of the man page examples too. Most users have issues (rightfully) with understanding how to chain things together, not with the concept of a single tool.

This example is actually the integration test.

man/tpm2_startauthsession.1.md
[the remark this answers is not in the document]
williamcroberts: Yeah, it eventually needs tpmNonce per [table 16](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf)

man/tpm2_startauthsession.1.md
williamcroberts: Tools, like tpm2_unseal() could actually check for the session type in the session file and issue an error before the RC attribute error that the tpm will return. That is much more user friendly IMO. Something like, "the session being used for this is a trial session, where a policy session was expected."

williamcroberts:
@martinezjavier thanks for the review. Ill re-spin these for Monday and then I'll start flipping all the other -S uses to use the file.

lib/tpm2_session.c
williamcroberts: ahh tpm2_session_restore() creates a session context but does not need to start the authsession. This essentially lets you create a session context without calling startauthsession().

tools/tpm2_policypcr.c
williamcroberts: Output all things wrong with arguments versus just one error at a time.

tools/tpm2_policypcr.c
williamcroberts: Ill just use rc here and https://github.com/intel/tpm2-tools/pull/763#discussion-diff-160921182R116

tools/tpm2_policypcr.c
williamcroberts: Yes

tools/tpm2_policypcr.c
williamcroberts: it doesn't have multiple returns per-the-spec, so offhand I don't see why not. I just think having keys in the output helps from a human perspective at the expense of making scripts slightly more clunky.
``
tpm2_policypcr <args>
0x123456  # <-- what is this?
``
vs
```
tpm2_policypcr <args>
policy_digest: 0x123456 # <-- Oh its a policy_digest!
```

man/tpm2_unseal.1.md
williamcroberts: We have it in the tcti/abrmd tests since this abrmd specific. So I didn't want to jam it into the generic test suite in case users were running against in-kernel RM.

williamcroberts: I'm leaving *to*, as *with* just seems awkward when reading it.

williamcroberts:
Fixes #752

Issue #737, opened 2017-12-19
Tools that don't use a SAPI context, shouldn't fail if can't access the TPM


martinezjavier:
For example, tpm2_rc_decode could be used even if there's no access to a TPM. But if the user can't access the TPM it will fail due an error to initialize a TCTI context:

```
$ tpm2_rc_decode 0x9a2                                               
ERROR: Failed to initialize device TCTI context: 0xa000a
```

PR #767, opened 2018-01-15
tools: avoid sapi init if tool doesn't need it

williamcroberts:
Avoid the sapi initialization if tools request that
they don't need it via a flags argument to
tpm2_options_new();

Fixes: #737

### DIFF
--- a/lib/tpm2_options.h
+++ b/lib/tpm2_options.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, Intel Corporation
+ * Copyright (c) 2016-2018, Intel Corporation
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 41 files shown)


lib/tpm2_options.h
```diff
@@ -41,12 +41,19 @@
...
 union tpm2_option_flags {
     struct {
+#define TPM2_OPTION_FLAG_VERBOSE       (1 << 0)
         UINT8 verbose : 1;
+#define TPM2_OPTION_FLAG_QUIET         (1 << 1)
         UINT8 quiet   : 1;
+#define TPM2_OPTION_FLAG_ENABLE_ERRATA (1 << 2)
         UINT8 enable_errata  : 1;
+#define TPM2_OPTION_NO_SAPI            (1 << 3)
+        UINT8 no_sapi : 1;
```
martinezjavier: @williamcroberts This was merged before I got time to review it :smiley:

It looks good to me, I just have one comment: could we define these constants together after the tpm2_option_flags_init() macro instead of interleaving them with the struct fields? I think that would make it much easier to read.

williamcroberts: Done, ill put them on a separate commit

PR #770, opened 2018-01-16
test/system/tests/print.sh: Add

harisokanovic:
Verify tpm2_print works with a sample TPMS_ATTEST quote file.

Signed-off-by: Haris Okanovic <haris.okanovic@ni.com>


test/system/tests/print.sh
```diff
@@ -0,0 +1,63 @@
...
+
+        tpm2_print -t TPMS_ATTEST -f "$testFileBasename.TPMS_ATTEST" >"$PRINT_TEMP_FILE" 2>&1
+        diff "$PRINT_TEMP_FILE" "$expectedFile"
+    else
+        echo 1>&2 "Could not find input file"
+        exit 1
+    fi
+done
+
+rm "$PRINT_TEMP_FILE"
```
williamcroberts: you can just set up an on exit handler to remove this.

test/system/tests/print.sh
```diff
@@ -0,0 +1,63 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+#;**********************************************************************;
+
+onerror() {
+    echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
+    if [ -e "$PRINT_TEMP_FILE" ]; then
+        cat "$PRINT_TEMP_FILE"
+        rm "$PRINT_TEMP_FILE"
+    fi
```
williamcroberts: This should be in your onexit handler

test/system/tests/print-files/quote.printed
```diff
@@ -0,0 +1,20 @@
+magic: ff544347
```
williamcroberts: This and the binary blob should go under test fixtures: test/system/fixtures.

However, the test is missing that we want to detect if/when the binary blob output changes and breaks print. I would just output that blob for the test, and run print on it, and use a python script to load the yaml and verify what you can, ie that certain fields exist, etc.

harisokanovic: Sure, I can add a case to grab a quote and validate it. However, I'd like to have at least one premade blob with known good value -- I.e. a quote blob who's digest was validated by hand in a hexdump.

williamcroberts: That sounds good.

williamcroberts:
@harisokanovic have you updated this at all or am I missing something?

harisokanovic:
@williamcroberts, I've been busy with other work the last few weeks. I'll post an update next chance I get. Feel free to close this PR if you need to clear it from your view. I can open a new one after I implement your suggestions.

PR #775, opened 2018-01-23
Revert "[REVERTME] abrmd: patch for tcti init"

williamcroberts:
This reverts commit 9b3aa75514d38b55180a0bbf8130b7a097980e0e.


martinezjavier:
@williamcroberts I really don't like this `Revert "[REVERTME] foo: bar". The tpm2-abrmd patch was added and reverted in less than 48 hs. I would prefer the CI pipeline to be broken instead of polluting the git log history with these commits.

Yes, it will block merging new code, but it will just be temporary (would be a different story if the tpm2-abrmd fix would take days or weeks to land, but that never is the case).

Or at the very least, can we use as subject line `travis: add tpm2-abrmd patch that does foobar to fix the CI pipeline" and `travis: remove tpm2-abrmd patch that does foobar since landed upstream` or something like that? If a new developer looks the git history, she/he would question the project's quality with the current subject lines and that won't be fair.

williamcroberts:
So your arguing that a third party would assume that a reverted [REVERTME] patch implies poor quality but a big red "build failing" symbol on the main page doesn't? We can't leave the CI broken, it messes up other developers PRs.

We can make the subject lines different. I just like a big fat [REVERTME] tag that I can grep through at release time and make sure they all have a matching revert or their is a ticketed issue tracking why it wasn't reverted.

martinezjavier:
> So your arguing that a third party would assume that a reverted [REVERTME] patch implies poor quality but a big red "build failing" symbol on the main page doesn't? We can't leave the CI broken, it messes up other developers PRs.

Ideally we shouldn't have any of these two issues. My point is that these commits pollutes the git log history for little gain, since PRs takes some days to be merged anyways since they need to be released.

In fact, I now wonder if is a good idea to track tpm2-tss master in tpm2-tools and tpm2-abrmd. I understand that it helps to expose issues earlier but OTOH it constantly breaks these projects when there are non-backwards compatible changes in tpm2-tss.

Maybe, instead the tpm2-tools master branch should be developed against the latest tpm2-tss stable branch (so the CI will test that setup) and have a dev branch that's developed against the latest tpm2-tss master.

That way the master branch should be stable and only the dev branch will break, so is up to the tpm2-{abrmd,tools} developers to fix those issues before a new tpm2-tss version is released and the tpm2-{abrmd,tools} master branches are moved to track the new tpm2-tss release. This will also hep with the releases, since at the end you want a tpm2-{abrmd,tools} release to depend on a stable tpm2-tss version.

This was indeed an issue for the 3.0 release, and we had to do some mangling to allow the master to be branched to the 3.X release (i.e: dropping tpm2_import that was already in master but couldn't make it to the 3.0 release due some tpm2-tss API changes).

@williamcroberts, @flihp  what's your opinion on this?

williamcroberts:
1. The "noise" of revert commits is 2.3%. If the number was higher, I would worry more.
    There are 15 'Revert ' commits in the repo, so 30 commits if you include the revert commit and the 
    commit that was reverted.
    At this moment there are 1289 commits on master.
    30/1289 * 100 = 2.3%.
    To compare Bionic on master has 262 revert commits so 524 total reverted commits. With a total count 
    of 23430. Which is 2.23%. tpm2-tools 2.3% revert metric seems reasonable given my oh-so-
    comprehensive check of other  projects :-P

2. I used to think developing against stable bases worked better, but large scale projects have showed 
    me you just miss bugs. I'd rather know when things break via the CI ASAP. Eventually, the tss and tools 
    projects will become more mature and breaking changes will be introduced less, thus avoiding the 3.X
    fiasco. Anyone that needs a feature on 3.X (thus compatible with tss versions, etc) can back-port any 
    feature they want (has to be backwards compatible) and I'll cut a release for them if it passes testing.
    There is nothing stopping me from cutting a 4.X branch off 3.X if we needed a stable base, but the 
    next major tss release is off of master, so long term, I want to make sure the pieces are going to work 
    together.  There is a lot of tradeoffs here, most of them are made due to limited man-power.

williamcroberts:
FYI: SE Linux userspace is 4.8% revert rate.

I also like the [REVERTME] "tag" as it indicates you know this is garbage and that it's a workaround and that you should see a revert of it in the history. If it's just a normal patch title, the reader could assume that you think the patch is good, when it's total and utter garbage for a work around.

martinezjavier:
Ok, you made some very good points.

PR #878, opened 2018-01-31
Fix tpm2_import and enable tests for import

idesai:


tools/tpm2_import.c
```diff
@@ -57,6 +57,9 @@
 #define SYM_KEY_SIZE 16
 #define max_buffer_size  1024
 
+#define RSA_2K_MODULUS_SIZE_IN_BYTES 256
+#define RSA_2K_PUBLIC_MODULUS_OFFSET 28
```
williamcroberts: Couldn't we in theory use the unmarshal library and read in the file into the structure in memory? Then we can use the structure offsets instead of jumping around in the file?

williamcroberts:
I am going to take this with a ticket to fix it up. See #888

Issue #512, opened 2017-09-20
Change tpm2_nv{define,read,write,release,readlock} tools -a option to not require hex values


martinezjavier:
The tpm2 NV tools require the user to pass the hexadecimal values for the hierarchy authorization used, but we converted most tools to allow a pretty name instead (i.e: tpm2_evictcontrol).

williamcroberts:
This kind of collides with #514. Some tools take `-A o` and `-A p` as options.

Issue #778, opened 2018-01-30
stack overflow with tpm2_createpolicy 


MDO-TW:
Hello,
A stack smashing is detected  with tpm2_createpolicy  when user configures a number  of PCR greater than 8 (This limit is defined in 10.9.5 TPML_DIGEST in TPM Library Part 2: Structures).
Example :
`tpm2_pcrlist -L sha256:1,2,3,4,5,6,7,8,9 -o /tmp/pcrs`
`tpm2_createpolicy -P -L sha256:1,2,3,4,5,6,7,8,9 -F /tmp/pcrs -f /tmp/pcrs_policy`

To avoid that, I add this patch to tpm2-tools 
[001-check-PCR-number.patch.log](https://github.com/tpm2-software/tpm2-tools/files/1677022/001-check-PCR-number.patch.log)

and  I add this patch to tpm2-tss
[001-define-TPML_DIGEST_MAX.patch.log](https://github.com/tpm2-software/tpm2-tools/files/1677021/001-define-TPML_DIGEST_MAX.patch.log)

Thank you for this usefull projects

williamcroberts:
@MDO-TW let me know if PR #780 addresses your problem.

MDO-TW:
Yes It do it
Thanks

PR #879, opened 2018-01-31
Friendly Options for Hierarchy Specification

williamcroberts:
Correct the handling of hierarchies in the NV related tools to use friendly names like "o" for the owner platform rather than the raw numerical format.

This also moves the parsing logic to a common routine, with unit tests.

Fixes: #512

### DIFF
--- a/lib/tpm2_hierarchy.c
+++ b/lib/tpm2_hierarchy.c
@@ -34,10 +34,28 @@
 #include <sapi/tpm20.h>
 
 #include "log.h"
+#include "tpm2_hierarchy.h"
 #include "tpm2_util.h"
 
+/**
+ * Parses a hierarchy value from an option argument.
+ * @param value
+ *  The string to parse, which can be a numerical string as
+ *  understood by strtoul() with a base of 0, or an:
+ *    - o - Owner hierarchy
+ *    - p - Platform hierarchy
+ *    - e - Endorsement hierarchy
+ *    - n - Null hierarchy
+ * @param hierarchy
+ *  The parsed hierarchy as output.
+ * @param flags
+ *  What hierarchies should be supported by
+ *  the parsing.
+ * @return
+ *  True on success, False otherwise.
+ */
 bool tpm2_hierarchy_from_optarg(const char *value,
-        TPMI_RH_PROVISION *hierarchy) {
+        TPMI_RH_PROVISION *hierarchy, tpm2_hierarchy_flags flags) {
 
     if (!value) {
         return false;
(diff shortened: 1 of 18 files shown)


martinezjavier:
@williamcroberts it would be nice to use more descriptive subject lines in your pull request instead of just pointing to an issue number (and also require something in the description section with a more elaborated explanation of the change).

I also don't think that you meant issue #778 (stack overflow with tpm2_createpolicy) but instead #512 (Change tpm2_nv{define,read,write,release,readlock} tools -a option to not require hex values).

lib/tpm2_hierarchy.c
```diff
@@ -0,0 +1,78 @@
...
+    bool is_n = !strcmp(value, "n");
+    if (is_n) {
+        *hierarchy = TPM2_RH_NULL;
+        return true;
+    }
+
+    bool result = tpm2_util_string_to_uint32(value, hierarchy);
+    if (!result) {
+        LOG_ERR("Incorrect hierarchy value, got: \"%s\", expected [o|p|e|n]"
+                "or a number",
```
martinezjavier: I would say: or an hexadecimal number.

man/common/hierarchy.md
```diff
@@ -0,0 +1,6 @@
+    Suported options are:
+      * **o** for **TPM_RH_OWNER**
+      * **p** for **TPM_RH_PLATFORM**
+      * **e** for **TPM_RH_ENDORSEMENT**
+      * **n** for **TPM_RH_NULL**
+      * **`<num>`** where a raw number can be used.
```
martinezjavier: maybe **`<hex>`** where an hexadecimal number can be used.

lib/tpm2_hierarchy.c
martinezjavier: BTW, should we check that the hex value is a correct TPM_RH_* an report that or let the command fail when is sent by the TPM?

man/tpm2_evictcontrol.1.md
```diff
@@ -18,9 +18,8 @@ be evicted.
 # OPTIONS
 
   * **-A**, **--auth**=_AUTH_:
-    The authorization used to authorize the commands. Valid choices are:
-    *  **o** for **TPM_RH_OWNER**
-    *  **p** for **TPM_RH_PLATFORM**
+    The authorization used to authorize the commands.
+    [hierarchy specifiers](common/hierarchy.md)
```
martinezjavier: For storage, there are only platform and owner hierarchies, I don't think you can use the endorsement and null hierarchies so this wouldn't be correct.

martinezjavier: The same comment applies to all commands that work with the NVRAM.

lib/tpm2_hierarchy.c
martinezjavier: My latest comment made me realize that probably we should check for the Handle value passed to the tools too. I've filled issue #880 for that.

williamcroberts: I'd say let the TPM check that.

man/common/hierarchy.md
williamcroberts: You can use octal, or base 10, or even higher bases as described in man strtoul()

man/tpm2_evictcontrol.1.md
williamcroberts: Yeah I was wondering about how we wanted to handle that. We make the common parsing routine take some flags for valid options, but at the same time, anything invalid will be rejected by the TPM.

man/common/hierarchy.md
martinezjavier: Ah, right. I forgot we supported any base.

lib/tpm2_hierarchy.c
martinezjavier: Well checking is trivial, and I think it would be a better user experience if the tool can catch those and report a user friendly error message instead of getting a more generic error from the TPM.

I'm very busy with some internal projects lately but I'll try to make some time to propose a patch.

williamcroberts: The more I think about this, if you're putting in the raw hex, you should know what your doing and let it hit the TPM or another layer. If someone is trying to do something odd, perhaps with an experimental TPM feature and needs to pass a value straight in, they should be able to.

I'd like to be mechanism here, not policy.

man/tpm2_createprimary.1.md
```diff
@@ -22,11 +22,12 @@ will create and load a Primary Object. The sensitive area is not returned.
 
   * **-H**, **--hierarchy**=_HIERARCHY_:
     Specify the hierarchy under which the object is created. This will also dictate which authorization secret (if any) must be supplied.
-    Supported options are:
+    Suported options are:
```
williamcroberts: Spelling here and throughout...

lib/tpm2_hierarchy.c
martinezjavier: Yes, I agree with you when talking about the hierarchies (we do have a symbolic representation so why anyone would use a hex value?) but I still think it should be nice to have for handles since users have no other option that use hex values.

Issue #769, opened 2018-01-16
tpm2_rc_decode


williamcroberts:
remove TCTI options from manpage as no TCTI is required
Also as a reminder, don't forget to update the changelog adding tpm2_print tool.

williamcroberts:
Also move it to aux directory

williamcroberts:
Make sure print has tests as in #770

williamcroberts:
Forgot to update tpm2_rc_decode to not use a sapi context, done it:
https://github.com/tpm2-software/tpm2-tools/pull/893

williamcroberts:
Still to do:
* drop common TCTI options from manpage for aux tools
* tcti options should cause a parser error in common tool parsing.

williamcroberts:
closed on #906

PR #882, opened 2018-02-01
liberror: implement in tools

williamcroberts:
Eventually, we would like to formalize an interface for handling
tpm error codes analagous to C's strtoerr. Until we have such
a formalization, let this live under lib/future in the tools
project.

The tools will be consuming this for the 4.0 release under
ticket #769:
  - https://github.com/tpm2-software/tpm2-tools/issues/769

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -113,6 +113,8 @@ lib_libcommon_a_SOURCES = \
     lib/tpm2_attr_util.h \
     lib/tpm2_errata.c \
     lib/tpm2_errata.h \
+    lib/tpm2_error.c \
+    lib/tpm2_error.h \
     lib/tpm2_header.h \
     lib/tpm2_hierarchy.c \
     lib/tpm2_hierarchy.h \
@@ -206,7 +208,8 @@ check_PROGRAMS = \
     test/unit/test_tpm2_errata \
     test/unit/test_tpm2_session \
     test/unit/test_tpm2_policy \
-    test/unit/test_tpm2_hierarchy
+    test/unit/test_tpm2_hierarchy \
+    test/unit/test_tpm2_error
 
 test_unit_tpm2_rc_decode_unit_CFLAGS  = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
 test_unit_tpm2_rc_decode_unit_LDADD   = $(CMOCKA_LIBS) $(LIB_COMMON)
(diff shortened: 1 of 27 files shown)


Makefile.am
```diff
@@ -138,6 +138,16 @@ lib_libcommon_a_SOURCES = \
     lib/tpm2_tcti_ldr.c \
```
martinezjavier: Is there any need  to have #769 twice in the commit message?

lib/future/README.md
```diff
@@ -0,0 +1,6 @@
+# Future Library
+
+This folder is intended for future additions to the TSS and is meant as a sandbox/playground
+until the TSS is ready to take it in.
```
martinezjavier: I'm not sure to like this future sub-dir. Did I understand correctly that the idea of future is to have code that may eventually made to the TSS specification? Shouldn't then be added to the tpm2-tss project?

Please elaborate a little more about the goal of this and the expected process would be to push this lower in the stack or into the TCG specs.

martinezjavier: Right now I see a lot of code with no users in the repo, so it's kind of hard to review... I think it would be easier if these can make as normal helpers in the tpm2-tools project (living under lib like all the other helpers) and the tools be ported to it. Don't even need to be on a flag day, but can be done incrementally IMHO.

Makefile.am
williamcroberts: Nope.

lib/future/README.md
williamcroberts: @martinezjavier as far as in-code users, the tools where it reports the error hex code will be updated to use this. As for the goal, the goal is to get it into the spec, but since that is a ways off and the TSS project has it's own goal bigger then this at the moment, and I want it in the next release, I was proposing the future sub-dir.

With that said, I am not tightly coupled to the future dir, and we can drop that and just pull it in, adhering to our coding standard and worry about a port later.

martinezjavier:
@willliamcroberts Ok, agreed with the goal but I would prefer the latter (having them with the other helpers and using the tools coding standard).

Once/if they made to the spec, then can be added to tpm-tss and dropped from tpm-tools.

I had a quick look yesterday and I like the approach, I'll have it a more detailed look tomorrow.

Makefile.am
williamcroberts: Actually it's github messing it up, the verbatim text is:
```
The tools will be consuming this for the 4.0 release under
ticket #769:
  - https://github.com/tpm2-software/tpm2-tools/issues/769
```

PR #889, opened 2018-02-08
cleanup error handling

kaccardi:
Add extremely minimal error handling to tpm2_tools so that if onrun() fails, it will at least indicate that it will. Also, modify tpm2_createprimary's onrun() function to return slightly more useful errors (one day maybe someone will use them). refactor onrun() a little bit to make it easier to read and correct inconsistency with flag naming.

### DIFF
--- a/tools/tpm2_createprimary.c
+++ b/tools/tpm2_createprimary.c
@@ -62,10 +62,9 @@ struct tpm_createprimary_ctx {
     char *context_file;
     TPM2_HANDLE handle2048rsa;
     struct {
-        UINT8 A : 1;
+        UINT8 H : 1;
         UINT8 g : 1;
         UINT8 G : 1;
-        UINT8 C : 1;
     } flags;
 };
 
@@ -202,7 +201,7 @@ static bool on_option(char key, char *value) {
         if (!res) {
             return false;
         }
-        ctx.flags.A = 1;
+        ctx.flags.H = 1;
         break;
     case 'P':
         res = tpm2_password_util_from_optarg(value, &ctx.session_data.hmac);
@@ -239,7 +238,6 @@ static bool on_option(char key, char *value) {
         if(ctx.context_file == NULL || ctx.context_file[0] == '\0') {
             return false;
         }
-        ctx.flags.C = 1;
         break;
     case 'L':
         ctx.in_public.publicArea.authPolicy.size = BUFFER_SIZE(TPM2B_DIGEST, buffer);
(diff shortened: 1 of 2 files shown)


tools/tpm2_tool.c
```diff
@@ -157,6 +158,10 @@ int main(int argc, char *argv[], char *envp[]) {
      * 'main'.
      */
     ret = tpm2_tool_onrun(sapi_context, flags) ? 1 : 0;
+    if (ret != 0) {
+            LOG_ERR("Unable to run %s: %s", argv[0], strerror(errno));
```
williamcroberts: errno is not guaranteed to be set correctly here by all the other tools. Plus we don't want this to be in the business of handling tool specific errors. Unable to run %s argv[0] is fine.

This could lead to the dreaded "Unable to run tpm2_foo: Success" message.

williamcroberts: Extra indent level on LOG_ERR

tools/tpm2_createprimary.c
```diff
@@ -293,24 +294,26 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
...
+int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
     UNUSED(flags);
-    int returnVal = 0;
+    int returnVal;
 
-    if(ctx.flags.A == 1 && ctx.flags.g == 1 && ctx.flags.G == 1) {
-        returnVal = create_primary(sapi_context);
+    if (VALID_CTX(ctx) != true) {
+        LOG_ERR("Invalid option flags");
+        return -EINVAL;
```
williamcroberts: Just return 1 here.

tools/tpm2_createprimary.c
```diff
@@ -293,24 +294,26 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
...
+    }
 
-        if (returnVal == 0 && ctx.flags.C) {
-            returnVal = files_save_tpm_context_to_path(sapi_context, ctx.handle2048rsa,
-                                                       ctx.context_file) != true;
-        }
+    returnVal = create_primary(sapi_context);
+    if (returnVal != 0 || ctx.flags.C != 1) {
```
williamcroberts: ```if (!returnVal || !ctx.flags.C)```

If you wanted to kill flags, you could check if ctx.context_file is set.

tools/tpm2_createprimary.c
```diff
@@ -293,24 +294,26 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
...
+	(ctx.flags.A && ctx.flags.g && ctx.flags.G)
 
+int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
     UNUSED(flags);
-    int returnVal = 0;
+    int returnVal;
 
-    if(ctx.flags.A == 1 && ctx.flags.g == 1 && ctx.flags.G == 1) {
-        returnVal = create_primary(sapi_context);
+    if (VALID_CTX(ctx) != true) {
```
williamcroberts: I don't know if this is in the coding standard, but prefer !VALID_CTX(ctx) here.

tools/tpm2_createprimary.c
```diff
@@ -293,24 +294,26 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
     return *opts != NULL;
 }
 
-int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
+#define VALID_CTX(ctx) \
```
williamcroberts: static inline, better type checking/error reporting.

tools/tpm2_createprimary.c
```diff
@@ -293,24 +294,26 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
...
+        return returnVal;
+    }
 
-        if(returnVal) {
-            return 1;
-        }
-    } else {
-        return 1;
+    if (files_save_tpm_context_to_path(sapi_context, ctx.handle2048rsa,
+                                       ctx.context_file) != true) {
```
williamcroberts:
```
bool res = files_save_tpm_context_to_path();
if (!res) {
    / /ERROR
}
```
But you could also go for the tail call here,

return !files_save_tpm_context_to_path();

tools/tpm2_createprimary.c
```diff
@@ -203,7 +203,7 @@ static bool on_option(char key, char *value) {
         if (!res) {
             return false;
         }
-        ctx.flags.A = 1;
+        ctx.flags.H = 1;
```
williamcroberts: this looks fine.

tools/tpm2_createprimary.c
kaccardi: @williamcroberts in this case change_primary() doesn't just return 0 or 1 so you can't do !returnVal. This is why in general I feel it's better to do an explicit check against value if returning an int rather than a bool.

williamcroberts:
@dantpm 

The static routine ```create_primary()``` should just be updated to return true/false as more granularity than that isn't being used for anything.

Issue #634, opened 2017-11-17
stdout as yaml


williamcroberts:
Make sure that ALL tools are doing tpm2_tooloutput in a yaml format and update all tests to not use grep/awk/sed to find values. Use here-doc python scripts that use PyYaml to validate/parse the yaml output for values.

williamcroberts:
@martinezjavier you'll be glad to know that an output of just a number is valid YAML. It's a scalar type output. So those tools really don't need any changes, but the docs likely need to be updated to include the output formatting.

martinezjavier:
@williamcroberts ah cool, I didn't know that.

williamcroberts:
@martinezjavier I am looking at libyaml and thinking maybe the tools should just hand back a list of libyaml events that can be emitted, and let tpm2_tool.c handle output.

Right now, if a tool fails, it could fail with partial, incorrect output.... just a thought.

martinezjavier:
@williamcroberts yeah, it's an interesting idea. I just don't know if it's going to make the output logic more complex...

williamcroberts:
@martinezjavier not sure, and not in this release. We can do that later with no major version bump.

PR #898, opened 2018-02-16
Yaml output

williamcroberts:
Audit the tools for printf usages and replace with tpm2_tool_output. Also verify that output is in a YAML compliant format. Add tests that verify that the output can be loaded via the YAML module in python.

Fixes: #634

### DIFF
--- a/lib/tpm2_alg_util.c
+++ b/lib/tpm2_alg_util.c
@@ -351,7 +351,7 @@ UINT8* tpm2_extract_plain_signature(UINT16 *size, TPMT_SIGNATURE *signature) {
         if (!buffer) {
             goto nomem;
         }
-        memcpy(buffer, &hmac_sig, *size);
+        memcpy(buffer, hmac_sig, *size);
         break;
     }
     case TPM2_ALG_ECDSA: {
--- a/man/tpm2_getcap.1.md
+++ b/man/tpm2_getcap.1.md
@@ -58,16 +58,32 @@ command.
     * handles-saved-session:
       Display handles about saved sessions.
 
+  * **-l**, **--list**:
+    List known supported capability names. These names can be
+    supplied as the argument to the **-c** option. Output is in a
+    YAML compliant list to stdout.
+
+    For example:
+```
+    - algorithms
+    - commands
+    - properties-fixed
+    ...
+```
+
 [common options](common/options.md)
 
 [common tcti options](common/tcti.md)
 
 # EXAMPLES
 
-```
-tpm2_getcap --capability="properties-fixed"
+To list the fixed properties of the TPM:
 
-```
+```tpm2_getcap --capability="properties-fixed"```
+
+To list the supported capability arguments to **-c**:
+
+```tpm2_getcap -l```
 
 # RETURNS
 
--- a/test/system/helpers.sh
+++ b/test/system/helpers.sh
@@ -99,3 +99,53 @@ hash_alg_supported() {
         fi
     done
 }
+
+#
+# Verifies that the contexts of a file path provided
+# as the first argument loads as a YAML file.
+#
+function yaml_verify() {
+python << pyscript
+from __future__ import print_function
+
+import sys
+import yaml
+
+with open("$1") as f:
+	try:
+		y = yaml.load(f)
+	except yaml.YAMLError as exc:
+		sys.exit(exc)
+pyscript
+}
+
+#
+# Given a file as argument 1, prints the value of the key
+# provided as argument 2 and optionally argument 3 (for nested maps).
+# Note that if key is a string, pass the quotes. This allows lookups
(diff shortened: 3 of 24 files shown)


williamcroberts:
Looks like 
```TPM2_PT_VENDOR_STRING_2:
  raw: 0x2054504D
  value: "TPM"
```
The value has a garbage byte, can't repro locally...hmmm

williamcroberts:
Ahh need to subtract ```i``` from ```j``` when setting the initial value as the byte pointer walks forward....

tools/tpm2_getcap.c
```diff
@@ -159,9 +159,8 @@ int sanity_check_capability_opts (void) {
```
martinezjavier: In the commit message:

s/cononical/canonical
s/Ie/i.e
s/worK/work

tools/tpm2_getcap.c
```diff
@@ -159,9 +159,8 @@ int sanity_check_capability_opts (void) {
 
     size_t i;
     for (i = 0; i < CAPABILITY_MAP_COUNT; ++i) {
-        int cmp = strncmp(capability_map [i].capability_string,
-                          options.capability_string,
-                          strlen(capability_map [i].capability_string));
+        int cmp = strcmp(capability_map [i].capability_string,
```
martinezjavier: Remove the space between `capability_map` and `[i]`. Nice catch BTW.

man/tpm2_getcap.1.md
```diff
@@ -58,6 +58,11 @@ command.
     * handles-saved-session:
       Display handles about saved sessions.
 
+  * **-l**, **--list**:
+    List known supported capability names. These names can be
+    supplied as the argument to the -c option. Output is in a
```
martinezjavier: \*\*-c\*\* option.

test/system/tests/dictionarylockout.sh
```diff
@@ -32,6 +32,30 @@
 #;**********************************************************************;
 ###this script use for test the implementation tpm2_dictionarylockout 
 
+function yaml_get() {
```
martinezjavier: Probably we should at some point move all these functions to a lib.sh or something and source it from the tests. I haven't checked, but from previous review I remember quite similar functions so there's no need for the code duplication.

test/system/helpers.sh
```diff
@@ -99,3 +99,22 @@ hash_alg_supported() {
         fi
     done
 }
+
+#
+# Verifies that the contexts of a file path provided
+# as the first argument loads as a YAML file.
+#
+function yaml_verify() {
```
martinezjavier: Nice, you are doing what I suggested :)

But I think this should be one of the first patches in the series, since other patches are duplicating this function in the tests. Also, it seems that Github is messing with the commit order, since I saw in previous patches that you were sourcing this file and using the function...

test/system/tests/quote.sh
```diff
@@ -51,14 +51,7 @@ Handle_ak_quote=0x81010016
 Handle_ek_quote=0x81010017
```
martinezjavier: Can you trim the quoted line in the commit message, we get the idea. No need to have such a long hex number :)

test/system/tests/nv.sh
```diff
@@ -58,26 +60,6 @@ onerror() {
 }
 trap onerror ERR
 
-function yaml_get() {
```
martinezjavier: This should also be in a separate patch. Basically I would do:

1) Factor out all the existing funcions in helper.sh as the first patch
2) Add any new functions that are going to be used
3) Do all the YAML conversions and sourcing helpers.sh in the tests.

test/system/tests/create.sh
```diff
@@ -49,27 +49,6 @@ cleanup() {
 }
 trap cleanup EXIT
 
-function yaml_get() {
```
martinezjavier: Same here, I think it will be easier to read if you first factor out yaml_get(), then on a subsequent patch add yaml_get_kv() and finally on another patch change the users.

lib/tpm2_alg_util.c
```diff
@@ -351,7 +351,7 @@ UINT8* tpm2_extract_plain_signature(UINT16 *size, TPMT_SIGNATURE *signature) {
         if (!buffer) {
             goto nomem;
         }
-        memcpy(buffer, &hmac_sig, *size);
```
martinezjavier: I would also have this at the beginning of the set since is a trivial fix (although maybe you did since as mentioned Github seems to mess with the commits order).

williamcroberts:
@martinezjavier the ordering on github drives me nuts, it uses the timestamp versus the actual commit parent order. I plan on re-writing the history on this and bring in the helper.sh changes first.

PR #899, opened 2018-02-16
ci: Have travis install libgcrypt dev package.

flihp:
This is a new dependency from the addition of the ESAPI library to
tpm2-tss.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,7 @@ addons:
   apt:
     packages:
     - autoconf-archive
-    - libcurl4-openssl-dev
     - libdbus-1-dev
-    - libgcrypt-dev
     - libglib2.0-dev
     - clang-3.8
     - pandoc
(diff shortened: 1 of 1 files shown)


flihp:
Ugh. I hadn't even considered this but it looks like the travis image won't install gcrypt dev packages and the openssl dev packages at the same time? Not sure this is really what's happening but from my first pass over the build output it looks like it may be.

flihp:
that's not it, the build is complaining about not being able to install `librtmp-dev` which is required by `libcurl4-openssl-dev`

flihp:
```
The following packages have unmet dependencies:
 libgcrypt20-dev : Conflicts: libgcrypt-dev
                   Conflicts: libgcrypt11-dev but 1.5.3-2ubuntu4.5 is to be installed
```

flihp:
looks like this issue: https://launchpad.net/ubuntu/+source/libgcrypt20/1.6.1-2 which oddly isn't an issue in the tpm2-tss and tpm2-abrmd builds. All evidence points to the VM travis image being the issue since the two other projects are using the container build.

Might be able to work around this by doing some `sudo apt-get` stuff in `before_install`. Alternatively it may be time to bite the bullet and port the travis-ci stuff over to the container build.

dantpm:
There may be a hysterical reason for it, but it’s unusual to need both as they provide similar functionality.

Perhaps the test suite needs openssl and the tpm2 software needs gcrypt?

Gcrypt is a lot more buggy than OpenSSL, but sometimes gcrypt is used because of the GNU license.

Dan

flihp:
This is fun: manually removing libgcrypt11-dev causes apt to remove libcurl4-openssl-dev. Definitely the right thing to do.

flihp:
This was an exercise in futility. See #900.

PR #900, opened 2018-02-17
Disable ESAPI & do build in a container

flihp:


.travis.yml
```diff
@@ -1,7 +1,4 @@
-dist: trusty
-# This is a lie: we don't need sudo but this is required to get an
-# Ubuntu image with a libc that isn't ancient, and with cmocka libs.
-sudo: required
+sudo: false
```
williamcroberts: I'm surprised sudo calls below work...

Issue #901, opened 2018-02-19
Question: Why there is no "tpm2_createek" command?


liuqun:
In TPM1.2 package tpm-tools, the EK creating command is named as "tpm_createek".
Why not add a command "tpm2_createek" here?
Does "tpm2_getpubek" handles all the jobs including creating a new EK which seems to be a job that "tpm2_createek" should do?

See: http://trousers.sourceforge.net/man/tpm_createek.8.html

Issue #421, opened 2017-08-22
Non-persistent AKs


srihanfb:
There doesn't seem to be an easy way to create a non-persistent AK under the EK. tpm2_create/load don't have an option to use PolicySecret for authorization. One way would be add that functionality to tpm2_create/load. An alternative approach might be to add an option to tpm2_getpubak to write the public and private portions to a file, and skip making the key persistent.

williamcroberts:
I'm planning the next release, I am not 100% sure what you want here, any way you could submit prototype patches?

williamcroberts:
@srihanfb 
Oh now that I am more familiar with these tools, yes I see what you mean. However, I don't see CreatePrimary returning the private portion of the AK.

I think what you would want is a -C option to save the context of the handle.

srihanfb:
@williamcroberts : Tss2_Sys_Create returns both the public and private parts of the key. For the AK we're creating a key under the endorsement hierarchy so we need Create and not CreatePrimary. I you look at tpm_getpubak you'll see that it uses Tss2_Sys_Create.

williamcroberts:
@srihanfb Yes I see that, so you also need the private key exported as well. Will add that in.

srihanfb:
@williamcroberts : yeah, I think we need an option to export the private key. One possibility is to have an option that just generates the AK, but doesn't actually load it, and exports both the private and public key. That's how I've been using it internally.

We also need a corresponding tool to load a key into the TPM. The existing tpm_load.c isn't sufficient because it doesn't create a session. I can file a separate issue for that if you like.

williamcroberts:
tpm2_load takes a session option via the `-S` option (currently, it will change).

Sessions are tricky, if your using an RM that isn't the latest ABRMD they won;t wonk with the tools. Abrmd currently has a bug preventing it from working with it too: https://github.com/tpm2-software/tpm2-abrmd/issues/317

PR #904, opened 2018-02-21
OpenSSL Format and getpubak/ek renames

williamcroberts:
Since we are going to a major version bump, and tpm2_getpubak and tpm2_getpubek had horrendous names that caused confusion, rename them to tpm2_createak and tpm2_createek respectively.

Change the -f option to -p, and use -f for specifying the format of the generated public key.

Fixes: #901
Fixes: #421

The session issues are on #917


williamcroberts:
Nice catch @liuqun I will make sure to update that.

williamcroberts:
@martinezjavier - Session handling, yet again, has reared it's ugly head. So some tools, and potential future tools, may need multiple sessions for auth. For instance tpm2_getpubak (tpm2_createak on this PR) needs multiple auths as indicated by the "password" options.

I'm thinking that the initial process of `-S` for session handle is a bit naive and limited. Perhaps, we use the existing password options and do something like `--password=session:path/to/session.dat`

This has a few effects:
1. Less options, at the expense of expanding the password mini-language
2. Works with existing tool options, no need to add new ones
3. Supports compound tools.

Thoughts?

martinezjavier:
@williamcroberts I'll need to spend some time reading the docs again to fully grasp the issue and the possible solutions (it seems authorization and sessions is something I keep forgetting the details). But here are some thoughts.

IIUC you want to use the same option for password, HMAC and policy sessions, right? Then I think it shouldn't be called `password` but `authorization`.

By multiple you mean that it may require different authorization methods or that should support more than one? In other words, it should be an exclusive or or an either? AFAICT the TPM2 spec supports the latter, so I think the tools should do the same. With your suggested approach, would we have something like:

```
--password=session:path/to/session.dat,hmac:secretpassword
```

Or how do you plan to support multiple authorization methods?

If possible I also prefer to keep parsing at a minimum. Even if our format is simple, parsing strings in C is very error prone. So I wonder if wouldn't be better to just keep different `-p` and `-S` options for password and sessions and let getopt to do the parsing for us.

williamcroberts:
@martinezjavier 
With respect to parsing,  the code already parses passwords. For example `-P"hex:abcd"`. So this would just be -P"session:session.dat"

With respect to why, tpm2_getpubak is a good example. It requires an owner and endorsement authorizations, so a single session option, like `-S` won't work as is. Either we would need:
1. More options
2. Smarten-up -S to have a map to the object, `-S"owner=owner.dat:hierarchy=hrchy.dat"`
3. Smarten up the existing "password" options
  1. Optionally rename them to authorization. I'm trying to keep non-backwards compat changes to a 
      minimum. 

I am leaning towards 3.

As far as using multiple methods for authorization, complex auths require a session AFAIK. The auth structure, TPMS_AUTH_COMMAND one sets the session handle to either:
1.  A valid session handle from create_session
2. TPM2_RS_PW (for password/hmac auth)

martinezjavier:
> With respect to parsing, the code already parses passwords. For example -P"hex:abcd". So this would just be -P"session:session.dat"

I don't mind to extend the command, but still I think it should not be called `password`, since as you said the structure is TPMS_AUTH_COMMAND and is the authorization in the session are of the command.

Backward incompatible change wouldn't be an issue, we already broke a lot of things and are even changing the tool names for these.

PR #905, opened 2018-02-21
tcti option cleanups

williamcroberts:
Commits that:
1. Make --tcti invalid (like -T) when a tool requests that no sapi be initialized
2. Update the tpm2_rc_decode manpage to remove the tcti options

Fixes: #769

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -279,13 +279,19 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv, char **envp,
      * grep -rn case\ \'[a-zA-Z]\' | awk '{print $3}' | sed s/\'//g | sed s/\://g | sort | uniq | less
      */
     struct option long_options [] = {
-        { "tcti",          required_argument, NULL, 'T' },
         { "help",          no_argument,       NULL, 'h' },
         { "verbose",       no_argument,       NULL, 'v' },
         { "quiet",         no_argument,       NULL, 'Q' },
         { "version",       no_argument,       NULL, 'V' },
         { "enable-errata", no_argument,       NULL, 'Z' },
+        /*
+         * tcti must be last, because when a tool requests that
+         * a sapi is not provided, we subtract 1 from the size
+         * of the array length so --tcti is not valid.
+         */
+        { "tcti",          required_argument, NULL, 'T' },
     };
+    size_t long_options_len =  ARRAY_LEN(long_options);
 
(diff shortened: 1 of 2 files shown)


lib/tpm2_options.c
```diff
@@ -279,23 +279,30 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv, char **envp,
...
         { "verbose",       no_argument,       NULL, 'v' },
         { "quiet",         no_argument,       NULL, 'Q' },
         { "version",       no_argument,       NULL, 'V' },
         { "enable-errata", no_argument,       NULL, 'Z' },
+        /*
+         * tcti must be last, because when a tool requests that
+         * a sapi is not provided, we subtract 1 from the size
+         * of the array length so --tcti is not valid.
+         */
+        { "tcti",          required_argument, NULL, 'T' },
```
martinezjavier: Even when are nicely documented, I always have the feeling that making these kind of assumptions is fragile. I would rather prefer to just keep all the common options but instead just check for `flags & TPM2_OPTIONS_NO_SAPI` in the `'T'` case. I've proposed PR #906 that does this if you want to evaluate that option.

Yes, that means that the tools that don't support `-T, --tcti` will still says that an argument is required if a user just passes `-T` with no argument. But it will print an error if the option is attempted to use correctly.

The option wouldn't even be documented in the tools' man pages so why an user will try to use it in the first place?

lib/tpm2_options.c
```diff
@@ -279,23 +279,30 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv, char **envp,
...
     };
+    size_t long_options_len =  ARRAY_LEN(long_options);
 
     const char *tcti_conf_option = NULL;
 
     /* handle any options */
     const char* common_short_opts = "T:hvVQZ";
     if (tool_opts && (tool_opts->flags & TPM2_OPTIONS_NO_SAPI)) {
```
martinezjavier: wonder if we should rename this to `TPM2_OPTIONS_NO_TCTI` ?

williamcroberts: Well it ultimately ends up with a NULL sapi pointer from the tools perspective

lib/tpm2_options.c
williamcroberts: 1. The docs are lacking
2. Users do all sorts of stuff
3. When TCTI is disabled, I want the missing -T and --tcti messages thrown by getopt.

Issue #909, opened 2018-02-23
ESAPI


williamcroberts:
Figure out how to get the ESAPI building with tools, so we can start to use that going forward instead of the SAPI directly.

This will hopefully reduce the size of the lib that we have to carry around the tools, decrease the time it takes to create new tools, and reduce our maintenance.

williamcroberts:
FYI details of the PR @flihp did to at least get tools building is here: https://github.com/tpm2-software/tpm2-tools/pull/900

PR #913, opened 2018-02-27
travis: use a custom docker container

williamcroberts:
Travis uses ancient docker and VM images. Switch to a custom
docker image that has all dependencies for the TSS stack pre-installed.

This has a few benefits:
1. Local testing by just running the docker command in the travis.yml file
2. Less dependency setup code in the travis.yml
3. ESAPI builds work on Travis now (the tools can start switching over)

This has a major draw backs:
1. Back on travis VM based builds, which can have long wait queues.
  A. A workaround to this, is you can run your full CI buildup locally
     so you can just merge outside of travis if need-be.

Fixes: #909

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 #;**********************************************************************;
 #
 # Copyright (c) 2017, Intel Corporation
@@ -29,10 +30,51 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
-#!/usr/bin/env bash
 
 # all command failures are fatal
 set -e
+
+WORKSPACE=`dirname $TRAVIS_BUILD_DIR`
+
+echo "Workspace: $WORKSPACE"
+
+source $TRAVIS_BUILD_DIR/.ci/download-deps.sh
+
+get_deps "$WORKSPACE"
+
+export LD_LIBRARY_PATH=/usr/local/lib/
+
+echo "echo changing to $TRAVIS_BUILD_DIR"
+# Change to the the travis build dir
+cd $TRAVIS_BUILD_DIR
+
+echo "starting dbus"
+mkdir -p /var/run/dbus
+if [ -e /var/run/dbus/pid ]; then
+  rm /var/run/dbus/pid
+fi
+
+# start dbus
+dbus-daemon --fork --system
+echo "dbus started"
+
+# let dbus have time to start up
+sleep 5
+
+echo "starting tpm server"
+# start the tpm server
+/ibmtpm974/src/tpm_server &
+echo "tpm server started"
+
+# start tpm2-abrmd
+
+# let the tpm simulator have time to start up before abrmd
+# tries to connect.
+sleep 5
+
+echo "starting abrmd server"
+tpm2-abrmd --allow-root --tcti=socket &
+echo "started abrmd server"
 
 if [ -d build ]; then
   rm -rf build
@@ -74,17 +116,6 @@ else #GCC
   echo "Exported ENABLE_COVERAGE=true"
   config_flags="--disable-hardening --enable-code-coverage"
 fi
-
-# Travis is using some EGLIBC version of libc
-# which appears to have some atexit() handling
-# issues when dlclose is used. Just disable it
-# for gcc builds, as clang DOES NOT run the
-# the system tests.
-# I cannot reproduce this on:
-#  OS: Ubuntu 16.04.3 LTS
-#   C: ldd (Ubuntu GLIBC 2.23-0ubuntu10) 2.23
-config_flags="--disable-dlclose $config_flags"
-echo "Enabling configure flag --disable-dlclose"
 
 # Bootstrap in the tpm2.0-tss tools directory
 ./bootstrap
--- a/.ci/download-deps.sh
+++ b/.ci/download-deps.sh
@@ -1,5 +1,6 @@
-#!/bin/bash
-
+#!/usr/bin/env bash
+#;**********************************************************************;
+#
 # Copyright (c) 2017, Intel Corporation
 # All rights reserved.
 #
@@ -13,6 +14,10 @@
 # this list of conditions and the following disclaimer in the documentation
 # and/or other materials provided with the distribution.
 #
+# 3. Neither the name of Intel Corporation nor the names of its contributors
+# may be used to endorse or promote products derived from this software without
+# specific prior written permission.
+#
 # THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 # AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 # IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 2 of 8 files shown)


.ci/docker.run
```diff
@@ -29,11 +30,44 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
-#!/usr/bin/env bash
 
 # all command failures are fatal
 set -e
 
+T=`dirname $0`
```
williamcroberts: Should probably use TRAVIS_BUILD_DIR.... or delete any other usages of TRAVIS_BUILD_DIR and use this "T" or TOP variable. (T is a convention from Android, open to change).

.travis.yml
```diff
@@ -33,50 +16,29 @@ env:
    - secure: "XrDhcREuntMn39qhSFBSSKZli9ZduGKCoq5rs/b+6e5zQIDwyDklwZIvgmrx2BJV5Sh7c5HB2Uvh9I7HdUOOOqDveZ9dcvjr2me7wfrhfMChB2miK318xVfyrK+6mkRnSMfuvG1CZiIcKYVStN1tX1uou9n3CnGz1ndmt5e4QwSQHR3K3g/HEXEgNjuei0OvtvE35/UaOpdxCoDlV3oExa3pU8mXLL77eskq35ebzwGsOu2zxD3wFhE1DF0O9yH4skH+fOH9ByxqK4n++uxbxXHW2oh4I/mEqr2wyyead/wuVa5UdxMDM7traUojlXy5O/Nn2Zfeky3ngSRQjzEYhKmiU2GiQ6ZBh78tmijPmKQpNOg2+0Uop7iT0SmefhappYOzc7KSTnYtWeEbZ4BXxeXzQZ98WsJj7Tmh258Y53KE2DDrmZYwk4EJuc6/4dv/HNMDdH8ZDzAdwMBi4lcsLnf8GQG/jfWYsrvCQQECkSfP2ICuRL9k/8pUUAnbp3X/WbCEMtaETy9STz7PNVqiWsSv/m9lBDVaHwZ7K87TdqhyDhYRDmPYP8qtu/M0tlNZ/M0VGo09xUgkb8Q2nNB7ahC7XYtBeHEIoIu6nf18gW8LgknpuhskF75ZHrKhh5VeN1zoPbG/oFYEk7ehNjGZC268d7jep5p5EaJzch5ai14="
```
williamcroberts: if this goes into the travis config on the server, this gets rid of this project specific setting

.travis.yml
```diff
@@ -33,50 +16,29 @@ env:
...
-    - sha256sum libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb | grep -q 99f550db61b0054715095fc77901280e81235900435f90b7db34af406f053832
-    - sha256sum libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb | grep -q e44b09b81717a9ae86ff17adae1729682cb00b5285710e991f7b61c8a351c744
-    - sudo dpkg -i libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb
-    - sudo dpkg -i libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb
+- docker pull shootingatshadow/tpm2-tss
 
 script:
-    - ./.ci/travis-build-and-run-tests.sh
+#
+# Docker starts you in a cloned repo of your project with the changed checkout out.
```
williamcroberts: Wow... try that sentence again.

martinezjavier:
@williamcroberts I'll review it tomorrow since I'm about to leave for today.

.ci/docker.run
williamcroberts: I think getting rid of travis-build dir makes sense, so everything works without it, smarter...

.travis.yml
```diff
@@ -33,50 +16,29 @@ env:
...
-    - sudo dpkg -i libcmocka0_1.0.1-2_amd64.deb
-    - sudo dpkg -i libcmocka-dev_1.0.1-2_amd64.deb
-    # openssl 1.0.2g
-    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb
-    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb
-    - sha256sum libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb | grep -q 99f550db61b0054715095fc77901280e81235900435f90b7db34af406f053832
-    - sha256sum libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb | grep -q e44b09b81717a9ae86ff17adae1729682cb00b5285710e991f7b61c8a351c744
-    - sudo dpkg -i libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb
-    - sudo dpkg -i libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb
+- docker pull shootingatshadow/tpm2-tss
```
williamcroberts: @martinezjavier we can add a fedora image here for better testing on your platform

.ci/download-deps.sh
```diff
@@ -24,15 +29,33 @@
...
-PATH=${PATH}:/usr/local/clang/bin
-(pushd ${TRAVIS_BUILD_DIR}/tpm2-tss && \
- make install && \
- popd)
+	echo "pwd clone abrmd: `pwd`"
+	git clone https://github.com/tpm2-software/tpm2-abrmd.git
+	pushd tpm2-abrmd
+	echo "pwd build abrmd: `pwd`"
+	./bootstrap
+	./configure --with-dbuspolicydir=/etc/dbus-1/system.d
```
williamcroberts: We put a conf file in the image in the right spot that doesn't have the "tss" user, so delete this.

.ci/docker.run
martinezjavier: @williamcroberts I like `builddir` because I'm used to autotools terminology, but I don't really have a strong opinion on the naming.

martinezjavier: > However, the get_deps() command knows that /workspace is the right spot.

Or pass an argument to get_deps()  so you don't need to hardcode it there?

.travis.yml
```diff
@@ -33,50 +16,29 @@ env:
...
-    - sha256sum libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb | grep -q e44b09b81717a9ae86ff17adae1729682cb00b5285710e991f7b61c8a351c744
-    - sudo dpkg -i libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb
-    - sudo dpkg -i libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb
+- docker pull tpm2software/tpm2-tss
 
 script:
-    - ./.ci/travis-build-and-run-tests.sh
+#
+# Docker starts you in a cloned repo of your project with the changed checkout out.
+# We want those changes IN the docker imager, so use the -v option to mount the
```
martinezjavier: s/imager/image right?

.travis.yml
```diff
@@ -33,50 +16,29 @@ env:
...
+# prpoject repo in the docker image.
```
martinezjavier: s/prpoject/project

.travis.yml
```diff
@@ -33,50 +16,29 @@ env:
...
-    - sudo dpkg -i libcmocka0_1.0.1-2_amd64.deb
-    - sudo dpkg -i libcmocka-dev_1.0.1-2_amd64.deb
-    # openssl 1.0.2g
-    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb
-    - wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb
-    - sha256sum libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb | grep -q 99f550db61b0054715095fc77901280e81235900435f90b7db34af406f053832
-    - sha256sum libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb | grep -q e44b09b81717a9ae86ff17adae1729682cb00b5285710e991f7b61c8a351c744
-    - sudo dpkg -i libssl1.0.0_1.0.2g-1ubuntu4.10_amd64.deb
-    - sudo dpkg -i libssl-dev_1.0.2g-1ubuntu4.10_amd64.deb
+- docker pull tpm2software/tpm2-tss
```
martinezjavier: I'm not well versed in neither Docker nor Travis, so forgive me if my questions are silly. But from where is still image pulled? I can't find it in the [Docker registry](https://hub.docker.com). 

Is that pulling from https://github.com/tpm2-software/tpm2-tss/blob/master/Dockerfile using some magic Travis + Github integration?

.travis.yml
```diff
@@ -33,50 +16,29 @@ env:
...
+# We want those changes IN the docker imager, so use the -v option to mount the
+# prpoject repo in the docker image.
+#
+# Also, pass in any env variables required for the build via .ci/docker.env file
+#
+# Execute the build and test procedure by running .ci/docker.run
+#
+- >
+  docker run --env-file .ci/docker.env \
+    -v `pwd`:/workspace/tpm2-tools shootingatshadow/tpm2-tss \
```
martinezjavier: What's `shootingatshadow/tpm2-tss` ?

martinezjavier:
@williamcroberts I'm not that familiar with "clould" technologies, but the changes look good to me.

I like the flexibility of being able to pull any Docker image. And yes, testing on Fedora as you said would obviously be interesting to me. But we are quite close to upstream so I think any image should work.

.ci/docker.run
williamcroberts: Yeah I want to get rid of all the hardcodes and then standardize around some variables. We essentially have two spots we care about:
1. Where we work from `/workspace`
2. Where the tools clone is mounted `/workspace/tpm2-tools`

We should probably pass `TRAVIS_BUILD_DIR` just for compliance reasons, but we can just set `WORKSPACE=/workspace` or something like that.

Let me work on this more today, it'll fail due to abrmd...

.travis.yml
williamcroberts: Travis does nothing here really. We pull the docker image from docker hub and pass control to it.

The image in this PR was coming from my dockerhub repo, but not it will be coming from:
https://hub.docker.com/r/tpm2software/tpm2-tss/

dockerhub is the github of docker images, you can push/pull them.

.travis.yml
williamcroberts: That's the image name, it's going to change to tpm2software/tpm2-tss

https://hub.docker.com/r/tpm2software/tpm2-tss/

.travis.yml
martinezjavier: @williamcroberts right, I read that from the docker-pull man page but I swear that I searched for `tpm2` and didn't find that image. I see it now, so I guess I'm just blind...

williamcroberts:
Last build passing was: https://travis-ci.org/tpm2-software/tpm2-tools/builds/347950240?utm_source=github_status&utm_medium=notification

The only change I did was squash some commits.... so lets see if it runs and passes OK

PR #923, opened 2018-03-05
[for 3.x] Backport various tools options fixes from master

martinezjavier:
This pull request contains backports with fixes for many issues with the tools' options.

It contains the commits in PR #747 and #751 that for some reasons never made it to the 3.x branch but should since are mostly fixes.


williamcroberts:
@martinezjavier can you just update the CHANGELOG with the important bits and then ill merge it as in. Let me know when you want a release.

martinezjavier:
@williamcroberts updated the changelog as suggested. Thanks.

PR #924, opened 2018-03-05
tools: remove incorrect 's' option from tpm2_startup optstring

martinezjavier:
The tool doesn't have this short option, so the opstring isn't correct.


martinezjavier:
```
** (process:5186): CRITICAL **: failed to allocate dbus proxy object: Error calling StartServiceByName for com.intel.tss2.Tabrmd: GDBus.Error:org.freedesktop.DBus.Error.Spawn.FileInvalid: Cannot do system-bus activation with no user
ERROR: tcti init allocation routine failed for library: "tabrmd" options: "(null)"
ERROR: Could not load tcti, got: "tabrmd"
```

PR #930, opened 2018-03-08
INSTALL.md: Document that Pandoc is Needed for Man Pages

### LINKED ISSUE #929: INSTALL.md: Document that Pandoc is Needed for Man Pages
INSTALL.md mentions that md2man-roff is needed to build man pages.
It also needs to document that pandoc is needed.
Otherwise the man pages are not built and configure and make run, but no man pages.

dantpm:
Add mention that pancod is needed (in addition to md2man-roff) to build man pages.

Fixes #929.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,7 +23,7 @@ To build and install the tpm2-tools software the following software is required:
   * Curl library and header files
 
 #### Optional Dependencies:
-  * To build the man pages you need [md2man-roff](https://github.com/sunaku/md2man)
+  * To build the man pages you need [pandoc](https://github.com/jgm/pandoc)
   * To enable the new userspace resource manager, one must get tpm2-tabrmd
     (**recommended**).
 


INSTALL.md
```diff
@@ -23,7 +23,7 @@ To build and install the tpm2-tools software the following software is required:
   * Curl library and header files
 
 #### Optional Dependencies:
-  * To build the man pages you need [md2man-roff](https://github.com/sunaku/md2man)
+  * To build the man pages you need [md2man-roff](https://github.com/sunaku/md2man) and pandoc
```
williamcroberts: nope just pandoc!

We might want to make INSTALL.md point to this part of the wiki or vice-versa:
https://github.com/tpm2-software/tpm2-tools/wiki/Getting-Started#installing

But that wiki installing page is way better. So we may want to copy that context to INSTALL.md and make
the wiki Installing page point to INSTALL.md file.

dantpm:
As requested, I replaced md2man-roff with pandoc in the list of prerequisites (instead of adding pandoc and keeping md2man-roff).

PR #931, opened 2018-03-08
tss2: Account for changes TPMA_* macros in upstream headers.

flihp:

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -74,7 +74,7 @@ echo "tpm server started"
 sleep 5
 
 echo "starting abrmd server"
-tpm2-abrmd --allow-root --tcti=socket &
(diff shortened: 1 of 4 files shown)


flihp:
strange, error messages are all related to dbus activation (dbus trying to launch an instance of the tabrmd). I'm assuming this isn't intentional ...

liuqun:
@flihp Maybe the error is cause since we rename libtcti_socket.so to libtcti_mssim.so. And tpm2-abrmd failed to access libtcti_socket.so

flihp:
thanks @liuqun, lots of churn to keep up with these last few weeks. Will have this building again in a bit.

flihp:
looks like I b0rked up the default config string for the mssim TCTI upstream

Issue #925, opened 2018-03-07
Distribute generated man pages in release tarball


mgerstner:
The current release tarballs require `pandoc` for building the man pages. This is quite a big dependency (pandoc requires a haskell stack). It would be easier for packaging to ship the generated man pages in the release tarball.

This can be achieved by adding `dist_man_MANS := $(man1_MANS)` to Makefile.am. However it conflicts with the current conditional `HAVE_PANDOC`. If pandoc isn't present then the man pages will not be installed in the context of the resulting release tarball, although they are already generated.

If have not found a clean solution to address this within automake. One approach might be to always expect `pandoc` to be present when building the source tarball. Maybe you have a good idea how to approach this.

williamcroberts:
On one hand, if your going to distribute part-of-the build, it would seem analogous to building the C files as well. But I get where you're coming from. When you're cross-compiling (which I have no idea how to do with autotools) it should be tetsting for host pandoc (if it isn't that's wrong), and then using the host to build the manpages. What build machine can't you just do "apt-get install pandoc"...

On the other hand, since it's not platform specific, just generate it and release it.

I wonder what @martinezjavier thinks, as he is a downstream packager, and has more understanding about these types of issues than I do, thoughts?

mgerstner:
@williamcroberts, it is actually an issue in our SUSE build environment. We do not have the pandoc and haskell packages in our enterprise repositories. Therefore I had to disable man page installation for tpm2-tools there for the moment.

My only options to get the man pages back there would be to add pandoc & friends to our enterprise repositories (big issue) or packaging the generated man pages separately myself (maintenance burden).

martinezjavier:
@mgerstner thanks for bringing this, I agree with you. In fact, we had the exact same issue in our enterprise repository. For now the man pages are added as a patch, but we need to find a better solution.

@williamcroberts maybe what we could do is to generate the distribution tarball using `make distcheck` instead of `make dist`? That way the man pages will be generated and added to the distribution tarball.

And remove the `HAVE_PANDOC` check in the `Makefile.am` as mentioned by @mgerstner.

What do you think?

williamcroberts:
I have 2 concerns with that:
1. Removing the PANDOC check. Someone could create a release tarball without the manpages.
2. distcheck will add binaries, right? Which I don't want to do.

What if we hooked make dist and checked for pandoc. We don't want someone to accidentally create a release tarball with out these.

[...]

PR #937, opened 2018-03-14
dist: include built manpages

williamcroberts:
Generate the man pages from markdown and distribute in the
dist tarball by making a dependency on the man pages via
the EXTRA_DIST variable.

Also, make "make dist" fail if pandoc is not installed:
$ make -j4 dist > /dev/null
You do not have pandoc, a requirement for the distribution of manpages
make[2]: *** [dist-hook] Error 1
make[1]: *** [distdir] Error 2
make: *** [dist] Error 2

Behavior under clean:
1. On a build tree, make clean removes the manpages
2. On a dist tarball, make clean leaves the manpages

Fixes: #925

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -326,6 +326,15 @@ if HAVE_PANDOC
     man/man1/tpm2_startup.1 \
     man/man1/tpm2_unseal.1 \
     man/man1/tpm2_verifysignature.1
+
+# If pandoc is enabled, we want to generate the manpages for the dist tarball
+EXTRA_DIST += $(man1_MANS)
+else
+# If pandoc is not enabled, we want to complain that you need pandoc for make dist,
+# so hook the target and complain.
+dist-hook:
+	@(>&2 echo "You do not have pandoc, a requirement for the distribution of manpages")
+	@exit 1
 endif
 
 MARKDOWN_COMMON_DEPS = \


williamcroberts:
@mgerstner and @martinezjavier this was the cleanest, and surprisingly simple way to get it done.

The behavior (surprisingly) was correct under clean as well, that wasn't by understanding autotools!

Thoughts?

williamcroberts:
Whoops! Are docker image is missing pandoc!

joshuagl:
Could be useful to host the Dockerfile (and other configuration) for the container in this repo?

williamcroberts:
@joshuagl the container is publicly available for download (look at .travis.yml) and runs the scripts in the .ci directory.

The dockerfile itself is not-up-to-date and incorrect. I just modify the image directly and push.

mgerstner:
Thank you very much for working on this issue.

Indeed a surprisingly simple solution. Autotools are beyond me sometimes ;-)

I have tested the change and it behaves like documented. Should work out well to solve the issue.

martinezjavier:
@mgerstner thanks for testing, I'm merging this now then.

martinezjavier:
@williamcroberts I think we should do the same for the 3.x branch.

williamcroberts:
@martinezjavier done on 3.X: https://github.com/tpm2-software/tpm2-tools/pull/943

PR #938, opened 2018-03-15
Upstream reorg

flihp:
This accounts for changes upstream to location of headers and some library prefix stuff.

### DIFF
--- a/lib/tpm2_tcti_ldr.h
+++ b/lib/tpm2_tcti_ldr.h
@@ -34,8 +34,8 @@
  * Loads a TCTI from a friendly name, library name, or path.
  * For example
  *  friendly:     path = tabrmd
- *  library name: path = libtcti-socket.so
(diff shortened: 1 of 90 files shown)


lib/tpm2_tcti_ldr.h
```diff
@@ -34,8 +34,8 @@
  * Loads a TCTI from a friendly name, library name, or path.
  * For example
  *  friendly:     path = tabrmd
- *  library name: path = libtcti-socket.so
- *  full path:    path = /home/user/lib/libtcti-custom.so
+ *  library name: path = libtss2-tcti-socket.so
```
joshuagl: Wasn't the socket TCTI recently renamed to mssim?

williamcroberts: Ill update this after we pull this in.

PR #939, opened 2018-03-15
Minor docs tweaks

joshuagl:
- Ensure github links point to the latest repos in the new tpm2-software organisation
- Fix TCTI common options for the socket->mssim rename
- Remove a TODO item which pointed to a bug which is closed WONTFIX


joshuagl:
Travis failures are due to the changes in tpm2-tss, should be fixed by merging #938

PR #942, opened 2018-03-16
all: Switch TPMA_OBJECT_SIGN to _ENCRYPT

AndreasFuchsSIT:
Removes all mentions of TPMA_OBJECT_SIGN, since upstream will remove them after https://github.com/tpm2-software/tpm2-tss/pull/868 is merged.

The build currently fails, because upstream removed tpm20.h though...

I guess I or anyone should rebase, once a patch for tpm20.h removal is in master...


joshuagl:
Builds are broken with master of tpm2-tss, we need this PR, #944 and #945 to fix the builds.

PR #944, opened 2018-03-16
fix up headers

williamcroberts:
tpm20.h was removed from the TSS. Change to the new tss2_sys.h
header file. This introduced issues in where we were missing
standard header files via include directives, as tpm20.h was
including them, correct these.

Also, the TPM2B type was removed, re-declare this in tpm2_util.h,
as that generic container type is convenient.

### DIFF
--- a/lib/files.c
+++ b/lib/files.c
@@ -32,8 +32,9 @@
 #include <inttypes.h>
 #include <stdbool.h>
 #include <stdio.h>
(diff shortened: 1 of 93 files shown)


joshuagl:
Builds are broken with master of tpm2-tss, we need this PR, #942 and #945 to fix the builds.

PR #945, opened 2018-03-16
tpm2_tool: define ABI version locally

tstruk:
TSS doesn't export the ABI defines in public heades anymore.
This is a followup for https://github.com/tpm2-software/tpm2-tss/pull/864


joshuagl:
Builds are broken with master of tpm2-tss, we need this PR, #942 and #945 to fix the builds.

dantpm:
This might fix the broken build, or at least help, for issue #946.  See also pull request #945.

PR #947, opened 2018-03-20
Various man page fixes

joshuagl:
This small series includes corrections to what I believe are errors/typos in several manual pages.

Issue #892, opened 2018-02-09
set heirarchy to owner by default for all tools, make -H optional.


kaccardi:
We could reduce complexity for the user by making some of these options that are required today optional, with a sane default. For heirarchy, could we set this by default to Owner, then remove the requirement that -H be specified on the command line.

williamcroberts:
Yes I think this makes sense. We should however set all tools to default to the owner hierarchy. Currently the tools that have hierarchy defaults are set to platform, which was a poor choice on my part.

if/when we do this:
1. update all the tools to use a default platform of owner
2. update -H to generate on if not selected.

kaccardi:
ok, I'll edit the title of this issue to encompass all the tools.

joshuagl:
I've started working on this, I'd appreciate if someone could assign to me.

williamcroberts:
@joshuagl I need to add you into one of the groups.

PR #948, opened 2018-03-20
Set heirarchy to owner by default for all tools, make -H optional

joshuagl:
The following series of patches implements a solution to issue #892. The changes within:

- changes all tools which take a hierarchy argument to make that argument optional and default to the owner hierarchy when no value is passed
- for tools which take a handle argument in order to specificy a persistent handle when making a transient object persistent no longer require a handle to be specified. 
  - The *tpm2_evictcontrol* tool will do the right thing and, when a handle to a transient object is specified without a matching permanent handle we now automatically try and find an open slot in the persistent handle namespace.
  - tools which take an optional handle to persist a transient object at now have the option of specifying **-** as a handle. When passed a handle of *(-** the tool will automatically find an open handle (as above) and use it, printing out the address of the used handle.

**Note**: this series also includes the changes from #947, as otherwise the man page changes within would need rebasing after #947 landed.

Fixes: #947
Fixes: #892

### DIFF
--- a/lib/tpm2_capability.c
+++ b/lib/tpm2_capability.c
@@ -34,6 +34,7 @@
 #include <tss2/tss2_sys.h>
 
 #include "log.h"
+#include "tpm2_util.h"
 
 #include "tpm2_capability.h"
 
@@ -66,4 +67,51 @@ bool tpm2_capability_get (TSS2_SYS_CONTEXT *sapi_ctx,
     }
 
     return true;
+}
+
+bool tpm2_capability_find_vacant_persistent_handle (TSS2_SYS_CONTEXT *sapi_ctx,
+        UINT32 *vacant) {
+
+    TPMS_CAPABILITY_DATA capability_data = TPMS_CAPABILITY_DATA_EMPTY_INIT;
+    bool ret = tpm2_capability_get(sapi_ctx, TPM2_CAP_HANDLES,
+                    TPM2_PERSISTENT_FIRST, TPM2_MAX_CAP_HANDLES,
+                    &capability_data);
+    if (!ret) {
+        return false;
+    }
+
+    bool handle_found = false;
+    UINT32 count = capability_data.data.handles.count;
+    if (count == 0) {
+        /* There aren't any persistent handles, so use the first */
+        *vacant = TPM2_PERSISTENT_FIRST;
+        handle_found = true;
+    } else if (count == TPM2_MAX_CAP_HANDLES) {
+        /* All persistent handles are already in use */
+        return false;
+    } else if (count < TPM2_MAX_CAP_HANDLES) {
+        /* iterate over used handles to ensure we're selecting
+            the next available handle. */
+        UINT32 i;
+        for (i = TPM2_PERSISTENT_FIRST;
+            i <= (UINT32)TPM2_PERSISTENT_LAST;
+            ++i) {
+            bool inuse = false;
+            UINT32 c;
+            for (c = 0; c < count; ++c) {
+                if (capability_data.data.handles.handle[c] == i) {
+                    inuse = true;
+                    break;
+                }
+            }
+
+            if (!inuse) {
+                *vacant = i;
+                handle_found = true;
+                break;
+            }
+        }
+    }
+
+    return handle_found;
 }
--- a/lib/tpm2_capability.c
+++ b/lib/tpm2_capability.c
@@ -1,0 +1,69 @@
+//**********************************************************************;
+// Copyright (c) 2018, Intel Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// 3. Neither the name of Intel Corporation nor the names of its contributors
+// may be used to endorse or promote products derived from this software without
+// specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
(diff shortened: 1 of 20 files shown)


joshuagl:
Possibly it would make sense for the tools to be more consistent by making *tpm2_evictcontrol * take a **-** value to the **H** option when we want to have the tool find an open handle in the persistent namespace?

lib/tpm2_capability.c
```diff
@@ -0,0 +1,117 @@
...
+}
+
+bool tpm2_capability_find_vacant_persistent_handle (TSS2_SYS_CONTEXT *sapi_ctx,
+        UINT32 *vacant) {
+
+    TPMS_CAPABILITY_DATA capability_data = TPMS_CAPABILITY_DATA_EMPTY_INIT;
+    bool ret = tpm2_capability_get(sapi_ctx, TPM2_CAP_HANDLES,
+                    TPM2_PERSISTENT_FIRST, TPM2_MAX_CAP_HANDLES,
+                    &capability_data);
+    bool handle_found = false;
```
williamcroberts: line 83, so you check ret first.

lib/tpm2_capability.c
```diff
@@ -0,0 +1,117 @@
...
+
+    if (!ret)
+        return false;
```
williamcroberts: brackets around everything

lib/tpm2_capability.h
```diff
@@ -0,0 +1,68 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+
+#ifndef LIB_TPM2_CAPABILITY_H_
+#define LIB_TPM2_CAPABILITY_H_
+
+
+/**
```
williamcroberts: You need a body describing what it does, here and throughout.

tools/tpm2_createak.c
```diff
@@ -556,6 +563,17 @@ int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
...
 
+    if (ctx.find_persistent_ak) {
+        bool ret = tpm2_capability_find_vacant_persistent_handle(sapi_context,
+                        &ctx.ak.in.handle);
+        if (!ret) {
+            LOG_ERR("ak-handle/k passed with a value of '-' but unable to find"
+                    " a vacant persistent handle!");
+            return 1;
+        }
+        tpm2_tool_output("AK persistent-handle: 0x%x\n", ctx.ak.in.handle);
```
williamcroberts: maybe `ak-persistent-handle: 0x1234`

I'm not a fan of spaces, even though JSON handles it fine. I might not have told you, but all outputs are in JSON, this way tools with compound outputs can have some sane way of parsing them.

tools/tpm2_hash.c
```diff
@@ -152,7 +152,7 @@ static bool on_option(char key, char *value) {
 bool tpm2_tool_onstart(tpm2_options **opts) {
 
     static struct option topts[] = {
-        {"hierarchy", required_argument, NULL, 'H'},
+        {"hierarchy", optional_argument, NULL, 'H'},
```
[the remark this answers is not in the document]
joshuagl: Thanks for the pointer, I'm not very familiar with getopt but by the time I'd finished opening the pull request I was beginning to lean to **-H-** for consistency with the later changes, I'll make this change.

tools/tpm2_evictcontrol.c
```diff
@@ -64,11 +65,13 @@ struct tpm_evictcontrol_ctx {
         UINT8 c : 1;
         UINT8 P : 1;
     } flags;
+    UINT8 persist_passed : 1;
```
williamcroberts: Another way could be setting P to two bits and setting it 2 if it was an implied -P, thoughts?

tools/tpm2_hash.c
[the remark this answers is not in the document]
williamcroberts: I didn't know it has an optional_argument parameter, so I had to go check it out.

tools/tpm2_createak.c
joshuagl: I hadn't realised that we were outputting JSON with a consideration to parsing the tools output. With that in mind particularly, I like your proposal.

lib/tpm2_capability.h
joshuagl: Will add.

lib/tpm2_capability.c
joshuagl: Of course, I usually prefer that style but it's a while since I wrote much C. Will fix.

lib/tpm2_capability.c
joshuagl: Sure, will fix.

tools/tpm2_hash.c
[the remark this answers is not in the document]
joshuagl: Actually, this is just plain wrong for the change as implemented. Right the -H option *requires* an argument. If H isn't passed we default to the owner hierarchy, if it is passed we expect a hierarchy value to be defined.

I think the right fix here is to change this back to required_argument. We *could* take -H- and implicitly set owner hierarchy, but that's already the default so I'm not sure it makes much sense.

tools/tpm2_evictcontrol.c
joshuagl: I'm not opposed. It seems marginally harder to grok, but not significantly so. Happy to make that change.

williamcroberts: Consider this optional, you can leave it as you like.

tools/tpm2_createak.c
williamcroberts: Even for tools that have a single output, it's considering a "JSON Scalar" at that point. So the JSON output is quite flexible.

williamcroberts: To be clear, in your case it's a JSON key value pair.

tools/tpm2_hash.c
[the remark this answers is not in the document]
williamcroberts: On top of that we are confusing `-H` when it's a hierarchy versus handle.

When it's specifying a handle, it should be required and have `-` argument to mean pick a persistent handle for me.

When it's the hierarchy, it should be for specifying the hierarchy and default to owner.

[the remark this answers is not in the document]
joshuagl: That's how this series implements things, for the most part. The only place we vary slightly is that for *tpm2_evictcontrol* if the handle passed via **-H** is a transient handle and no **-p** is specified we pick a persistent slot to persist the transient handle at.

We could:
- leave this as implemented
- require **-H-** be passed to pick a persistent slot
- optionally take **-H-** to pick a persistent slot, otherwise fall back on the logic as implemented (picking a persistent slot to use when **tpm2_evictcontrol -H** is invoked with a transient handle)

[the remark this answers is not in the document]
williamcroberts: Well `H` or `c` (which c is just a context blob where it get's the value of `H` from) need to be specified no matter what. So I think we just leave this as implemented, or make `-p-` mean pick a slot, thoughts?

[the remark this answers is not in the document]
joshuagl: I'm leaning to leave this as implemented.

joshuagl:
A few extra changes have been pushed:

1. *tpm2_evictcontrol* has been fixed to pass the tests (it was failing in how it was used in tests/hmac.sh)
2. two new commits introduce extra tests for *tpm2_evictcontrol*, I should probably add more tests to test the new ways the tools can be invoked
3. *tests/hash.sh* was updated as part of 2859152 to explicitly pass a null hierarchy. With a non-null hierarchy the tpm2_hash command emits a ticket as well as the hashed value, which confuses the YAML parser.

williamcroberts:
This all looks good, just need to rebase and re-push.

williamcroberts:
Also mark the issues this fixes by following this:
https://help.github.com/articles/closing-issues-using-keywords/

Something like:
Fixes: #947 
Fixes: #892 

Actually that above in this conversation should do it. Now when we merge this, it will auto-magically close those issues referenced via Fixes.

williamcroberts:
Stupid merge commit from github... rebased and moved to: https://github.com/tpm2-software/tpm2-tools/pull/950

PR #949, opened 2018-03-20
bootstrap: generate library sources

williamcroberts:
Generate the library sources and create variables for them.

This will help make the Makefile smaller and easier to
maintain.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -1,6 +1,26 @@
 #! /bin/sh
+set -e
 
+# generate list of source files for use in Makefile.am
+# if you add new source files, you must run ./bootstrap again
+src_listvar () {
+    basedir=$1
+    suffix=$2
+    var=$3
+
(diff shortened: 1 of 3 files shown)


bootstrap
```diff
@@ -2,5 +2,26 @@
...
+    basedir=$1
+    suffix=$2
+    var=$3
+
+    find "${basedir}" -name "${suffix}" | LC_ALL=C sort | tr '\n' ' ' | (printf "${var} = " && cat)
+    echo ""
+}
+
+VARS_FILE=src_vars.mk
+AUTORECONF=${AUTORECONF:-autoreconf}
```
joshuagl: We have this line twice now.

williamcroberts: Need to correct that.

probably want a set -e in this as well, so the script dies on failure.

As far as why they list them, Auto tools doesn't support dynamically adding the source, so you need to do it like this. In Android, you can dynamically add them, and in regular make you can as well with just a normal wildcard expression.

PR #954, opened 2018-03-25
tpm2_certify: fix "'type' may be used uninitialized" warning

casantos:
get_key_type does not assign a value to 'type' on error conditions,
which makes the compiler believe that it will be used uninitialized in
the LOG_ERR call, in set_scheme. In practice this will never happen
because set_scheme returns immediately if get_key_type retuns false but
the compiler does not understand that logic and fails because warnings
are treated as errors.

Signed-off-by: Carlos Santos <casantos@datacom.ind.br>


casantos:
I noticed that this change was committed to master but did not appear in any of the subsequent 3.x.y versions.

PR #959, opened 2018-03-26
Change more tools to default to owner hierarchy

joshuagl:
My previous attempt to set the default hierarchy for all tools to the owner hierarchy in #950 turned out not to include all tools. This series fixes up some tools I missed.

### DIFF
--- a/man/tpm2_nvreadlock.1.md
+++ b/man/tpm2_nvreadlock.1.md
@@ -24,6 +24,8 @@ is released on subsequent restart of the machine.
     specifies the handle used to authorize:
     * **0x40000001** for **TPM_RH_OWNER**
     * **0x4000000C** for **TPM_RH_PLATFORM**
+    Defaults to **0x40000001**, **TPM_RH_OWNER**, when no value has been
+    specified.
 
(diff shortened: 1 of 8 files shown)


man/tpm2_nvreadlock.1.md
```diff
@@ -24,6 +24,8 @@ is released on subsequent restart of the machine.
     specifies the handle used to authorize:
     * **0x40000001** for **TPM_RH_OWNER**
     * **0x4000000C** for **TPM_RH_PLATFORM**
```
martinezjavier: I wonder why this man page still uses hex numbers instead of the pretty names `o` and `p` like all the other NVRAM related tools. AFAICT all use the same `tpm2_hierarchy_from_optarg()` helper function so it just seems like a left over in the man page.

Maybe fixing that too while you are there?

joshuagl: You're right, tmp2_readlock is using `tpm2_hierarchy_from_optarg()`. I've pushed an extra commit which updates the man page to use the pretty names.

PR #962, opened 2018-03-28
Implicit H follow on

joshuagl:
Following on from #950 and #959 this series fixes **tpm2_createprimary** and **tpm2_loadexternal** to not require an explicit hierarchy option be passed, using the default of **TPM2_OWNER** when one isn't specified.

Further we add some tests for the default owner hierarchy and persistent handle finding changes made in #950.

### DIFF
--- a/test/system/tests/createak.sh
+++ b/test/system/tests/createak.sh
@@ -31,6 +31,8 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
 
+source helpers.sh
+
 onerror() {
     echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
     exit 1
@@ -38,7 +40,7 @@ onerror() {
 trap onerror ERR
 
 cleanup() {
-    rm -f ek.pub ak.pub ak.name ak.name
+    rm -f ek.pub ak.pub ak.name ak.name ak.log
 
     # Evict persistent handles, we want them to always succeed and never trip
     # the onerror trap.
@@ -54,6 +56,11 @@ cleanup
 
 tpm2_createek -Q -H 0x8101000b -g rsa -p ek.pub
 
-tpm2_createak -Q -E 0x8101000b  -k 0x8101000c -g rsa -D sha256 -s rsassa -p ak.pub  -n ak.name
+tpm2_createak -Q -E 0x8101000b -k 0x8101000c -g rsa -D sha256 -s rsassa -p ak.pub -n ak.name
+
+# Find a vacant persistent handle
+tpm2_createak -E 0x8101000b -k - -g rsa -D sha256 -s rsassa -p ak.pub -n ak.name > ak.log
+phandle=`yaml_get_kv ak.log \"ak\-persistent\-handle\"`
+tpm2_evictcontrol -Q -a o -H $phandle
 
 exit 0
--- a/test/system/tests/createek.sh
+++ b/test/system/tests/createek.sh
@@ -31,6 +31,8 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
 
+source helpers.sh
+
 onerror() {
     echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
     exit 1
@@ -38,7 +40,7 @@ onerror() {
 trap onerror ERR
 
 cleanup() {
-    rm -f ek.pub
+    rm -f ek.pub ek.log
 
     # Evict persistent handles, we want them to always succeed and never trip
     # the onerror trap.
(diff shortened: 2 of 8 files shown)


test/system/tests/createak.sh
```diff
@@ -54,6 +54,11 @@ cleanup
 
 tpm2_createek -Q -H 0x8101000b -g rsa -p ek.pub
 
-tpm2_createak -Q -E 0x8101000b  -k 0x8101000c -g rsa -D sha256 -s rsassa -p ak.pub  -n ak.name
+tpm2_createak -Q -E 0x8101000b -k 0x8101000c -g rsa -D sha256 -s rsassa -p ak.pub -n ak.name
+
+# Find a vacant persistent handle
+tpm2_createak -E 0x8101000b -k - -g rsa -D sha256 -s rsassa -p ak.pub -n ak.name > ak.log
+phandle=`grep "ak-persistent-handle: " ak.log | awk '{print $2}'`
```
williamcroberts: I like to get better validation that this is indeed YAML compliant. I've been writing HERE doc python scripts that use the yaml module to import the output. I think it's in helpers.sh. You may need to add a new routine if your format doesn't follow those.

The only time we don't check, is if the tool has simple yaml scalar output, for example:
```
tpm2_xxx
42
```
That's a scalar output in YAML.

test/system/tests/createek.sh
```diff
@@ -50,4 +50,10 @@ cleanup
 
 tpm2_createek -H 0x81010005 -g rsa -p ek.pub
 
+cleanup
+
+tpm2_createek -H - -g rsa -p ek.pub > ek.log
+phandle=`grep "persistent-handle: " ek.log | awk '{print $2}'`
```
williamcroberts: same here and throughout.

test/system/tests/createak.sh
joshuagl: yaml_verify from helpers.sh works for the output of the tools I'm modifying here, thanks for pointing it out.

joshuagl:
Yes, it's documented in _tpm2_createak.1.md_ as:

> If a value of - is passed the tool will find a vacant persistent handle to use and print out the automatically selected handle.

and words to similar effect in _tpm2_createek.1.md_, _tpm2_getmanufec.1.md_ and _tpm2_evictcontrol.1.md_.

test/system/tests/createak.sh
```diff
@@ -54,6 +56,13 @@ cleanup
 
 tpm2_createek -Q -H 0x8101000b -g rsa -p ek.pub
 
-tpm2_createak -Q -E 0x8101000b  -k 0x8101000c -g rsa -D sha256 -s rsassa -p ak.pub  -n ak.name
+tpm2_createak -Q -E 0x8101000b -k 0x8101000c -g rsa -D sha256 -s rsassa -p ak.pub -n ak.name
+
+# Find a vacant persistent handle
+tpm2_createak -E 0x8101000b -k - -g rsa -D sha256 -s rsassa -p ak.pub -n ak.name > ak.out
+phandle=`grep "ak-persistent-handle: " ak.out | awk '{print $2}'`
```
williamcroberts: Can you use something like yaml_get_kv() from helpers.sh here to get the value versus awk

This will take care of verifying via yaml.load() as well as fetching the value.

williamcroberts: here and throughout the changes

test/system/tests/createak.sh
```diff
@@ -31,14 +31,16 @@
...
+
 onerror() {
     echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
     exit 1
 }
 trap onerror ERR
 
 cleanup() {
-    rm -f ek.pub ak.pub ak.name ak.name ak.log
+    rm -f ek.pub ak.pub ak.name ak.name ak.out
```
martinezjavier: you are changing `.log` for `.out`. I don't really have a preference but I see the other tests use `.log` so I would just keep it for consistency.

test/system/tests/getmanufec.sh
```diff
@@ -31,6 +31,8 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
```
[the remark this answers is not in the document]
joshuagl: Sure, will do. That should give me more space to write a short log and make the git log output look less cluttered.

test/system/tests/createak.sh
joshuagl: Agreed.

test/system/tests/createak.sh
joshuagl: Will do. The only place that causes a problem is in _getmanufec.sh_ which does a `tpm2_listpersistent | grep -q $phandle`. The problem is that the YAML conversion in yaml_get_kv converts the hex handle to a scalar, so we end up passing grepping for an int value.

There are two ways to resolve this:
1. quote the hex values we output so that they are always treated as a string
2. use a different YAML loader in yaml_get_kv function which doesn't do the conversion

Any thoughts?

joshuagl: The third and simplest option is of course to drop the line `tpm2_listpersistent | grep -q $phandle`

joshuagl: I went with the option of dropping that line from the test.

williamcroberts:
Failed: ```./tests/getmanufec.sh: line 102: yaml_get_kv: command not found```

PR #964, opened 2018-03-28
man: clean up tcti description

williamcroberts:
When the TCTI configuration got simpler, the wording
never followed. Clean up the TCTI description so
it's more understandable.

### DIFF
--- a/man/common/tcti.md
+++ b/man/common/tcti.md
@@ -1,90 +1,94 @@
-# TCTI ENVIRONMENT
+# TCTI Configuration
 
-This collection of environment variables that may be used to configure the
-various TCTI modules available.
+The TCTI or "Transmission Interface" is the communication mechanism with the
+TPM. TCTIs can be changed for communication with TPMs across different
+mediums.
 
-The values passed through these variables can be overridden on a per-command
-basis using the available command line options, see the _TCTI_OPTIONS_ section.
+To control the TCTI, the tools respect:
 
-The variables respected depend on how the software was configured.
+  1. The command line option **-T** or **--tcti**
+  2. The environment variable: _TPM2TOOLS\_TCTI\_NAME_.
 
-  * _TPM2TOOLS\_TCTI\_NAME_:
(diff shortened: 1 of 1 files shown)


man/common/tcti.md
```diff
@@ -1,90 +1,106 @@
...
-	but it could be a TPM simulator or TPM device itself.
+**Note:** The command line option always overrides the environment variable.
 
-  The current known TCTIs are:
+The current known TCTIs are:
 
-	* tabrmd - The new resource manager, called
-	           [tabrmd](https://github.com/tpm2-software/tpm2-abrmd).
+  * tabrmd - The new resource manager, called
+             [tabrmd](https://github.com/tpm2-software/tpm2-abrmd).
```
joshuagl: Does it still qualify as new?

man/common/tcti.md
```diff
@@ -1,90 +1,106 @@
...
+      --tcti:bus_type=session
+      ```
 
-## TCTI Defaults
-When a TCTI is not specified, the default TCTI is searched for using dlopen(3) semantics.
-The tools will search for *tabrmd*, *device* and *mssim* TCTIs **IN THAT ORDER** and
-**USE THE FIRST ONE FOUND**. You can query what TCTI will be chosen as the default by
-using the `-v` option to print the version information. The "default-tcti" key-value pair
-will indicate which of the aforementioned TCTIs is the default.
+      Specify the device tcti and use the default config:
```
joshuagl: Having a device related example under the **abrmd** section confused me, should this be moved up to the **device** section? Or perhaps the goal is to have a section of worked examples at the end of the TCTI section? If so we should add a separator/header for the section.

man/common/tcti.md
williamcroberts: I guess not.

man/common/tcti.md
williamcroberts: No that should be local to device.

PR #972, opened 2018-04-03
Tpm2 hmac authentication

idesai:
Enabling HMAC authentication for the tools using passwords.

Since there are existing issues with tpm2-abrmd regarding session continuation, this feature can only be tested directly with the TPM. This necessitates issue#971.

In order to test.
Setup NV Index  -- tpm2_nvdefine -x 0x1500016 -a 0x40000001 -s 32 -t 0x2020006 -I "hmactest" --tcti=mssim:tcp://127.0.0.1
Setup owner password -- tpm2_changeauth -o test --tcti=mssim:tcp://127.0.0.1

Test#1 - Write with NV Index password
tpm2_nvwrite  -x 0x1500016 -a 0x1500016  -P hmactest test.nv --tcti=mssim:tcp://127.0.0.1

Test#2 - Write with TPM owner password
tpm2_nvwrite  -x 0x1500016 -a 0x40000001  -P test test.nv --tcti=mssim:tcp://127.0.0.1

For PCR Policy Test
tpm2_pcrlist -Q -L sha1:"0,1,2,3" -o pcrs.file
tpm2_createpolicy -Q -P -L sha1:"0,1,2,3" -F pcrs.file -f pcrs.policy
tpm2_nvdefine -Q -x 0x1500016 -a 0x40000001 -s 32 -L pcrs.policy -t "policyread|policywrite"
tpm2_nvwrite -Q -x 0x1500016 -a 0x1500016 -L sha1:"0,1,2,3" -F pcrs.file test.nv

### DIFF
--- a/lib/tpm2_auth_util.c
+++ b/lib/tpm2_auth_util.c
@@ -34,43 +34,104 @@
 #include <tss2/tss2_sys.h>
 
 #include "log.h"
+#include "pcr.h"
 #include "tpm2_auth_util.h"
+#include "tpm2_nv_util.h"
+#include "tpm2_policy.h"
 #include "tpm2_session.h"
 #include "tpm2_util.h"
 
+typedef enum tpm2_session_type tpm2_session_type;
+enum tpm2_session_type {
+    tpm2_session_fail = 0,
+    tpm2_session_password,
+    tpm2_session_hmac,
+    tpm2_session_ea
+};
+
 #define HEX_PREFIX "hex:"
-#define HEX_PREFIX_LEN sizeof(HEX_PREFIX) - 1
+#define HEX_PREFIX_LEN (sizeof(HEX_PREFIX) - 1)
 
 #define STR_PREFIX "str:"
-#define STR_PREFIX_LEN sizeof(STR_PREFIX) - 1
+#define STR_PREFIX_LEN (sizeof(STR_PREFIX) - 1)
+
+#define HMAC_PREFIX "hmac:"
+#define HMAC_PREFIX_LEN (sizeof(HMAC_PREFIX) - 1)
 
 #define SESSION_PREFIX "session:"
-#define SESSION_PREFIX_LEN sizeof(SESSION_PREFIX) - 1
+#define SESSION_PREFIX_LEN (sizeof(SESSION_PREFIX) - 1)
 
-static bool handle_hex(const char *password, TPMS_AUTH_COMMAND *auth) {
+#define PCR_PREFIX "pcr:"
+#define PCR_PREFIX_LEN (sizeof(PCR_PREFIX) - 1)
 
-    /* if it is hex, then skip the prefix */
-    password += HEX_PREFIX_LEN;
+static tpm2_session_type handle_hex(const char *password, TPMS_AUTH_COMMAND *auth) {
 
     auth->hmac.size = BUFFER_SIZE(typeof(auth->hmac), buffer);
     int rc = tpm2_util_hex_to_byte_structure(password, &auth->hmac.size, auth->hmac.buffer);
     if (rc) {
         auth->hmac.size = 0;
+        return tpm2_session_fail;
+    }
+
+    return tpm2_session_password;
+}
+
+static tpm2_session_type copy_password(TPM2B_AUTH *hmac, const char *password) {
+    /*
+     * Per the man page:
+     * "a return value of size or more means that the output was truncated."
+     */
+    size_t wrote = snprintf((char *)hmac->buffer, sizeof(hmac->buffer), "%s", password);
+    if (wrote >= sizeof(hmac->buffer)) {
+        hmac->size = 0;
+        memset(hmac->buffer, 0, sizeof(hmac->buffer));
         return false;
     }
 
+    hmac->size = wrote;
     return true;
 }
 
-static bool handle_session(const char *path, TPMS_AUTH_COMMAND *auth,
-        tpm2_session **session) {
+static tpm2_session_type handle_hmac(TSS2_SYS_CONTEXT *sapi, const char *password, TPMS_AUTH_COMMAND *auth,
+        tpm2_session **session, tpm2_auth_cb *cb) {
 
-    /* if it is session, then skip the prefix */
-    path += SESSION_PREFIX_LEN;
+    bool result = copy_password(&auth->hmac, password);
+    if (!result) {
+        return tpm2_session_fail;
+    }
+
+    tpm2_session_data *d = tpm2_session_data_new(TPM2_SE_HMAC);
+    if (!d) {
+        LOG_ERR("Could not allocate new session data");
+        return tpm2_session_fail;
+    }
+
+    if (cb && cb->hmac.init) {
+        result = cb->hmac.init(d);
+        if (!result) {
+            LOG_ERR("hmac callback failed");
+            tpm2_session_data_free(&d);
+            return tpm2_session_fail;
+        }
+    }
+
+    *session = tpm2_session_new(sapi, d);
+    if (!*session) {
+        LOG_ERR("Could not start new session");
+        return tpm2_session_fail;
+    }
+
+    auth->sessionHandle = tpm2_session_get_handle(*session);
+
+    return tpm2_session_hmac;
+}
+
+static tpm2_session_type handle_session(const char *path, TPMS_AUTH_COMMAND *auth,
+        tpm2_session **session) {
 
     *session = tpm2_session_restore(path);
     if (!*session) {
-        return false;
+        return tpm2_session_fail;
     }
 
     auth->sessionHandle = tpm2_session_get_handle(*session);
@@ -80,13 +141,65 @@ static bool handle_session(const char *path, TPMS_AUTH_COMMAND *auth,
         LOG_ERR("A trial session cannot be used to authenticate, "
                 "Please use an hmac or policy session");
         tpm2_session_free(session);
-        return false;
+        return tpm2_session_fail;
     }
 
-    return true;
+    return tpm2_session_ea;
 }
 
-static bool handle_str(const char *password, TPMS_AUTH_COMMAND *auth) {
+static tpm2_session_type handle_pcr(TSS2_SYS_CONTEXT *sapi, const char *spec, TPMS_AUTH_COMMAND *auth,
+        tpm2_session **session) {
+
+    /*
+     * Handle the parsing of a PCR spec which is pcr:f=<path> or pcr:p=<pcr spec>
+     */
+    TPML_PCR_SELECTION pcrs = { .count = 0 };
+    const char *pcr_file = NULL;
+    if (!strncmp(spec, "f=", 2)) {
+        pcr_file = &spec[2];
+    } else if(!strncmp(spec, "p=", 2)) {
+        const char *pcr_spec = &spec[2];
+        bool result = pcr_parse_selections(pcr_spec, &pcrs);
+        if (!result) {
+            LOG_ERR("Could not parse PCR selections, got: \"%s\"", pcr_spec);
+            return tpm2_session_fail;
+        }
+    } else {
+        LOG_ERR("Expected either a file (f=) or pcr (p=) based PCR spec,"
+                "got: \"%s\"", spec);
+        return tpm2_session_fail;
+    }
+
+    tpm2_session_data *d =
+            tpm2_session_data_new(TPM2_SE_POLICY);
+    if (!d) {
+        LOG_ERR("oom");
+        return tpm2_session_fail;
+    }
+
+    tpm2_session *s = tpm2_session_new(sapi, d);
+    if (!s) {
+        LOG_ERR("Could not start tpm session");
+        return tpm2_session_fail;
+    }
+
+    bool result = tpm2_policy_build_pcr(sapi, s,
+            pcr_file,
+            &pcrs);
+    if (!result) {
+        LOG_ERR("Could not build a pcr policy");
+        tpm2_session_free(&s);
+        return tpm2_session_fail;
+    }
+
+    auth->sessionHandle = tpm2_session_get_handle(s);
+    auth->sessionAttributes |= TPMA_SESSION_CONTINUESESSION;
+
+    *session = s;
+    return tpm2_session_ea;
+}
+
+static tpm2_session_type handle_str(const char *password, TPMS_AUTH_COMMAND *auth) {
 
     /* str may or may not have the str: prefix */
     bool is_str_prefix = !strncmp(password, STR_PREFIX, STR_PREFIX_LEN);
@@ -94,26 +207,17 @@ static bool handle_str(const char *password, TPMS_AUTH_COMMAND *auth) {
         password += STR_PREFIX_LEN;
     }
 
-    /*
-     * Per the man page:
-     * "a return value of size or more means that the output was truncated."
-     */
-    size_t wrote = snprintf((char *)&auth->hmac.buffer, BUFFER_SIZE(typeof(auth->hmac), buffer), "%s", password);
-    if (wrote >= BUFFER_SIZE(typeof(auth->hmac), buffer)) {
-        auth->hmac.size = 0;
-        return false;
-    }
-
-    auth->hmac.size = wrote;
-
-    return true;
+    bool result = copy_password(&auth->hmac, password);
+    return result ? tpm2_session_password : tpm2_session_fail;
 }
 
-bool tpm2_auth_util_from_optarg(const char *password, TPMS_AUTH_COMMAND *auth,
-        tpm2_session **session) {
+tpm2_session_type tpm2_auth_util_from_optarg2(TSS2_SYS_CONTEXT *sapi, const char *password, TPMS_AUTH_COMMAND *auth,
+        tpm2_session **session, tpm2_auth_cb *cb) {
 
     bool is_hex = !strncmp(password, HEX_PREFIX, HEX_PREFIX_LEN);
     if (is_hex) {
+
+        password += HEX_PREFIX_LEN;
         return handle_hex(password, auth);
     }
 
@@ -123,9 +227,310 @@ bool tpm2_auth_util_from_optarg(const char *password, TPMS_AUTH_COMMAND *auth,
             LOG_ERR("Tool does not support sessions for this auth value");
             return false;
         }
+
+        password += SESSION_PREFIX_LEN;
         return handle_session(password, auth, session);
     }
 
+    bool is_hmac = !strncmp(password, HMAC_PREFIX, HMAC_PREFIX_LEN);
+    if (is_hmac) {
+        if (!session) {
+            LOG_ERR("Tool does not support HMAC for this auth value");
+            return false;
+        }
+
+        password += HMAC_PREFIX_LEN;
+        return handle_hmac(sapi, password, auth, session, cb);
+    }
+
+    bool is_pcr = !strncmp(password, PCR_PREFIX, PCR_PREFIX_LEN);
+    if (is_pcr) {
+        if (!session) {
+            LOG_ERR("Tool does not support PCR for this auth value");
+            return false;
+        }
+
+        password += PCR_PREFIX_LEN;
+        return handle_pcr(sapi, password, auth, session);
+    }
     /* must be string, handle it */
     return handle_str(password, auth);
+}
+
+bool tpm2_auth_util_from_optarg(const char *password, TPMS_AUTH_COMMAND *auth,
+        tpm2_session **session) {
+    return tpm2_auth_util_from_optarg2(NULL, password, auth, session, NULL) != tpm2_session_fail;
+}
+
+bool tpm2_auth_util_from_options(TSS2_SYS_CONTEXT *sapi, tpm2_auth *auth, tpm2_auth_cb *cb,
+        bool support_sessions) {
+
+    if (!auth->cnt) {
+        /*
+         * default to the empty password.
+         * The password and handle were set in the init macro.
+         */
+        auth->auth_list.count++;
+        return true;
+    }
+
+    auth->cb = *cb;
+
+    unsigned i;
+    for (i=0; i < auth->cnt; i++) {
+
+        const char *o = auth->optargs[i];
+        TPMS_AUTH_COMMAND *a = &auth->auth_list.auths[i];
+        tpm2_session **s = support_sessions ? &auth->sessions[i] : NULL;
+        tpm2_session_type stype = tpm2_auth_util_from_optarg2(sapi, o, a, s, cb);
+        if (stype == tpm2_session_fail) {
+            tpm2_auth_util_free(sapi, auth);
+            return false;
+        } else if(stype == tpm2_session_hmac) {
+            auth->hmac_indexes |= 1 << i;
+        }
+
+        auth->auth_list.count++;
+    }
+
+    return true;
+}
+
+bool tpm2_read_transient_persistent_obj_name(TSS2_SYS_CONTEXT *sapi_context,
+    TPM2_HANDLE entity_handle, TPM2B_NAME *entity_name) {
+
+    TPM2B_PUBLIC object_public;
+    TSS2L_SYS_AUTH_RESPONSE sessions_data_out;
+    TPM2B_NAME qualified_name = TPM2B_TYPE_INIT(TPM2B_NAME, name);
+    TSS2_RC rval = TSS2_RETRY_EXP( Tss2_Sys_ReadPublic(sapi_context, entity_handle,
+                                        NULL, &object_public, entity_name, &qualified_name,
+                                        &sessions_data_out));
+    if (rval != TSS2_RC_SUCCESS) {
+        LOG_PERR(Tss2_Sys_ReadPublic, rval);
+        return false;
+    }
+    return true;
+}
+
+bool tpm2_hmac_auth_get_entity_name(TSS2_SYS_CONTEXT *sapi_context, TPM2_HANDLE entity_handle,
+    TPM2B_NAME *entity_name) {
+
+    TPM2B_NV_PUBLIC nv_public = TPM2B_EMPTY_INIT;
+    bool res;
+    uint8_t temp[4];
+    unsigned i;
+    //Get Entity names from handles
+    switch(entity_handle & TPM2_HR_RANGE_MASK) {
+        case TPM2_HR_NV_INDEX:
+            res = tpm2_util_nv_read_public2(sapi_context,entity_handle, &nv_public,
+                entity_name);
+            if (!res) {
+                LOG_ERR("Failed reading NV public when calculating name");
+                return false;
+            }
+            break;
+        case TPM2_HR_TRANSIENT:
+        case TPM2_HR_PERSISTENT:
+            res = tpm2_read_transient_persistent_obj_name(sapi_context,
+                    entity_handle, entity_name);
+            if (!res) {
+                LOG_ERR("Failed reading Object public when calculating name");
+                return false;
+            }
+            break;
+        case TPM2_HR_PCR:
+        case TPM2_HR_HMAC_SESSION:
+        case TPM2_HR_POLICY_SESSION:
+        case TPM2_HR_PERMANENT:
+            entity_name->size = sizeof(TPM2_HANDLE);
+            memcpy(temp, &entity_handle, sizeof(TPM2_HANDLE));
+            for (i = 0; i < sizeof(TPM2_HANDLE); i++) {
+               entity_name->name[i] = temp[3-i];
+            }
+            break;
+        default:
+            LOG_ERR("Handle not valid for HMAC auth");
+            return false;
+    }
+    return true;
+}
+
+// TODO Figure out how I work and refactor me
+#include <openssl/bio.h>
+#include <openssl/buffer.h>
+#include <openssl/evp.h>
+#include <openssl/hmac.h>
+#include <openssl/sha.h>
+
+static bool command_authentication_hmac(tpm2_session *hmac_session, uint8_t *cp_hash,
+    TPMS_AUTH_COMMAND *session_data, TPMS_AUTH_RESPONSE *response_data, char *optargs) {
+
+    if (response_data->nonce.size) {
+        tpm2_session_update_nonce_older(hmac_session, &response_data->nonce);
+        session_data->hmac.size = strlen(optargs);
+        memcpy(session_data->hmac.buffer,optargs,strlen(optargs));
+    }
+
+    const TPM2B_NONCE *nonce_tpm =
+        tpm2_session_get_nonce_tpm(hmac_session);
+
+    uint8_t *to_hmac_buffer = malloc(SHA256_DIGEST_LENGTH + nonce_tpm->size
+                          + sizeof(TPMA_SESSION));
+    if (!to_hmac_buffer) {
+        LOG_ERR("oom");
+        return false;
+    }
+
+    memcpy(to_hmac_buffer, cp_hash, SHA256_DIGEST_LENGTH);
+
+    memcpy(to_hmac_buffer + SHA256_DIGEST_LENGTH, nonce_tpm->buffer, nonce_tpm->size);
+
+    memcpy(to_hmac_buffer + SHA256_DIGEST_LENGTH + nonce_tpm->size,
+           &session_data->sessionAttributes, sizeof(TPMA_SESSION));
+
+    unsigned int temp_hash_len = 0;
+    HMAC(EVP_sha256(), session_data->hmac.buffer, session_data->hmac.size,
+        to_hmac_buffer, SHA256_DIGEST_LENGTH + nonce_tpm->size + sizeof(TPMA_SESSION),
+        session_data->hmac.buffer, &temp_hash_len);
+    session_data->hmac.size = temp_hash_len;
+    free(to_hmac_buffer);
+
+    return true;
+}
+
+uint8_t *get_cp_hash(TSS2_SYS_CONTEXT *sapi_context,
+    TPM2B_NAME entity_1_name, TPM2B_NAME entity_2_name) {
+
+    SHA256_CTX sha256;
+    int is_success = SHA256_Init(&sha256);
+    if (!is_success) {
+        LOG_ERR("SHA256_Init failed when calculating HMAC");
+        return NULL;
+    }
+
+    uint8_t cmd_code[4];
+    TSS2_RC rval = Tss2_Sys_GetCommandCode(sapi_context, cmd_code);
+    if (rval != TSS2_RC_SUCCESS) {
+        LOG_PERR(Tss2_Sys_GetCommandCode, rval);
+        return NULL;
+    }
+
+    is_success = SHA256_Update(&sha256, cmd_code, sizeof(cmd_code));
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+
+    is_success = SHA256_Update(&sha256, entity_1_name.name, entity_1_name.size);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+    is_success = SHA256_Update(&sha256, entity_2_name.name, entity_2_name.size);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+
+    size_t command_params_size;
+    const uint8_t *command_params;
+    rval = Tss2_Sys_GetCpBuffer(sapi_context, &command_params_size, &command_params);
+    if (rval != TPM2_RC_SUCCESS) {
+        LOG_PERR(Tss2_Sys_GetCpBuffer, rval);
+        return NULL;
+    }
+
+    is_success = SHA256_Update(&sha256, command_params, command_params_size);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+
+    uint8_t *cp_hash = malloc(SHA256_DIGEST_LENGTH);
+    if (!cp_hash) {
+        LOG_ERR ("OOM");
+        return NULL;
+    }
+
+    is_success = SHA256_Final(cp_hash, &sha256);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Final failed when calculating HMAC");
+        free(cp_hash);
+        return NULL;
+    }
+
+    return cp_hash;
+}
+
+static bool tpm2_hmac_auth_get_command_buffer_hmac(TSS2_SYS_CONTEXT *sapi_context, TPMS_AUTH_COMMAND *auth,
+    tpm2_session *hmac_session, TPM2B_NAME entity_1_name,
+    TPM2B_NAME entity_2_name, TPMS_AUTH_RESPONSE *response_data, char *optargs) {
+
+    uint8_t *cp_hash = get_cp_hash(sapi_context, entity_1_name, entity_2_name);
+    if (!cp_hash) {
+        LOG_ERR("Command Parameter hash failed");
+        return false;
+    }
+
+    command_authentication_hmac(hmac_session, cp_hash, auth, response_data, optargs);
+    free(cp_hash);
+    return true;
+}
+
+static bool tpm2b_auth_update_for_hmac(TSS2_SYS_CONTEXT *sapi,
+        tpm2_auth_cb *cb, tpm2_session *s, TPMS_AUTH_COMMAND *a,
+        void *udata, TPMS_AUTH_RESPONSE *response_data, char *optargs) {
+
+    const TPM2_HANDLE *h = tpm2_session_get_auth_handles(s);
+
+    unsigned i;
+    TPM2B_NAME names[2];
+    for (i=0; i < 2; i++) {
+        bool result = tpm2_hmac_auth_get_entity_name(sapi, h[i], &names[i]);
+        if (!result) {
+            return false;
+        }
+    }
+
+    if (cb && cb->hmac.update) {
+        bool result = cb->hmac.update(sapi, udata);
+        if (!result) {
+            return false;
+        }
+    }
+
+    return tpm2_hmac_auth_get_command_buffer_hmac(sapi, a, s, names[0], names[1],
+            response_data, optargs);
+}
+
+bool tpm2b_auth_update(TSS2_SYS_CONTEXT *sapi, tpm2_auth *auth, void *udata) {
+
+    unsigned i;
+    for (i=0; i < auth->cnt; i++) {
+        if (auth->hmac_indexes & (1 << i)) {
+            bool result = tpm2b_auth_update_for_hmac(sapi, &auth->cb,
+                    auth->sessions[i], &auth->auth_list.auths[i], udata,
+                    &auth->resp_list.auths[i], (char *)auth->optargs[i]+HMAC_PREFIX_LEN);
+            if (!result) {
+                return false;
+            }
+        }
+    }
+
+    return true;
+}
+
+bool tpm2_auth_util_free(TSS2_SYS_CONTEXT *sapi, tpm2_auth *auth) {
+
+    bool result = true;
+    unsigned i;
+    for (i=0; i < auth->cnt; i++) {
+        tpm2_session *s = auth->sessions[i];
+        if (s) {
+            result &= tpm2_session_close(sapi, s);
+            tpm2_session_free(&s);
+        }
+    }
+
+    return result;
 }
--- a/lib/tpm2_nv_util.h
+++ b/lib/tpm2_nv_util.h
@@ -52,19 +52,23 @@
  * @return
  *  True on success, false otherwise.
  */
-static inline bool tpm2_util_nv_read_public(TSS2_SYS_CONTEXT *sapi_context,
-        TPMI_RH_NV_INDEX nv_index, TPM2B_NV_PUBLIC *nv_public) {
-
-    TPM2B_NAME nv_name = TPM2B_TYPE_INIT(TPM2B_NAME, name);
+static inline bool tpm2_util_nv_read_public2(TSS2_SYS_CONTEXT *sapi_context,
+        TPMI_RH_NV_INDEX nv_index, TPM2B_NV_PUBLIC *nv_public, TPM2B_NAME *nv_name) {
 
     TSS2_RC rval = TSS2_RETRY_EXP(Tss2_Sys_NV_ReadPublic(sapi_context, nv_index, NULL, nv_public,
-            &nv_name, NULL));
+            nv_name, NULL));
     if (rval != TSS2_RC_SUCCESS) {
         LOG_PERR(Tss2_Sys_NV_ReadPublic, rval);
         return false;
     }
 
     return true;
+}
+
+static inline bool tpm2_util_nv_read_public(TSS2_SYS_CONTEXT *sapi_context,
+        TPMI_RH_NV_INDEX nv_index, TPM2B_NV_PUBLIC *nv_public) {
+    TPM2B_NAME nv_name = TPM2B_TYPE_INIT(TPM2B_NAME, name);
+    return tpm2_util_nv_read_public2(sapi_context, nv_index, nv_public, &nv_name);
 }
 
 /**
--- a/lib/tpm2_session.c
+++ b/lib/tpm2_session.c
@@ -48,6 +48,7 @@ struct tpm2_session_data {
     TPMT_SYM_DEF symmetric;
     TPMI_ALG_HASH authHash;
     TPM2B_NONCE nonce_caller;
+    TPM2_HANDLE auth_handles[2];
 };
 
 struct tpm2_session {
@@ -56,12 +57,14 @@ struct tpm2_session {
 
     struct {
         TPMI_SH_AUTH_SESSION session_handle;
-        TPM2B_NONCE nonceTPM;
+        TPM2B_NONCE nonceOlder;
     } output;
 
     struct {
         TPM2B_NONCE nonceNewer;
         const char *path;
+        bool started_session;
+        bool is_closed;
     } internal;
 };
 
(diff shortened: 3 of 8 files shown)


tools/tpm2_nvwrite.c
```diff
@@ -147,8 +150,11 @@ static bool nv_write(TSS2_SYS_CONTEXT *sapi_context) {
 
 static bool on_option(char key, char *value) {
     bool result;
-
+    ctx.hmac_auth = true;//DEBUG
```
martinezjavier: This seems to be a left over, right?

tools/tpm2_nvwrite.c
```diff
@@ -233,9 +239,10 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
         { "offset",         required_argument, NULL, 'o' },
         { "set-list",       required_argument, NULL, 'L' },
         { "pcr-input-file", required_argument, NULL, 'F' },
+        { "hmac-auth",      no_argument, NULL, 'X' },
```
martinezjavier: You should also add this new option to the man page.

martinezjavier:
@idesai thanks for working on this. A few general comments about the commits so I'll comment here:

* Please add a commit message explaining why a change is made, that makes reviewing the patches much more easier.
* Use a proper subject line, commit 3, 4 and 5 have the same subject line so that makes quite hard to tell what they are about.
* Instead of using an "HMAC-auth" prefix, use as prefix the part of the code that you are changing, e.g: instead of "HMAC-auth: Enable hmac authentication option and start hmac authorization session - tpm2_nvwrite" use "tpm2_write: Add option to authenticate using HMAC and start HMAC authorization session".

tools/tpm2_nvwrite.c
```diff
@@ -249,6 +256,24 @@ int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
...
     int rc = 1;
 
+    /* set up hmac authentication session */
+    if (ctx.hmac_auth && ctx.session_data.hmac.size) {
+
+        tpm2_session_data *session_data =
+                tpm2_session_data_new(TPM2_SE_HMAC);
+        if (!session_data) {
+            LOG_ERR("oom");
+            return 1;
```
martinezjavier: Maybe return rc?

tools/tpm2_nvwrite.c
```diff
@@ -249,6 +256,24 @@ int tpm2_tool_onrun(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
...
+        }
+        ctx.hmac_session = (tpm2_session *)tpm2_session_new(sapi_context,
```
martinezjavier: I don't think the cast is needed.

tools/tpm2_nvwrite.c
martinezjavier: And also you should add a unit test for HMAC auth.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+    return true;
+}
+
+bool tpm2_get_entity_name(TSS2_SYS_CONTEXT *sapi_context, TPM2_HANDLE entity_handle, 
+    TPM2B_NAME *entity_name) {
+    TPM2B_NV_PUBLIC nv_public = TPM2B_EMPTY_INIT;
+    bool res;
+    uint8_t temp[4];
+    //Get Entity names from handles
+    switch(entity_handle & 0xFF000000) {
```
martinezjavier: Please don't use magic numbers, I believe this should be `TPM2_HR_RANGE_MASK`.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+        case 0x01000000: res = tpm2_util_nv_read_public(sapi_context,
```
martinezjavier: This should be `TPM2_HR_NV_INDEX`

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+        case 0x01000000: res = tpm2_util_nv_read_public(sapi_context, 
+                                                        entity_handle, 
+                                                        &nv_public,
+                                                        entity_name);
+                    if (!res) {
+                        LOG_ERR("Failed reading NV public when calculating name");
+                        return false;
+                    }
+                    break;
+        case 0x80000000:
```
martinezjavier: TPM2_HT_TRANSIENT

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+        case 0x81000000: res = tpm2_read_transient_persistent_obj_name(sapi_context,
```
martinezjavier: TPM2_HR_PERSISTENT

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+                                    entity_handle, entity_name);
+                         if (!res) {
+                            LOG_ERR("Failed reading Object public when calculating name");
+                            return false;
+                        }
+                        break;
+        case 0x00000000:
```
martinezjavier: TPM2_HR_PCR

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+        case 0x02000000:
+        case 0x03000000:
+        case 0x40000000: entity_name->size = sizeof(TPM2_HANDLE);
+                         memcpy(temp, (uint8_t *)&entity_handle, sizeof(TPM2_HANDLE));
```
martinezjavier: I don't think the cast is needed here.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+        case 0x80000000:
+        case 0x81000000: res = tpm2_read_transient_persistent_obj_name(sapi_context,
+                                    entity_handle, entity_name);
+                         if (!res) {
+                            LOG_ERR("Failed reading Object public when calculating name");
+                            return false;
+                        }
+                        break;
+        case 0x00000000:
+        case 0x02000000:
```
martinezjavier: TPM_HT_HMAC_SESSION ?

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+        case 0x03000000:
```
martinezjavier: TPM_HT_POLICY_SESSION ?

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+        case 0x40000000: entity_name->size = sizeof(TPM2_HANDLE);
```
martinezjavier: TPM_HT_PERMANENT

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
...
+                         memcpy(temp, (uint8_t *)&entity_handle, sizeof(TPM2_HANDLE));
+                         for(unsigned i=0; i<sizeof(TPM2_HANDLE); i++) {
```
martinezjavier: Please use a space between the operator and operands.

lib/tpm2_hmac_auth.c
```diff
@@ -31,9 +31,78 @@
 
 #include <tss2/tss2_sys.h>
 
+#include <openssl/bio.h>
+#include <openssl/buffer.h>
+#include <openssl/evp.h>
+#include <openssl/sha.h>
```
martinezjavier: Can you please check that this builds with both OpenSSL 1.1.0 and OpenSSL < 1.1.0 ? It's annoying but there were a lot of API changes so we needed #ifdefery in other part of the tools.

lib/tpm2_hmac_auth.c
```diff
@@ -31,9 +31,78 @@
...
+        return NULL;
+    }
+
+    is_success = SHA256_Update(&sha256, command_params, command_params_size);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+
+    uint8_t *cpHash = (unsigned char*)malloc(SHA256_DIGEST_LENGTH);
```
martinezjavier: There's no need to cast here.

lib/tpm2_hmac_auth.c
```diff
@@ -31,9 +31,78 @@
...
+    if (!cpHash) {
+        LOG_ERR ("OOM");
+        return NULL;
+    }
+    is_success = SHA256_Final(cpHash, &sha256);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Final failed when calculating HMAC");
```
martinezjavier: You should free the cpHash here.

lib/tpm2_hmac_auth.c
```diff
@@ -158,4 +158,49 @@ bool tpm2_get_entity_name(TSS2_SYS_CONTEXT *sapi_context, TPM2_HANDLE entity_han
     return true;
 }
 
+bool command_authentication_hmac(tpm2_session *hmac_session,
+                                 uint8_t *cpHash,
+                                 TPMS_AUTH_COMMAND *session_data) {
+
+    uint8_t *to_hmac_buffer = (unsigned char*)malloc(SHA256_DIGEST_LENGTH
```
martinezjavier: Again the cast isn't needed here.

tools/tpm2_nvwrite.c
```diff
@@ -61,13 +61,16 @@ struct tpm_nvwrite_ctx {
...
+    tpm2_session *hmac_session;
+    bool hmac_auth;
     TPML_PCR_SELECTION pcr_selection;
     struct {
         UINT8 L : 1;
     } flags;
 };
 
 static tpm_nvwrite_ctx ctx = {
+    .hmac_auth = false,
```
martinezjavier: I wonder if we want the default to be true and instead have an option to use plaintext passwords? My understanding is that not using HMAC sessions is insecure so probably makes sense to do it that way?

martinezjavier:
@idesai are you planning to do the same for the other tools too? and also for the 3.x branch? If not, can you please file issues so we can keep track of those?

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,206 @@
...
+
+    SHA256_CTX sha256;
+    int is_success = SHA256_Init(&sha256);
+    if(!is_success) {
+        LOG_ERR("SHA256_Init failed when calculating HMAC");
+        return NULL;
+    }
+
+    uint8_t cmd_code[4];
+    TSS2_RC rval = Tss2_Sys_GetCommandCode(sapi_context, cmd_code );
```
williamcroberts: space at end of cmd_code

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,206 @@
...
+#include <openssl/evp.h>
+#include <openssl/sha.h>
+
+#include "log.h"
+#include "tpm2_error.h"
+#include "tpm2_nv_util.h"
+#include "tpm2_openssl.h"
+#include "tpm2_session.h"
+
+uint8_t *tpm2_get_cpHash(TSS2_SYS_CONTEXT *sapi_context, TPM2B_NAME entity_1_name,
```
williamcroberts: get_cp_hash, we don't camel case.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,206 @@
...
+    is_success = SHA256_Update(&sha256, entity_2_name.name, entity_2_name.size);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+
+    size_t command_params_size;
+    const uint8_t *command_params;
+    rval = Tss2_Sys_GetCpBuffer( sapi_context, &command_params_size, &command_params);
+    if( rval != TPM2_RC_SUCCESS ) {
```
williamcroberts: weird spaces in lines 81 and 82

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,206 @@
...
+        LOG_ERR("Failed to retrieve cp Buffer:%08X", rval);
```
williamcroberts: Use LOG_PERR on tpm calls.

lib/tpm2_hmac_auth.c
williamcroberts: Well if you use a cpp compiler it complains. Even so the cast should have been uint8_t * not unsigned char *.

Just drop the cast.

lib/tpm2_hmac_auth.c
williamcroberts: yep mem leak.

lib/tpm2_session.c
```diff
@@ -85,6 +85,14 @@ void tpm2_session_set_nonce_caller(tpm2_session_data *data, TPM2B_NONCE *nonce)
     data->nonce_caller = *nonce;
 }
 
+UINT16 tpm2_session_get_nonceTPM_size(tpm2_session *session) {
```
williamcroberts: no camelCase here and throughout

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,206 @@
...
+
+    uint8_t *to_hmac_buffer = (unsigned char*)malloc(SHA256_DIGEST_LENGTH
+                            + tpm2_session_get_nonceTPM_size(hmac_session)
+                            + sizeof(TPMA_SESSION));
+    if (!to_hmac_buffer) {
+        LOG_ERR("oom");
+        return false;
+    }
+
+    memcpy(to_hmac_buffer, cpHash, SHA256_DIGEST_LENGTH);
```
williamcroberts: Looks like SHA256 is the hard coded choice. Could we make the API specify the algorithm and then just set a default in the caller.

something like this in the tool:
```
hmac_doit(a, b, c);
```
static hmac_doit(int a, int b int c, algorithm alg) {
   do stuff;
}

int hmac_doit(int a, int b int c) {
   return hmac_doit_alg(a, b, c, SHA256);
}
```

This way later if we need to, we can support HMAC in algorithm flexible way by just exposing the algorithm variant.

File: tools/tpm2_nvwrite.c [hunk 26]
```diff
@@ -145,8 +180,11 @@ static bool nv_write(TSS2_SYS_CONTEXT *sapi_context) {
 
 static bool on_option(char key, char *value) {
     bool result;
-
+    ctx.hmac_auth = true;//DEBUG
     switch (key) {
+    case 'X':
```
williamcroberts: We should use the auth language for this.

```
# compute the hmac ticket
--auth-xxx=hmac:
# use a precomputed ticket
--auth-xxx=hmac:path/to/file
```

tools/tpm2_nvwrite.c
```diff
@@ -61,13 +62,16 @@ struct tpm_nvwrite_ctx {
     UINT16 offset;
     char *raw_pcrs_file;
     tpm2_session *policy_session;
+    tpm2_session *hmac_session;
+    bool hmac_auth;
```
williamcroberts: I don't think we need these. You should extend [tpm2_auth_util_from_optarg](https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_auth_util.c#L112) to support the hmac keyword. Store the type of session in the session *, and smarten up the rest of the session logic to avoid calling session save on hmac sessions.

tools/tpm2_nvwrite.c
idesai: @martinezjavier @williamcroberts , then we may need to account for if hmac sessions are not available. any thoughts?

idesai:
@martinezjavier @williamcroberts,
_"are you planning to do the same for the other tools too? and also for the 3.x branch? If not, can you please file issues so we can keep track of those?"_
Yes, However i want to have this code review complete prior to pushing it on other tools. Thanks for all the reviews.

tools/tpm2_nvwrite.c
martinezjavier: @idesai oh, isn't HMAC supported on all TPM2 devices or am I not following?

martinezjavier:
> Yes, However i want to have this code review complete prior to pushing it on other tools. Thanks for all the reviews.

@idesai awesome, thanks a lot for working on this!

tools/tpm2_nvwrite.c
williamcroberts: When are hmac sessions not available?

I think the default auth should always be a NULL password. Then users use the auth mini-language to specify how they want to auth.

ie:
```
--auth-xxx password
Create an hmac ticket
--auth-xxx hmac:

Comment (idesai) [reply tools/tpm2_nvwrite.c]: @martinezjavier @williamcroberts by sessions not available, I meant max allowed sessions already resident on the TPM. Starting a new session for hmac would result in TSS2_RC "out of memory"

File: lib/tpm2_hmac_auth.c [hunk 28]
```diff
@@ -125,35 +125,37 @@ bool tpm2_get_entity_name(TSS2_SYS_CONTEXT *sapi_context, TPM2_HANDLE entity_han
...
-        case 0x03000000:
-        case 0x40000000: entity_name->size = sizeof(TPM2_HANDLE);
-                         memcpy(temp, (uint8_t *)&entity_handle, sizeof(TPM2_HANDLE));
-                         for(unsigned i=0; i<sizeof(TPM2_HANDLE); i++) {
-                            entity_name->name[i] = temp[3-i];
-                         }
-                         break;
-        default:         LOG_ERR("Handle not valid for HMAC auth");
-                         return false;
+    switch((entity_handle & TPM2_HR_RANGE_MASK)>>TPM2_HR_SHIFT) {
```
martinezjavier: No need for that bit shifting I believe, you could just do:

```
switch (entity_handle & TPM2_HR_RANGE_MASK) {
        case TPM2_PT_TPM2_HR_NV_INDEX:
        ....
}
```

tools/tpm2_nvwrite.c
williamcroberts: Oh yeah, well that's always a potential issue when using the TPM directly. It's a user problem if their using the TPM without an RM and get this error code.

martinezjavier: @idesai ah, got it. Hmm, I don't think the tools should handle that. I mean, you could also attempt to create/load more objects than TPM_PT_HR_TRANSIENT_MIN and the TPM (and the tools) would just return a TPM_RC_OBJECT_MEMORY in that case.

So I think that we should just make HMAC session auth the default and if the user is getting this, then it should either access the TPM through a resource manager or use plain text passwords. Just like you need to either manage your objects or use a resource manager.

What do you think?

martinezjavier: @williamcroberts I see that you agree (you answered while I was writing mine :P)

tools/tpm2_nvwrite.c
idesai: @martinezjavier was hitting some weird issues with command option not being parsed/ recognized. Yes it is a left over.

tools/tpm2_nvwrite.c
idesai: @martinezjavier @williamcroberts, cool lets treat it as such. 
I like the tpm2_auth_util_from_optarg way of doing this which should address it.
-P hmac:<Password>

idesai: The tpm2_auth_util_from_optarg was added recently, So I have to rebase this PR.

tools/tpm2_nvwrite.c
idesai: @williamcroberts there is one problem with this approach. If hmac-auth was the only authentication session the suggested approach would work. The way tpm2_auth_util_from_optarg is it takes only one session as an input. For example if i have a policy session requiring hmac as the only form of authentication then I would need to do something like this:
tpm2_nvwrite -P session:HEX -P hmac:PASSWORD
Any thoughts?

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,209 @@
+//**********************************************************************;
+// Copyright (c) 2015-2018, Intel Corporation
```
williamcroberts: new file, so just 2018 here.

lib/tpm2_hmac_auth.h
```diff
@@ -0,0 +1,40 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+#ifndef TPM2_HMAC_AUTH_H_
+#define TPM2_HMAC_AUTH_H_
+
+#include "tpm2_session.h"
+
+bool tpm2_get_command_buffer_hmac(TSS2_SYS_CONTEXT *sapi_context,
```
williamcroberts: here and throughout doxygen style comments on what these methods do.

lib/tpm2_hmac_auth.h
```diff
@@ -0,0 +1,40 @@
+//**********************************************************************;
+// Copyright (c) 2015-2018, Intel Corporation
```
williamcroberts: new file, this will apply throughout. No 2015 here.

lib/tpm2_session.h
```diff
@@ -76,6 +76,9 @@ void tpm2_session_set_key(tpm2_session_data *data, TPMI_DH_OBJECT key);
  */
 void tpm2_session_set_nonce_caller(tpm2_session_data *data, TPM2B_NONCE *nonce);
 
+UINT16 tpm2_session_get_nonce_tpm_size(tpm2_session *session);
+UINT8 *tpm2_session_get_nonce_tpm(tpm2_session *session);
```
williamcroberts: document these, above it is an example.

lib/tpm2_session.c
```diff
@@ -86,6 +86,14 @@ void tpm2_session_set_nonce_caller(tpm2_session_data *data, TPM2B_NONCE *nonce)
     data->nonce_caller = *nonce;
 }
 
+UINT16 tpm2_session_get_nonce_tpm_size(tpm2_session *session) {
+    return session->internal.nonceNewer.size;
```
williamcroberts: You only need one of these functions. And that's one that returns a pointer to the nonceNewer struct.

tools/tpm2_nvwrite.c
williamcroberts: So you're saying you need both the sessionHandle and the hmac data in the TPMS_AUTH_COMMAND data?
Right now that is not supported.

You could do the `-P` approach as you outlined. But the tool would need to be smart enough to complain if you try to set the same field twice, ie:
* 2+ passwords
* 2+ hmac's,
* 2+ sessions
* 1 password and 1 hmac 

You could also use a comma to separate them:
`-P session:<file>,hmac:<password>`

I think I like the multiple -P option approach that you outline. I would make the tpm2_auth_ routine the one smart enough to check if it's set before setting it, so we don't have to duplicate that logic everywhere.

lib/tpm2_hmac_auth.h
williamcroberts: Also naming convention. tpm2_hmac_auth.c is the name of the file, thus all exposed routines should be prefixed with tpm2_hmac_auth_

So this would be `tpm2_hmac_auth_get_cmd_buf()`

lib/tpm2_nv_util.h
```diff
@@ -1,5 +1,5 @@
 //**********************************************************************;
-// Copyright (c) 2017, Intel Corporation
+// Copyright (c) 2015-2018, Intel Corporation
```
williamcroberts: 2017-2018

man/tpm2_nvwrite.1.md
```diff
@@ -40,6 +40,12 @@ If _FILE_ is not specified, it defaults to stdin.
     should follow the authorization formatting standards, see section
     "Authorization Formatting".
 
+  * **-X**, **--hmac-auth**
+    Enforces HMAC authentication method instead of clear text password. It 
+    uses the password value provided with auth-hierarchy option to calculate the 
+    HMAC value. It ensures clear text passwords are not visible on the TPM 
+    interfaces during authentication.
+
```
williamcroberts: this should be going away

lib/tpm2_nv_util.h
dantpm: I was told that changes to copyright need to be in a separate commit.  Is this true?

idesai:
Rebase...hmm...
@martinezjavier @williamcroberts, Hopefully the commits are cleaner now

lib/tpm2_hmac_auth.c
martinezjavier: @idesai you forgot to answer this question.

lib/tpm2_hmac_auth.c
martinezjavier: @idesai and also didn't do the suggested change.

tools/tpm2_nvwrite.c
```diff
@@ -74,15 +77,15 @@ static tpm_nvwrite_ctx ctx = {
     .auth = {
             .session_data = TPMS_AUTH_COMMAND_INIT(TPM2_RS_PW),
             .hierarchy = TPM2_RH_OWNER
-    }
+    },
+    .hmac_auth = false
```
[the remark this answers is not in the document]
williamcroberts: We shouldn't even need this variable. Currently with these changes were going to leave the password logic the same. Ie default to a password. We're going to extend the password langauge to support `hmac:<password>`.

Currently `-P foo` in all actuality defaults to `-P str:foo`.
Once we get this working, we can default `-P foo` to `-P hmac:foo` relatively easily.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+            }
+            break;
+        default:
+            LOG_ERR("Handle not valid for HMAC auth");
+            return false;
+    }
+    return true;
+}
+
+bool command_authentication_hmac(tpm2_session *hmac_session,
```
williamcroberts: static?

lib/tpm2_session.c
```diff
@@ -86,6 +86,10 @@ void tpm2_session_set_nonce_caller(tpm2_session_data *data, TPM2B_NONCE *nonce)
     data->nonce_caller = *nonce;
 }
 
+void tpm2_session_get_nonce_tpm(tpm2_session *session, TPM2B_NONCE *nonce_tpm) {
+    *nonce_tpm = session->internal.nonceNewer;
```
williamcroberts: this is pass by copy, is this what you want?
```c
return session->internal.nonceNewer;
```
I would do:
```c
const TPM2B_NONCE * tpm2_get_nonce_tpm(tpm2_session *session) {
    return (const TPM2B_NONCE *) &session->internal.nonceNewer;
}
```

Looks like you just bring it back for  a `memcpy()` when generating the HMAC, so this avoids a copy.

williamcroberts: I don't think you even need that `const` cast, my local testing shows no compiler complaints.

lib/tpm2_hmac_auth.h
```diff
@@ -0,0 +1,40 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+#ifndef TPM2_HMAC_AUTH_H_
+#define TPM2_HMAC_AUTH_H_
+
+#include "tpm2_session.h"
+
+bool tpm2_hmac_auth_get_command_buffer_hmac(TSS2_SYS_CONTEXT *sapi_context,
```
williamcroberts: doxygen comment doc here

lib/tpm2_hmac_auth.h
```diff
@@ -0,0 +1,40 @@
...
+	TPM2_HANDLE auth_handle, tpm2_session *hmac_session,
+	TPMS_AUTH_COMMAND *session_data, TPM2B_NAME entity_1_name,
+	TPM2B_NAME entity_2_name);
+
+bool tpm2_hmac_auth_get_entity_name(TSS2_SYS_CONTEXT *sapi_context, TPM2_HANDLE entity_handle,
```
williamcroberts: doxygen comment doc here

lib/tpm2_nv_util.h
```diff
@@ -48,12 +48,10 @@
  *  True on success, false otherwise.
  */
 static inline bool tpm2_util_nv_read_public(TSS2_SYS_CONTEXT *sapi_context,
-        TPMI_RH_NV_INDEX nv_index, TPM2B_NV_PUBLIC *nv_public) {
-
-    TPM2B_NAME nv_name = TPM2B_TYPE_INIT(TPM2B_NAME, name);
+        TPMI_RH_NV_INDEX nv_index, TPM2B_NV_PUBLIC *nv_public, TPM2B_NAME *nv_name) {
```
williamcroberts: Document this new parameter. Can you allow NULL here? This way the nvread/write callers can just specify NULL and not have to add a variable that is never really used?
```c
TPM2B_NAME _nv_name = TPM2B_TYPE_INIT(TPM2B_NAME, name);
if (!nv_name) {
   nv_name = &_nv_name;
}

TSS2_RC rval = TSS2_RETRY_EXP(Tss2_Sys_NV_ReadPublic(sapi_context, nv_index, NULL, nv_public,
            nv_name, NULL));
...
```

lib/tpm2_session.h
```diff
@@ -76,6 +76,15 @@ void tpm2_session_set_key(tpm2_session_data *data, TPMI_DH_OBJECT key);
  */
 void tpm2_session_set_nonce_caller(tpm2_session_data *data, TPM2B_NONCE *nonce);
 
+/**
+ * Gets the nonceNewer parameter.
```
williamcroberts: nonceNewer is the internal name. The caller might not know that. What is this nonce value for. Describe it more high level terms.

Retrieves the nonce value as returned via starting a new session. This nonce value can be used for blah blah blah.

man/tpm2_nvwrite.1.md
```diff
@@ -40,6 +40,11 @@ If _FILE_ is not specified, it defaults to stdin.
     should follow the authorization formatting standards, see section
     "Authorization Formatting".
 
+  * **-X**, **--hmac-auth**
```
williamcroberts: We're not doing this, delete this.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+#include "tpm2_nv_util.h"
+#include "tpm2_openssl.h"
+#include "tpm2_session.h"
+
+uint8_t *tpm2_get_cp_hash(TSS2_SYS_CONTEXT *sapi_context, TPM2B_NAME entity_1_name,
+    TPM2B_NAME entity_2_name) {
+
+    SHA256_CTX sha256;
+    int is_success = SHA256_Init(&sha256);
+    if(!is_success) {
```
williamcroberts: `if (`
Missing a space here and in some other areas as well.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+        LOG_ERR("SHA256_Init failed when calculating HMAC");
+        return NULL;
+    }
+
+    uint8_t cmd_code[4];
+    TSS2_RC rval = Tss2_Sys_GetCommandCode(sapi_context, cmd_code);
+    if(rval != TPM2_RC_SUCCESS) {
```
williamcroberts: space on if and should use TSS_RC_SUCCESS. TPM_RC_* is for direct from TPM RC codes, where as this came from the TSS calls, and thus should use the TSS_RC_* returns. This is really a semantic difference, as the values are the same.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+
+    uint8_t *cp_hash = malloc(SHA256_DIGEST_LENGTH);
+    if (!cp_hash) {
+        LOG_ERR ("OOM");
+        return NULL;
+    }
```
williamcroberts: you have newlines between your code blocks except for here, add it.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+        LOG_ERR ("SHA256_Final failed when calculating HMAC");
+        free(cp_hash);
+        return NULL;
+    }
+
+    return cp_hash;
+}
+
+bool tpm2_read_transient_persistent_obj_name(TSS2_SYS_CONTEXT *sapi_context,
+    TPM2_HANDLE entity_handle, TPM2B_NAME *entity_name) {
```
williamcroberts: newline here

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+    TPM2B_PUBLIC object_public;
+    TSS2L_SYS_AUTH_RESPONSE sessions_data_out;
+    TPM2B_NAME qualified_name = TPM2B_TYPE_INIT(TPM2B_NAME, name);
+    TSS2_RC rval = TSS2_RETRY_EXP( Tss2_Sys_ReadPublic(sapi_context, entity_handle,
+                                        NULL, &object_public, entity_name, &qualified_name,
+                                        &sessions_data_out));
+    if (rval != TPM2_RC_SUCCESS) {
```
williamcroberts: TSS2_RC_SUCCESS

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+    is_success = SHA256_Update(&sha256, entity_2_name.name, entity_2_name.size);
+    if (!is_success) {
+        LOG_ERR ("SHA256_Update failed when calculating HMAC");
+        return NULL;
+    }
+
+    size_t command_params_size;
+    const uint8_t *command_params;
+    rval = Tss2_Sys_GetCpBuffer(sapi_context, &command_params_size, &command_params);
+    if (rval != TPM2_RC_SUCCESS) {
```
williamcroberts: TSS2_RC_SUCCESS

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+        LOG_PERR(Tss2_Sys_ReadPublic, rval);
+        return false;
+    }
+    return true;
+}
+
+bool tpm2_hmac_auth_get_entity_name(TSS2_SYS_CONTEXT *sapi_context, TPM2_HANDLE entity_handle,
+    TPM2B_NAME *entity_name) {
+    TPM2B_NV_PUBLIC nv_public = TPM2B_EMPTY_INIT;
```
williamcroberts: newline

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+    session_data->hmac.size = temp_hash_len;
+    free(to_hmac_buffer);
+
+    return true;
+}
+
+bool tpm2_hmac_auth_get_command_buffer_hmac(TSS2_SYS_CONTEXT *sapi_context,
+    TPM2_HANDLE auth_handle, tpm2_session *hmac_session,
+    TPMS_AUTH_COMMAND *session_data, TPM2B_NAME entity_1_name,
+    TPM2B_NAME entity_2_name) {
```
williamcroberts: newline

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+    if (!auth_handle) {
+        LOG_ERR("Authentication handle cannot be NULL");
+        return false;
+    }
+
+    command_authentication_hmac(hmac_session,
+        tpm2_get_cp_hash(sapi_context, entity_1_name, entity_2_name), session_data);
```
williamcroberts: If `tpm2_get_cp_hash()` fails, `command_authentication_hmac()` will deref NULL. You need to separate these and check for NULL return from `tpm2_get_cp_hash()`

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************;
+#ifndef HMAC_AUTH_H
+#define HMAC_AUTH_H
```
williamcroberts: Get rid of these header ifndef, define, endif's not needed. Likely just a copy paste error.

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,208 @@
...
+        LOG_ERR("Authentication handle cannot be NULL");
+        return false;
+    }
+
+    command_authentication_hmac(hmac_session,
+        tpm2_get_cp_hash(sapi_context, entity_1_name, entity_2_name), session_data);
+    return true;
+}
+
+#endif /* HMAC_AUTH_H */
```
williamcroberts: remove me

tools/tpm2_nvwrite.c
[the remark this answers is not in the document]
idesai: Ok I have a version of this ready. I will make a PR soon as a I flush out an issue with hmac calculation.

idesai:
@williamcroberts @martinezjavier Can you please take a look at the latest commits. I added the support for hmac authorizations that can be enforced specifying the hmac string to auth/password utility like -P hmac:<password>

martinezjavier:
> @williamcroberts @martinezjavier Can you please take a look at the latest commits. I added the support for hmac authorizations that can be enforced specifying the hmac string to auth/password utility like -P hmac:

I can take a look, but I see that you didn't address comments that were made for the previous version?

Something that isn't easy in the Github workflow is to know what specific version of a patch-set one is reviewing or what changed between revisions...

tools/tpm2_nvwrite.c
idesai: @martinezjavier, this is no longer an option. The choice to use HMAC-auth instead of plaintext password is now specified in the auth utility. So you can specify like -P hmac:<Text>

tools/tpm2_nvwrite.c
idesai: @martinezjavier the man page for authorization is now updated. Not really sure if i can add a unit test that will work. Reason being, the session continuation is broken with user space resource manager and kernel resource manager is not available with simulator. Currently am testing it working directly with the TPM.

martinezjavier:
@idesai yeah, my comment wasn't referring to this PR but was just complaining about the Github workflow in general.

What I usually do for non-trivial PR that require many iterations is to edit the PR subject each time that I push a new version and add a `v2` prefix.

I also add a new comment mentioning that I pushed a new version and add a changelog, i.e:

`
Changes since v1:

- Replaced Foo by Bar
- Changed X to do Y instead.
- Etc
`

tools/tpm2_nvwrite.c
idesai: Fixed

tools/tpm2_nvwrite.c
idesai: Fixed

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,92 @@
+//**********************************************************************;
+// Copyright (c) 2015, Intel Corporation
```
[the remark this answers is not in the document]
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: @martinezjavier. From looking at the change notes, The specific openssl api calls made here do not need the #ifdefery like other tools do.

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

tools/tpm2_nvwrite.c
idesai: @martinezjavier @williamcroberts I added a new routine tpm2_auth_util_from_optarg_new to parse hmac: string. If this is good then we could change it in other tools.

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_session.c
idesai: Fixed

tools/tpm2_nvwrite.c
idesai: Implemented

tools/tpm2_nvwrite.c
idesai: Implemented

lib/tpm2_hmac_auth.c
```diff
@@ -0,0 +1,206 @@
...
+    SHA256_CTX sha256;
+    int is_success = SHA256_Init(&sha256);
+    if(!is_success) {
+        LOG_ERR("SHA256_Init failed when calculating HMAC");
+        return NULL;
+    }
+
+    uint8_t cmd_code[4];
+    TSS2_RC rval = Tss2_Sys_GetCommandCode(sapi_context, cmd_code );
+    if( rval != TPM2_RC_SUCCESS ) {
```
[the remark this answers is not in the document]
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.h
idesai: fixed

lib/tpm2_hmac_auth.h
idesai: Fixed

lib/tpm2_session.h
idesai: done

lib/tpm2_session.c
idesai: Fixed

lib/tpm2_nv_util.h
idesai: @williamcroberts @danintel will make a separate commit.

man/tpm2_nvwrite.1.md
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_session.c
idesai: Implemented

lib/tpm2_hmac_auth.h
idesai: done

lib/tpm2_hmac_auth.h
idesai: done

lib/tpm2_session.h
idesai: fixed

man/tpm2_nvwrite.1.md
idesai: Fixed. Updated the authorization.md file instead to indicate hmac auth mechanism.

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: fixed

lib/tpm2_hmac_auth.c
idesai: Implemented

lib/tpm2_hmac_auth.c
idesai: Fixed

lib/tpm2_hmac_auth.c
idesai: Fixed

williamcroberts:
<snip>
 <https://github.com/williamcroberts>  @martinezjavier
> <https://github.com/martinezjavier>  Can you please take a look at the latest
> commits. I added the support for hmac authorizations that can be enforced
> specifying the hmac string to auth/password utility like -P hmac:
> 
> I can take a look, but I see that you didn't address comments that were made for
> the previous version?
> 
> Something that isn't easy in the Github workflow is to know what specific version
> of a patch-set one is reviewing or what changed between revisions...

Maybe we should document that you need a [V1] in the PR message and we can increment it
And do a shortlog of V1, v2 changes in the message body.

Also if you git amend --reset-author it updates the timestamp which helps me know if it's new.
I can look at my comment time versus patch time.

> 
<snip>

lib/tpm2_auth_util.c
```diff
@@ -94,6 +97,12 @@ static bool handle_str(const char *password, TPMS_AUTH_COMMAND *auth) {
         password += STR_PREFIX_LEN;
     }
 
+    bool is_hmac_prefix = !strncmp(password, HMAC_SESSION_PREFIX,
+        HMAC_SESSION_PREFIX_LEN);
+    if (is_hmac_prefix) {
+        password += HMAC_SESSION_PREFIX_LEN;
+    }
```
williamcroberts: You're polluting the routine handle_str with hmac logic to re-use the code block on 109. The right way to reuse that code block is to move it to a seperate routine and call it.

You should have a routine called handle_hmac() that does this hmac prefix jump, and then call that copy routine.

lib/tpm2_session.c
```diff
@@ -86,6 +86,10 @@ void tpm2_session_set_nonce_caller(tpm2_session_data *data, TPM2B_NONCE *nonce)
     data->nonce_caller = *nonce;
 }
 
+TPM2B_NONCE * tpm2_session_get_nonce_tpm(tpm2_session *session) {
```
williamcroberts: I think you want this cost TPM2B_NONCE

lib/tpm2_session.h
```diff
@@ -32,6 +32,9 @@
 
 #include <tss2/tss2_sys.h>
 
+#define MAX_AUTH_SESSIONS 3
```
williamcroberts: this is the ARRAY_LEN() of type TSS2L_SYS_AUTH_COMMAND AUTHS field... so one could use that.

idesai:
@williamcroberts thanks for all the help to put more structure for the hmac authentication code.

williamcroberts:
We can close this. We're going to flip over to ESAPI so we don't have to write this code. @AndreasFuchsSIT this should make you smile.

PR #973, opened 2018-04-03
travis: get 3.x building

williamcroberts:
Travis stopped working on 3.X as the deb packages are
no longer at the link specified. Rather then having to wget
a ton of dependencies, that tend to cause intermittent
build failures, use the docker container approach on the
master branch.


.ci/docker.run
```diff
@@ -0,0 +1,158 @@
...
+
+# let the tpm simulator have time to start up before abrmd
+# tries to connect.
+sleep 5
+
+echo "starting abrmd server"
+tpm2-abrmd --tcti=socket &
+echo "started abrmd server"
+
+if [ -d build ]; then
```
martinezjavier: Wouldn't be better to just pass the `--rm` option to the `docker run` command so we know that the container image is cleaned up on exit?

That way you don't need to care about previous state.

williamcroberts: It is. This is for me when running in an interactive container instance numerous times.

PR #978, opened 2018-04-11
INSTALL.md: add a build from source example 

kaiserkyu:
Add a complete build example of tpm2-tools with its dependencies on fedora 27 ready to run

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -29,6 +29,13 @@ To build and install the tpm2-tools software the following software is required:
 
 ### Typical Distro Dependency Installation
 
+Here we are going to satisfy tpm2-tools dependencies with:
+* tpm2-tss: <https://github.com/tpm2-software/tpm2-tss>
+* tpm2-abrmd: <https://github.com/tpm2-software/tpm2-abrmd>
+* TPM simulator: <https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm974.tar.gz>
+
+Which are necessary for the build example section at the bottom of this file, we need to satisfy the dependencies for each item named above except for the simulator.
+
 #### Ubuntu 16.04
 
 Satisfying the dependencies for tpm2-tools falls into two general steps, stuff
@@ -63,6 +70,33 @@ source, building and installing them. They can be located here:
   * SAPI - The low level system API: <https://github.com/tpm2-software/tpm2-tss>
   * ABRMD (**recommended but optional**) - Which is the userspace resource
     manager: <https://github.com/tpm2-software/tpm2-abrmd>
+
+
+
+#### Fedora
+
+
+In case you want to build from source the next command block should cover all the dependencies for tpm2-tools, the low level system API (tpm2-tss) and the userspace resource manager (tpm2-abrmd).  
+
+```
+$ sudo dnf -y update && sudo dnf -y install automake libtool \
+autoconf autoconf-archive libstdc++-devel gcc pkg-config \
+uriparser-devel libgcrypt-devel dbus-devel glib2-devel \
+compat-openssl10-devel libcurl-devel PyYAML
+
+```
+Some distros have the dependencies already packaged, you can simply install the package that contains the needed build requirements.
+
+```
+$ sudo dnf builddep tpm2-tools
+
+```   
+
+The package installed above contains all the dependencies for tpm2-tools included the projects mentioned at the beginning of this section (tpm2-tss and tpm2-abrmd)
+
+For more detailed information about the dependencies of tpm2-tss and tmp2-abrmd, please consult the corresponding links for each project.
+
+
 
 ### tpm2-tools SAPI and ABRMD Dependency Version Chart
 
@@ -100,10 +134,92 @@ For Example:
 ./tools/tpm2_getrandom 4
 ```
 
-## Installing
+### Building from source example
 
-For those who wish to install, one can execute:
+
+Now we can start building the projects, there are four major steps for building the projects from source:
+#### Bootstrapping the build
+
+With the bootstrap command we run a script which generates the list of source files along with the configure script.
+
+In the project directory:
+
+```
+$  ./bootstrap
+```
+
+#### Configuring the build
+
+Here we run the configure script, this generates the makefiles needed for the compilation.
+
+```
+$ ./configure
+```
+
+Depending of the project, you can add additional information to the configure script, please refer to the links provided below for more information about the custom options.
+
+* Default values for GNU installation directories: <https://www.gnu.org/prep/standards/html_node/Directory-Variables.html>
+* Custom options for tpm2-tss: <https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md>
+* Custom options for tpm2-abrmd: <https://github.com/tpm2-software/tpm2-abrmd/blob/master/INSTALL.md>
+
+#### Compiling the Libraries
+
+We use the make command for compile the code.
+
+```
+$ make -j$(nproc)
+```
+
+
+#### Installing the Libraries
+
+Once we have finish building the projects it's time to install them.
 
 ```
 $ sudo make install
 ```
+Now putting all together:
+
+* ##### Tpm2-tss
+```
+$ git clone git://github.com/tpm2-software/tpm2-tss.git
+$ cd tpm2-tss
+$ ./bootstrap
+$ ./configure --prefix=/usr
+$ make -j5
+$ sudo make install
+```
+
+* ##### Tpm2-abrmd
+```
+$ git://github.com/tpm2-software/tpm2-abrmd.git
+$ cd tpm2-abrmd
+$ ./bootstrap
+$ ./configure --with-dbuspolicydir=/etc/dbus-1/system.d
+--with-udevrulesdir=/usr/lib/udev/rules.d
+--with-systemdsystemunitdir=/usr/lib/systemd/system
+--libdir=/usr/lib64 --prefix=/usr
+$ make -j5
+$ sudo make install
+```
+
+* ##### Tpm2-tools
+```
+$ git clone git://github.com/tpm2-software/tpm2-tools.git
+$ cd tpm2-tools
+$ ./bootstrap
+$ ./configure --prefix=/usr
+$ make -j5
+$ sudo make install
+```
+* ##### TPM simulator
+```
+$ mkdir ibmtpm && cd ibmtpm
+$ wget https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm974.tar.gz
+$ tar -zxvf ibmtpm974.tar.gz
+$ cd src
+$ make -j5
+```
+
+
+And it's done, you are ready to run the projects.


INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
 ./tools/tpm2_getrandom 4
 ```
 
+### Build example
```
martinezjavier: I would separate the *installing build dependencies* on a certain distro from the more general *build from source* that should be basically the same for every distro. There's already a *Typical Distro Dependency Installation* section where it explains how to install the dependencies for Ubuntu, so I would just add another sub-section for Fedora.

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+
+#### Fedora 27
```
martinezjavier: Just say Fedora, otherwise this doc will be outdated soon (Fedora 28 is going to be released in a of weeks).

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+* TPM simulator: <https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm974.tar.gz>
+
+First we need to satisfy the dependencies for each item named above except for the simulator, in the next code block we install a mix of packages that should get the work done.
+
+
+```
+$ sudo dnf -y update && sudo dnf -y install automake libtool \
+autoconf autoconf-archive libstdc++-devel gcc pkg-config \
+uriparser-devel libgcrypt-devel dbus-devel glib2-devel \
+compat-openssl10-devel libcurl-devel PyYAML
```
martinezjavier: I think this could just be replaced with ```sudo dnf builddep tpm2-tools```

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+
+```
+For more detailed information about the dependencies of tpm2-tss and tmp2-abrmd, please consult the links provided above.
+
+Now we can start building the packages
```
martinezjavier: There are no packages, maybe say components or projects?

INSTALL.md
martinezjavier: Actually, I think it should be ```sudo dnf builddep tpm2-tss``` otherwise the distro package `tpm2-tss` will be installed.

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+
+Now we can start building the packages  
+
+* ##### Tpm2-tss
+```
+$ git clone git://github.com/tpm2-software/tpm2-tss.git
+$ cd tpm2-tss
+$ ./bootstrap
+$ ./configure --prefix=/usr
+$ make -j5
```
martinezjavier: You are missing `sudo make install`, otherwise the libraries won't be installed in a path where the linker can find them.

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+* ##### Tpm2-abrmd
+```
+$ git://github.com/tpm2-software/tpm2-abrmd.git
+$ cd tpm2-abrmd
+$ ./bootstrap
+$ ./configure --with-dbuspolicydir=/etc/dbus-1/system.d
+--with-udevrulesdir=/usr/lib/udev/rules.d
+--with-systemdsystemunitdir=/usr/lib/systemd/system
+--libdir=/usr/lib64 --prefix=/usr
+$ make -j5
```
martinezjavier: same here, please add a `sudo make install` command.

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+```
+
+* ##### Tpm2-tools
+```
+$ git clone git://github.com/tpm2-software/tpm2-tools.git
+$ cd tpm2-tools
+$ ./bootstrap
+$ ./configure --prefix=/usr
+$ make -j5
```
martinezjavier: And probably makes sense here too.

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+* ##### TPM simulator
+```
+$ mkdir ibmtpm && cd ibmtpm
+$ wget https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm974.tar.gz
+$ tar -zxvf ibmtpm974.tar.gz
+$ cd src
+$ make -j5
+```
+Depending of the version of fedora you are running you might have problems with the make command:
+![TPM simulator issue](/fedora-shared/tpm_simulator.png)
```
martinezjavier: This is a broken link. But I don't think these workarounds should be mentioned here, could you instead post your patch to the TPM simulators developers explaining that it's broken on recent Fedora releases (probably due having a newer  GCC as you said).

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+This issue is present on fedora 26/27, it could be an issue with an older toolchain on the simulator than the one used by fedora 26/27, also, the -Werror wont let any build warning pass, so we need to patch this.
+The patch we are going to use is this: https://hastebin.com/raw/ovayaxamim
```
martinezjavier: I believe the links in hastebin are temporary, so this will become a broken link as well. But as mentioned, better to post that patch to the TPM simulator developers.

INSTALL.md
```diff
@@ -105,6 +105,81 @@ For Example:
...
+$ sudo curl https://hastebin.com/raw/ovayaxamim -o fix_build.patch
+$ sudo patch -p1 < ./fix_build.patch
+$ cd src
+$ make -j5
+```
+
+And it's done, you are ready to run and test.
+
+
 ## Installing
```
martinezjavier: I believe think this *Installing* section can be removed after your changes.

INSTALL.md
martinezjavier: This would posting the patch to ibmtpm20tss-users@lists.sourceforge.net I think.

martinezjavier:
@dantpm I honestly don't understand what are you getting at...

I personally won't ever say no to more documentation, for me the more the better. In this case @kaiserkyu (who it seems happens to be a Fedora user) is proposing to better document how to build and install the tools from source. As I mentioned in my review, that should be mostly the same on all distros so I suggested to him to move the Fedora specific part to the section where we already are mentioning other distros (Ubuntu).

williamcroberts:
<snip>
> the same on all distros so I suggested to him to move the Fedora specific part to
> the section where we already are mentioning other distros (Ubuntu).

I agree. This comment from Javier makes perfect sense. General instructions up front for
Everybody. Distro specific on the backend. I would even recommend making a docker
container and adding it to .travis.yml. Then you have a docker image ready for folks
to use and we increase our coverage on various distros.

We could even add commands on how to get these docker containers.

I'm not saying we have to do all this to take a doc patch, it's just food for thought.

> 
<snip>

martinezjavier:
> We could even add commands on how to get these docker containers.

Absolutely.

dantpm:
@martinezjavier I'm all for more documentation too--it was very painful to get this stuff working when simple required things aren't mentioned (such as how to specify the tcti).  I agree that the Fedora part should be separate.

I also think the tpm2-tss stuff belongs in  tpm2-tss, not here.

I will try again to add a Ubuntu documentation section after this is integrated and hope it will be accepted this time.

martinezjavier:
> I also think the tpm2-tss stuff belongs in tpm2-tss, not here.

I disagree. I think it doesn't hurt to have instructions on how to build the dependencies in one place, instead of the user having to go back and forth between the INSTALL files in the different projects.

I can't count how many times I've to assist users in the mailing lists due problems building and installing the tpm2-{tss,abrmd,tools}. So I really think we need something like the proposed PR.

And it's true that the simulator isn't part of tpm2-software projects, but we use it for testing and development purposes too so I don't see why it can't be mentioned. The only drawback with documentation is that it can become stale, but if that's the case and nobody step up to fix it, you can always remove it later.

dantpm:
Thanks for all your assistance by the way.

I find it surprising that we're answering the same questions again and again instead of documenting it.  This should help a lot.

For what it's worth, I still think the tss documentation belongs in tss. The three repos, tss, abrmd, and tools are like a layer cake and should be done one-by-one. One should not duplicate how to build tss under tools. But I may lose this argument.

INSTALL.md
herrold: Please no, here -- this practice is suitable for a developmental machine, perhaps, but not for much more.  

Clearly a 'spec' file for Fedora exists -- use it and make packages

martinezjavier: @herrold of course I know that there are packages for tpm2-{tss,abrmd,tools} available in Fedora, I'm the one who maintains them...

Point is what this PR suggests to add information on how to **build from source**.  For whatever reasons (testing the latest master, development purposes, etc) people may want to build from source and if there are already instructions for that, then I think that we should include information on how what was built can be installed.

martinezjavier:
@kaiserkyu I still didn't read your new patches, but you shouldn't add fixups on top but instead rebase on a new set of patches and force a git push so Github updates the patch-set.

williamcroberts:
> -----Original Message-----
> Sent: Friday, April 13, 2018 10:43 AM
> Cc: Roberts, William C <william.c.roberts@intel.com>; Comment
> <comment@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-tools] INSTALL.md: add a build example on
> fedora (#978)
> 
> @kaiserkyu <https://github.com/kaiserkyu>  I still didn't read your new patches,
> but you shouldn't add fixups on top but instead rebase on a new set of patches
> and force a git push 

FYI: You'll need to force push back to your fork branch the PR is originating from:
git push -f fork branch

so Github updates the patch-set.
>

williamcroberts:
@kaiserkyu Also --reset-author helps when your ammending commits, as it updates the time stamp so we know these have changed. Let us know when you upload new commits and are ready for a review.

INSTALL.md
```diff
@@ -68,6 +68,25 @@ source, building and installing them. They can be located here:
 
 To get md2man-roff, see there page at: <https://github.com/sunaku/md2man>
```
martinezjavier: @kaiserkyu I would change your commit subject line and message to say that you are adding a `build from source` example rather than something that's Fedora specific.

INSTALL.md
```diff
@@ -68,6 +68,25 @@ source, building and installing them. They can be located here:
...
 
+#### Fedora
+
+Here we are going to satisfy tpm2-tools dependencies with:
+* tpm2-tss: <https://github.com/tpm2-software/tpm2-tss>
+* tpm2-abrmd: <https://github.com/tpm2-software/tpm2-abrmd>
+* TPM simulator: <https://downloads.sourceforge.net/project/ibmswtpm2/ibmtpm974.tar.gz>
+
+Which are necessary for the build example section at the bottom of this file, we need to satisfy the dependencies for each item named above except for the simulator, in the next code block we install a mix of packages that should get the work done.
+
```
martinezjavier: Please move all the above either to the ***Building*** section or before the ***Ubuntu 16.04*** section.

martinezjavier:
@kaiserkyu what @williamcroberts said, please let us know once you had done the suggested changes.

I think that basically what we want is:

* General information about the dependencies.
* How to install those dependencies on common distros (we can have a per distro section there)
* How to install those dependencies from source in case that someone wants to build on a distro where the dependencies have not been packaged.
* How to install the `tpm2-tools` from source, as others mentioned this is only for development or testing purposes since users should install from the distro packages instead, if these are available.

You can use as a reference the `tpm2-tss` INSTALL document that I think has managed to explain this clearly:

https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md

kaiserkyu:
I've made the changes suggested by @martinezjavier, i think it covers all the points

martinezjavier:
@kaiserkyu I think is great now, thanks a lot for addressing the issues we pointed out! Only yesterday we had yet another report of the tpm2-abrmd not working which was just a problem with the reporter build environment (https://github.com/tpm2-software/tpm2-abrmd/issues/432), so I really think we should merge this as soon as possible.

One last thing, your changes not conflicts with commit https://github.com/tpm2-software/tpm2-tools/commit/a9ab692201c2e1bd1ca8236bf7fc432708911234 ("INSTALL.md: Remove mention of md2man-roff"). Could you please rebase and resolve the conflict?

You need basically to do:

```
$ git remote update
$ git rebase origin/master
Solve conflict and then
$ git add INSTALL.md
$ git push -f kaiserkyu Develop (or whatever your remote is named).
```

kaiserkyu:
@martinezjavier thanks for the review, i'm looking forward to keep working on the project

PR #980, opened 2018-04-11
[RFC] Initial ESAPI Support

AndreasFuchsSIT:
This patch introduces the step-by-step transition to ESAPI into the tools.
It introduces the basic recurring infrastructure in main.c and moves the
first command tpm2_pcrlist over to esapi.

Please let me know what to change...

For the future, I guess the handling of data formats for session and transient objects will be the biggest obstacles... Please let me know, if we want to make them big one-step changes...

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,7 +170,6 @@ tools_tpm2_quote_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_changeauth_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_createek_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_createak_CFLAGS = $(AM_CFLAGS) -DSAPI
-tools_tpm2_hash_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_activatecredential_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_makecredential_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_nvlist_CFLAGS = $(AM_CFLAGS) -DSAPI
--- a/Makefile.am
+++ b/Makefile.am
@@ -172,7 +172,6 @@ tools_tpm2_createek_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_createak_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_activatecredential_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_makecredential_CFLAGS = $(AM_CFLAGS) -DSAPI
-tools_tpm2_nvlist_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_nvread_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_nvreadlock_CFLAGS = $(AM_CFLAGS) -DSAPI
 tools_tpm2_nvwrite_CFLAGS = $(AM_CFLAGS) -DSAPI
--- a/Makefile.am
+++ b/Makefile.am
@@ -39,13 +39,13 @@ INCLUDE_DIRS = -I$(top_srcdir)/tools -I$(top_srcdir)/lib
 LIB_COMMON := lib/libcommon.a
 
 AM_CFLAGS := \
-    $(INCLUDE_DIRS) $(EXTRA_CFLAGS) $(TSS2_SYS_CFLAGS) $(CRYPTO_CFLAGS) \
-    $(CODE_COVERAGE_CFLAGS)
+    $(INCLUDE_DIRS) $(EXTRA_CFLAGS) $(TSS2_ESYS_CFLAGS) $(TSS2_SYS_CFLAGS) \
+    $(CRYPTO_CFLAGS) $(CODE_COVERAGE_CFLAGS)
 
 AM_LDFLAGS   := $(EXTRA_LDFLAGS) $(CODE_COVERAGE_LIBS)
 
 LDADD = \
-    $(LIB_COMMON) $(TSS2_SYS_LIBS) $(CRYPTO_LIBS) -ldl
+    $(LIB_COMMON) $(TSS2_ESYS_LIBS) $(TSS2_SYS_LIBS) $(CRYPTO_LIBS) -ldl
 
 # keep me sorted
 bin_PROGRAMS = \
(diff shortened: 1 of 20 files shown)


martinezjavier:
@AndreasFuchsSIT thanks a lot for working on this. I only did a quick look to the code but I've some questions that are more about the approach we want to take rather than implementation details.

In my opinion, the tpm2-tools project serves 3 purposes:

1) It can be use to quickly prototype using the TPM (i.e: just a bunch of shell scripts for proof-of-concept, etc).
2) To test and get familiar with the TPM.
3) As a reference on how to use the TSS API for people that want to write their own applications.

Now, the TSS has (depending on how you look it) 6 software layers: RM, AB, TCTI, SAPI, ESAPI and FAPI. So my understanding is that this gives flexibility to application writers to target different layers of the stack, depending on what's the best trade-off for them between a high level of abstraction and fine grained control of the TPM commands sent to the TPM.

With that in mind, the question is which layers should we support in the tpm2-tools? We have been  using SAPI basically because it was the only one implemented. Now that we have ESAPI, this PR implies that we will target that layer instead. Will we stuck with that or the plan is to migrate to FAPI once that's properly specified and implemented?

Another thing is that the granularity and naming for the tools was based on the TPM commands. So for example tpm2_createprimary maps to TPM2_CreatePrimary command and so on. Tools that don't map a TPM command are in the `aux` directory. Now, you choose a good example because there isn't a TPM2_PCRList command, probably the tool should be renamed to tpm2_pcr_read.

Anyway, I don't have an answer to this questions but I think is good to first decide what should be the purpose of the tools and which TSS layer (or layers) are supposed to target.

martinezjavier:
@AndreasFuchsSIT thanks a lot for your explanation, yes I agree with you to target the ESAPI then. I'm a little busy today with pending tasks but I'll take a look to the code tomorrow.

williamcroberts:
@martinezjavier I should have looped you in on the discussion I had with @AndreasFuchsSIT. But he, eloquently explained why we want to make the move.

I don't have any plans on moving to FAPI.

Makefile.am
```diff
@@ -109,7 +110,7 @@ tools_tpm2_clearlock_SOURCES = tools/tpm2_clearlock.c $(TOOL_SRC)
 tools_tpm2_create_SOURCES = tools/tpm2_create.c $(TOOL_SRC)
 tools_tpm2_createprimary_SOURCES = tools/tpm2_createprimary.c $(TOOL_SRC)
 tools_tpm2_getcap_SOURCES = tools/tpm2_getcap.c $(TOOL_SRC)
-tools_tpm2_pcrlist_SOURCES = tools/tpm2_pcrlist.c $(TOOL_SRC)
+tools_tpm2_pcrlist_SOURCES = tools/tpm2_pcrlist.c $(ETOOL_SRC)
```
williamcroberts: It would be a little more work, but could we just create a union with both contexts and send them to the tools?
Or maybe have a tool say, give me esapi and send it the esapi vs sapi context?

I don't want to have near duplicate files hanging around.

AndreasFuchsSIT: you mean something like this and a bunch of other ifdefs thoughout the tpm2-tools file ?
```
#ifdef SAPI
#define TOOL_CONTEXT TSS2_SYS_CONTEXT
#elif defined ESAPI
#define TOOL_CONTEXT ESYS_CONTEXT
#else
#error "Need to specify -DSAPI or -DESAPI !"
#endif
```

williamcroberts: No, pass this to the tools "on run" routine.
```c
union {
 ESAPI *esapi;
 SAPI *sapi;
} tpm_context;
```
You'll have to update tpm2_tool.c to instantiate both of these and update the interface to all tools to pass this. Then in the tools just assign the original variable name from the union.

If you don't want to instantiate both, you can just have tpm2_options_new() set a flag to indicate to give me esapi, and then pick based on that.

AndreasFuchsSIT:
@williamcroberts I was hoping to avoid a branch as much as possible.

My basic idea was:
1. Get one example in place (this PR)
2. Move all tools that use neither sessions nor transient objects (several PRs)
3. Move all tools that use fixed handles only (e.g. Hierarchies) without session support
4. Move all tools that use sequences without session support
5. Move all tools that use NV without session support
6. Add session support

I guess we can do 1. and 2. on master. The question is if you'd like to see the intermediate steps after 3., 4., 5. in a branch or if you can live without session support in master for some time ?

Makefile.am
AndreasFuchsSIT: I don't think I understand you. Do you wish to keep both implementations (SAPI and ESAPI) side-by-side, or are you requesting less duplicate code for the transition period from SAPI to ESAPI ?

In the latter case, I'd prefer to have each of the ESAPI files look like they shall look in the end, so we don't need to touch them again once SAPI is eliminated; i.e. they all use ESYS_CONTEXT.
The tpm2_tools.c can would be adapted to SAPI vs ESAPI using #ifdefs that we then remove after transition is complete ?

williamcroberts:
<snip>
> 
> I don't think I understand you. Do you wish to keep both implementations (SAPI
> and ESAPI) side-by-side, or are you requesting less duplicate code for the
> transition period from SAPI to ESAPI ?

The more I think about this, I can't have this in an incomplete state on master.
We need to have this on a separate branch and take it all at once, or I need
To branch 4.X now, as I may need to cut a 4.x branch at some point soon.

With that said, I care less on how it's actually done.

> 
> In the latter case, I'd prefer to have each of the ESAPI files look like they shall look
> in the end, so we don't need to touch them again once SAPI is eliminated; i.e.
> they all use ESYS_CONTEXT.
> The tpm2_tools.c can would be adapted to SAPI vs ESAPI using #ifdefs that we
> then remove after transition is complete ?
> 
<snip>

tools/main.c
```diff
@@ -0,0 +1,186 @@
...
+    .tssVersion = 108, \
+}
+
+static ESYS_CONTEXT* esys_ctx_init(TSS2_TCTI_CONTEXT *tcti_ctx) {
+
+    TSS2_ABI_VERSION abi_version = SUPPORTED_ABI_VERSION;
+    ESYS_CONTEXT *esys_ctx;
+
+    TSS2_RC rval = Esys_Initialize(&esys_ctx, tcti_ctx, &abi_version);
+    if (rval != TPM2_RC_SUCCESS) {
```
williamcroberts: TSS2_RC_SUCCESS

AndreasFuchsSIT:
Ok, give me a branch then... :-D

williamcroberts:
> Ok, give me a branch then... :-D


https://github.com/tpm2-software/tpm2-tools/tree/esapi-integration

AndreasFuchsSIT:
Ok, so this is the current status...
The system-test on nv will be fixed by this PR https://github.com/tpm2-software/tpm2-tss/pull/962.

I don't know why clang dies here though. Works on my machine...

Especially patch tpm2_hash shows the approach going forward:
- Duplicate functions with a _sapi suffix to keep the whole thing running.

I guess I'll run into problems once tests with sessions and objects show up.
If you have any idea, let me know...

williamcroberts:
> I don't know why clang dies here though. Works on my machine...

Maybe it’s a clang version or env issue? Did you try building and running in the docker image
locally?

AndreasFuchsSIT:
Ok, the only remaining problem seems to be lack of https://github.com/tpm2-software/tpm2-tss/pull/962

williamcroberts:
Im closing this as you have that branch you're working on.

Issue #966, opened 2018-03-29
Unify options to specify parent and current/child object


joshuagl:
When tools take a parent and/or current object we can simplify the options by having **-C** always be the parent object and **-c** the child object.
Rather than having a separate parameter for different object types (file vs. handle) we can reduce the number of related options from 4 to 2  by defining semantics for specifying handle vs. file and have a library function to determine the object type and load it appropriately.

For example:
```
0x1234 - specifies a handle
foo.dat - specifies a context file
file:0x1234 - specifies a file
```

joshuagl:
I'm most of the way there on a series to implement this:
https://github.com/joshuagl/tpm2-tools/commits/issue966

A few actions remain:

- [ ] fix examples using modified commands in all manual pages
- [ ] ensure common fragment is correctly included in manual pages of modified commands
- [ ] figure out what to do where -c/-C are used in other tools, ensure -c/-C usage is at least consistent for tools of the same family
- [ ] fix the cases where tests are using awk rather than `yaml_get_kv`
  - [ ] createak.sh
  - [ ] createek.sh
  - [ ] getmanufec.sh
- [ ] ensure commit logs reflect the patches (type and function were renamed)

Whilst the patch is in review, or even after merge, I need to:
- [ ] update the options matrix wiki page

williamcroberts:
@joshuagl I have a python script here that finds conflicting issues:
https://github.com/tpm2-software/tpm2-tools/pull/1007

joshuagl:
Wiki page will be updated as part of #957, closing this ticket.

PR #982, opened 2018-04-12
Unify options to specify parent and current/child object

joshuagl:
The goal of this series is to begin formalising around a standard set of short options, specifically that where tools take a parent and/or current object we can simplify the options by having -C always be the parent object and -c the child object.

The first patch adds a function to tpm2_util to determine the type of a context object from the value of the option argument, 4 cases are handles:

1. 0x1234 - specifies a handle
2. 123456 - specifies a handle if it parses to a uint32 with `tpm2_util_string_to_uint32()`
3.  foo.dat - specifies a context file
4. file:0x1234 - specifies a file

Case 2 was added specifically to support the tests where the YAML parser used in yaml_get_kv coerces the hexadecimal representation into an integer representation.

When reviewing please double-check my thought processes on what constitutes a parent object (-C) and what constitutes a child object (-c). I wasn't always certain from reading our documentation and in a few cases changed my mind before the final PR.

Resolves issue #966


joshuagl:
Whoops, I broke one of the unit tests with Clang. Looking into it...

lib/tpm2_util.c
```diff
@@ -41,6 +41,13 @@
 #include "tpm2_tool.h"
 #include "tpm2_util.h"
 
+#define HANDLE_PREFIX "0x"
+#define HANDLE_PREFIX_LEN strlen(HANDLE_PREFIX)
+#define FILE_PREFIX "file:"
+#define FILE_PREFIX_LEN strlen(FILE_PREFIX)
+#define FILE_SUFFIX ".dat"
+#define FILE_SUFFIX_LEN strlen(FILE_SUFFIX)
```
williamcroberts: ETOOMANY restrictions....

These should work:
```
1234
0x4D2
foo
foo.bin
foo.dat
file:1234
```

```c
if (!start_with_file) {
    tpm2_util_string_to_uint32(optvalue, &handle);
   // fall through on fail to file handler
}

load_tpm_context_from_file(optvalue);

return;
```

lib/tpm2_util.c
```diff
@@ -345,3 +352,40 @@ void tpm2_util_public_to_yaml(TPM2B_PUBLIC *public) {
         tpm2_tool_output("\n");
     }
 }
+
+tpm2_context_object_type tpm2_util_string_extract_object_name_type(const char *objectstr,
```
williamcroberts: This should probably load up an object if it's not a handle. Maybe give it a struct like so:
```c
struct loaded_object {
    HANDLE handle;
    const char *path;
}
```
Make a complimentary routine:
```c
tpm2_object_save(loaded_object *o) {
    if ( o->path) {
        // save context
    }
}
```
That's just a suggestion, im not 100% on it.

man/tpm2_createak.1.md
```diff
@@ -43,8 +43,9 @@ loaded-key:
     Specifies the current owner authorization.
     Same formatting as the endorse password value or **-e** option.
 
-  * **-E**, **--ek-handle**=_EK\_HANDLE_:
-    Specifies the persistent handle of the EK.
+  * **-C**, **--ek-context**=_EK\_CONTEXT\_OBJECT_:
+    Specifies the object context of the EK. Either a file or a handle's hex id.
```
williamcroberts: We support arbitrary bases for the handle number as understood by strtoul with a base of 0.

joshuagl:
Thanks for the review @williamcroberts, I'll work through your suggestions next week and come up with a new series.

man/tpm2_unseal.1.md
```diff
@@ -21,13 +21,10 @@ alive and pass that session using the **--input-session-handle** option.
 
 # OPTIONS
 
-  * **-H**, **--item**=_ITEM\_HANDLE_:
+  * **-c**, **--item-context**=_ITEM\_CONTEXT\_OBJECT_:
```
williamcroberts: item should probably go to object.

joshuagl:
Updated to address review comments, rebased on master and with @williamcroberts changes from #1008 pulled in.

williamcroberts:
This PR gets us much closer:
```
group: object:
	tools: ['tpm2_load', 'tpm2_createprimary', 'tpm2_unseal', 'tpm2_loadexternal', 'tpm2_makecredential', 'tpm2_readpublic', 'tpm2_activatecredential', 'tpm2_create', 'tpm2_changeauth']
	conflicts: set(['c', 'o'])
group: context:
	tools: ['tpm2_evictcontrol', 'tpm2_flushcontext']
	conflicts: set(['c', 'handle', 'H', 'context'])
```

lib/tpm2_util.c
```diff
@@ -41,6 +41,9 @@
 #include "tpm2_tool.h"
 #include "tpm2_util.h"
 
+#define FILE_PREFIX "file:"
+#define FILE_PREFIX_LEN strlen(FILE_PREFIX)
```
williamcroberts: this should be (sizeof(FILE_PREFIX) - 1) to take advantage of known size at compile time. But I won't nack this series for that. It's a minor optimization.

man/common/ctxobj.md
```diff
@@ -0,0 +1,10 @@
+# Context Object Format
+
+The type of a context object, whether it is a handle or file name, is
+determined according to the following logic:
+
+  * if the argument begins with the prefix "file:" it will be treat as a
```
williamcroberts: treated

man/common/ctxobj.md
```diff
@@ -0,0 +1,10 @@
...
+    context file
+  * if the argument can be loaded as a number by the strtoul() function
+    it will be treat as a handle
+  * otherwise the object is treat as a file
```
williamcroberts: treated

man/common/ctxobj.md
```diff
@@ -0,0 +1,10 @@
+# Context Object Format
+
+The type of a context object, whether it is a handle or file name, is
+determined according to the following logic:
+
+  * if the argument begins with the prefix "file:" it will be treat as a
+    context file
+  * if the argument can be loaded as a number by the strtoul() function
```
williamcroberts: No need to mention strtoul() here. Just if its a number should be sufficient.

man/common/ctxobj.md
williamcroberts: Add examples of each one.

man/tpm2_activatecredential.1.md
```diff
@@ -14,25 +14,25 @@ key.
...
 missing, assume NULL. Currently only support using TCG profile compliant EK as
 the keyHandle.
 
 # OPTIONS
 
 These options control the object verification:
 
-  * **-H**, **--handle**=_HANDLE_:
-    _HANDLE_ of the object associated with the created certificate by CA.
+  * **-c**, **--context**=_CONTEXT\_OBJECT_:
```
williamcroberts: I think we need a better long option name... Maybe something like: **--object**=_OBJ\_CTX\_OR\_HANDLE_

joshuagl:
I've updated to address review comments, except the remaining inconsistencies and the unit test -- as `tpm2_object_load()` will attempt to load a context file from a path with `files_load_tpm_context_from_path()` I'm not sure how easy it will be to unit test. I'll think on that.

williamcroberts:
> 
> I've updated to address review comments, except the remaining inconsistencies
> and the unit test -- as tpm2_object_load() will attempt to load a context file from
> a path with files_load_tpm_context_from_path() I'm not sure how easy it will be
> to unit test. I'll think on that.

You should be able to use cmocka to hook that since it's in the static lib. Look for wrap in the
Makefile for examples how it's been done. The you can just do whatever you need there and
lie to the caller.

The goal here is to test the parsing logic.

williamcroberts:
@joshuagl 
I had a thought (yes it hurt), The parent handle/context_file option should be capital and the child should be lowercase.

williamcroberts:
> @williamcroberts <https://github.com/williamcroberts>  that was actually my
> intent, though I guess I messed up in some places?

You may have the capital stuff, I didn't look that close. I just wanted to record that
Fleeting thought before I forgot.

>  Once I finish up the final
> pieces you've requested and rebase on master I'll review each of the tools I've
> changed and attempt to ensure we're using c for child context and C for parent
> context.
> 
> Apologies for the delay in response to this PR.

Sounds good, no worries.

joshuagl:
PR has been updated to address the remaining inconsistencies and rebase on master. The unit test is still outstanding but has been filed as issue #1037.

PR #997, opened 2018-04-20
test: Fix scan-build Warnings

### LINKED ISSUE #996: Scan-build makes CI die
since addition of scan-build in 53882c15af007a8f37951fb393656c556070f79c, compilation dies on travis-CI:
```
  CC       test/unit/test_unit_test_tpm2_header-test_tpm2_header.o
../test/unit/test_files.c:138:5: warning: Value stored to 'res' is never read
    res = files_read_bytes(f, found, sizeof(found));
    ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
../test/unit/test_tpm2_session.c:86:1: warning: Potential leak of memory pointed to by 'e'
}
^
1 warning generated.
FAIL: test/unit/test_tpm2_session
```
I hope these warnings are the reason for the Unit-Test-Failure...

danintel:
Fix scan-build warnings about a memory leak and a variable set but not used
that causes the compilation to fail.

Fixes #996.

### DIFF
--- a/test/unit/test_tpm2_session.c
+++ b/test/unit/test_tpm2_session.c
@@ -62,13 +62,14 @@ struct expected_data {
         TPM2_RC rc;
     } output;
 };
+static expected_data *e; /* XXX: Test kludge to quiet scan-build mem leak */
 
 static inline void set_expected(TPMI_DH_OBJECT key, TPMI_DH_ENTITY bind,
         TPM2B_ENCRYPTED_SECRET *encrypted_salt, TPM2_SE session_type,
         TPMT_SYM_DEF *symmetric, TPMI_ALG_HASH auth_hash,
         TPM2B_NONCE *nonce_caller, TPMI_SH_AUTH_SESSION handle, TPM2_RC rc) {
 
-    expected_data *e = calloc(1, sizeof(*e));
+    e = calloc(1, sizeof(*e));
     assert_non_null(e);
 
     e->input.key = key;
--- a/test/unit/test_files.c
+++ b/test/unit/test_files.c
@@ -136,6 +136,7 @@ static void test_file_read_write_bytes(void **state) {
 
     UINT8 found[1024] = { 0 };
     res = files_read_bytes(f, found, sizeof(found));
+    assert_true(res);
 
     assert_memory_equal(expected, found, sizeof(found));
 }


test/unit/test_tpm2_session.c
```diff
@@ -211,11 +212,15 @@ static void test_tpm2_session_defaults_good(void **state) {
     assert_int_equal(auth_hash, TPM2_ALG_SHA256);
 
     tpm2_session_free(&s);
+    free (e);
```
williamcroberts: I think we would need to use the test functions that have a setup and teardown, as any assert() will not run the cleanup code. just put this pointer in *state and pass it through.

Something changed is causing a seg fault in the travis logs.

williamcroberts:
FAIL: test/unit/test_tpm2_session

danintel:
I am cancelling this pull request.
It looks like scan-build is a red herring.
The real problem is some other regression.
scan-build passes--the failure is at test execution time (and it fails only for clang, not gcc).

See this log, for example:
https://travis-ci.org/tpm2-software/tpm2-tools/builds/369308446?utm_source=github_status&utm_medium=notification

These are minimal diffs to remove two compiler warnings:

```
1  test/unit/test_files.c
@@ -136,6 +136,7 @@ static void test_file_read_write_bytes(void **state) {
 
     UINT8 found[1024] = { 0 };
     res = files_read_bytes(f, found, sizeof(found));
+    assert_true(res);
 
     assert_memory_equal(expected, found, sizeof(found));
 }
    
3  test/unit/test_tpm2_session.c
@@ -62,13 +62,14 @@ struct expected_data {
         TPM2_RC rc;
     } output;
 };
+static expected_data *e; /* XXX: Test kludge to quiet scan-build mem leak */
 
 static inline void set_expected(TPMI_DH_OBJECT key, TPMI_DH_ENTITY bind,
         TPM2B_ENCRYPTED_SECRET *encrypted_salt, TPM2_SE session_type,
         TPMT_SYM_DEF *symmetric, TPMI_ALG_HASH auth_hash,
         TPM2B_NONCE *nonce_caller, TPMI_SH_AUTH_SESSION handle, TPM2_RC rc) {
 
-    expected_data *e = calloc(1, sizeof(*e));
+    e = calloc(1, sizeof(*e));
     assert_non_null(e);
 
     e->input.key = key;
```

PR #998, opened 2018-04-23
Revert "Add scan-build Static Code Analysis to Travis"

williamcroberts:
This reverts commit 53882c15af007a8f37951fb393656c556070f79c.


williamcroberts:
@danintel Looks like you did indeed take the build down. Not sure why, as your patch was initially marked as passing. But I'll revert this since it unblocks travis and we can try to get this in again.

PR #1002, opened 2018-04-24
tpm2_nv(read|write): use a default buffer size if TPM reports 0.

dunk-emesent:
On a NUC5i7RYH, the TPM reports it's max NV buffer size as 0, which
means we can neither read nor write. In this case, default to
MAX_NV_BUFFER_SIZE.

### DIFF
--- a/tools/tpm2_nvread.c
+++ b/tools/tpm2_nvread.c
@@ -123,6 +123,9 @@ static bool nv_read(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
     if (max_data_size > MAX_NV_BUFFER_SIZE) {
         max_data_size = MAX_NV_BUFFER_SIZE;
     }
+    if (max_data_size == 0) {
+        max_data_size = MAX_NV_BUFFER_SIZE;
+    }
 
     UINT8 *data_buffer = malloc(data_size);
     if (!data_buffer) {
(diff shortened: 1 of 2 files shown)


tools/tpm2_nvread.c
```diff
@@ -123,6 +123,9 @@ static bool nv_read(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
     if (max_data_size > MAX_NV_BUFFER_SIZE) {
         max_data_size = MAX_NV_BUFFER_SIZE;
     }
+    if (max_data_size == 0) {
```
williamcroberts: else if

tools/tpm2_nvread.c
```diff
@@ -123,6 +123,9 @@ static bool nv_read(TSS2_SYS_CONTEXT *sapi_context, tpm2_option_flags flags) {
...
+        max_data_size = MAX_NV_BUFFER_SIZE;
```
williamcroberts: I've heard that picking something along the lines of 512 is a safe bet that it will be supported versus going to the MAX size. So I think we should set this to 512. Make a define for this. `#define DEFAULT_NV_BUFFER_SIZE 512`

tools/tpm2_nvwrite.c
```diff
@@ -119,6 +119,9 @@ static bool nv_write(TSS2_SYS_CONTEXT *sapi_context) {
     if (max_data_size > MAX_NV_BUFFER_SIZE) {
         max_data_size = MAX_NV_BUFFER_SIZE;
     }
+    if (max_data_size == 0) {
```
williamcroberts: else if

tools/tpm2_nvwrite.c
```diff
@@ -119,6 +119,9 @@ static bool nv_write(TSS2_SYS_CONTEXT *sapi_context) {
...
+        max_data_size = MAX_NV_BUFFER_SIZE;
```
williamcroberts: Use 512.

williamcroberts:
Moved to #1012

PR #1003, opened 2018-04-24
Enable scan-build

### LINKED ISSUE #999: restore scan-build
See: https://github.com/tpm2-software/tpm2-tools/pull/998

danintel:
Enable scan-build compile-time static analysis.
Rebuild after running scan-build as test_tpm2_session fails
when build under scan-build.

Fixes #999.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -120,6 +120,16 @@ fi
 mkdir ./build
 pushd ./build
 
+# Run scan-build for gcc only.
+# Scan-build does not work with clang because of asan linking errors.
+if [[ "$CC" == gcc* ]]; then
+    scan-build ../configure --enable-unit $config_flags
+    scan-build --status-bugs make -j$(nproc)
+
+    # scan-build causes test_tpm2_session to fail, so
+    # rebuild after running scan-build.
+fi
+
 ../configure --enable-unit $config_flags
 make -j$(nproc)
 dbus-launch make -j$(nproc) check


joshuagl:
I would have thought the failure of *test_tpm2_session* after *scan-build* indicates that there's a bug in *test_tpm2_session* related to the runtime checks used by the clang static analyser.

Rather than this workaround, can we find what the issue is and fix it? I would have thought a good first step would be to perform the *scan-build* locally then manually invoke *test_tpm2_session* and see what messages are printed.

danintel:
@joshuagl I filed this bug for the scan-build problem building test_tpm2_session, see issue #1004.

This is just a first step to stop the bleeding.  I've noticed while working on this that new scan-build warnings are introduced with other integrations (such as set-but-not-used variables) while I was fixing old scan-build warnings.

williamcroberts:
Why did it die before? I'm a bit leery pulling this in without understanding why it broke in the first place.

danintel:
Test test_tpm2_session failed when compiled with scan-build enabled.

williamcroberts:
@danintel yes we know that, why did it fail? I thought scan-build only performs static analysis, but it was dying in tests?

danintel:
@williamcroberts Yes, scan-build should only perform static analysis. Apparently  it has a side effect or bug of modifying the object so it doesn't work in this one case. That's why I rebuild without scan-build and why I filed issue #1004.

Issue #957, opened 2018-03-26
Make option short forms consistent


joshuagl:
Overhaul the options the tools take so that the short form of options are consistent and carry the same meaning across tools.

- [x] document current option flows on the wiki
  - https://github.com/tpm2-software/tpm2-tools/wiki/3.0.x-options-matrix
- [x] wiki page recording unified options and their short forms as an easy reference point for future tools to follow the patterns
- [ ] fix short options to unified and consistent options across tools
  - [ ] Fix tools
  - [ ] Update man pages

williamcroberts:
Cool you got check boxes!

joshuagl:
I found them in the markdown help when editing the wiki -- `[ ]`.

williamcroberts:
I think `-A` is only for specifying attributes and all the `--pwdk`` type things are now `--auth-` so we want to get this matrix updated with those changes.

williamcroberts:
So I see some patterns emerging:
1. Most tools specify auth values for 1 or more objects, my patch corrects those
2. Most tools specify hierarchy, my patch makes this clean, not sure if **-a** is best.
3. Most tools specify a parent and current object in either:
  * Handle
  * Context blob
4. Others I haven't looked at.

Maybe we can reign in item 3 by combining these concepts. What if we do
**-C** is always the parent object and **-c** is always the child object.
And then figure out if they specified a handle or a file blob. We can figure this out
and allow a way for coercion.
```
0x1234 - Treated as handle
foo.dat - Treated as context file
file:0x1234 - Treated as file
```
This would collapse 4 options from tools into 2.

We also may want to consider getting the options consistent not for ALL the tools, but for the groups as laid out by:
https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf

You'll see them organized by "Session Commands", "Object Commands", etc. I think we should be consistent in groups.

Thoughts?

joshuagl:
Thanks for taking a look. I like all of your suggestions a lot. Keeping options consistent for groups of tools, as laid out in the commands spec, seems like a much more attainable goal and should certainly be the default position - nothing wrong with aiming higher than we can reach though.

I saw your patch land and have a WIP change to the wiki page to update the options following that change, will get it completed ASAP and start working on a proposal based on your suggestions above.

joshuagl:
[...]

PR #1008, opened 2018-04-24
progress on "object" class options

williamcroberts:
This PR helps to make progress towards #957 which rectifies the option conflicts in the tools.

The script isn't perfect. It uses sets, for the diff, so you lose the coupling of short/long option, however, if things differ, it will complain about at least one of them. It doesn't look outside of the option group to see if unknown/conflicting options are used in other groups outside of it's own. So you manually need to add it to it's ignore set.

The script however, does point out conflicting areas in groups. When resolving a conflict, the human needs to see if it should be ignored or modified based on more intelligence then I can whip up in a reasonable amount of time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changelog
 ### next
+  * tpm2_activatecredential: Option `--Password` changes to `--auth-key`.
   * system tests are now run with make check when --enable-unit is used in configure.
   * tpm2_unseal: Option `--pwdk` changes to `--auth-key`.
   * tpm2_sign: Option `--pwdk` changes to `--auth-key`.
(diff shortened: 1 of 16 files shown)


williamcroberts:
For the object class of tools it gets us to:
```
group: object:
	tools: ['tpm2_load', 'tpm2_createprimary', 'tpm2_unseal', 'tpm2_loadexternal', 'tpm2_makecredential', 'tpm2_readpublic', 'tpm2_activatecredential', 'tpm2_create', 'tpm2_changeauth']
	conflicts: set(['C', 'handle', 'H', 'object', 'c', 'item'])
```
Which I think will be handled by PR #982. item should probably go to object.

joshuagl:
Nice! I'll pull these into PR #982 when I submit the updated series tomorrow.

williamcroberts:
These are all on #982

PR #1009, opened 2018-04-24
tcti: Update mssim TCTI usage to use new key/value pair conf string.

flihp:
The TCTI library isn't using URIs for the config string any longer.
Instead we're using a string of key/value pairs with 'host' and
'port' as the keys.


williamcroberts:
Weird I thought I tested it with the TCTI changes, I must have screwed something up.

Issue #976, opened 2018-04-10
Objects created with tpm2-tools 3.X can't be used with tpm2-tools master


martinezjavier:
I know that we are breaking backward compatibility in the tools names, options, etc but I think that at the very least we should maintain backward compatibility for the created objects.

That is, keys created with a version of the tools should keep working with newer versions. If there's a bug on how the data is serialized in tpm2-tools 3.X, then I think we should fix that instead so is compatible with the upcoming 4.X release.

So I did this using the tpm2-tools 3.0.3 and tpm2-tss 1.4.0:

```
$ export TPM2TOOLS_TCTI_NAME=device
$ export TPM2TOOLS_DEVICE_FILE=/dev/tpmrm0

$ tpm2_createprimary -H o -g sha256 -G ecc -C primary.context
ObjectAttribute: 0x00030072 

CreatePrimary Succeed ! Handle: 0x80ffffff

$ tpm2_pcrlist -L sha1:0,1,2,3,4 -o pcr.bin
sha1 :
  0  : c72ec9e6cbc2b6a95f334dddd6513981da00f0c2
  1  : 2455942631ba0d81130c0b31aebfd1775e29fe8a
  2  : b2a83b0ebf2f8374299a5b2bdfc31ea955ad7236
  3  : b2a83b0ebf2f8374299a5b2bdfc31ea955ad7236
  4  : 678f6bb33400c4686099379791f0d5ad150b994f

$ tpm2_createpolicy -P -L sha1:0,1,2,3,4 -F pcr.bin -f pcr.policy

$ tpm2_create -g sha256 -G keyedhash -u obj.pub -r obj.priv -I- -c primary.context -L pcr.policy   -A 'sign|fixedtpm|fixedparent|sensitivedataorigin' <<< "secret"
algorithm:
  value: sha256
  raw: 0xb
attributes:
  value: fixedtpm|fixedparent|userwithauth
  raw: 0x52
type: 
  value: keyedhash
  raw: 0x8
  keyedhash: 77aebc7a94d986b67a86e3044c52da22437a45eeacdc40fb617d24ab9728c6e6
authorization policy: c972f4350dcd2bcd1393b54b0e1f310a0ab7244c5a4b2c4445fbaa1ff44a7795

$ tpm2_load -c primary.context  -u obj.pub -r obj.priv -n obj.name -C load.context

Load succ.
LoadedHandle: 0x80fffffe


[...]

PR #1014, opened 2018-04-26
Fix save and load for TPM2B_PRIVATE

williamcroberts:

### DIFF
--- a/tools/tpm2_create.c
+++ b/tools/tpm2_create.c
@@ -213,14 +213,14 @@ int create(TSS2_SYS_CONTEXT *sapi_context)
     tpm2_util_public_to_yaml(&outPublic);
 
     if (ctx.flags.o) {
-        bool res = files_save_public(&outPublic, ctx.opu_path);
+        bool res = files_save_private(&outPrivate, ctx.opr_path);
         if(!res) {
             return -3;
         }
     }
 
     if (ctx.flags.O) {
-        bool res = files_save_bytes_to_file(ctx.opr_path, outPrivate.t.buffer, outPrivate.t.size);
+        bool res = files_save_private(ctx.opr_path, &outPrivate);
         if (!res) {
             return -4;
         }
--- a/lib/files.c
+++ b/lib/files.c
@@ -586,7 +586,8 @@ bool files_load_bytes_from_file_or_stdin(const char *path, UINT16 *size, BYTE *b
         return files_save_bytes_to_file(path, buffer, offset); \
     }
 
-#define LOAD_TYPE(type, name) \
+#define LOAD_TYPE(type, name) LOAD_TYPE_FAIL(type, name, on_load_fail_default)
+#define LOAD_TYPE_FAIL(type, name, fail) \
     bool files_load_##name(const char *path, type *name) { \
     \
         UINT8 buffer[sizeof(*name)]; \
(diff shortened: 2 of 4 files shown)


martinezjavier:
@williamcroberts how did you test it? doesn't compile for me. The fix is trivial though:

```
diff --git a/tools/tpm2_create.c b/tools/tpm2_create.c
index d896e66cd6a2..41d7b428deed 100644
--- a/tools/tpm2_create.c
+++ b/tools/tpm2_create.c
@@ -220,7 +220,7 @@ int create(TSS2_SYS_CONTEXT *sapi_context)
     }
 
     if (ctx.flags.O) {
-        bool res = files_save_private(ctx.opr_path, &outPrivate);
+        bool res = files_save_private(&outPrivate, ctx.opr_path);
         if (!res) {
             return -4;
         }
```

It does fix the issue though since after storing the size field, I can load the key and unseal using master:

```
$ export TPM2TOOLS_TCTI=device:/dev/tpmrm0

$ tpm2_load -c primary.context  -u obj.pub -r obj.priv -n obj.name -C load.context                                                                                                
handle: 0x80fffffe
$ tpm2_unseal -c load.context -L sha1:0,1,2,3,4 -F pcr.bin
secret

```

martinezjavier:
@williamcroberts pleaes add a Fixes: #976 to your commit message.

tools/tpm2_create.c
```diff
@@ -220,7 +220,7 @@ int create(TSS2_SYS_CONTEXT *sapi_context)
     }
 
     if (ctx.flags.O) {
-        bool res = files_save_bytes_to_file(ctx.opr_path, outPrivate.t.buffer, outPrivate.t.size);
+        bool res = files_save_private(ctx.opr_path, &outPrivate);
```
martinezjavier: I would just include this part and the others that fix the issue and avoid the other bits that deal with the wrong size. This was clearly a bug so I would just fix it. Although the others changes cause no harm so up to you.

williamcroberts:
<snip>
> 
> > @@ -220,7 +220,7 @@ int create(TSS2_SYS_CONTEXT *sapi_context)
>      }
> 
>      if (ctx.flags.O) {
> -        bool res = files_save_bytes_to_file(ctx.opr_path, outPrivate.t.buffer,
> outPrivate.t.size);
> +        bool res = files_save_private(ctx.opr_path, &outPrivate);
> 
> @williamcroberts <https://github.com/williamcroberts>  right, sorry for
> explaining so badly. What I tried to say is that this line is what fixes the
> serialization issue since by also storing the size field of the structure.
> 
> And most of the changes in lib/files.c are to cope with loading an object of wrong
> size, and I said that I wouldn't include that.
> 

Oh, so what about a file saved via the broken version, the LOAD won't work, I thought
You didn't want to break that?

martinezjavier:
> Oh, so what about a file saved via the broken version, the LOAD won't work, I thought
> You didn't want to break that?

@williamcroberts  well, I thought that we shouldn't break backward compatibility on purpose for the saved files between major releases, but I think it's different if the store of the private portion was already broken. In other words, I think it's different fixing a serialization bug than changing how things are stored between releases.

Also, AFAICT no major distro has already shipped tpm2-tools 3.X on a release. Fedora 28 and openSUSE Leap 15.0 will be released soon with tpm2-tools 3.X and Ubuntu Bionic still ships the 2.X version. So I think it's still safe to just fix this bug without the need to do a conversion on load.

I remember we found others serialization bugs in the past and didn't add code to cope with that, but I don't have a strong position on this. As I said, the code causes no harm other than maybe having code that may not be used.

williamcroberts:
<snip

> 
> Also, AFAICT no major distro has already shipped tpm2-tools 3.X on a release.
> Fedora 28 and openSUSE Leap 15.0 will be released soon with tpm2-tools 3.X and
> Ubuntu Bionic still ships the 2.X version. So I think it's still safe to just fix this bug
> without the need to do a conversion on load.

SGTM

> 
> I remember we found others serialization bugs in the past and didn't add code to
> cope with that, but I don't have a strong position on this. As I said, the code
> causes no harm other than maybe having code that may not be used.
> 

Speaking of that, I think the next release should really be 3.X with these fixes, and
Then a 3.1 working on the newer version of the TSS code.

Do you want to spin these and become the stable branch guy?

martinezjavier:
> Speaking of that, I think the next release should really be 3.X with these fixes, and
> Then a 3.1 working on the newer version of the TSS code.

Not sure I understood correctly, I think we agree that a 3.0.4 release will be 3.0.3 + all the fixes in the 3.X branc (plus your TPM2B_PRIVATE save / load fix). 

And 3.1 will be a release from the master branch once tpm2-tss has a release?

> Do you want to spin these and become the stable branch guy?

Do you mean making a 3.0.4 tag, creating and signing the tarballs, etc? Sure thing, I can do the 3.0.4 on Monday. I can follow the RELEASE.md and ask you for assistance if something isn't clear.

williamcroberts:
> -----Original Message-----
> Sent: Friday, April 27, 2018 2:50 PM
> Subject: Re: [tpm2-software/tpm2-tools] Fix save and load for TPM2B_PRIVATE
> (#1014)
> 
> 	Speaking of that, I think the next release should really be 3.X with these
> fixes, and
> 	Then a 3.1 working on the newer version of the TSS code.
> 
> Not sure I understood correctly, I think we agree that a 3.0.4 release will be 3.0.3
> + all the fixes in the 3.X branc (plus your TPM2B_PRIVATE save / load fix).

3.0.4 will have all the patches on 3.X

The next 3.X release IMO, will be 3.1.0 re-worked ontop of the newest TSS version (master)
currently, but that will be getting RCs at some point.

> 
> And 3.1 will be a release from the master branch once tpm2-tss has a release?
> 
> 	Do you want to spin these and become the stable branch guy?
> 
> Do you mean making a 3.0.4 tag, creating and signing the tarballs, etc? Sure thing,
> I can do the 3.0.4 on Monday. I can follow the RELEASE.md and ask you for
> assistance if something isn't clear.

Yep.

martinezjavier:
I'm closing this pull request since it has been superseded by #1018 "[v2] tools: Fix save and load for TPM2B_PRIVATE".

Issue #971, opened 2018-04-02
Test infrastructure changes to exercise tests on TPM directly (or without RM)


idesai:
1. Identify tools and tests that can run on the TPM directly 
2. Test infrastructure changes to be able to create tests to run directly on the TPM.

flihp:
@liuqun provided a few patches to the tabrmd that used some of the the autotools parallel test harness features to accomplish this: https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html#Parallel-Test-Harness. It required some coordination across the autoconf and automake bits but it's pretty clean. Might be a good reference for this work.

martinezjavier:
@flihp I think you meant @jiazhang0? I just want to see if I'm looking at the right commit, are you referring to this one https://github.com/tpm2-software/tpm2-abrmd/commit/64fc43a7b2353c6789ab746370ca89cd8cd9d615 ?

flihp:
doh! Should have checked the history of https://github.com/tpm2-software/tpm2-abrmd/blob/master/scripts/int-hardware-setup.sh before posting. Thanks for the correction @martinezjavier, you got the commit right and everything.

williamcroberts:
Not doing this

PR #1015, opened 2018-04-26
test: fix no-abrmd testing option

williamcroberts:
One can specify no-abrmd as argument 1 to the start_up test
function to get a test env that uses the simulatory directly.


williamcroberts:
Fixes: #971

PR #1017, opened 2018-04-29
Adding support for importing RSA2K keys to tpm2_import tool

idesai:


test/integration/tests/import.sh
```diff
@@ -70,7 +73,29 @@ openssl enc -in plain.enc -out plain.dec.ssl -d -K `xxd -p sym.key` -iv 0 \
...
+
+openssl rsa -in private.pem -out public.pem -outform PEM -pubout
+openssl rsautl -encrypt -inkey public.pem -pubin -in plain.txt -out plain.rsa.enc
+
+tpm2_rsadecrypt -c import_rsa_key.ctx -I plain.rsa.enc -o plain.rsa.dec
+
+diff plain.txt plain.rsa.dec
+if [ $? != 0 ];then
+echo "TEST: tpm2_import failed for Asymmetric Key"
 exit 1
```
williamcroberts: You shouldn't need this if block. The tests have the trap ERR set, so on error that trap handler gets called and prints the line number of the failed command and exits 1.

tools/tpm2_import.c
```diff
@@ -288,9 +297,38 @@ static bool calc_sensitive_unique_data(void) {
...
+    (X).publicArea.objectAttributes |= TPMA_OBJECT_SIGN_ENCRYPT;\
+    (X).publicArea.objectAttributes &= ~TPMA_OBJECT_FIXEDTPM;\
+    (X).publicArea.objectAttributes &= ~TPMA_OBJECT_FIXEDPARENT;\
+    (X).publicArea.objectAttributes &= ~TPMA_OBJECT_SENSITIVEDATAORIGIN;\
+    (X).publicArea.authPolicy.size = 0;\
+    (X).publicArea.parameters.rsaDetail.symmetric.algorithm = TPM2_ALG_NULL;\
+    (X).publicArea.parameters.rsaDetail.scheme.scheme = TPM2_ALG_NULL;\
+    (X).publicArea.parameters.rsaDetail.keyBits = RSA_2K_MODULUS_SIZE_IN_BITS;\
+    (X).publicArea.parameters.rsaDetail.exponent = 0x0;\
+    (X).publicArea.unique.rsa.size = RSA_2K_MODULUS_SIZE_IN_BYTES;\
```
williamcroberts: These should probably be static inlines, but don't worry about that for now, since this patch follows the convention laid out.

test/integration/tests/import.sh
idesai: @williamcroberts IIUC I need to replace the commented block with "trap onerror ERR" or is it that, it is set by default so just remove line 97-99

williamcroberts: It's already set, so you just need diff. Ill fix this up for you today and get in the repo. You did enough here.

williamcroberts:
I'll also look into why this is failing for clang, and get an update to the travis build scripts to output the fail logs. This passes with clang + asan locally.

test/integration/tests/import.sh
idesai: hehe.. thanks 👍

danintel:
At least for me it helped to upgrade from clang 3.8 to to clang 5.0.  Ubuntu has package clang-5.0 that installs /usr/bin/clang-5.0 and friends, /usr/bin/*-5.0, etc (leaving /usr/bin/clang at 3.8).

williamcroberts:
merged on https://github.com/tpm2-software/tpm2-tools/pull/1020

PR #1018, opened 2018-04-30
[v2] tools: Fix save and load for TPM2B_PRIVATE

martinezjavier:
The tpm2_create tools is only storing the TPM2B_PRIVATE .buffer field but
not it's .size field. This causes the private keys stored with the stable
version of the tools, to fail loading when using the latest master branch.

Fix this by storing the whole TPM2B_PRIVATE object, including size field.

Fixes: #976

### DIFF
--- a/lib/files.c
+++ b/lib/files.c
@@ -606,6 +606,9 @@ bool files_load_bytes_from_file_or_stdin(const char *path, UINT16 *size, BYTE *b
         return rc == TPM_RC_SUCCESS; \
     }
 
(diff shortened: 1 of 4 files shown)


martinezjavier:
Changes since v1:

- Remove code to attempt to fix TPM2B_PRIVATE objects stored without their .size field.
- Fix build error due wrong parameters passed to files_save_private().
- Add Fixes tag
- Add commit message explaining the rationale for the change.

PR #1022, opened 2018-05-01
Import work

williamcroberts:
Updated: 5/1/2018 @ 11:43AM Pacific

Adds tests for import that:
* Sign with the tpm and verify with openssl
* Sign with openssl and verify with the tpm
  * Required -r raw support to be fixed
  * Required conversion routine from plain signature to tss

Additional Changes:
* Changes private key input to openssl PEM file and  supports encrypted private keys.
* Drops a dead function prototype as well
* Drops dead crufty code
* Update tpm2_verifysignature to support `-f --format` instead of `-r --raw`.
  * This is the generic plumbing for taking in a signing scheme, even though only rsassa is supported.
* Add more unit tests

Post-Merge TODO:
Update Changelog to include the tpm2_verifysignature option rename

### DIFF
--- a/lib/tpm2_convert.c
+++ b/lib/tpm2_convert.c
@@ -1,5 +1,6 @@
 //**********************************************************************;
 // Copyright (c) 2017, SUSE Linux GmbH
+// Copyright (c) 2018, Intel Corporation
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
@@ -211,6 +212,39 @@ bool tpm2_convert_sig(TPMT_SIGNATURE *signature, tpm2_convert_sig_fmt format, co
     }
     default:
         LOG_ERR("Unsupported signature output format.");
+        return false;
+    }
+}
+
+static bool sig_load(const char *path, TPMI_ALG_SIG_SCHEME sig_alg,
+        TPMI_ALG_HASH halg, TPMT_SIGNATURE *signature) {
+
+    signature->sigAlg = sig_alg;
+
+    switch (sig_alg) {
+        case TPM2_ALG_RSASSA:
+            signature->signature.rsassa.hash = halg;
+            signature->signature.rsassa.sig.size = sizeof(signature->signature.rsassa.sig.buffer);
+            bool res = files_load_bytes_from_path(path,
+                    signature->signature.rsassa.sig.buffer,
+                    &signature->signature.rsassa.sig.size);
+            return res;
+        default:
+            LOG_ERR("Unsupported signature input format.");
+            return false;
+    }
+}
+
+bool tpm2_convert_sig_load(const char *path, tpm2_convert_sig_fmt format, TPMI_ALG_SIG_SCHEME sig_alg,
+        TPMI_ALG_HASH halg, TPMT_SIGNATURE *signature) {
+
+    switch(format) {
+    case signature_format_tss:
+        return files_load_signature(path, signature);
+    case signature_format_plain:
+        return sig_load(path, sig_alg, halg, signature);
+    default:
+        LOG_ERR("Unsupported signature input format.");
         return false;
     }
 }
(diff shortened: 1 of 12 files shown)


lib/tpm2_convert.c
```diff
@@ -214,3 +214,38 @@ bool tpm2_convert_sig(TPMT_SIGNATURE *signature, tpm2_convert_sig_fmt format, co
...
+    }
+}
+
+bool tpm2_convert_sig_load(const char *path, tpm2_convert_sig_fmt format, TPMI_ALG_SIG_SCHEME sig_alg,
+        TPMI_ALG_HASH halg, TPMT_SIGNATURE *signature) {
+
+    switch(format) {
+    case signature_format_tss:
+        return files_load_signature(path, signature);
+    case signature_format_plain: {
```
williamcroberts: No brackets here

lib/tpm2_convert.c
```diff
@@ -214,3 +214,38 @@ bool tpm2_convert_sig(TPMT_SIGNATURE *signature, tpm2_convert_sig_fmt format, co
         return false;
     }
 }
+
+bool sig_load(const char *path, TPMI_ALG_SIG_SCHEME sig_alg,
```
williamcroberts: Static

lib/tpm2_convert.c
```diff
@@ -214,3 +214,38 @@ bool tpm2_convert_sig(TPMT_SIGNATURE *signature, tpm2_convert_sig_fmt format, co
...
+    case signature_format_tss:
+        return files_load_signature(path, signature);
+    case signature_format_plain: {
+        return sig_load(path, sig_alg, halg, signature);
+    }
+    default:
+        LOG_ERR("Unsupported signature output format.");
+        return false;
+    }
+
```
williamcroberts: No newline

tools/tpm2_verifysignature.c
```diff
@@ -153,7 +154,10 @@ static bool init(TSS2_SYS_CONTEXT *sapi_context) {
     }
 
     if (ctx.flags.sig) {
-        bool res =  files_load_signature(ctx.sig_file_path, &ctx.signature);
+
+        TPMI_ALG_SIG_SCHEME sig_alg = ctx.flags.raw ? TPM2_ALG_RSASSA : TPM2_ALG_ERROR;
+        tpm2_convert_sig_fmt fmt = ctx.flags.raw ? signature_format_plain : signature_format_tss;
+        bool res = tpm2_convert_sig_load(ctx.sig_file_path, fmt, sig_alg, ctx.halg, &ctx.signature);
```
williamcroberts: Ensure that ctx.halg is specified with -r.

Should we change -r to -f format and handle what they pass as an argument.

idesai: @williamcroberts, i think so too. Also may be we should call the fmt variable signature_format_ssl_rsa instead of signature_format_plain

idesai:
@williamcroberts, Agreed, we should add support in tpm2_convert for ASN1_parse(BIO *bp, const unsigned char *pp, long len, int indent); from https://github.com/openssl/openssl/blob/master/include/openssl/asn1.h

lib/tpm2_convert.c
```diff
@@ -214,3 +215,36 @@ bool tpm2_convert_sig(TPMT_SIGNATURE *signature, tpm2_convert_sig_fmt format, co
...
+bool tpm2_convert_sig_load(const char *path, tpm2_convert_sig_fmt format, TPMI_ALG_SIG_SCHEME sig_alg,
+        TPMI_ALG_HASH halg, TPMT_SIGNATURE *signature) {
+
+    switch(format) {
+    case signature_format_tss:
+        return files_load_signature(path, signature);
+    case signature_format_plain:
+        return sig_load(path, sig_alg, halg, signature);
+    default:
+        LOG_ERR("Unsupported signature output format.");
```
williamcroberts: input format

PR #1027, opened 2018-05-03
tests: Allow to run the integration tests using a real TPM device

martinezjavier:
Currently, the integration tests can't only be run either against in the
following configurations:

1) tpm2-abrmd + TPM simulator
2) the TPM simulator directly

But this doesn't cover some configurations that were previously supported:

3) tpm2-abrmd + the default TPM device (/dev/tpm0)
4) tpm2-abrmd + a TPM device that's not the default (i.e: /dev/tpm1)
5) a TPM device directly (i.e: /dev/tpmrm0)

This patch allows the integration tests to be run using all the possible
configurations, by setting TPM2_* environment variables instead of the
(unused) no-abrmd option.

For example:

1) $ TPM2_ABRMD=/path/to/tpm2-abrmd TPM2_SIM=/path/to/tpm_server ./tests/pcrlist.sh
2) $ TPM2_SIM=/path/to/tpm_server ./tests/pcrlist.sh
3) $ TPM2_ABRMD=/path/to/tpm2-abrmd ./tests/pcrlist.sh
4) $ TPM2_ABRMD=/path/to/tpm2-abrmd TPM2_DEVICE=/dev/tpm1 ./tests/pcrlist.sh
5) $ TPM2_DEVICE=/dev/tpmrm0 ./tests/pcrlist.sh

### DIFF
--- a/test/integration/helpers.sh
+++ b/test/integration/helpers.sh
@@ -226,9 +226,20 @@ function start_sim() {
 
 function start_abrmd() {
 
-	if [ $UID -eq 0 ]; then
-		tpm2_tabrmd_opts="--allow-root $tpm2_tabrmd_opts"
+    if [ -z "$TPM2_SIM" ]; then
+	local tcti="device"
+	if [ -n "$TPM2_DEVICE" ]; then
+	    tcti="device:$TPM2_DEVICE"
 	fi
+
+	local name="com.intel.tss2.Tabrmd.device"
+	tpm2_tabrmd_opts="--session --dbus-name=$name --tcti=$tcti"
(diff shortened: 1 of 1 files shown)


williamcroberts:
We planned on using "no-abrmd" when running in-progress code/tests for hmac auth. But @flihp has fixed the lingering session issues in tools/abrmd.

These env vars are nice for those who wish to run the test scripts outside of the harness, but most
of them were written without worry about how many objects were loaded and thus need an RM, but I guess if your running them manually, you know what your doing.

martinezjavier:
@williamcroberts yeah, the idea is to give all the flexibility that we used to have with the TPM2_TCTI* env vars.

PR #1028, opened 2018-05-03
tpm2_createprimary allow primary key to be created with sign bit

williamcroberts:
Allow a primary key to be created with the sign attribute

### DIFF
--- a/lib/tpm2_attr_util.c
+++ b/lib/tpm2_attr_util.c
@@ -495,6 +495,7 @@ static bool common_strtoattr(char *attribute_list, void *attrs, dispatch_table *
 
 bool tpm2_attr_util_nv_strtoattr(char *attribute_list, TPMA_NV *nvattrs) {
 
+    memset(nvattrs, 0, sizeof(*nvattrs));
     return common_strtoattr(attribute_list, nvattrs, nv_attr_table, ARRAY_LEN(nv_attr_table));
 }
 
--- a/lib/tpm2_attr_util.c
+++ b/lib/tpm2_attr_util.c
@@ -500,6 +500,7 @@ bool tpm2_attr_util_nv_strtoattr(char *attribute_list, TPMA_NV *nvattrs) {
 
 bool tpm2_attr_util_obj_strtoattr(char *attribute_list, TPMA_OBJECT *objattrs) {
 
+    memset(objattrs, 0, sizeof(*objattrs));
     return common_strtoattr(attribute_list, objattrs, obj_attr_table, ARRAY_LEN(obj_attr_table));
 }
 
--- a/tools/tpm2_createprimary.c
+++ b/tools/tpm2_createprimary.c
@@ -62,6 +62,11 @@ struct tpm_createprimary_ctx {
         UINT8 g :1;
         UINT8 G :1;
     } flags;
+
+    struct {
+        char *attrs;
+        char *alg;
+    } optargs;
 };
 
 static tpm_createprimary_ctx ctx = {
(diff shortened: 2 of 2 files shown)


tools/tpm2_createprimary.c
```diff
@@ -91,17 +96,20 @@ static bool set_name_alg(TPMI_ALG_HASH halg, TPM2B_PUBLIC *public) {
...
     case TPM2_ALG_RSA: {
         TPMS_RSA_PARMS *r = &public->publicArea.parameters.rsaDetail;
-       r->symmetric.algorithm = TPM2_ALG_AES;
-       r->symmetric.keyBits.aes = 128;
-       r->symmetric.mode.aes = TPM2_ALG_CFB;
-       r->scheme.scheme = TPM2_ALG_NULL;
-       r->keyBits = 2048;
-       r->exponent = 0;
-       public->publicArea.unique.rsa.size = 0;
+        if (public->publicArea.objectAttributes & TPMA_OBJECT_SIGN_ENCRYPT) {
```
williamcroberts: I have no idea what these algorithms are really for, but you can't have RSA + AES with the attributes on sign

lib/tpm2_attr_util.c
```diff
@@ -495,6 +495,7 @@ static bool common_strtoattr(char *attribute_list, void *attrs, dispatch_table *
 
 bool tpm2_attr_util_nv_strtoattr(char *attribute_list, TPMA_NV *nvattrs) {
 
+    memset(nvattrs, 0, sizeof(*nvattrs));
```
martinezjavier: @williamcroberts do you think these two first commits should also go to the 3.X stable branch?

tools/tpm2_createprimary.c
martinezjavier: @williamcroberts I'm also not completely sure what these algorithms are for. By reading the specs it seems that they are used to define the algorithm and scheme used that can be used by restricted keys and the symmetric algorithm used to decrypt a child key when loading to the TPM.

You can find information about it in:

`12.2 Public Area Structures` and `12.2.3.5 TPMS_RSA_PARMS` in `Trusted Platform Module Library Part 2: Structures` and `25 Object Attributes` in `Trusted Platform Module Library Part 1: Architecture`. 

Could you please mention in the commit message what's the goal of this change? What exactly did you try that didn't work, etc.

tools/tpm2_createprimary.c
```diff
@@ -91,17 +96,20 @@ static bool set_name_alg(TPMI_ALG_HASH halg, TPM2B_PUBLIC *public) {
...
+            r->symmetric.algorithm = TPM2_ALG_NULL;
```
martinezjavier: Do you also need `r->symmetric.scheme = TPM_ALG_NULL` here? `12.2.3.5 TPMS_RSA_PARMS` in `Trusted Platform Module Library Part 2: Structures` says:

```
NOTE   When both sign and decrypt are SET, restricted shall be CLEAR
       and scheme shall be TPM_ALG_NULL.
```

lib/tpm2_attr_util.c
williamcroberts: Probably

tools/tpm2_createprimary.c
williamcroberts: I tried to make a primary object I could use for signing and verifying:
This command wouldn't work: tpm2_createprimary -Q -G rsa -g sha256 -a o -C parent.ctx -A "fixedtpm|fixedparent|sensitivedataorigin|userwithauth|decrypt|sign"

Based on what you're saying above, maybe it's if not restricted.... but the check in the simulator was on sign.

tools/tpm2_createprimary.c
williamcroberts: I'm not sure. The simulator didn't kick back what I sent

tools/tpm2_createprimary.c
martinezjavier: @williamcroberts I could misunderstood the spec too. But it would be nice if you can add that information in the commit message (command executed, errors shown, etc) to have a better understanding on why the change is needed.

idesai:
@williamcroberts @martinezjavier this is my inference on the discussion here. 
Restricted keys are special, in that, they either allow at least signing or decryption but not both (in line with common practice of not having the same key for signing and decryption, it is simply enforced here with restricted keys). 
Additionally they operate only on TPM generated structures, which if sensitive, are further encrypted and integrity protected with HMAC. The secrets to these operations are derived from parent seed for child objects (like @martinezjavier mentioned) and is a one time random for duplication blobs like the one you can see in import tool. Anyway.. this is where the symmetric keys are used. 
Hence specifying that the key type is not restricted we are mandating, by specifying the symmetric algorithm is NULL, that it cannot protect TPM generated structures. Since there is no scheme of MAC or signing associated with NULL types, the parameter is ignored. @williamcroberts you can set the scheme to NULL as follow on to specifying the symmetric algorithm NULL.

idesai:
@williamcroberts so there are these structures
1. rsaDetail.scheme which should be TPM2_ALG_RSAPSS or TPM2_ALG_RSASSA if sign bit is set and it should be set to NULL if encrypt is also set for a restricted key.
2. rsaDetail.symmetric, there is a typo in the previous comments from @idesai (editing my previous comment) and @martinezjavier that was referring to r->symmetric.scheme; it instead is r->symmetric.algorithm, this should be NULL for keys with sign and encrypt set

PR #1043, opened 2018-05-17
3.X: update to TSS version 2.0

williamcroberts:


martinezjavier:
@williamcroberts I'll have time to review this next week but for now some general comments. We should probably create a new 3.1 branch for this to leave 3.X as a place to backport fixes and do new releases for people using tpm2-tools 3.0 + tss 1.4.

I also wonder who would be the target of this release, since I guess that either people would use the tpm2-tools 3.0 + tss 1.4 releases as early adopters or the upcoming tpm2-tools 4.0 + tss 2.0.

I'm not saying that's not useful, but just wondering if the cost of maintaining won't be bigger than the benefit.

williamcroberts:
@martinezjavier the way stable branches was going to work was have all 3.X releases on one branch and tag as you go. If you need a minor release, and the branch HEAD has moved past, then you can create a dedicated branch from the existing tag as you go.

4.0 may not be ready

martinezjavier:
> @martinezjavier the way stable branches was going to work was have all 3.X releases on one branch and tag as you go. If you need a minor release, and the branch HEAD has moved past, then you can create a dedicated branch from the existing tag as you go.

So you are saying that we should create a 3.0.x branch instead then? That would work too although it find that counter intuitive. If you are tracking a stable branch, then you would expect that the branch name / URI doesn't  change.

martinezjavier:
> 4.0 may not be ready

What's needed to cut a new release from the master branch instead? I see that we have two open issues with the `enhancement` label. Both related to have a consistent set of options between tools.

https://github.com/tpm2-software/tpm2-tools/labels/enhancement

williamcroberts:
https://github.com/tpm2-software/tpm2-tools/milestone/6

martinezjavier:
> https://github.com/tpm2-software/tpm2-tools/milestone/6

Thanks. Yes, there's a lot of stuff missing for 4.0...

I'm just worried about having such a big change and breaking API/ABI compatibility between (what would look like for most people) as a minor release. Another option is to call this 4.0 (tpm2-tools 3.X + tss 2.0) and make tpm2-tools master + tss 2.0 version 5.

I know that we are not breaking backward compatibility at the CLI level, but distros would have to bump TSS to a new major release if they want to update to 3.1.

martinezjavier:
@mgerstner what's your opinion on this since you also maintain tpm2-{tools,tss,abrmd} packages IIUC?

williamcroberts:
> So you are saying that we should create a 3.0.x branch instead then? That would work too, although I find it counter intuitive. If you are tracking a stable branch, then you would expect that the branch name / URI doesn't change.

The branch that the 3.0.X tags are on, is a branch called 3.X. Where the X can be anything. If you need a long running patch release branch, git checkout -b 3.0.X origin/3.0.2 && git push origin 3.0.X, then work off of that. We don't plan on supporting any-long running patch releases at the moment.

> I'm just worried about having such a big change and breaking API/ABI compatibility between (what would look like for most people) as a minor release. Another option is to call this 4.0 (tpm2-tools 3.X + tss 2.0) and make tpm2-tools master + tss 2.0 version 5.

> I know that we are not breaking backward compatibility at the CLI level, but distros would have to bump TSS to a new major release if they want to update to 3.1.

The project claims to be following semantic versioning, so we shouldn't stray. Your logic would mean that when libc get's a major bump, everyone linking to it should bump their major numbers.

martinezjavier:
> The branch that the 3.0.X tags are on, is a branch called 3.X. Where the X can be anything. If you need a

I disagree that 3.X can be **anything**, it already has some meaning for people since it's a **stable** branch. Which means that only bug fixes from master should be cherry-picked and no new features or changes that could cause regressions are made.

> long running patch release branch, git checkout -b 3.0.X origin/3.0.2 && git push origin 3.0.X, then work off of that. We don't plan on supporting any-long running patch releases at the moment.

Sure, as an upstream maintainer you are free to only care about the master branch and not maintain stable branches. But people can volunteer to cherry-pick the fixes from the master branch and do new releases as long as they are interested. I don't see why we would do more difficult for these people to maintain the stable branches by doing big changes that clearly not belong to stable branches.

> The project claims to be following semantic versioning, so we shouldn't stray. Your logic would mean

Where such a claim? The project [RELEASE.md](https://github.com/tpm2-software/tpm2-tools/blob/master/RELEASE.md) says:

```
The general release process will be to fork a branch at each major release followed by
a ".X". For example, the major version 2 branches will be 2.X
```

You are clearly doing a major release here since a minor release would only be about bugfixes in a stable branch, so following the RELEASE.md document this should have its own branch (probably 4.X).

> when libc get's a major bump, everyone linking to it should bump their major numbers.

I think is more constructive if we keep the conversations without taking examples to the extreme. But yes, I wouldn't do a change in a stable branch that requires a new version of glibc (or any other library).

Now since you mentioned glibc, this is what its [release process policy (https://sourceware.org/glibc/wiki/Release) says:

```
Normally, a release branch is forked off master after the stable version is tagged
(usually, commits intended to update the version appear for few days after the
version is tagged and the release branch would diverge only right before version.h
is updated to reflect start of development for next main version), and consists purely
or almost purely of cherry-picked fixes committed to the master branch.

Each branch is maintained by a single release manager, who has the responsibility
of maintaining the particular stable branch, has final say on what goes in. The branch
is expected to receive many liberally merged bugfixes from master at the beginning
when many new bugs are found, then become more conservative and decrease the
rate of back-ports as most major problems are solved and the branch is being
propagated to more mature distribution versions. Usually, the interested committers
have discretion over which bugfixes to pick for back-porting, but if discussion arises,
general consensus of the community is sought, the default choice being to err on the
conservative side. Patch backports to stable branches are discussed on libc-stable,
and any patch on master that doesn't change ABI or API is immediately suitable for
backporting to a stable branch. It is polite to post your backport to libc-stable (in
parallel with your commit) and explain the reason for the backport; this helps distro
maintainers decide if they want to sync to get the fix.
```
Which is the same process I say we should follow.

williamcroberts:
`I disagree that 3.X can be anything, it already has some meaning for people since it's a stable branch. Which means that only bug fixes from master should be cherry-picked and no new features or changes that could cause regressions are made.

X can be anything, and by virtue of semantic versioning, we get the exact property you desire here. If we add a new, backwards compatible feature the minor version get's bumped, if we make a bug fix, the patch level goes up.

>Sure, as an upstream maintainer you are free to only care about the master branch and not maintain stable branches. But people can volunteer to cherry-pick the fixes from the master branch and do new releases as long as they are interested. I don't see why we would do more difficult for these people to maintain the stable branches by doing big changes that clearly not belong to stable branches.

 We haven't had the need for long running patch level only release branches, which we can create trivially if needed. We have had the need for long running stable branches, which we have, where minor and patch level changes can be done on the `<major>.X` branch.

> Where is such a claim? The project RELEASE.md says:
```
$ grep -rn semantic
RELEASE.md:23:Releases shall be tagged following semantic version guidelines found at:
```
"Releases shall be tagged following semantic version guidelines found at: http://semver.org/"

# I think is more constructive if we keep the conversations without taking examples to the extreme. But yes, I wouldn't do a change in a stable branch that requires a new version of glibc (or any other library).

This is not the extreme, this is the exact scenario that is occuring. Just because libsapi is changing their version number, which requires internal code changes, doesn't change our version number as it does not affect any exposed API in a non-backwards compatible fashion.

On glibc's release process, that's pretty much the process we're doing. It is very close.
We branch off of master, with a branch called `<major>.X`, and tag when ready after going through the RC process. We then take changes on the stable branch that don't break backwards compatibility, ie minor and patch level changes only.


`

martinezjavier:
> "Releases shall be tagged following semantic version guidelines found at: http://semver.org/"

Sorry, I missed that. Thanks a lot for pointing out.

So I think we agree on the process but disagree on what constitutes a backwards-compatible change. I think is OK to add new functionality as a minor change but I don't think that a change to the internal dependency is minor, even when there are no user visible changes.

@williamcroberts I know that the text in semver says the same than you, but I see several issues and open PR in the semver Github repo discussing this, so is not that everyone agrees with the text.

It seems to me that semver is more focused to environments where dependencies are static linked into the binaries or when each program pulls their own dependencies, but it doesn't work so well when shared libraries are used since in that case transitive dependency matters in my opinion. As mentioned distros usually don't update software components between major versions during the lifecycle of a release, so that means that a distro that has tpm2-tools 3.0.X may not able to update to the 3.1.X release. That doesn't seem correct to me since it should just be a minor backward compatible release.

Anyway, maybe I'm the one who's wrong on this. But I would like to know the opinions of others before this PR is merged.

flihp:
I was just reading through https://github.com/semver/semver/issues/148 to get my head around the disagreement here.

williamcroberts:
> Anyway, maybe I'm the one who's wrong on this. But I would like to know the opinions of others before this PR is merged.

You're wrong about being wrong :-P. This is an issue that needs to have a defined, repeatable process around it. The assumption that you want, which may not always hold, is that a stable release of 3.X should always work against the target branch of libsapi that it initially was forked to support. I think that would be an amazing property, as it keeps the dependency hunt minimal.

Do we go wild here:
1. ifdef the code? Id like to avoid this as the maintenance burden is high.
2. Do something else with minor version?
3. Do a major release and say, we do semver + we major bump if any TSS dependency has a major bump?

We just need to document the practice, and uphold it. I'm just trying to uphold the current terms that we laid out for users.

martinezjavier:
> I was just reading through semver/semver#148 to get my head around the disagreement here.

@flihp I bevieve https://github.com/semver/semver/issues/341 better reflect what's being discussed here. For example, tpm2-tss depends on uriparser and libgcrypt. Would you consider using symbols from a new major release of these and bump the dependency requirements on a stable branch?

@williamcroberts yes, my assumption is that a major tpm2-tss release will always be based on the same major tpm2-tss release. Since tpm2-tss libraries also follow the version semantics from semver, this means that the tpm2-tools dependency can always be updated.

> 3. Do a major release and say, we do semver + we major bump if any TSS dependency has a major bump?

I believe this is the best option.

williamcroberts:
Whatever we do should be consistent between the projects.

williamcroberts:
@martinezjavier we need to define the line, what about our dependency on OpenSSL, Python (for testing), LibC, etc. Do we just limit it to TSS?

flihp:
Just to be specific we're not talking about only major changes in the tpm2-tss libraries but in *any* dependency right?

williamcroberts:
> Just to be specific we're not talking about only major changes in the tpm2-tss libraries but in any dependency right?
Yes, and AFIAK specifically a major version bump of a dependency.

martinezjavier:
@williamcroberts @flihp yes, I believe it applies to any dependency.

In fact, I even disagree with [semver](https://semver.org/) in that MINOR releases that `add functionality in a backwards-compatible manner` are suitable for a stable branch, since it may introduce new bugs. But that's not a hill I'm willing to die on.

flihp:
> For example, tpm2-tss depends on uriparser and libgcrypt. Would you consider using symbols from a > new major release of these and bump the dependency requirements on a stable branch?

I would consider this, but that doesn't mean it's the right thing to do. The ambiguity in semver is what we need to iron out here and since it will affect the other tss2 repos we should probably include the other maintainers.

williamcroberts:
> -----Original Message-----
> Sent: Monday, May 21, 2018 3:06 PM
> Subject: Re: [tpm2-software/tpm2-tools] 3.X: update to TSS version 2.0 (#1043)
> 
> @williamcroberts <https://github.com/williamcroberts>  @flihp
> <https://github.com/flihp>  yes, I believe it applies to any dependency.
> 
> In fact, I even disagree with semver <https://semver.org/>  in that MINOR
> releases that add functionality in a backwards-compatible manner are suitable for
> a stable branch, since it may introduce new bugs. 

Bug fixes can introduce bugs. Looking at the TPM in the wrong way can introduce a bug...

But that's not a hill I'm willing to

martinezjavier:
> Bug fixes can introduce bugs. Looking at the TPM in the wrong way can introduce a bug...

Indeed, it's just that new features usually add more code than bug fixes so are more likely to be affected. But yes, you are correct on this and please ignore my comment since is unrelated to our current discussion :smiley:

mgerstner:
> @mgerstner what's your opinion on this since you also maintain tpm2-{tools,tss,abrmd} packages IIUC?

Sorry for the late reply, I was on vacation.

Well this is a long discussion by now and I hope I followed through correctly even without diving into the semantic versioning specification. It looks like once you give the version number a meaning, everybody has a different opinion of it ;-)

As an end user of the tpm2 projects I would find it nice to quickly identify which versions of them work correctly together. Given that each project is versioned more or less independently this is already out of the question, I guess.

Looking at tpm2-tools as an isolated project it sounds like a minor version bump should be enough. Since the tpm2 projects are closely related I have at the same time the feeling that it is more than a minor change when a new major version of the tss is required. I don't think this would be the case if only a new version of, say, libcurl would be required.

The safe and clearly defined way is surely to look at each project independently i.e. making this only a minor version bump for tpm2-tools. Even if it might feel wrong from some people.

williamcroberts:
Please see the version matrix:
https://github.com/tpm2-software/tpm2-tools/wiki/Dependency-Matrix#dependency-version-matrix

PR #1055, opened 2018-06-06
ci: use latest, 2.0.0_rc3, tpm2-tss RC

joshuagl:


joshuagl:
Probably makes more sense to keep tracking master of the TSS & RM for the master branch of the tools.

Issue #1057, opened 2018-06-06
dlopen()'ing raw .so files considered bad


twoerner:
I've built the latest tpm2 software:
- tpm2-tss @ https://github.com/tpm2-software/tpm2-tss/commit/fafc92108e40374457d4de144d2b6a4b7735c6ed
- tpm2-abrmd @ https://github.com/tpm2-software/tpm2-abrmd/commit/7285806ad83a8b5bc0bdd8a3fef1e7cc0882ddf1
- tpm2-tools @ https://github.com/tpm2-software/tpm2-tools/commit/6b257d1e63d4d32999d4af23d3dc4dc5d887fd77

and I'm still seeing:

        root@raspberrypi3:~# tpm2_pcrlist -v
        tool="tpm2_pcrlist" version="3.0.2-535-g6b257d1" tctis="dynamic" tcti-default=tabrmd dlclose=enabled

        root@raspberrypi3:~# tpm2_pcrlist   
        ERROR: Could not dlopen library: "libtss2-tcti-tabrmd.so"
        ERROR: Could not load tcti, got: "tabrmd"

        root@raspberrypi3:~# tpm2_pcrlist -T device:/dev/tpm0
        ERROR: Could not dlopen library: "libtss2-tcti-device.so"
        ERROR: Could not load tcti, got: "device"

From what I can tell, patches were added to tpm2-abrmd and tpm2-tools to support `dlopen()`'ing versioned `*.so` files (i.e. `*.so.0`), maybe something was missed or needs to be added so the entire TPM2 stack supports it?

twoerner:
Please see https://lists.01.org/pipermail/tpm2/2018-May/000700.html for more details.

flihp:
Hi @twoerner. Are you seeing the same issue with tabrmd? The 1 line change there:
https://github.com/tpm2-software/tpm2-abrmd/commit/d2f0b2d1f0f25f1ef466c88484160ed8edc975f6
was identical to the change made in the tools by @williamcroberts here:
https://github.com/tpm2-software/tpm2-tools/commit/e1fc8d4dda9ee6e64884e475bb4f3401d127c91d

According to the hash you've provided for the tools repo you're using e1fc8d4dd should be in your tree so maybe just the error messages haven't been updated?

williamcroberts:
@twoerner I see the problem:
I updated it here:
[...]

PR #1058, opened 2018-06-08
tcti_ldr: load TCTI's by SONAME, not raw .so filename

joshuagl:
The raw .so file always points to the latest shared library object and is
usually split into the development package in distributions.

Factor TCTI dlopen'ing into a shared function that loads the TCTI by
versioned SONAME.

Resolves issue #1057


joshuagl:
The original Travis run failed to clone from GitHub, retriggering.

PR #1063, opened 2018-06-13
tpm2_create and tpm2_createprimary: move to common pub init

williamcroberts:
The public structure for a tpm object needs many values set
when creating an object. Start to move away from the hardcodes,
while preserving the old defaults, and allow for more complex
algorithm specifications.

For instance to create an AES256 key, mode CBC:
tpm2_create -C context.out -G 'symcipher:symalg=aes,symkeybits=256,modesym=cbc' -u key.pub -r key.priv

TODO:
- [x] Tests
- [x] Docs
- [x] Create a ticket to come back and complete:
   a. Setting all the subordinate fields
   b. YAML outputting the right data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Changelog
 ### next
+  * tpm2_hash: -g changes to -G
   * tpm2_encryptdecrypt: Support IVs via -i and algorithm modes via -G.
   * tpm2_hmac: drop -g, just use the algorithm associated with the object.
   * tpm2_getmanufec: -g changes to -G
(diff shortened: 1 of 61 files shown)


williamcroberts:
I don't like this, I think we need to go to something like:

AES_256_CBC, AES_128_OCB, etc....

williamcroberts:
We can see how brittle tpm2_import still is, not sure why it's failing, but it's in the inner integrity check, we can just open another bug against import if we take this before figuring out the problem. Import has been a thorn in my side!

williamcroberts:
Also, look at tpm2_encryptdecrypt. It takes a mode, but it's set to tpm2_alg_null. IIUC, this means use the mode as specified. It would be nice to have a mode option, and be able to create aes256:null objects. Which the default should probably go to that, This way you can use the differing modes with the default aes key.

tpm2_encryptdecrypt also needs an IV input, to support the other modes.

williamcroberts:
Probably fuzz this with AFL...

Issue #1039, opened 2018-05-16
capital options for parent, lowercase for object


williamcroberts:
Parent auths and options should be uppercase, current object should be lowercase.

Noticed so far:
* -c/-C and -K on tpm2_certify
* -K on tpm2_createprimary

Possibly change:
* -e for hierarchy auth for tpm2_activatecredential
* -e for hierarchy auth for tpm2_createak

PR #1065, opened 2018-06-14
Capital options for parent, lower for current object

joshuagl:
Fix issues noted in #1039 where the incorrect case is used for short options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Changelog
 ### next
   * tpm2_activatecredential: -e becomes -E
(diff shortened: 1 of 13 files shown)


joshuagl:
Original PR failed integration checks. It turns out there were still tests using the **-g** option to `tpm2_quote` even though it was idempotent. My final commit in this series removes g from the shortopts array but didn't update the integration tests. These have now been updated, I spent some time reading about PCRs but I'm not convinced the updated integration tests are effective and testing the same thing. I need to spend some time properly learning about this part of TPM.

PR #1066, opened 2018-06-14
Reduce need for duplicate args in tpm2_nvwrite/tpm2_nvread

### LINKED ISSUE #941: nv read and write default the auth index
Default -a to the index so commands like this don't need to specify `-x` this:
```
tpm2_nvwrite -Q -x 0x1500015 -a 0x1500015 -P "index" nv.test_w
tpm2_nvread -Q -x 0x1500015 -a 0x1500015 -P "index"
```
This might mean that having "owner" as the default doesn't make sense here? Not sure

With nv commands you can authorize against:
1. The owner hierarchy
2. The platform hierarchy
3. The nv index itself

joshuagl:
Update tpm2_nvwrite and tpm2_nvread such that when -x is specified and -a is empty we try and authorise using the NV_INDEX value passed to -x. Fixes #941

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 ### next
+  * tpm2_nvwrite and tpm2_nvread: when -P is "index" -a is optional and defaults to
+    the NV_INDEX value passed to -x.
+  * Load TCTI's by SONAME, not raw .so file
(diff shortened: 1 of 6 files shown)


williamcroberts:
Moved and rebased ontop with conflict fixed...
https://github.com/tpm2-software/tpm2-tools/pull/1069

PR #1068, opened 2018-06-15
add bash completion support for all the tools 

kaiserkyu:
This is the bash completion support for the available tools, what i did was take advantage of the short summary that the tools print when you invoked them without options, this was a problem because some tools got defined actions without options, so i added optional arguments (summary and manpage) for the -h and --help options in order to force the printing of this short summary, this output generate the options for each tool. For now it completes all the available options and some arguments for them, any suggestion for improvements or other features will be helpful.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -22,6 +22,10 @@ PKG_CHECK_MODULES([TSS2_SYS], [tss2-sys >= 2.0.0])
 PKG_CHECK_MODULES([CRYPTO], [libcrypto >= 1.0.2g])
 PKG_CHECK_MODULES([CURL], [libcurl])
 
+PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], ,
+  bashcompdir="${sysconfdir}/bash_completion.d")
+AC_SUBST(bashcompdir)
+
 AC_ARG_ENABLE([unit],
             [AS_HELP_STRING([--enable-unit],
                             [build cmocka unit tests (default is no)])],
--- a/dist/bash-completion/tpm2-tools/tpm2_completion.bash
+++ b/dist/bash-completion/tpm2-tools/tpm2_completion.bash
@@ -1,0 +1,153 @@
+#!/bin/sh
+# bash completion for tmp2-tools                                 -*- shell-script -*-
+
+
+_tpm2_tools()
+{
+
+
+
+    local cur prev words cword
+    _init_completion || return
+    local common_options=(-h --help -v --version -V --verbose -Q --quiet -Z --enable-errata -T --tcti=)
+    local aux1=$( ${COMP_WORDS[0]} -h no-man 2>/dev/null )
+    local aux2=$( echo "${aux1}" | tr "[]|" " " | awk '{if(NR>2)print}' | tr " " "\n")
+    suggestions=("${aux2[@]}" "${common_options[@]}") #generate all the opts for the tool
+    local halg=(0x4 sha1 0xB sha256 0xC sha384 0xD sha512 0x12 sm3_256)
+    local public_object_alg=(0x1 rsa 0x8 keyedhash 0x23 ecc 0x25 symcipher)
+    local signing_alg=(0x5 hmac 0x14 rsassa 0x16 rsapss 0x18 ecdsa 0x1A ecdaa 0x1B sm2 0x1C ecschnorr)
+    local signing_schemes=(0x5 hmac 0x14 rsassa 0x15 rsaes 0x16 rsapss 0x17 oeap)
+    local tcti_opts=(device: mssim: abrmd:)
+
+
+    case $prev in
+      -g)
+          if [[ "${COMP_WORDS[0]}" != "tpm2_createek" && "${COMP_WORDS[0]}" != "tpm2_getmanufec" && "${COMP_WORDS[0]}" != "tpm2_createak" ]]; then
+            suggestions=( $( compgen -W "${halg[*]}" -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          else
+            suggestions=( $( compgen -W "${public_object_alg[*]}" -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          fi
+          return;;
+      -G)
+          if [[  "${COMP_WORDS[0]}" != "tpm2_import"  && "${COMP_WORDS[0]}" != "tpm2_quote" ]]; then
+            suggestions=( $( compgen -W "${public_object_alg[*]}" -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          else
+            suggestions=( $( compgen -W "${halg[*]}" -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          fi
+          return;;
+      -D)
+          if [[  "${COMP_WORDS[0]}" == "tpm2_createak" ]]; then
+            suggestions=( $( compgen -W "${halg[*]}" -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          else
+            _filedir
+          fi
+          return;;
+
+      -h | --help)
+          suggestions=( $( compgen -W 'summary manpage' -- "$cur" ) )
+          COMPREPLY=("${suggestions[@]}")
+          return;;
+      -T)
+          suggestions=( $( compgen -W "${tcti_opts[*]}" -- "$cur" ) )
+          COMPREPLY=("${suggestions[@]}")
+          [[ $COMPREPLY == *: ]] && compopt -o nospace
+          return;;
+      -f)
+          if [[ "${COMP_WORDS[0]}" == "tpm2_activatecredential" || "${COMP_WORDS[0]}" == "print" ]]; then
+            _filedir
+          elif [[ "${COMP_WORDS[0]}" == "tpm2_quote" || "${COMP_WORDS[0]}" == "tpm2_sign" ]]; then
+            suggestions=( $( compgen -W 'plain tss' -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          elif [[ "${COMP_WORDS[0]}" == "tpm2_verifysignature" ]]; then
+            suggestions=( $( compgen -W "${signing_schemes[*]}" -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          fi
+          return;;
+      -s)
+          if [[ "${COMP_WORDS[0]}" == "tpm2_createak" ]]; then
+            suggestions=( $( compgen -W "${signing_alg[*]}" -- "$cur" ) )
+            COMPREPLY=("${suggestions[@]}")
+          elif [["${COMP_WORDS[0]}" == "tpm2_verifysignature" ]];then
+            _filedir
+          fi
+          return;;
+      -u | -r)
+          if [[ "${COMP_WORDS[0]}" == "tpm2_load" || "${COMP_WORDS[0]}" == "tpm2_loadexternal" ]]; then
+            _filedir
+          fi
+          return;;
+      -I)
+          if [[ "${COMP_WORDS[0]}" != "tpm2_nvdefine" ]]; then
+            _filedir
+          fi
+          return;;
+      -k | -K)
+          if [[ "${COMP_WORDS[0]}" == "tpm2_import" ]]; then
+            _filedir
+          fi
+          return;;
+      -i)
+          if [[ "${COMP_WORDS[0]}" == "tpm2_send" ]]; then
+            _filedir
+          fi
+          return;;
+      -m)
+          if [[ "${COMP_WORDS[0]}" != "tpm2_quote" ]]; then
+            _filedir
+          fi
+          return;;
+      -t)
+          if [[ "${COMP_WORDS[0]}" == "tpm2_sign" ]]; then
+            _filedir
+          fi
+          return;;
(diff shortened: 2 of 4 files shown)


lib/tpm2_options.c
```diff
@@ -267,35 +269,24 @@ static void show_version (const char *name) {
...
     }
 
-    command_copy = strdup(command);
-    printf("Usage: %s%s%s\n", basename(command_copy),
-           tool_opts->callbacks.on_opt ? " [<options>]" : "",
-           tool_opts->callbacks.on_arg ? " <arguments>" : "");
-    free(command_copy);
+    printf("usage: %s%s%s\n", command,
+           tool_opts->callbacks.on_opt ? " [OPTIONS]" : "",
+           tool_opts->callbacks.on_arg ? " ARGUMENTS" : "");
```
williamcroberts: Why did this change? Is this the proper output syntax?

lib/tpm2_options.c
```diff
@@ -424,13 +429,25 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv,
...
 
     if (show_help) {
-        bool did_manpager = execute_man(argv[0]);
-        if (!did_manpager) {
+        bool did_manpager = true;
+        bool is_manpage = true;
+        bool is_short = true;
+        if (help_arg) {
+            is_manpage = strcmp(help_arg,"manpage");
+            is_short = strcmp(help_arg,"summary");
```
williamcroberts: isn't `--help=summary` and `--help-manpage` mutually exclusive?

You're already doing the strcmp on line 354, maybe just set a value for show_help and ditch help_arg:
0 - no help
1 - manpage
2 - summary

default:
show_help = 0

Set it:
show_help |= !strcmp(argv[optind],"manpage") << 0
show_help |= !strcmp(argv[optind],"summary") << 1

lib/tpm2_options.c
```diff
@@ -424,13 +429,25 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv,
...
+        }
+        if (!is_manpage || help_arg == NULL) {
```
williamcroberts: if it's not manpage, execute the manpager? Seems wrong....

williamcroberts:
I saw two issues:
The first one:
I did:
```bash
cp  dist/bash-completion/tpm2-tools/tpm2_completion.bash ~/.bash_completion
```
Re-started the shell and got the below errors:
```
dirname: missing operand
Try 'dirname --help' for more information.
find: ‘./.local/lib’: Permission denied
```
The second one:
It hangs when trying to tab-complete if the TCTI is not started. I think it's because I didn't have abrmd started and it was waiting. You'll need to SKIP tcti initialization if I am right.

lib/tpm2_options.c
kaiserkyu: thanks for the review, i didn't changue that, maybe the branch wasn't update, i'll fix that

lib/tpm2_options.c
kaiserkyu: yeah, it seems wrong, but the idea was to leave the manpage as default, if the -h or --help are invoked with no arguments, then the manpage has the priority but now that you mention maybe the argument manpage is not necessary

kaiserkyu:
@williamcroberts i'll be back with the fixups for the issues you mentioned, thanks for the review 👍

kaiserkyu:
@williamcroberts i have solve the issues you pointed, but still unclear the permissions issues you had when trying to use the completion, please let me know if it happens again

williamcroberts:
@kaiserkyu I'll look tomorrow and see if I can get it working. Thanks!!

williamcroberts:
@kaiserkyu tooka look running it:
on current master `tpm2_pcrlist -h` tries to output help.

with your patch: `tpm2_pcrlist -h` runs the actual tool.

We can't break -h. Although tpm2_pcrlist complains about no man page installed and doesn't even try and run the built-in help output (seems like a bug).

Also, I got no output with it installed and attempting to tab comple. I actually installed via make install and fired up a new shell to test.

kaiserkyu:
@williamcroberts sorry for the problem with the -h output, when show_help is 0 (show_help=1 for summary arg and show_help=2 for manpage arg) it skip the help output because my if condition is wrong, i didn't notice that i left `if (show_help)` so the entire code block is skipped

williamcroberts:
Ahh. I thought --help=no-man would dump what you need for tpm2_pcrlist. I'll take. Look soon, can you roll patches for any option changes you need against current master?

kaiserkyu:
@williamcroberts Ok, i will apply the changes on current master 👍

kaiserkyu:
@williamcroberts I've applied the changes you made last time, i added a few fixes for the behavior of -h/--help arguments, now is working fine. please let me know if you can make it work

williamcroberts:
@kaiserkyu I'll look at this soon. Thanks. @martinezjavier any chance you could give it a spin if you have time? It might take me a few days to get to it.

martinezjavier:
@williamcroberts @kaiserkyu sure thing, I'll do it tomorrow.

martinezjavier:
@kaiserkyu I've tested this today. In general it works pretty well for me, so nice work!

I found a couple of things though, I'll comment in-line in your patches where appropriate.

configure.ac
```diff
@@ -22,6 +22,10 @@ PKG_CHECK_MODULES([TSS2_SYS], [tss2-sys >= 2.0.0])
 PKG_CHECK_MODULES([CRYPTO], [libcrypto >= 1.0.2g])
 PKG_CHECK_MODULES([CURL], [libcurl])
 
+PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], ,
+  bashcompdir="${sysconfdir}/bash_completion.d")
```
martinezjavier: On Fedora this sets `bashcompdir` to `/usr/share/bash-completion/completions/`. I also thought that was the correct path but when testing your patch it didn't work.

By looking at the code in `/usr/share/bash-completion/bash_completion`, it seems that `__load_completion()` is called lazily when completion has to be made for a command line tool. Now, this searches for a completion script in `/usr/share/bash-completion/completions/$cmd` or `/usr/share/bash-completion/completions/$cmd.bash`.

For example for `tpm2_create` it will try to source a `/usr/share/bash-completion/completions/tpm2_create` or `/usr/share/bash-completion/completions/tpm2_create.bash`.

So it seems that its designed to have a completion that matches the command name. This would work if we had subcommands (i.e: tpm2 create, tpm2 load, etc) but it won't work in our case since every command is named differently.

Instead, for "global" completion that covers different commands, the `/etc/bash_completion.d` directory should be used instead. I see that the files there are sourced unconditionally.

So I think your `configure.ac` snippet should instead be:

```
PKG_CHECK_VAR(bashcompdir, [bash-completion], [compatdir], ,
  bashcompdir="${sysconfdir}/bash_completion.d")
AC_SUBST(bashcompdir)
```
In other words, using `compatdir` instead of `completionsdir` by looking at `/usr/share/pkgconfig/bash-completion.pc`:

```
prefix=/usr
compatdir=/etc/bash_completion.d
completionsdir=${prefix}/share/bash-completion/completions
helpersdir=${prefix}/share/bash-completion/helpers

Name: bash-completion
Description: programmable completion for the bash shell
URL: https://github.com/scop/bash-completion
Version: 2.8
```

@kaiserkyu I'm not a bash completion expert though so please let me know if I got something wrong.

dist/bash-completion/tpm2-tools/tpm2_completion.bash
```diff
@@ -0,0 +1,153 @@
...
+         return;;
+
+    esac
+
+    if [[ "$cur" == -* ]]; then #start completion
+        _exclude_completed_opts
+        COMPREPLY=( $( compgen -W '$( echo ${suggestions[@]//<value>} )' -- "$cur" ) )
+        [[ $COMPREPLY == *= ]] && compopt -o nospace
+        return
+    fi
```
martinezjavier: I would also add here:

```
 COMPREPLY=( $( compgen -W '$( echo ${suggestions[@]//<value>} )' -- "$cur" ) )
```

Because when I tested, I typed `tpm2_create` then a space and then pressed a tab but nothing happened. I had to look at your code to understand that I first had to type `-` for the options to be listed. That makes sense for someone who knows the tool but someone using for the first type wouldn't know if there's a subcommand or just options.

Adding the mentioned line would complete a `-` for the user.

dist/bash-completion/tpm2-tools/tpm2_completion.bash
```diff
@@ -0,0 +1,153 @@
...
+    suggestions=("${aux2[@]}" "${common_options[@]}") #generate all the opts for the tool
+    local halg=(0x4 sha1 0xB sha256 0xC sha384 0xD sha512 0x12 sm3_256)
+    local public_object_alg=(0x1 rsa 0x8 keyedhash 0x23 ecc 0x25 symcipher)
+    local signing_alg=(0x5 hmac 0x14 rsassa 0x16 rsapss 0x18 ecdsa 0x1A ecdaa 0x1B sm2 0x1C ecschnorr)
+    local signing_schemes=(0x5 hmac 0x14 rsassa 0x15 rsaes 0x16 rsapss 0x17 oeap)
+    local tcti_opts=(device: mssim: abrmd:)
+
+
+    case $prev in
+      -g)
```
martinezjavier: I see that `tpm2_create -g` and tab completes the algorithms which is great. Do you know if it would be possible to also do the same for the long option? For example `tpm2_create --halg=` make to autocomplete?

If you don't know it's OK, that can always be added later, I just thought it would be nice to support both short and long options.

dist/bash-completion/tpm2-tools/tpm2_completion.bash
```diff
@@ -0,0 +1,153 @@
...
+
+
+
+    local cur prev words cword
+    _init_completion || return
+    local common_options=(-h --help -v --version -V --verbose -Q --quiet -Z --enable-errata -T --tcti=)
+    local aux1=$( ${COMP_WORDS[0]} -h summary 2>/dev/null )
+    local aux2=$( echo "${aux1}" | tr "[]|" " " | awk '{if(NR>2)print}' | tr " " "\n")
+    suggestions=("${aux2[@]}" "${common_options[@]}") #generate all the opts for the tool
+    local halg=(0x4 sha1 0xB sha256 0xC sha384 0xD sha512 0x12 sm3_256)
```
martinezjavier: I wonder if want to list the hex values, probably the pretty names are enough. What do you think @williamcroberts ?

martinezjavier:
@kaiserkyu please squash your 3 commits in 1. Having it as separate commits makes it hard to review.

You can do something like the following:

```
$ git rebase -i 9d27b267af0

And then

pick 8774e4175a6 add bash completion support for all the tools
fixup c497ea10d60 # This is a combination of 2 commits. # This is the 1st commit message:
fixup 10bf33b899e add bash completion support for all the tools

```

Also, there's a bug in the command options changes since I see the following:

```
$ tpm2_pcrlist 
ERROR: Tss2_Sys_GetCapability(0x80005) - sys:A pointer is NULL that isn't allowed to be NULL.
ERROR: Unable to run tpm2_pcrlist
```

That doesn't happen with master without your changes.

williamcroberts:
> -----Original Message-----
> Sent: Tuesday, July 3, 2018 3:17 AM
> Subject: Re: [tpm2-software/tpm2-tools] add bash completion support for all the
> tools (+2 squashed commits) (#1068)
> 
> @martinezjavier commented on this pull request.
> 
> ________________________________
> 
> In dist/bash-completion/tpm2-tools/tpm2_completion.bash
> <https://github.com/tpm2-software/tpm2-
> tools/pull/1068#discussion_r199755635> :
> 
> > +#!/bin/sh
> +# bash completion for tmp2-tools                                 -*- shell-script -*-
> +
> +
> +_tpm2_tools()
> +{
> +
> +
> +
> +    local cur prev words cword
> +    _init_completion || return
> +    local common_options=(-h --help -v --version -V --verbose -Q --quiet -Z --
> enable-errata -T --tcti=)
> +    local aux1=$( ${COMP_WORDS[0]} -h summary 2>/dev/null )
> +    local aux2=$( echo "${aux1}" | tr "[]|" " " | awk '{if(NR>2)print}' | tr " " "\n")
> +    suggestions=("${aux2[@]}" "${common_options[@]}") #generate all the opts
> for the tool
> +    local halg=(0x4 sha1 0xB sha256 0xC sha384 0xD sha512 0x12 sm3_256)
> 
> I wonder if want to list the hex values, probably the pretty names are enough.
> What do you think @williamcroberts <https://github.com/williamcroberts>  ?

I think the friendly names are enough. I think (not sure) but 4.0 might break away from
allowing raw hex, or least stating its possible.

kaiserkyu:
@martinezjavier thanks for the review, i'll work on the fixes and will let you know when their are ready, about the completion of the long options, i tried to add support for them but i was having problems because there isn't a space between '=' and the argument, but i will try other solutions 👍

williamcroberts:
This looks like you addressed @martinezjavier's comments, but I grabbed this and can't get it to work. @kaiserkyu what OS are you testing this on, and do you know what I need to do to get it working for Ubuntu?

martinezjavier:
@kaiserkyu I've created PR #1100 where I addressed the issues pointed out in my review. Could you please take a look to the version of your patch there in case I did something wrong?

That works for me and I think it can be merged now after @williamcroberts reviews / test it.

williamcroberts:
closed on #1100

PR #1070, opened 2018-06-18
tpm2_import: support specifying parent key via context file

### LINKED ISSUE #1023: tpm2_import specify handle via context file.
The parent handle is required to be persistent in the tpm2_import tool. @joshuagl has been working on rectifying options, so wait until that work is done, so we know what option to choose for a parent handle.

See:
https://github.com/tpm2-software/tpm2-tools/pull/982/files

joshuagl:
Update to support specifying the parent key as either a handle *or* a context file. Change the option long and short forms:
-H -> -C
--parent-key-handle -> --parent-key-context

Fixes #1023

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Changelog
 ### next
+  * tpm2_import: support specifying parent key with a context file,
+    --parent-key-handle/-H becomes --parent-key/-C
   * tpm2_nvwrite and tpm2_nvread: when -P is "index" -a is optional and defaults to
     the NV_INDEX value passed to -x.
   * Load TCTI's by SONAME, not raw .so file
--- a/man/tpm2_import.1.md
+++ b/man/tpm2_import.1.md
@@ -21,30 +21,30 @@ OPTIONS
 
 These options control the key importation process:
 
-  * `-G`, **--import-key-alg**=_ALGORITHM_:
-    The hash algorithm to use. Algorithms should follow the
-    " formatting standards, see section "Algorithm Specifiers".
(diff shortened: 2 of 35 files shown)


man/tpm2_import.1.md
```diff
@@ -21,30 +21,31 @@ OPTIONS
 
 These options control the key importation process:
 
-  * `-G`, **--import-key-alg**=_ALGORITHM_:
+  * **-G**, **--import-key-alg**=_ALGORITHM_:
     The hash algorithm to use. Algorithms should follow the
```
williamcroberts: Not form your patch, but I think this description is wrong. -G should be the Algorithm for the key iteself, -g is the name algorithm which is a hashing algorithm.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Changelog
 ### next
+  * tpm2_import: support specifying parent key with a context file,
+    --parent-key-handle/-H becomes --parent-key-context/-C
```
williamcroberts: We should probably do `--parent-key`, since it can be either a context or a handle.

joshuagl: agreed, will change.

man/tpm2_import.1.md
joshuagl: had to dig through code and docs to confirm, but I believe you're right. I'll fix and resubmit.

joshuagl:
Review comments addressed and added a bonus patch to tidy up some manual page formatting that I noticed whilst fixing the description for `tpm2_import`'s **-G** option.

PR #1078, opened 2018-06-22
Revert "tpm2_verifysignature: -f/--format becomes -i/--input-format"

williamcroberts:
This reverts commit d8d87a3876ecab882ec1986cf24f31eda792c45f.


williamcroberts:
@joshuagl Sorry I missed this looking over your PR. I think we want `-f` to stay a formatting option where possible AFAICT. Especially in signature to verify, since you verify a signature.

joshuagl:
Apologies, I probably should have started a discussion about that rather than blindly sending a patch.

Issue #1097, opened 2018-07-11
tpm2_unseal regression on 3.1.0


martinezjavier:
The following commands used to work with tpm2-tools 3.0:

```
$ tpm2_createprimary -Q -H e -g sha256 -G ecc -C context.p_sha256_ecc
$ tpm2_create -Q -g sha256 -G keyedhash -u opu_sha256_keyedhash -r opr_sha256_keyedhash \
 -I- -c context.p_sha256_ecc -A 'sign|fixedtpm|fixedparent|sensitivedataorigin' <<< secret
$ tpm2_load -Q -c context.p_sha256_ecc -u opu_sha256_keyedhash -r opr_sha256_keyedhash -n \
name.load_sha256_ecc-sha256_keyedhash -C ctx_load_out_sha256_ecc-sha256_keyedhash
$ tpm2_unseal -c ctx_load_out_sha256_ecc-sha256_keyedhash
secret
```

But since 3.1.0 it fails with:

```
$ tpm2_unseal -c ctx_load_out_sha256_ecc-sha256_keyedhash
ERROR: Sys_Unseal failed. Error Code: 0x12f
ERROR: Unseal failed!
ERROR: Unable to run tpm2_unseal

$ tpm2_rc_decode 0x12f
error layer
  hex: 0x0
  identifier: TSS2_TPM_RC_LAYER
  description: Error produced by the TPM
format 0 error code
  hex: 0x2f
  name: TPM2_RC_AUTH_UNAVAILABLE
  description: authValue or authPolicy is not available for selected entity.
```

It works correctly if a PCR selection is provided:

```
$ tpm2_createprimary -Q -H e -g sha256 -G ecc -C context.p_sha256_ecc
$ tpm2_pcrlist -Q -L sha1:0,1,2,3 -o pcr.bin
$ tpm2_createpolicy -Q -P -L sha1:0,1,2,3 -F pcr.bin -f policy.data
$ tpm2_create -Q -g sha256 -G keyedhash -u opu_sha256_keyedhash -r opr_sha256_keyedhash -I- \
-c context.p_sha256_ecc -L policy.data -A 'sign|fixedtpm|fixedparent|sensitivedataorigin' \
<<< secret
$ tpm2_load -Q -c context.p_sha256_ecc -u opu_sha256_keyedhash -r opr_sha256_keyedhash - \
name.load_sha256_ecc-sha256_keyedhash -C ctx_load_out_sha256_ecc-sha256_keyedhash
[...]

PR #1098, opened 2018-07-14
Revert "objectattrs: clear before or'ing in values"

martinezjavier:
This reverts commit e103bbf5117b0b62b358fd15f18f848854fcb0ee.

The tpm2-tools 3.1.0 release contains a backward incompatible change that
was introduced by commit e103bbf5117 ("objectattrs: clear before or'ing
in values"), that changed the way that object attributes were specified.

Before there were a set of default attributes and the user could specify
additional attributes to be used, but after the mentioned commit the user
must specify all attributes.

This is a user visible change that changes the tools semantics, so is not
a suitable change for a MINOR version number increment, according to the
Semantic Versioning document (https://semver.org) since it breaks rule 2:

2.MINOR version when you add functionality in a backwards-compatible manner

Fixes: #1097


williamcroberts:
Ahh nice.

PR #1102, opened 2018-07-17
Fix no TCTI initialized and correct travis

williamcroberts:
The following commit:
https://github.com/tpm2-software/tpm2-tools/commit/2fb68758e01bca412f9001ad53374beb862c752f

Introduced in a bug in TCTI initialization, which then skips providing a SAPI context
to the tools. Correct the check so TCTI initialization occurs.

Also, it looks like Travis stopped getting the exit code propagated back, fix this with some set -e in the scripts.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -31,6 +31,8 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
 #;**********************************************************************;
 
+set -e
+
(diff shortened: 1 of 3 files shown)


williamcroberts:
How did that commit slip through Travis?

williamcroberts:
It failed:
https://travis-ci.org/tpm2-software/tpm2-tools/jobs/404362681#L2403

Not sure why this didn't get propagated back to Travis, let me look.

Issue #1099, opened 2018-07-16
master branch regression: tools don't print a short summary when executed without options


martinezjavier:
On tpm2-tools 3.X executing the tools without any options prints a short summary to help users:

```
$ tpm2_create
Usage: tpm2_create [<options>]
Where <options> are:
    [ -H | --parent=<value>] [ -P | --pwdp=<value>] [ -K | --pwdk=<value>] [ -g | --halg=<value>]
    [ -G | --kalg=<value>] [ -A | --object-attributes=<value>] [ -I | --in-file=<value>] [ -L | --policy-file=<value>]
    [ -u | --pubfile=<value>] [ -r | --privfile=<value>] [ -c | --context-parent=<value>] [ -S | --input-session-handle=<value>]
```

But on the master branch, this feature doesn't work anymore:

```
$ tpm2_create
ERROR: Must specify parent object via -C
ERROR: Unable to run tpm2_create
```

joshuagl:
There was an intentional behaviour change in [12bfda8](https://github.com/tpm2-software/tpm2-tools/commit/12bfda8) with the (relevant part of the) commit message:

> Drop the help flag. Either we show help based on -h|--help option or if a tool fails along the way in option handling.

Yet the tool here fails with an error and doesn't show the help message, therefore this does appear to be a bug.

@williamcroberts can you help us understand why the behaviour was changed and we chose to no longer show the help when a tool is run without options? It's fairly common behaviour, though not universal.

williamcroberts:
Some tools execute without options, and adding that flag just kind of complicates things on when we should print versus when we shouldn't. Rather, just print on -h or on error early enough in the execution process. If the tool fails because, for instance, the keyahndle is not loaded, the tpm error is displayed, but theirs no need to dump the options out.

williamcroberts:
Currently it skips on run, because dumping help in some cases won't help the user. It's really, if we're still verifying options in the "onrun" portion.

So we could modify the return values for tpm2_onrun, so -1 show help, 0 success, 1 fail.

This way tools like tpm2_create, and others can say, show short-help please.

Or we could keep this behavior and say "see help" maybe a LOG_HELP() message?

joshuagl:
That's an elegant solution with a minimal impact on code complexity and readability -- I like it.
I have the beginnings of a patch to make that change, will likely get a PR out tomorrow.

PR #1110, opened 2018-07-20
Print usage information when incorrect options are passed to tools

joshuagl:
This PR updates the tpm2_tools tpm2_tool_onrun() method, the main interface for tools, such that if -1 is returned the tools usage will be printed.

Addresses #1099

### DIFF
--- a/tools/tpm2_tool.h
+++ b/tools/tpm2_tool.h
@@ -59,7 +59,9 @@ bool tpm2_tool_onstart(tpm2_options **opts) __attribute__((weak));
  * @param flags
  *  Flags that tools may wish to respect.
  * @return
- *  0 on success.
+ *  0 on success
+ *  1 on failure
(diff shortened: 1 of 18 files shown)


tools/tpm2_tool.h
```diff
@@ -59,7 +59,9 @@ bool tpm2_tool_onstart(tpm2_options **opts) __attribute__((weak));
  * @param flags
  *  Flags that tools may wish to respect.
  * @return
- *  0 on success.
+ *  0 on success
+ *  1 on failure
+ * -1 to show usage
```
williamcroberts: I think this is the standard, should we document this in the mapage? You can do this as a follow up to the patch series.

joshuagl: As implemented the user doesn't see this return code, we set rc = 0 after printing usage in 2d80bdf.

PR #1115, opened 2018-07-23
tests: fix create.sh and createprimary.sh

williamcroberts:
Restore functionality of these tests.

### DIFF
--- a/test/integration/helpers.sh
+++ b/test/integration/helpers.sh
@@ -53,7 +53,7 @@ pyscript
 populate_hash_algs() {
 	algs=`mktemp`
 	`tpm2_getcap -c algorithms > "$algs"`
-	filter_algs_by "$algs" 'details["hash"] and not details["method"] and not details["symmetric"] and not details["signing"] $1'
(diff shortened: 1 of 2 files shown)


joshuagl:
create.sh, extended-sessions.sh and unseal.sh are failing with this change. Though the summary suggests that 4 tests failed, those 3 are the only ones I can see marked FAIL:

williamcroberts:
I'm going to kill off https://github.com/tpm2-software/tpm2-tools/pull/1115/commits/6de743f0c01ef06ab7e98ad6fe2aac2bb37f227d.

keyedhash is implying a NULL scheme where as hmac and xor are implying keyedhash with an xor|hmac scheme, and then the algorithm to use.

Trying to imply the correct way to map keyedhash onto the correct algorithm, is quite complicated. Likely more then we should be doing. The SchemeChecks in the TPM are quite a maze.

It's not impossible to do this, I spent the morning seeing what we can do, but considering that it could lead the user to creating an object way differently then they desire, and a revert of that commit only causes createprimary.sh tests to fail, we'll just update createprimary.sh.

Issue #1089, opened 2018-07-03
error: value computed is not used [-Werror=unused-value]


yunyings:
When packaging latest 3.1.0 for Fedora, got build errors:
'_In file included from tools/tpm2_getmanufec.c:42:0:
tools/tpm2_getmanufec.c: In function 'Base64Encode':
/usr/local/ssl/include/openssl/bio.h:596:34: **error: value computed is not used [-Werror=unused-value]**
 \# define BIO_flush(b)            (int)BIO_ctrl(b,BIO_CTRL_FLUSH,0,NULL)
                                  ^
tools/tpm2_getmanufec.c:290:5: note: in expansion of macro 'BIO_flush'
     **BIO_flush(bio);**
     ^
/usr/local/ssl/include/openssl/bio.h:589:34: **error: value computed is not used [-Werror=unused-value]**
 \# define BIO_set_close(b,c)      (int)BIO_ctrl(b,BIO_CTRL_SET_CLOSE,(c),NULL)
                                  ^
tools/tpm2_getmanufec.c:292:5: note: in expansion of macro 'BIO_set_close'
     **BIO_set_close(bio, BIO_NOCLOSE);**
     ^
cc1: all warnings being treated as errors
make: *** [tools/tpm2_getmanufec.o] Error 1_
'

Should we add "(void)" where it calls "BIO_flush()"/"BIO_set_close()" in tpm2_getmanufec.c?

joshuagl:
I'm unable to reproduce this on Fedora 28. I'm building the 3.X branch with *--enable-hardening* and a patch to add -Werror=unused to EXTRA_CFLAGS [here](https://github.com/joshuagl/tpm2-tools/commit/1368bb3f0c0523845ddf43974c9b2f7f5848e599).
I can see that the CFLAGS in he generated Makefile tcontain `-Werror=unused`:

> $ grep -ir CFLAGS Makefile | grep unused
> EXTRA_CFLAGS =  -Wall -Wextra -Werror -Werror=unused -Wformat -Wformat-security -Wstack-protector -fstack-protector-all -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -fPIC -fPIE -D_GNU_SOURCE -fdata-sections -ffunction-sections

My builds succeed.

joshuagl:
[...]

PR #1121, opened 2018-07-25
Fixes for building on CentOS 7

joshuagl:
Whilst trying to replicate #1089 I have build both the 3.X and master branches on CentOS 7. The following two patches were required to build master on CentOS 7.


williamcroberts:
Maybe we should have a CENTOS container we build and test in... we need to look at how much time we get allotted on travis for free... maybe do a random pick docker image...

Issue #1123, opened 2018-07-26
Unsealing data despite PCR policy


havolarkis:
Folowing the examples in branch 3.X test scripts 'test/system/test_tpm2_unseal.sh' for sealing I've got some undesirable behavior. It is possible to unseal the secret data wihtout passing the PCR list and still get it. 

```
# tpm2_pcrlist -Q -L sha1:0,1,2,3 -o pcr.bin
# tpm2_createpolicy -Q -P -L sha1:0,1,2,3 -F pcr.bin -f policy.data
# tpm2_create -Q -g sha256 -G keyedhash -u opu_sha256_keyedhash -r opr_sha256_keyedhash -I- -c context.p_sha256_ecc -L policy.data -A 'sign|fixedtpm|fixedparent|sensitivedataorigin'
# tpm2_load -Q -c context.p_sha256_ecc -u opu_sha256_keyedhash -r opr_sha256_keyedhash -n name.load_sha256_ecc-sha256_keyedhash -C ctx_load_out_sha256_ecc-sha256_keyedhash
# tpm2_unseal -c ctx_load_out_sha256_ecc-sha256_keyedhash -L sha1:0,1,2,3 -F pcr.bin
12345678
# tpm2_pcrextend 3:sha1=6c10289a8da7f774cf67bd2fc8502cd4b585346a
# tpm2_unseal -c ctx_load_out_sha256_ecc-sha256_keyedhash
12345678
# tpm2_unseal -c ctx_load_out_sha256_ecc-sha256_keyedhash -L sha1:0,1,2,3 -F pcr.bin
ERROR: Failed parse_policy_type_and_send_command
ERROR: Building PCR policy failed: 0x1c4
```

If I do not pass the option `-L sha1:0,1,2,3 -F pcr.bin`, I still get the secret sealed. Am I missing something? Is it some sort of TPM2 feature?

I am on a CentOS-7.5
ibmtpm1119.tar.gz (tpm server)
tpm2-abrmd-1.1.0
tpm2-tools-3.0.1
tpm2-tss-1.3.0

diabonas:
[...]

Issue #1079, opened 2018-06-22
fix tpm2_import


williamcroberts:
PR #1063 breaks tpm2_import yet again. Figure out why it broke, and fix it.

--Update -- We know why, we need to:
- [x] - fix to existing
- [x] - Support as-many-as-possible Parent algorithms
- [x] - Support name algorithm for created object (#1120)
- [x] - Support different parent name algorithms (used for hmac) (#1120)
- [x] - Support to-be-created object attributes (41b14163d2)
- [ ] - ~~Move to common public init~~ N/A - we load them from a on disk structure or PEM file.
- [x] - move OSSL to compat library to handle EVP ctx init differences, etc. (#1120)

joshuagl:
The `import.sh` test fails with this sequence of commands:

> tpm2_createprimary -Q -G 1 -g 0xb -a o -o parent.ctx
> tpm2_evictcontrol -Q -a o -c parent.ctx -p 0x81010005
> dd if=/dev/urandom of=sym.key bs=1 count=16 2>/dev/null
> tpm2_readpublic -Q -c 0x81010005 --out-file parent.pub
> tpm2_import -Q -G aes -k sym.key -C 0x81010005 -K parent.pub -q import_key.pub -r import_key.priv

I tried manually executing this sequence of commands and replacing the tools at each step with ones built with a copy of the code before PR #1063 landed. The issue seems to be related specifically to behavioural changes in `tpm2_createprimary`:

> $ ./tools/tpm2_createprimary -Q -G 1 -g 0xb -a o -o parent.ctx
> $ ./tools/tpm2_evictcontrol -Q -a o -c parent.ctx -p 0x81010005
> $ ./tools/tpm2_readpublic -Q -c 0x81010005 --out-file parent.pub
> $ ./tools/tpm2_import -Q -G aes -k sym.key -C 0x81010005 -K parent.pub -q import_key.pub -r import_key.priv
> ERROR: Tss2_Sys_Import(0x3D5) - tpm:parameter(3):structure is the wrong size
> ERROR: Unable to run ./tools/tpm2_import

whereas with an older `tpm2_createprimary`:

> $ ~/Desktop/tpm2-tools/tools/tpm2_createprimary -Q -G 1 -g 0xb -a o -o parent.ctx
> $ ./tools/tpm2_evictcontrol -Q -a o -c parent.ctx -p 0x81010005
[...]

PR #1126, opened 2018-07-30
[for 3.X] tpm2_create: Use better object attributes defaults for authentication

martinezjavier:
The tpm2_create tool allows to define a policy session or a password for
authentication. By default no policy session is used and the password is
empty, which means that this empty password is used for authentication.

So the default object attribute flag userWithAuth is set in order to use
the empty password. This isn't a good default though if a policy is set,
since in this case the policy session has to be used for authentication
instead of an empty password.

If a policy is defined, the userWithAuth bit has to be clear unless the
user defines a password so in that case authentication would happen only
using the policy session or the defined password.

### DIFF
--- a/tools/tpm2_create.c
+++ b/tools/tpm2_create.c
@@ -243,6 +243,7 @@ static bool on_option(char key, char *value) {
             return false;
         }
         ctx.flags.K = 1;
+        ctx.in_public.publicArea.objectAttributes |= TPMA_OBJECT_USERWITHAUTH;
         break;
     case 'g':
         ctx.nameAlg = tpm2_alg_util_from_optarg(value);
@@ -281,6 +282,9 @@ static bool on_option(char key, char *value) {
             return false;
         }
         ctx.flags.L = 1;
+        if (!ctx.flags.K) {
+             ctx.in_public.publicArea.objectAttributes &= ~TPMA_OBJECT_USERWITHAUTH;
+        }
         break;
     case 'S':
         if (!tpm2_util_string_to_uint32(value, &ctx.session_data.sessionHandle)) {
--- a/test/system/test_tpm2_unseal.sh
+++ b/test/system/test_tpm2_unseal.sh
@@ -109,4 +109,55 @@ unsealed=`tpm2_unseal -c $file_unseal_key_ctx -L ${alg_pcr_policy}:${pcr_ids} -F
 
 test "$unsealed" == "$secret"
 
+# Test that unseal fails if a PCR policy isn't provided
+
+trap - ERR
+
+tpm2_unseal -c $file_unseal_key_ctx 2> /dev/null
+if [ $? != 1 ]; then
+  echo "tpm2_unseal didn't fail without a PCR policy!"
+  exit 1
+fi
+
+# Test that unseal fails if PCR state isn't the same as the defined PCR policy
+
+pcr_extend=$(echo $pcr_ids | cut -d ',' -f1)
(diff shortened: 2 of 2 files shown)


martinezjavier:
Tests fail but it seems that's not my fault:

```
ECcert.bin ECcert2.bin differ: char 1, line 1
cmp ECcert.bin ECcert2.bin on line 68 failed: 1
test_tpm2_getmanufec.sh ... FAILED
```

martinezjavier:
> Tests fail but it seems that's not my fault:

Ok, the `test_tpm2_getmanufec.sh` passed this time so it seems that was a transient error.

diabonas:
As stated in https://github.com/tpm2-software/tpm2-tools/pull/1127#issuecomment-408835398, this pull request works as expected for the `3.X` branch.

martinezjavier:
@diabonas thanks for the explanation of the issue and for testing!

tools/tpm2_create.c
```diff
@@ -281,6 +282,9 @@ static bool on_option(char key, char *value) {
             return false;
         }
         ctx.flags.L = 1;
+        if (!ctx.flags.K)
```
williamcroberts: brackets

martinezjavier:
@williamcroberts @diabonas I've pushed an update that doesn't use `adminWithPolicy` at all. So basically what we do is:

* Set `userWithAuth` by default (neither policy nor password specified). This is the *binding* data against a TPM case.
* Set `userWithAuth` if a user provided a password (regardless if it's specified a policy or not), that allows to authenticate using either a password or a policy for both USER and ADMIN roles. This is the *sealing* data against a TPM PCR state case. But leaves a password in case a user wants to unseal even when a PCR state changed.
* Clear `userWithAuth` if a user didn't provide a password (default) but provided a policy. That fixes issue #1123 to not authenticate using an empty password.

Does that sound sane to you? Thanks a lot for your feedback and suggestions.

BTW, I didn't change the integration tests from my latest version so the above seems to work correctly.

martinezjavier:
@williamcroberts a nice thing is that I think we didn't break any backward compatibility here, since authenticating using an empty password when sealing data against a PCR state is clearly something that no user could ever want.

williamcroberts:
@martinezjavier that seems good for (checks branch for PR) 3.X. On Master, we'll want to issue an error when no password and policy specified.

martinezjavier:
@diabonas great, thanks again for testing!

@williamcroberts yes, agreed. I think that I should wait though for you to fix issue #1079 before porting the patch to master, since as mentioned by @diabonas we need to set `alg_primary_key=rsa2048:null:aes128cfb` in `test/integration/tests/unseal.sh`, the current `alg_primary_key=ecc` doesn't work.

joshuagl:
3.1.2 is still in an overly long rc0, I think it's probably worth spinning an rc1 with this added.

williamcroberts:
> -----Original Message-----
> From: Joshua Lock [mailto:notifications@github.com]
> Sent: Tuesday, July 31, 2018 1:14 PM
> Cc: Roberts, William C <william.c.roberts@intel.com>; State change
> <state_change@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-tools] [for 3.X] tpm2_create: Use better
> object attributes defaults for authentication (#1126)
> 
> 3.1.2 is still in an overly long rc0, I think it's probably worth spinning an rc1 with
> this added.

SGTM

Issue #1129, opened 2018-07-30
tpm2_import


williamcroberts:
Support ECC parent key

williamcroberts:
Duplicate of #1128

Issue #1128, opened 2018-07-30
tpm2_import: support ECC


williamcroberts:
Support Importing ECC keys

williamcroberts:
fixed: 90e5d99ed3771

PR #1130, opened 2018-07-30
Enhancments to tpm2-import

williamcroberts:
Enhance tpm2-import:
1. Internal refactoring to reduce usages of globals.
2. Support setting name hash algorithm via -g.
3. Support varying parent name-alg and symmetric parameters.
4. Support importation of RSA 1024 keys (likely works for 4096 but needs testing)
5. Support importation of AES 128/192/256 keys

This makes progress on #1079, but doesn't resolve everything. Issues #1129 and #1128 have been opened to address adding in missing ECC support.

It might be awhile before I can get back to tpm2_import, but this makes major progress and should be included in master.

### DIFF
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -912,12 +912,14 @@ static bool load_rsa_key(const char *private_path,
     RSA_get0_key(k, &n, NULL, NULL);
 #endif
 
-    int priv_bytes = BN_num_bytes(p);
-    if (priv_bytes > priv->size) {
-        LOG_ERR("Expected prime \"p\" to be less than or equal to %u,"
-                " got: %d", priv->size, priv_bytes);
+    unsigned priv_bytes = BN_num_bytes(p);
+    if (priv_bytes > sizeof(priv->buffer)) {
+        LOG_ERR("Expected prime \"p\" to be less than or equal to %lu,"
+                " got: %u", sizeof(priv->buffer), priv_bytes);
         goto out;
     }
+
+    priv->size = priv_bytes;
 
     int success = BN_bn2bin(p, priv->buffer);
     if (!success) {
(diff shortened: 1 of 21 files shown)


tools/tpm2_import.c
```diff
@@ -870,12 +870,12 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
...
       { "import-key-private", required_argument, NULL, 'r'},
-      { "import-key-public",  required_argument, NULL, 'q'},
+      { "import-key-public",  required_argument, NULL, 'u'},
       { "object-attributes",  required_argument, NULL, 'A'},
       { "halg",               required_argument, NULL, 'g'},
     };
 
-    *opts = tpm2_options_new("G:k:C:K:q:r:A:g:", ARRAY_LEN(topts), topts, on_option,
+    *opts = tpm2_options_new("G:k:C:K:u:r:A:g:", ARRAY_LEN(topts), topts, on_option,
                              NULL, 0);
```
joshuagl: Need to update the tpm2_import manual page for this change

tools/tpm2_import.c
```diff
@@ -982,6 +982,42 @@ static bool load_key(
     return true;
 }
 
+/**
+ * Check all options and report as many errors as possible via LOG_ERR.
+ * @return
+ *  0 on success, 1 on failure.
```
joshuagl: Might be nice to return -1 instead of 1 for errors in this function, that way the usage information will be shown?

tools/tpm2_import.c
williamcroberts: Good catch, I saw your comment and I was like no i did that...nope I updated the changelog...

tools/tpm2_import.c
williamcroberts: Yeah that's a good idea.

PR #1133, opened 2018-08-01
Ensure Python 3 compatibility of print function

diabonas:
This is just an extremely trivial pull request to add parentheses to the Python `print` function so that it works with Python 3 as well as Python 2. This is necessary to run the integration tests e.g. on Arch Linux, which uses Python 3 as the default interpreter. I think I found all instances where the old syntax was still used, while some scripts like [test/integration/tests/getcap.sh](https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/tests/getcap.sh#L58) already used the new syntax anyway.

With this change, the test suite compiles and runs successfully on Arch Linux. The resulting files still work with Python 2 as well.

### DIFF
--- a/scripts/utils/analyze.py
+++ b/scripts/utils/analyze.py
@@ -12,6 +12,7 @@ a human will need to make a plan of action to correct the tools so they conform.
 The tool exits with a status of non-zero on error, making this useful for checking
 PRs and can thus be added to travis.
 '''
+from __future__ import print_function
 
 import glob
 import os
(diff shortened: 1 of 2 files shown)


joshuagl:
This change looks OK to. I don't think we want to go as far as assuming that python==python3, afaik Arch are the only distro who have made that change.

martinezjavier:
> This change looks OK to. I don't think we want to go as far as assuming that python==python3, afaik Arch are the only distro who have made that change.

@joshuagl agreed, but on the other hand we should assume that by using python in the shebang, the scripts are be compatible with both python2 and 3 as explained in [PEP 0394](https://www.python.org/dev/peps/pep-0394/).

My understanding is that Fedora is switching to Python 3 as default soon too, but I don't know for what release is planned.

joshuagl:
Thanks for pointing out PEP 0394 @martinezjavier, I'd somehow missed that. I agree we should follow the Python recommendation and maintain compatibility with both versions as long as python is used in the shebang.

Issue #1134, opened 2018-08-02
build failure with tpm2-tools


twoerner:
Commit 45a1dd2bf41557286bb5b5c8ee64b4cd2f506994 to tpm2-tools causes the following build failure:

```
| In file included from ../tpm2-tools/tools/tpm2_import.c:52:
| ../tpm2-tools/tools/tpm2_import.c: In function 'load_rsa_key':
| ../tpm2-tools/tools/tpm2_import.c:917:17: error: format '%lu' expects argument of type 'long unsigned int', but argument 5 has type 'unsigned int' [-Werror=format=]
|          LOG_ERR("Expected prime \"p\" to be less than or equal to %lu,"
|                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
|                  " got: %u", sizeof(priv->buffer), priv_bytes);
|                              ~~~~~~~~~~~~~~~~~~~~
| ../tpm2-tools/lib/log.h:57:69: note: in definition of macro 'LOG_ERR'
|  #define LOG_ERR(fmt, ...) _log(log_level_error, __FILE__, __LINE__, fmt, ##__VA_ARGS__)
|                                                                      ^~~
| ../tpm2-tools/tools/tpm2_import.c:917:69: note: format string is defined here
|          LOG_ERR("Expected prime \"p\" to be less than or equal to %lu,"
|                                                                    ~~^
|                                                                    %u
[...]

PR #1135, opened 2018-08-02
tpm2_import: fix size_t printf format specifiers

williamcroberts:
Commit:
45a1dd2bf415 "tpm2_import: drop hardcoded prime size in bytes"

Uses %lu to print the returns from sizeof(), when it should be
using %zu for size_t. This causes build issues on 32 bit machines.

Fixes: #1134


williamcroberts:
@twoerner can you verify this patch corrects your issue?

PR #1136, opened 2018-08-03
RELEASE: add explanation on how to edit the CHANGELOG file for releases

martinezjavier:
So far there wasn't consistency on how the CHANGELOG was edited to reflect
the release candidates and final releases. Add some words on how this must
be done so the CHANGELOG remains consistent between releases and branches.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -18,6 +18,32 @@ When a release is cut, the process is the same as an rc, with the exception that
 not marked as "pre-release" on github. The release notes should include everything from
 the last release to the latest release.
 
+### Updating the CHANGELOG for release candidates and final releases
+
+When a first release candidate is cut, a new entry will be added to the CHANGELOG file. This
+entry will have the release candiate version and the date on which the release candidate was
+released. The entry will list all the changes that happened between the latest final release
+and the first release candidate.
+
+The commit that made the change will be tagged with a release version and a -rc0 suffix as an
+indication that is not a final release but a first release candidate. If after a week the -rc
+has no changes, then a final release can be made as explained above. But if changes are made,
+then a new releases candidate will be released and the -rc suffix will be incremented.
+
+For each release candidate, the changes that happened between the previous release candidate
+will be appended to the CHANGELOG entry, and both the release candidate version and the date
+of the release will be updated to reflect the latest release candidate.
+
+The commit that updated the CHANGELOG entry will be tagged as the latest release candidate.
+
+When a final version will be released, there will not be changes to append to the CHANGELOG
+entry since otherwise a new release candidate should be cut. So only the release version and
+date should be updated in the CHANGELOG.
+
+The commit that updated the CHANGELOG entry will be tagged as the final release.
+
+For a final release, change the version to the final release version (i.e: 3.0.5-rc3 -> 3.0.5) and update the date. The commit for this change will be tagged as $version.
+
 ### Version Information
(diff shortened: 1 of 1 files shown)


joshuagl:
LGTM

RELEASE.md
```diff
@@ -18,6 +18,32 @@ When a release is cut, the process is the same as an rc, with the exception that
...
+
+For each release candidate, the changes that happened between the previous release candidate
+will be appended to the CHANGELOG entry, and both the release candidate version and the date
+of the release will be updated to reflect the latest release candidate.
+
+The commit that updated the CHANGELOG entry will be tagged as the latest release candidate.
+
+When a final version will be released, there will not be changes to append to the CHANGELOG
+entry since otherwise a new release candidate should be cut. So only the release version and
+date should be updated in the CHANGELOG.
```
williamcroberts: I was thinking you would squash the RC`<N>` into one summary with the version and date. I would imagine folks would want to diff version X to version Y change logs without the noise of RC's.

RELEASE.md
```diff
@@ -18,6 +18,32 @@ When a release is cut, the process is the same as an rc, with the exception that
...
+
+The commit that updated the CHANGELOG entry will be tagged as the final release.
+
+For a final release, change the version to the final release version (i.e: 3.0.5-rc3 -> 3.0.5) and update the date. The commit for this change will be tagged as $version.
+
```
williamcroberts: We should probably document the format semantics of:
```md
<VERSION> - <DATE>
  * item 1
  * item 2
  * ...
```

RELEASE.md
martinezjavier: @williamcroberts agreed, and that's what I tried to write in those paragraph. So what I tried to say is the following:

1) Current release is `4.0` and a new `4.1` release will be made. The changes since `4.0` where `foo` and `bar`. So for `-rc0` we will add the following will be added to the `CHANGELOG`:

```
4.1-rc0 - 2018-08-03
* foo
* bar
```
2) Some issues are found on `-rc0`, so a change `baz` was made. So we will cut another `-rc1` and the `CHANGELOG` entry will be modified to (`baz` appended and `<version, date>` tuple updated).

```
4.1-rc1 - 2018-08-10
* foo
* bar
* baz
```

After a week, no new changes are made so we can cut a final `4.1` release. The entry is the `CHANGELOG` is modified again but this time only the `<version,date>` tuple. Since there were no changes nothing will be appended (and if we had changes to append, then that would imply a new `-rc2`):

```
4.1 - 2018-08-18
* foo
* bar
* baz
```

So users will only see a delta from `4.0` to `4.1`, without any noise of release candidate versions. But if someone downloads a `-rc` tarball, she would see a delta from `4.0` to whatever `-rc` was downloaded. If the final tarball is released, then only the final release will be present.

Does that makes sense? Please let me know were I used a wrong wording since I tried to explain this in my patch.

RELEASE.md
martinezjavier: @williamcroberts agreed.

RELEASE.md
williamcroberts: @martinezjavier re-reading that and it's clear to me. Not sure why I thought it wasn't.

RELEASE.md
williamcroberts: Ill apply this change on-top of your work.

PR #1138, opened 2018-08-04
tpm2_verifysignature enhancmeents

williamcroberts:
* Stop outputting the message bytes
* Make -t optional
* Issue a warning when -t is specified and the object hierarchy is NULL.


joshuagl:
Nice set of clean ups, LGTM.

Issue #1139, opened 2018-08-06
Path with "aux" causes problems in Windows


gilb3rt:
Issue https://github.com/swcarpentry/DEPRECATED-bc/issues/463 concluded that Windows has reserved "aux" for some reason. The newly added "tools/aux" directory causes my sync to fail on Windows with the error "fatal: cannot create directory at 'tools/aux': Invalid argument".

PR #1140, opened 2018-08-06
Fix for #1139 Windows aux issue

gilb3rt:
Solves #1139.


joshuagl:
This fails because you haven't updated the Makefiles after moving the source files. This would be cleaner as one patch which:
- uses git mv to rename the aux directory (`git mv tools/aux tools/misc`)
- change all uses of aux in *Makefile.am* to misc

williamcroberts:
The commit looks fine, but please update the commit message to something useful.
```
windows: change aux to misc for build fix

Step 1 describe the problem (do not actually include the steps in the message)
on Windows aux as a path component is reserved (yada yada, point to some doc).

Step 2: describe the fix
Change aux to misc to avoid this issue.

Fixes #1138
```

williamcroberts:
Moved to: https://github.com/tpm2-software/tpm2-tools/pull/1144

PR #1152, opened 2018-08-16
man: remove mentioned but unused -S options

williamcroberts:
-S is no longer used in tools, it's supported as a password
option, thus remove the mentions from the manpage.


joshuagl:
LGTM

PR #1153, opened 2018-08-16
tpm2_loadexternal: enhance support for object loading

williamcroberts:
tpm2_loadexternal was quite limited in functionality, with respect to:
1. ability to load pem files
   a. Affected the ability to actually load an external sensitive object
2. Auth values
3. Policy support
4. Object Attributes

This patch aims to bring:
1. RSA PEM support for public and sensitive portions of the object.
2. Auth values supported via the -p option.
3. Policy support via the -L option.
4. Object Attributes via the -A option.

Status:
It currently loads objects from PEM files, however when using a password, using the
objects in subsequent commands is failing.


test/integration/tests/loadexternal.sh
```diff
@@ -92,4 +93,22 @@ tpm2_create -Q -C $Handle_parent -g $alg_create_obj -G $alg_create_key -u $file_
...
+
+echo "hello world" > plain.txt
+openssl rsautl -encrypt -inkey public.pem -pubin -in plain.txt -out plain.rsa.enc
+
+tpm2_loadexternal -a o -p foo -r private.pem -o key.ctx
+
+#
+# This command is failing with:
+# ERROR: Tss2_Sys_RSA_Decrypt(0x12F) - tpm:error(2.0): authValue or authPolicy is not
+# available forselected entity
```
williamcroberts: I *think* I know what is missing. I forgot to initialize the auth structure with the handle for the password session.

williamcroberts:
@idesai this should do what you want, let me know.

williamcroberts:
@idesai - That seems like a good note to add, along with changing the default hierarchy to NULL.

PR #1154, opened 2018-08-16
tpm2_rsadecrypt: fix -p option

williamcroberts:
In the getopt short option string, a capital P was in place
for lowercase p resulting in the password option not being
processed. Fix this and add passwords to the tests to prevent
regressions.


joshuagl:
Good catch, thanks!

Issue #1171, opened 2018-09-05
tpm2-tools 3.X doesn't honor the TPM2TOOLS_* env vars to configure the TCTI


martinezjavier:
In master, the environment variables that are used to choose a TCTI and their properties were changed to use a single `TPM2TOOLS_ENV_TCTI` env var.

But this change landed into the `3.X` branch as well as commit ac9a5d787753 ("update to TSS version 2.0"). The idea of tpm2-tools 3.X was to maintain backward compatibility for the CLI users, but changing how the TCTI is definitely a regression in that sense and is against the [semantic versioning](https://semver.org/) used by the project.

For example an issue was filed in the clevis project (that makes uses of the tpm2-tools): https://github.com/latchset/clevis/issues/71

PR #1173, opened 2018-09-06
lib/tpm2_options: restore TCTI configuration environment variables

joshuagl:
The port to TSS2.0 introduced a new unified environment variable to
configure a TCTI, TPM2TOOLS_ENV_TCTI. Unfortunately this also unwittingly
removed the old-style environment variable per TCTI configuration options,
which is a behavioural regression for the 3.x series of tpm2-tools.

Restore the original TPM2TOOLS_* environment variables in addition to the
new style single environment variable.

Fixes issue #1171


joshuagl:
Looks like my integration test is faulty, but I'd appreciate some review of the code whilst I figure it out

lib/tpm2_options.c
```diff
@@ -200,24 +203,93 @@ static tcti_conf tcti_get_config(const char *optstr) {
...
+     * 
+     * Max IPV6 IP address, 45 characters   (45)
+     * Ports are 16bit int, 5 characters    (5)
+     * "host=", 5 characters                (5)
+     * "port=", 5 characters                (5)
+     * strlen = 60
+     */
+    size_t optlen = 60;
+    const char *host, *port;
+    char *ret = malloc(optlen);
```
williamcroberts: if ret null.... oom

lib/tpm2_options.c
```diff
@@ -200,24 +203,93 @@ static tcti_conf tcti_get_config(const char *optstr) {
...
+     * "host=localhost,port=2321" for the mssim tcti
+     * 
+     * Max IPV6 IP address, 45 characters   (45)
+     * Ports are 16bit int, 5 characters    (5)
+     * "host=", 5 characters                (5)
+     * "port=", 5 characters                (5)
+     * strlen = 60
+     */
+    size_t optlen = 60;
+    const char *host, *port;
```
williamcroberts: we dont do N variable declarations on a line.

lib/tpm2_options.c
joshuagl: god catch, will fix

lib/tpm2_options.c
joshuagl: got it

lib/tpm2_options.c
```diff
@@ -136,13 +140,25 @@ void tpm2_options_free(tpm2_options *opts) {
...
 struct tcti_conf {
-    const char *name;
-    const char *opts;
+    char *name;
+    char *opts;
 };
 
+/*
+ * Some tcti names changed in TSS 2.0, so in order to not break the
+ * the expected options of the 3.X tools series map:
```
martinezjavier: Double `the` in the comment.

Issue #1175, opened 2018-09-07
tpm2_getcap prints properties as TPM2_PT_* instead of TPM_PT_* in 3.X branch


vilemmarsik:
For my TPM2 device, identified in dmesg as "tpm_tis MSFT0101:00: 2.0 TPM (device-id 0xFE, rev-id 2)", kernel 3.10.0-944.el7.x86_64, and tpm2-tools-3.0.4-1.el7.x86_64, tpm2_getcap prints "TPM_PT_..." instead of "TPM2_PT_...", e.g.:
TPM_PT_HR_NV_INDEX:          0x0000000c
TPM_PT_HR_LOADED:            0x00000000
...
Regular expressions  in test/system fail on this.
Adding a patch to make things work both with "TPM2_*" and "TPM_*".

vilemmarsik:
My patch:
```
diff --git a/test/system/test_tpm2_dictionarylockout.sh b/test/system/test_tpm2_dictionarylockout.sh
index a90854a..5a1f46e 100755
--- a/test/system/test_tpm2_dictionarylockout.sh
+++ b/test/system/test_tpm2_dictionarylockout.sh
@@ -40,16 +40,16 @@ trap onerror ERR
 
 tpm2_dictionarylockout -s -n 5 -t 6 -l 7
 
-if [ "$(tpm2_getcap -c properties-variable | grep TPM2_PT_MAX_AUTH_FAIL | sed -e 's/TPM2_PT_MAX_AUTH_FAIL: \+//')" != "0x00000005" ];then
+if [ "$(tpm2_getcap -c properties-variable | grep TPM2\\?_PT_MAX_AUTH_FAIL | sed -e 's/TPM2\?_PT_MAX_AUTH_FAIL: \+//')" != "0x00000005" ];then
  echo "Failure: setting up the number of allowed tries in the lockout parameters"
  exit 1
 fi
 
-if [ "$(tpm2_getcap -c properties-variable | grep TPM2_PT_LOCKOUT_INTERVAL | sed -e 's/TPM2_PT_LOCKOUT_INTERVAL: \+//')" != "0x00000006" ];then
+if [ "$(tpm2_getcap -c properties-variable | grep TPM2\\?_PT_LOCKOUT_INTERVAL | sed -e 's/TPM2\?_PT_LOCKOUT_INTERVAL: \+//')" != "0x00000006" ];then
  echo "Failure: setting up the lockout period in the lockout parameters"
 fi
 
[...]

PR #1177, opened 2018-09-11
tpm2_getcap: restore tool output to print properties with TPM_PT prefix

martinezjavier:
Commit ac9a5d787753 ("update to TSS version 2.0") updated tpm2-tss version
to 2.0. On tpm2-tss, the TPM_PT_* constants where renamed to TPM2_PT_* but
that commit also changed the output of the tpm2_getcap tool to match the
name of the properties.

Unfortunately this is a bacward incompatible change that could break users
that were relying on the tpm2_getcap tool output.

Restore the older output and also fix the tests that check these values.

Fixes: #1175

Reported-by: Vilem Marsik <vmarsik@redhat.com>


joshuagl:
LGTM

PR #1187, opened 2018-10-03
Add support for moreData on GetCapability

### LINKED ISSUE #1050: tpm2_clear only in master
I assume there's a reason the tpm2_clear tool is only in master right now and not in the 3.x releases?  Was wondering if someone could bring me up to speed real quick as to why that's the case, as I haven't been staying on top of current development.

Thanks!

### LINKED ISSUE #1150: tpm2_getcap -c properties-fixed
I'm able to run tpm2_getcap with other arguments, but "properties-fixed" throws:

tpm2_getcap -c properties-fixed -V
WARN on line: "65" in file: "lib/tpm2_capability.c": More data to be queried: capability: 0x6, property: 0x100

ERROR on line: "166" in file: "tools/tpm2_tool.c": Unable to run ./tpm2_getcap

marcel-hamer:
Add additional calls to Tss2_Sys_GetCapability when moreData is available.
Additional data is appended into the result capability data after each read out.

Closes #1150

### DIFF
--- a/lib/tpm2_capability.c
+++ b/lib/tpm2_capability.c
@@ -30,6 +30,7 @@
 //**********************************************************************;
 
 #include <stdbool.h>
+#include <string.h>
 
 #include <tss2/tss2_sys.h>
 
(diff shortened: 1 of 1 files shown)


lib/tpm2_capability.c
```diff
@@ -33,38 +33,97 @@
...
 
+#include <string.h>
+
 #include "log.h"
 #include "tpm2_util.h"
 
 #include "tpm2_capability.h"
 
+static int tpm2_get_property_data_size(TPM2_CAP capability) {
+    int size;
```
williamcroberts: size_t or unsigned? I don't see why this needs to be signed.

single line of whitespace between function and body.

```C
static size_t tpm2_get_property_data_size(...) {

    size_t size;
...

File: lib/tpm2_capability.c [hunk 2]
```diff
@@ -33,38 +33,97 @@
 
 #include <tss2/tss2_sys.h>
 
+#include <string.h>
```
williamcroberts: This should go under line 32.

lib/tpm2_capability.c
```diff
@@ -33,38 +33,97 @@
...
-        LOG_WARN("More data to be queried: capability: 0x%x, property: "
-                 "0x%x\n", capability, property);
-        return false;
-    }
+    TPMI_YES_NO more_data;
+
+    TPMS_CAPABILITY_DATA local_data;
+
+    UINT32 offset = 0;
+    UINT32 property_size = tpm2_get_property_data_size(capability);
```
williamcroberts: See your coercing the int to UINT32 here, so probably just want to return UINT32.

lib/tpm2_capability.c
```diff
@@ -33,38 +33,97 @@
...
+
+    void *data_dest = (void *) &capability_data->data + sizeof(UINT32);
+    void *data_src  = (void *) &local_data.data + sizeof(UINT32);
```
williamcroberts: Why + `sizeof(UINT32)`, to jump over the size field in the subordinate list structure? Just pick any one of the TPML counts and begining data fields, and comment that this is OK because they all line up rather than a magic + sizeof(UINT32).

lib/tpm2_capability.c
```diff
@@ -33,38 +33,97 @@
...
+                    &more_data,
+                    &local_data,
+                    NULL));
+
+        if (rval != TPM2_RC_SUCCESS) {
+            LOG_ERR("Failed to GetCapability: capability: 0x%x, property: 0x%x",
+                  capability, property);
+            LOG_PERR(Tss2_Sys_GetCapability, rval);
+            return false;
+        } else {
```
williamcroberts: else isn't needed since the if returns.

lib/tpm2_capability.c
```diff
@@ -33,38 +33,97 @@
...
+            UINT32 count;
+
+            data_dest += offset * property_size;
+
+            memcpy(&count, &local_data.data, sizeof(UINT32));
```
williamcroberts: You can literally pick any TPML_XXX from the data union to get the count and then you can ditch this variable, copy and confusing code all together.


See structure definitions:

```C
/* Definition of TPMU_CAPABILITIES Union <OUT> */
typedef union {
    TPML_ALG_PROPERTY algorithms;
    TPML_HANDLE handles;
    TPML_CCA command;
    TPML_CC ppCommands;
    TPML_CC auditCommands;
    TPML_PCR_SELECTION assignedPCR;
    TPML_TAGGED_TPM_PROPERTY tpmProperties;
    TPML_TAGGED_PCR_PROPERTY pcrProperties;
    TPML_ECC_CURVE eccCurves;
    TPML_INTEL_PTT_PROPERTY intelPttProperty;
} TPMU_CAPABILITIES;

/* Definition of TPMS_CAPABILITY_DATA Structure <OUT> */
typedef struct {
    TPM2_CAP capability;     /* the capability */
    TPMU_CAPABILITIES data;  /* the capability data */
} TPMS_CAPABILITY_DATA;
```

lib/tpm2_capability.c
```diff
@@ -33,38 +33,97 @@
...
+
 #include "log.h"
 #include "tpm2_util.h"
 
 #include "tpm2_capability.h"
 
+static int tpm2_get_property_data_size(TPM2_CAP capability) {
+    int size;
+
+    switch (capability) {
```
williamcroberts: switch and return sizeof the field in the subordinate union member.

```C
TPMU_CAPABILITIES dummy;
switch (capability) {
case TPM2_CAP_ALGS:
    return sizeof(dummy.algorithms.algProperties[0]);
case ...;
    return ...'
...
```
Now you never have to worry about the spec changing a type, it'll always work or not compile.

lib/tpm2_capability.c
```diff
@@ -115,3 +174,4 @@ bool tpm2_capability_find_vacant_persistent_handle (TSS2_SYS_CONTEXT *sapi_ctx,
 
     return handle_found;
 }
+
```
williamcroberts: unrelated whitespace change

marcel-hamer:
@williamcroberts Thanks for the quick and thorough review. I haven't commented on all your requests separately, but I think I have made all the requested changes in the latest commit.

lib/tpm2_capability.c
```diff
@@ -38,33 +39,96 @@
...
-                 "0x%x\n", capability, property);
-        return false;
-    }
+    TPMI_YES_NO more_data;
+
+    TPMS_CAPABILITY_DATA local_data;
+
+    UINT32 result_count;
+    UINT32 offset = 0;
+    size_t property_size = tpm2_get_property_data_size(capability);
```
williamcroberts: if property_size is 0, that's a failure case.

lib/tpm2_capability.c
```diff
@@ -38,33 +39,96 @@
...
+
+    /* All count and data fields line up within the TPMU_CAPABILITIES union
+        so it it safe to pick one here. */
```
williamcroberts: This isn't how we do multiple line comments:
```C
/*
 * All count ...
 */
 ```

File: lib/tpm2_capability.c [hunk 11]
```diff
@@ -38,33 +39,96 @@
...
+        if (rval != TPM2_RC_SUCCESS) {
+            LOG_ERR("Failed to GetCapability: capability: 0x%x, property: 0x%x",
+                  capability, property);
+            LOG_PERR(Tss2_Sys_GetCapability, rval);
+            return false;
+        }
+
+        result_count = local_data.data.algorithms.count;
+
+        data_dest += offset * property_size;
```
williamcroberts: The fact that it's letting you do pointer arithmetic on a `void *` blows my mind. I think you want `data_dest` and `data_src` to be of type `BYTE *` Unless you can point to the C99 spec that says this arithmetic is defined in the standard as one memory address. Normal pointer arithmetic moves the address ahead by sizeof the dereferenced type, which in void doesn't have one.

lib/tpm2_capability.c
marcel-hamer: Yes, I will add this check.

lib/tpm2_capability.c
marcel-hamer: I double checked this with other code in the file [here](https://github.com/tpm2-software/tpm2-tools/blob/be47324238c21b6ab55f8080a01b9f312b9783d1/lib/tpm2_capability.c#L157), it was therefor I wrote it like this.

Will change this to the proposed style.

lib/tpm2_capability.c
marcel-hamer: I think at least the GCC compiler accepts this, but I agree it is not that nice to do pointer arithmetic on a void *.  So I will make this more clear as well by modifying the pointer type.

marcel-hamer:
@williamcroberts Added the proposed changes in a new commit.

PR #1196, opened 2018-10-08
tpm2_evictcontrol.1.md: Updated Examples to match current tool options.

idesai:


williamcroberts:
LGTM, i'd like for @joshuagl to take a look since hes been driving the option nightmare recently.

PR #1203, opened 2018-10-18
tests/hash.sh: fix sha256 hash size

igoropaniuk:
`xxd` expects 32 bytes value (sha256 size) instead 256.

`Signed-off-by: Igor Opaniuk <igor.opaniuk@gmail.com>`

### DIFF
--- a/test/integration/tests/hash.sh
+++ b/test/integration/tests/hash.sh
@@ -72,7 +72,7 @@ echo "T0naX0u123abc" > $hash_in_file
 tpm2_hash -a p -G sha256 -Q -o $hash_out_file -t $ticket_file < $hash_in_file
 
 expected=`sha256sum $hash_in_file | awk '{print $1}'`
-actual=`cat $hash_out_file | xxd -p -c 256`
+actual=`cat $hash_out_file | xxd -p -c 32`
 
 test "$expected" == "$actual"
 


test/integration/tests/hash.sh
```diff
@@ -72,7 +72,7 @@ echo "T0naX0u123abc" > $hash_in_file
 tpm2_hash -a p -G sha256 -Q -o $hash_out_file -t $ticket_file < $hash_in_file
 
 expected=`sha256sum $hash_in_file | awk '{print $1}'`
-actual=`cat $hash_out_file | xxd -p -c 256`
```
martinezjavier: hmm, that's correct. How did this didn't fail?

williamcroberts:
```
$ cat hash.dat | xxd -p -c 256 
fa53f49e265b52890e3a7d36b911ec5cef503d655d48ed6c74ae47da9bde80d2
wcrobert@wcrobert-MOBL1:~$ 
wcrobert@wcrobert-MOBL1:~$ cat hash.dat | xxd -p -c 32
fa53f49e265b52890e3a7d36b911ec5cef503d655d48ed6c74ae47da9bde80d2
```
per the man page:
```
       -c cols | -cols cols
              format  <cols> octets per line. Default 16 (-i: 12, -ps: 30, -b:
              6). Max 256.
```

igoropaniuk:
@williamcroberts @martinezjavier
Just a bit of clarification here. The reason for  introducing this fix was that while  running this test with our TEE-based implementation of TPM2.0 logic, the final `$actual` value was filled with a bunch of blanks after the hash value.

For instance, here the output of `test_tpm2_hash.sh` with `set -x`.
selection_006 and pay attention at the second value in the `test` command.

I assumed this could be because of wrong size specified for `xxd`, as for sha1 hash instead of `256` there is a proper value in bytes for SHA1 hash (`actual=`cat $hash_out_file | xxd -p -c 20`, line [63](https://github.com/igoropaniuk/tpm2-tools/blob/2c288464dd1044245321dc0e9bc3b78cd5349317/test/integration/tests/hash.sh#L63))

It could be an issue with TPM logic itself (this is what I'm investigating now), so by creating this PR I just wanted to be sure that it's not a simple typo here (as for sha1 there is `20` bytes instead of `256`)

test/integration/tests/hash.sh
igoropaniuk: @martinezjavier @williamcroberts 


Seems that it's a bit strange behavior with `xxd`:
selection_008

As you see `xxd` adds spaces after just after the actual value.

My version:
```
# xxd --help
BusyBox v1.28.3 (2018-10-12 16:01:33 EEST) multi-call binary.

Usage: xxd [OPTIONS] [FILE]

Hex dump FILE (or stdin)

	-g N		Bytes per group
	-c N		Bytes per line
	-p		Show only hex bytes, assumes -c30
	-l LENGTH	Show only first LENGTH bytes
	-s OFFSET	Skip OFFSET bytes

```

igoropaniuk:
@williamcroberts Just a little suggestion: could you please avoid closing PRs until the discussion discussion is still ongoing? It's really hard to follow it after it's closed
Thanks

williamcroberts:
Okay those details in the commit message would help. Authoring complete commit messages helps. Update your commit message to describe an actual problem or have an open bug ticket referenced with details. Granted it helps if a commit message describes the problem and the solution in the body.

Issue #390, opened 2017-08-02
tpm2_getmanufec: fix memory leak reported via asan


williamcroberts:
Building with ASAN enabled results in:

```
$ tpm2_getmanufec -g rsa -O -N -U -E ECcert.bin -f test_ek.pub -S https://ekop.intel.com/ekcertservice/
WARN: TLS communication with the said TPM manufacturer server setup with SSL_NO_VERIFY!


=================================================================
==12207==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x4b9408 in __interceptor_malloc (/home/wcrobert/workspace/tpm2.0-tools/tools/tpm2_getmanufec+0x4b9408)
    #1 0x7f23ab155e77 in CRYPTO_malloc (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x62e77)

Indirect leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x4b9408 in __interceptor_malloc (/home/wcrobert/workspace/tpm2.0-tools/tools/tpm2_getmanufec+0x4b9408)
    #1 0x7f23ab155e77 in CRYPTO_malloc (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x62e77)
```

williamcroberts:
```
==29488==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x4b96b8 in __interceptor_malloc (/home/wcrobert/workspace/tpm2-tools/tools/tpm2_getmanufec+0x4b96b8)
    #1 0x7f49c7990e77 in CRYPTO_malloc (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x62e77)

Indirect leak of 56 byte(s) in 1 object(s) allocated from:
    #0 0x4b96b8 in __interceptor_malloc (/home/wcrobert/workspace/tpm2-tools/tools/tpm2_getmanufec+0x4b96b8)
    #1 0x7f49c7990e77 in CRYPTO_malloc (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x62e77)

SUMMARY: AddressSanitizer: 80 byte(s) leaked in 2 allocation(s).
```
I think this is an issue in OpenSSL not my code. Maybe I can post to the mailing list and see if anyone of those guys know.

[...]

PR #1239, opened 2018-11-13
ESAPI port

joshuagl:
Here's my mostly complete initial port to ESAPI.

Before we can land this in master we need to resolve the issue Bill described on the mailing list with regards transient handles: https://lists.01.org/pipermail/tpm2/2018-November/000971.html

Once  this has landed in master there are various low-hanging fruit we can work on to improve the codebase following the port:
* we should be able to remove much of the crypo handling code, because ESAPI handles a lot of this for us
* we should be able to drastically reduce the session handling logic, we can use ESYS_TR_Serialize() to serialise an ESYS_TR to disk
* Probably more


joshuagl:
Whoops, scan-build found 8 bugs that need addressing too...

joshuagl:
Rebased on lastest master and resolved scan-build reported issues. Unfortunately several integration tests are failing under clang, so I'll need to reproduce locally and debug next week.

joshuagl:
The clang branch is seeing the same issue as reported in #390, which I think is an issue in the older OpenSSL in our container.

tools/tpm2_startauthsession.c
```diff
@@ -120,21 +120,13 @@ int tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
...
     }
 
-    ESYS_TR session_handle = tpm2_session_get_handle(s);
-    TPMI_SH_AUTH_SESSION tpm_handle;
-    bool result = tpm2_util_esys_handle_to_sys_handle(ectx, session_handle,
-                    &tpm_handle);
-    if (!result) {
-        goto out;
+    if (!ctx.output.path || ctx.output.path[0] == '\0') {
+        ctx.output.path = "session.dat";
```
williamcroberts: session.ctx

man/tpm2_load.1.md
```diff
@@ -35,8 +39,9 @@ into the TPM.
   * **-n**, **--name**=_NAME\_DATA\_FILE_:
     An optional file to save the name structure of the object.
 
-  * **-o**, **--out-context**=_CONTEXT\_FILE_:
-    An optional file to save the object context to.
+  * **-o**, **--out-context**=_CONTEXT\_FILE\_NAME_:
+    The file name of the saved object context, optional. If unspecified defaults
+    to *load.ctx*.
```
williamcroberts: object.ctx

man/tpm2_loadexternal.1.md
```diff
@@ -19,11 +19,13 @@ TPM. The command allows loading of a public area or both a public and a
 sensitive area.
 
 # Output
-The tool outputs the handle and name of the loaded object in a YAML format.
+The tool outputs the name of the loaded object in a YAML format and saves a
+context file for future interactions with the object. The context file name
+defaults to *external.ctx* and can be specified with **-o**.
```
williamcroberts: object.ctx

The problem will be clobbering something that's there.... this even occurs in the current one if they don't specify the name. Perhaps a routine that looks at what's there and does something like:
```
object.ctx
object1.ctx
object2.ctx
```

joshuagl: Great idea, I've got this mostly coded up but will do some more testing in the new year before I push here.

joshuagl: Implemented in fc6484a

joshuagl:
Ugh. Two strict-overflow failures on GCC.

Fortunately the tests all pass on clang with a switch from openssl to gnucrypt for the TSS crypto backend, more work to be done next week but this is very close.

williamcroberts:
Cool. I wonder if now that the build matrix is clang + gnucrypt, if we can
Re-enable the test for getmanufec.

joshuagl:
Sadly not, the leak in getmanufec is in code where we call OpenSSL directly, so switching the TSS to gnucrypt won't avoid it.

joshuagl:
getmanufec leak has been addressed elsewhere in #1270 

The wider leaks causing clang issues may be in the TSS, I've filed tpm2-software/tpm2-tss#1239 to track this and implemented a change to use the gcrypt backend when building the TSS with CC=clang (6b591af).

At this point I'd really like to get this branch merged into master and iterate once there.

joshuagl:
I've found the source of the leak and proposed a fix in tpm2-software/tpm2-tss#1242. This branch has been rebased to remove the switch to gcrypt and add some minor code quality enhancements.

CI will fail but I think it's good to go once the proposed tpm2-tss fix lands ... 🤞

PR #1240, opened 2018-11-16
tpm2_auth_util: cleanup session auth value parsing

williamcroberts:
Rather than have the general entry point deal with the parsing
pf the +, which would break passwords with a + in them, check
if it is a session, copy and modify the argument spliting the
string on the + character and call into the password handler for
the remaining password string if present.

This will support passwords with strings, and remove some
complexity from the code (confusing loop).

### DIFF
--- a/lib/tpm2_auth_util.c
+++ b/lib/tpm2_auth_util.c
@@ -28,6 +28,7 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 // THE POSSIBILITY OF SUCH DAMAGE.
 //**********************************************************************;
+#include <limits.h>
 #include <stdbool.h>
 #include <string.h>
 #include <stdlib.h>
@@ -48,7 +49,7 @@
 #define SESSION_PREFIX "session:"
 #define SESSION_PREFIX_LEN sizeof(SESSION_PREFIX) - 1
 
-static bool handle_hex(const char *password, TPMS_AUTH_COMMAND *auth) {
+static bool handle_hex_password(const char *password, TPMS_AUTH_COMMAND *auth) {
 
     /* if it is hex, then skip the prefix */
     password += HEX_PREFIX_LEN;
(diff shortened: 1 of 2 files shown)


lib/tpm2_auth_util.c
```diff
@@ -110,48 +87,77 @@ static bool handle_str(const char *password, TPMS_AUTH_COMMAND *auth) {
...
+        return false;
+    }
 
-        if (result) {
-            continue;
-        } else {
-            free((char*)password_string_ptr);
-            return false;
+    /*
+     * Sessions can have an associated password/auth value denoted adter
```
idesai: typo -- after

lib/tpm2_auth_util.c
```diff
@@ -110,48 +87,77 @@ static bool handle_str(const char *password, TPMS_AUTH_COMMAND *auth) {
...
+     * a + sign, ie session.ctx+foo, deal with it by
```
idesai: is the order restricted here. As in can one do hex:password+session:session.ctx?

williamcroberts: No that wouldn't work. the flow is session->password as in the session handler we check if their is a +, and if it contains an entry, call the handle_password routine which ONLY handles passwords.

idesai: Cool. Can you please add this info to https://github.com/tpm2-software/tpm2-tools/blob/master/man/common/authorizations.md?

man/common/authorizations.md
```diff
@@ -29,4 +29,23 @@ HMAC tickets can be presented as hex escaped passwords.
 
 When using a policy session to authorize the use of an object, prefix the option argument
 with the *session* keyword.  Then indicate a path to a session file that was created
-with tpm2_startauthsession(1).
+with tpm2_startauthsession(1). Optionally, if the session requires an auth value to be
+sent with the session handle (eg policy secret), then append a + and a string as described
```
idesai: eg policypassword

williamcroberts: Darn.. always switch those two!

idesai: @williamcroberts this PR is definitely much better way of handling this

PR #1243, opened 2018-11-16
Added support for QNX build

safayetahmedatge:
Differences on QNX:

*   Have to define `_QNX_SOURCE` to enable parts of the API.

*   Dynamic-library API (e.g. `dl_open`) is a part of libc. There is no `libdl`.

*   There is no `fmemopen`, `open_memstream`, or equivalent functionality

`Makefile.am`:
Replaced hard-coded `LDFLAGS`, "`-ldl`", with variable, `LIBDL_LDFLAGS`.

`configure.ac`:
Added `$host_os` case statement to detect QNX targets and modify `CFLAGS` and set `LIBDL_LDFLAGS`.

`lib/tpm2_options.c`:
Added `#include <libgen.h>` explicitly for `basename`.
Replaced `linux/limits.h` with `limits.h`. Appears to be a valid change in Ubuntu, Alpine, MinGW64, and QNX.

`lib/tpm2_hash.c`: Modified code to work without `fmemopen`.
- Most of the code previously in `tpm2_hash_file` is now in `tpm2_hash_common` with additional changes.
- `tpm2_hash_common` has the same function header as `tpm2_hash_compute_data`, and an additional `infilep` argument.
- `tpm2_hash_common` uses the`infilep` argument as the data source if non-NULL or uses `inbuffer` and `inbuffer_len`.
- Both `tpm2_hash_file` and `tpm2_hash_compute_data` wrap `tpm2_hash_common`, setting `inbuffer` and `inbuffer_len` to zero, or setting `infilep` to zero, respectively.

Signed-off-by: Safayet Ahmed <safayet.ahmed@ge.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -45,7 +45,7 @@ AM_CFLAGS := \
 AM_LDFLAGS   := $(EXTRA_LDFLAGS) $(CODE_COVERAGE_LIBS)
 
 LDADD = \
-    $(LIB_COMMON) $(TSS2_SYS_LIBS) $(CRYPTO_LIBS) -ldl
+    $(LIB_COMMON) $(TSS2_SYS_LIBS) $(CRYPTO_LIBS) $(LIBDL_LDFLAGS)
 
 # keep me sorted
 bin_PROGRAMS = \
@@ -99,6 +99,7 @@ bin_PROGRAMS = \
     tools/tpm2_startup \
     tools/tpm2_unseal \
     tools/tpm2_verifysignature
+
 
 noinst_LIBRARIES = $(LIB_COMMON)
 lib_libcommon_a_SOURCES = $(LIB_SRC)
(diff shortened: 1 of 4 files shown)


safayetahmedatge:
I looked at the reported failure, but it appears to be incomplete.

> CCLD tools/tpm2_verifysignature
> scan-build: 1 bug found.
> scan-build: Run 'scan-view /tmp/scan-build-2018-11-16-205038-2134-1' to examine bug reports.
> The command "./.ci/travis.run" exited with 1.

Can someone explain what went wrong?

williamcroberts:
Please split this patch into the logical changes:
1. for Makefile ldl change
2. one for limits
3. one for hashing

safayetahmedatge:
Thanks @williamcroberts . I'll split the patch into the three logical changes. Also, I see the bug. Looking into it.

PR #1250, opened 2018-11-23
tests/output_formats.sh: use /dev/urandom

diabonas:
The integration test `output_formats.sh` uses `/dev/random` to retrieve 4096 random bytes. As `/dev/random` blocks if there is not enough entropy available, this can take a very long time one some systems. Using the non-blocking `/dev/urandom` [is perfectly adequate](https://unix.stackexchange.com/questions/324209/when-to-use-dev-random-vs-dev-urandom) and [already done by all other integration tests](https://github.com/tpm2-software/tpm2-tools/blob/5505dc9893406ba6297f3b1711616f5ee5687760/test/integration/tests/encryptdecrypt.sh#L86).

### DIFF
--- a/test/integration/tests/output_formats.sh
+++ b/test/integration/tests/output_formats.sh
@@ -81,7 +81,7 @@ trap cleanup EXIT
 
 start_up
 
-head -c 4096 /dev/random > $file_hash_input
+head -c 4096 /dev/urandom > $file_hash_input
 
 tpm2_createek -Q -G $alg_ek -p "$file_pubek_orig" -c $handle_ek
 


diabonas:
Note:  The CI build is currently broken, see #1251 for a fix.

williamcroberts:
Can you please rebase ontop of master and re-push, I merged #1251

diabonas:
Done, but now the unrelated unit test `test/unit/test_tpm2_alg_util` fails - possibly because https://github.com/tpm2-software/tpm2-tss/pull/1211 got merged upstream? The same error appears for #1253.

safayetahmedatge:
@diabonas thanks for pointing this out. I think I can confirm this. Specifically, I think the problem is at the following line:
https://github.com/tpm2-software/tpm2-tools/blob/db51752736cd7b3d32694e6d227fabac29d7f18b/test/unit/test_tpm2_alg_util.c#L53
When we stringify `value` with `str(value)`, we create strings that look like `"((TPM2_ALG_ID) 0x0001)"`, which `tpm2_alg_util_from_optarg` can't handle. When I replaced the line above with the following, the tests no longer fail:

        char str_value[7] = "0x0000"; \
        int  str_value_len; \
        TPM2_ALG_ID from_hex_str; \
        str_value_len = snprintf(   str_value, \
                                    sizeof(str_value), \
                                    "0x%04x", \
                                    value); \
        assert_int_equal(6, str_value_len); \
        from_hex_str = tpm2_alg_util_from_optarg(str_value, flags); \

The above code might not be the cleanest way to fix this.

diabonas:
@safayetahmedatge Looks good, I confirm that this fixes the problem with the current master of tpm2-tss without breaking compatibility with the released version 2.1.0. It seems to be the correct approach as https://github.com/tpm2-software/tpm2-tss/commit/c1ddad6fbdd56a381afd1f9b00d36052eb0b3181 adds the explicit casts to `TPM2_ALG_ID` you mentioned, so we cannot directly use the expansion of the macros as a string any more. This patch would be worth a pull request to fix the CI build :)

Side note: I don't think the explicit specification (and verification) of the string length is necessary, something like
```
        char str_value[256]; \
        snprintf(str_value, sizeof(str_value), "0x%X", value); \
        TPM2_ALG_ID from_hex_str = tpm2_alg_util_from_optarg(str_value, flags);    \
```
should be enough, see e.g. https://github.com/tpm2-software/tpm2-tools/blob/db51752736cd7b3d32694e6d227fabac29d7f18b/test/unit/test_tpm2_alg_util.c#L125

safayetahmedatge:
@diabonas, done! (https://github.com/tpm2-software/tpm2-tools/pull/1254) Looks like it passed all checks.

PR #1251, opened 2018-11-23
Fix Travis CI build

diabonas:
Since https://github.com/tpm2-software/tpm2-tss/pull/1199 was merged upstream, the Gnulib include directory [needs to be specified during bootstrap of tpm2-tss](https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md#bootstrapping-the-build).


diabonas:
I have no idea why the Clang build failed, but it was a [test failure of `tests/import.sh`](https://travis-ci.org/tpm2-software/tpm2-tools/jobs/458917427#L1615) that should be completely unrelated to the Travis configuration. Maybe just an intermittent glitch?

PR #1255, opened 2018-12-07
Various tweaks

joshuagl:
A few minor changes from my esapi-integration branch which aren't ESAPI related


joshuagl:
I've fixed the tpm_kdfa -> tpm2_kdfa rename and the patch to show test-suite.log (we're in the root of the tpm2-tools repo, not its parent).

williamcroberts:
> I've fixed the tpm_kdfa -> tpm2_kdfa rename and the patch to show test-suite.log (we're in the root of the tpm2-tools repo, not its parent).

Ill take that as is, but I was wondering if a script under .ci that uses `find` would be more robust to deal with things in separate build directories, etc.

PR #1256, opened 2018-12-07
tpm2_sign: fix option error output

williamcroberts:
The error:
ERROR: Expected options (k or c) and (m or D) and s

Is both incorrect and confusing. Correct the "k" output, as "k"
is no longer supported and output them one at a time so the user
doesn't need to figure out what they were missing.

### DIFF
--- a/tools/tpm2_sign.c
+++ b/tools/tpm2_sign.c
@@ -118,8 +118,24 @@ static bool sign_and_save(TSS2_SYS_CONTEXT *sapi_context) {
 
 static bool init(TSS2_SYS_CONTEXT *sapi_context) {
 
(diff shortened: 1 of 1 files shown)


tools/tpm2_sign.c
```diff
@@ -118,9 +118,25 @@ static bool sign_and_save(TSS2_SYS_CONTEXT *sapi_context) {
...
+        option_fail = true;
+    }
+
+    if (!ctx.flags.s) {
+        LOG_ERR("Expected option s");
+        option_fail = true;
+    }
+
+    if (option_fail) {
+        return true;
```
williamcroberts: return false!

PR #1263, opened 2018-12-26
fixed support option -G 'sha256'

u7702045:
change optarg value compare from tpm2_alg_util_flags_sig to tpm2_alg_util_flags_hash


u7702045:
i guess tpm2_quote have typo on 'G' option

williamcroberts:
Thanks for the fix!

PR #1265, opened 2018-12-31
build: fixup  libressl support and add libressl-2.7

alonbl:
Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/lib/tpm2_openssl.h
+++ b/lib/tpm2_openssl.h
@@ -33,6 +33,10 @@
 #include <openssl/err.h>
 #include <openssl/hmac.h>
 
+#if (OPENSSL_VERSION_NUMBER < 0x1010000fL && !defined(LIBRESSL_VERSION_NUMBER)) || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x20700000L) /* OpenSSL 1.1.0 */
+#define LIB_TPM2_OPENSSL_OPENSSL_PRE11
+#endif
+
 /**
  * Function prototype for a hashing routine.
  *
--- a/tools/tpm2_import.c
+++ b/tools/tpm2_import.c
@@ -29,6 +29,25 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //**********************************************************************;
 
+//**********************************************************************;
+// Copyright 1999-2018 The OpenSSL Project Authors. All Rights Reserved.
+// Licensed under the Apache License 2.0 (the "License"). You may not use
+// this file except in compliance with the License. You can obtain a copy
+// in the file LICENSE in the source distribution or at
+// https://www.openssl.org/source/license.html
+//
+// EME-OAEP as defined in RFC 2437 (PKCS #1 v2.0)
+//
+// See Victor Shoup, "OAEP reconsidered," Nov. 2000, <URL:
+// http://www.shoup.net/papers/oaep.ps.Z> for problems with the security
+// proof for the original OAEP scheme, which EME-OAEP is based on. A new
+// proof can be found in E. Fujisaki, T. Okamoto, D. Pointcheval, J. Stern,
+// "RSA-OEAP is Still Alive!", Dec. 2000, <URL:http://eprint.iacr.org/2000/061/>.
+// The new proof has stronger requirements for the underlying permutation:
+// "partial-one-wayness" instead of one-wayness. For the RSA function, this
+// is an equivalent notion.
+//**********************************************************************;
+
 #include <errno.h>
 #include <stdbool.h>
 #include <stdlib.h>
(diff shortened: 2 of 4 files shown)


tools/tpm2_import.c
```diff
@@ -107,6 +107,71 @@ static int RSA_set0_key(RSA *r, BIGNUM *n, BIGNUM *e, BIGNUM *d) {
     return 1;
 }
 #endif
+#if defined(LIBRESSL_VERSION_NUMBER)
+static int RSA_padding_add_PKCS1_OAEP_mgf1(unsigned char *to, int tlen,
+        const unsigned char *from, int flen, const unsigned char *param, int plen,
+        const EVP_MD *md, const EVP_MD *mgf1md) {
```
williamcroberts: I'm not sure if we can legally pull this in, as it's a copy paste from OSSL:
https://github.com/openssl/openssl/blob/master/crypto/rsa/rsa_oaep.c#L41
They're both Apache 2.0...

The import.sh test is failing:
https://travis-ci.org/tpm2-software/tpm2-tools/jobs/473918265#L2055

Also, we should probably have a test that runs against libressl, so we don't regress this support in the future.

williamcroberts: > The import.sh test is failing:
> https://travis-ci.org/tpm2-software/tpm2-tools/jobs/473918265#L2055

Direct leak of 1344 byte(s) in 6 object(s) allocated from:
    #0 0x4dddb8 in malloc (/workspace/tpm2-tools/build/tools/tpm2_import+0x4dddb8)
    #1 0x7f24b65dbe77 in CRYPTO_malloc (/lib/x86_64-linux-gnu/libcrypto.so.1.0.0+0x62e77)

alonbl: What do you suggest regarding RSA_padding_add_PKCS1_OAEP_mgf1? I do not think there are lots of methods to implement this logic :)

alonbl: I checked this again, and there is no reason for import.sh to fail if using openssl as this patch is void. I can confirm it is working in my computer. Can you please look at the patch and tell me if I am wrong?

williamcroberts: > What do you suggest regarding RSA_padding_add_PKCS1_OAEP_mgf1? I do not think there are lots of methods to implement this logic :)

You can't just copy and paste without attribution.

As i've been informed by others:
I think that below your BSD header, if you would a line like "This file incorporates work covered by the following copyright and permission notice:" and then reproduce the header from rsa_oaep.c (below), you'd be fine:

/*
 * Copyright 1999-2018 The OpenSSL Project Authors. All Rights Reserved.
 *
 * Licensed under the Apache License 2.0 (the "License").  You may not use
 * this file except in compliance with the License.  You can obtain a copy
 * in the file LICENSE in the source distribution or at
 * https://www.openssl.org/source/license.html
 */

/* EME-OAEP as defined in RFC 2437 (PKCS #1 v2.0) */

/*
 * See Victor Shoup, "OAEP reconsidered," Nov. 2000, <URL:
 * http://www.shoup.net/papers/oaep.ps.Z> for problems with the security
 * proof for the original OAEP scheme, which EME-OAEP is based on. A new
 * proof can be found in E. Fujisaki, T. Okamoto, D. Pointcheval, J. Stern,
 * "RSA-OEAP is Still Alive!", Dec. 2000, <URL:
 * http://eprint.iacr.org/2000/061/>. The new proof has stronger requirements
 * for the underlying permutation: "partial-one-wayness" instead of
 * one-wayness.  For the RSA function, this is an equivalent notion.
 */
---

williamcroberts: > I checked this again, and there is no reason for import.sh to fail if using openssl as this patch is void. I can confirm it is working in my computer. Can you please look at the patch and tell me if I am wrong?

You must be wrong if Travis is failing. You can use docker locally to enter the build environment that travis is using.

Additionally, we need to run the test suite against a libressl build to verify that nothing regresses.

alonbl: ok, all done, I added the comment + figured out that we use clang.
thanks!

lib/tpm2_openssl.h
```diff
@@ -33,6 +33,10 @@
 #include <openssl/err.h>
 #include <openssl/hmac.h>
 
+#if (OPENSSL_VERSION_NUMBER < 0x1010000fL && !defined(LIBRESSL_VERSION_NUMBER)) || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x20700000L) /* OpenSSL 1.1.0 */
+#define LIB_TPM2_OPENSSL_OPENSSL_PRE11
```
williamcroberts: this makes life so much easier, nice.

PR #1275, opened 2019-01-11
lib/tpm2_auth_util: add support for reading auth passwords from a file

joshuagl:
Add support for reading an authorisation password from a file. This helps
keep the authorisation information private from UNIX utilities such as ps.

### DIFF
--- a/man/common/authorizations.md
+++ b/man/common/authorizations.md
@@ -9,10 +9,12 @@ Authorization for use of an object in TPM2.0 can come in 3 different forms:
 
 ## Passwords
 
-Passwords are interpreted in two forms, string and hex-string. A string password is not
+Passwords are interpreted in three forms; string, hex-string or a file. A string password is not
 interpreted, and is directly used for authorization. A hex-string password is converted from
 a hexidecimal form into a byte array form, thus allowing passwords with non-printable
 and/or terminal un-friendly characters.
+A file form should be the path of a file containing a password in string or hex-string format to be read by the tool.
+Storing passwords in files prevents information leakage, passwords passed as options can be read from the process list.
 
 By default passwords are assumed to be in the string form. Password form is specified
 with special prefix values, they are:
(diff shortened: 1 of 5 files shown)


man/common/authorizations.md
```diff
@@ -9,17 +9,21 @@ Authorization for use of an object in TPM2.0 can come in 3 different forms:
 
 ## Passwords
 
-Passwords are interpreted in two forms, string and hex-string. A string password is not
+Passwords are interpreted in three forms; string, hex-string and file. A string password is not
```
martinezjavier: shouldn't this be and **a** file ?

test/integration/tests/changeauth.sh
```diff
@@ -48,6 +53,7 @@ tpm2_changeauth -o $ownerPasswd -e $endorsePasswd -l $lockPasswd
 
 tpm2_changeauth -O $ownerPasswd -E $endorsePasswd -L $lockPasswd -o $new_ownerPasswd -e $new_endorsePasswd -l $new_lockPasswd
 
-tpm2_clear -L $new_lockPasswd
```
martinezjavier: I wonder if we should leave this too to make sure that there's no regression when not providing the password with a file. I think that calling clear twice won't be an issue.

man/common/authorizations.md
joshuagl: Well spotted, I think this should probably be "string, hex-string or a file". Will address.

test/integration/tests/changeauth.sh
joshuagl: That seems reasonable, I'll add this back in.

joshuagl: I had to add in an extra *tpm2_changeauth* before making an extra *tpm2_clear*

martinezjavier: > I had to add in an extra _tpm2_changeauth_ before making an extra _tpm2_clear_

Sounds good to me, I think the more coverage the better.

PR #1290, opened 2019-01-25
tests/changeauth: test changeauth on object

williamcroberts:
Test that changing the authvalue on an object works.

### DIFF
--- a/test/integration/tests/changeauth.sh
+++ b/test/integration/tests/changeauth.sh
@@ -34,7 +34,7 @@
 source helpers.sh
 
 cleanup() {
-    rm secret.txt
+    rm secret.txt key.ctx key.pub key.priv primary.ctx
 }
 trap cleanup EXIT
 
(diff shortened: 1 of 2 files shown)


joshuagl:
Build failure is due to tpm2-software/tpm2-tss#1238 reported as tpm2-software/tpm2-tss#1255 and proposed fix in tpm2-software/tpm2-tss#1256

williamcroberts:
@joshuagl somehow this test, and my subsequent changes in tpm2_ptool.py, both were working yesterday and I am not sure how. Today, everything blew up, and my investigation turned out that in tpm2_changeauth, two bugs existed:

First, the switch statement to see if it's NV/TRANSIENT/PCR was wrong, because on the PERSISTENT case the handle is 0 and only the ESYS_TR handle is present. So I had to modify that switch.

Second, the object loading code, when given an actual TPM handle (like for a persistent handle), never sets the esys TR object handle which is required to do that conversion.

Iv'e included these updates in this PR... still need to do full testing.

PR #1293, opened 2019-01-28
build: Use dbus-run-session in automake test harness.

flihp:
This commit removes the manual launching of the dbus session bus
instance in the docker.run file. This is replaced by adding the
'dbus-run-session' command to the AM_TEST_ENVIRONMENT. This allows
callers to simply execute 'make check' and the test harness will set up
a session bus instance automagically.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -35,23 +35,6 @@ set -e
 
 source $TRAVIS_BUILD_DIR/.ci/docker-prelude.sh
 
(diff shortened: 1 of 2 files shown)


williamcroberts:
Can you push this to pkcs11 as well?

diabonas:
> Can you push this to pkcs11 as well?

I added this to https://github.com/tpm2-software/tpm2-pkcs11/pull/118.

PR #1298, opened 2019-01-31
Bugfix: tpm2_createek tool auto-save context used incorrect path

jetwhiz:
In the latest version of tpm2_createek, it appears that saving the context for non-persistent handles is not targeting the correct variable.  

According to the documentation, the "-c" option, which sets `ctx.ctx_obj.path`, specifies the context output file to use or "ek.ctx" if none is given.  The current implementation, however, will use `ctx.out_file_path` (set by "-p") which is supposed to be the public key output file.  This PR is designed to correct this.

### DIFF
--- a/tools/tpm2_createek.c
+++ b/tools/tpm2_createek.c
@@ -174,38 +174,44 @@ static bool create_ek_handle(ESYS_CONTEXT *ectx) {
             LOG_PERR(Esys_FlushContext, rval);
             return false;
         }
-    } else if (ctx.ctx_obj.path) {
+    } else {
+        /* If it wasn't persistent, save a context for future tool interactions */
+        char *filename = NULL;
+
+        if (!ctx.ctx_obj.path) {
+            /* Ensure the context file path is unique, we don't want to clobber
+            * existing files.
+            */
+            bool ok = files_get_unique_name("ek.ctx", &filename);
+            if (!ok) {
+                return false;
+            }
+        } else {
+            /* Make a copy of specified path so we can free properly below */
+            filename = strdup(ctx.ctx_obj.path);
+            if (!filename) {
+                LOG_ERR("oom");
+                return false;
+            }
+        }
+
         bool result = files_save_tpm_context_to_path(ectx,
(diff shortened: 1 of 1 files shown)


williamcroberts:
I think this is more of a documentation fix. The tools shouldn't be producing output to a file via fixed name, and they should clobber. Theirs definitely a bug here as your commit mentions, but lets fix that bug and correct the doc.

jetwhiz:
Hi @williamcroberts, 

I agree about the tool clobbering existing files, to keep it in line with how the other tools behave.  Should I update this PR just to remove the `ctx.out_file_path`-specific issues, without adding the fixed name fix-ups to `ctx.ctx_obj.path` (and update the documentation), or does this need more discussion?

williamcroberts:
Oh I didn't realize that generate and create name portion was part of it or I likely would have balked at it. I'm trying to think of popular commandline tools on linux that have similar behavior.

joshuagl:
I can't think of a popular tool with similar behaviour, but overwriting a context file is an irksome way to shoot oneself in the foot. If we are going to change the behaviour I think we should probably default to requiring a filename be passed for the context file (rather than having that argument be optional) as more required options seems friendlier than unexpected behaviour. What do you think?

williamcroberts:
> I can't think of a popular tool with similar behaviour, but overwriting a context file is an irksome way to shoot oneself in the foot. If we are going to change the behaviour I think we should probably default to requiring a filename be passed for the context file (rather than having that argument be optional) as more required options seems friendlier than unexpected behaviour. What do you think?

I agree, the context file would become a mandatory option so you can't shoot yourself in the foot.

jetwhiz:
Should I update this PR to make providing a context option mandatory and fix the issue involving the ctx.out_file_path?

martinezjavier:
> I agree, the context file would become a mandatory option so you can't shoot yourself in the foot.

agree with this as well.

martinezjavier:
> Should I update this PR to make providing a context option mandatory and fix the issue involving the ctx.out_file_path?

I think your fix is correct. We can make the option to be mandatory and explicit in a follow-up patch, since that's really independent of your fix.

PR #1299, opened 2019-01-31
Add ability to run tpm2_makecredential without a TPM

jetwhiz:
Performing the Make Credential routine does not require usage of a TPM, though the current implementation relies on leveraging the TPM to perform this function.  This PR adds functionality to tpm2_makecredential that allows it to perform this routine without needing a TPM to be present, supporting RSA right now. 

A large portion of this PR involves moving shared code from tpm2_import.c into a new library called tpm2_identity_util.c, since the TPM-less makecredential also will use these functions.  Some of these functions were renamed to be compliant with the library naming conventions (e.g., begin with "tpm2_identity_util_XX"). 

The function `tpm2_identity_util_encrypt_seed_with_public_key` was also generalized to accept either IDENTITY mode or DUPLICATE mode (so it still supports tpm2_import). 

In order to toggle between the TPM and TPM-less versions, we have added a new global -X/--no-tpm option for any tools that can operate w/o a TPM (though right now makecredential is the only such tool).

### DIFF
--- a/lib/tpm2_identity_util.c
+++ b/lib/tpm2_identity_util.c
@@ -1,0 +1,482 @@
+//**********************************************************************;
+// Copyright (c) 2017, Intel Corporation
+// Copyright (c) 2019 Massachusetts Institute of Technology
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//**********************************************************************
+
+#include <errno.h>
+#include <stdint.h>
+#include <string.h>
+
+#include <openssl/aes.h>
+#include <openssl/bn.h>
+#include <openssl/evp.h>
+#include <openssl/err.h>
+#include <openssl/hmac.h>
+#include <openssl/obj_mac.h>
+#include <openssl/pem.h>
+#include <openssl/rand.h>
+#include <openssl/rsa.h>
+
+#include <tss2/tss2_mu.h>
+#include <tss2/tss2_sys.h>
+
+#include "files.h"
+#include "log.h"
+#include "tpm2_alg_util.h"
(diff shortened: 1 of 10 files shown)


tools/tpm2_makecredential.c
```diff
@@ -114,6 +120,130 @@ static bool write_cred_and_secret(const char *path, TPM2B_ID_OBJECT *cred,
...
+    tpm2_util_hexdump(encrypted_sensitive.buffer, encrypted_sensitive.size);
+    tpm2_tool_output("\n");
+
+    tpm2_tool_output("outer_hmac %u: ", outer_hmac.size);
+    tpm2_util_hexdump(outer_hmac.buffer, outer_hmac.size);
+    tpm2_tool_output("\n");
+
+    tpm2_tool_output("encrypted_seed: ");
+    tpm2_util_hexdump(encrypted_seed.secret, encrypted_seed.size);
+    tpm2_tool_output("\n");
```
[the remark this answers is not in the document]
jetwhiz: These were mostly for debugging, so they have been removed

jetwhiz:
Thank you for the feedback, @williamcroberts!  The latest commit cleans up the code as you requested, and the YAML outputs have been removed since they are not necessary.  Also, it looks like the the CLANG CI is happier and shows _passing_ status.

lib/tpm2_identity_util.c
```diff
@@ -0,0 +1,480 @@
+//**********************************************************************;
+// Copyright (c) 2017, Intel Corporation
```
martinezjavier: Should't this be MIT? Or did you take this code from Intel?

lib/tpm2_identity_util.c
```diff
@@ -0,0 +1,480 @@
...
+    
+    switch (alg) {
+    case TPM2_ALG_RSA:
+        result = encrypt_seed_with_tpm2_rsa_public_key(protection_seed, 
+                parent_pub, label, labelLen, encrypted_protection_seed);
+        break;
+    case TPM2_ALG_ECC:
+        LOG_ERR("Algorithm '%s' not supported yet", tpm2_alg_util_algtostr(alg,
+            tpm2_alg_util_flags_any));
+        result = false;
```
martinezjavier: Do you have plans to support ECC in the short term? Just out of curiosity.

lib/tpm2_identity_util.h
```diff
@@ -0,0 +1,140 @@
+//**********************************************************************;
+// Copyright (c) 2017, Intel Corporation
```
martinezjavier: Same comment than before, this need to have the proper copyright.

lib/tpm2_options.c
```diff
@@ -338,13 +338,14 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv,
         { "quiet",         no_argument,       NULL, 'Q' },
         { "version",       no_argument,       NULL, 'v' },
         { "enable-errata", no_argument,       NULL, 'Z' },
+        { "no-tpm",        no_argument,       NULL, 'X' },
```
martinezjavier: I'm not sure if I like the `no-tpm` option name. Maybe `use-openssl`? or `openssl-backend`?

lib/tpm2_identity_util.c
martinezjavier: I see now that you took some bits from `tpm2_import.c`, I should had go through all the patch before commenting. In that case I think you should add both Copyrights since you are also adding new code.

lib/tpm2_identity_util.c
```diff
@@ -0,0 +1,480 @@
...
+        goto err;
+    for (i = 0; i < mdlen; i++)
+        seed[i] ^= seedmask[i];
+
+    OPENSSL_free(dbmask);
+    return 1;
+
+ err:
+    OPENSSL_free(dbmask);
+    return 0;
```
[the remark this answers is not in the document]
jetwhiz: This code was just moved over from the tpm2_import tool, but I agree that it is more readable with this change.  The function has been updated to address this.

lib/tpm2_identity_util.c
```diff
@@ -0,0 +1,480 @@
...
+}
+#endif
+
+static TPM2_KEY_BITS get_pub_asym_key_bits(TPM2B_PUBLIC *public) {
+
+    TPMU_PUBLIC_PARMS *p = &public->publicArea.parameters;
+    switch(public->publicArea.type) {
+    case TPM2_ALG_ECC:
+        /* fall-thru */
```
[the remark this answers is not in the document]
jetwhiz: This was moved over from tpm2_import as well, but that's what it appears is happening here.  According [to this blog](https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/) it seems that `fall-thru` is also acceptable (at least up to `-Wimplicit-fallthrough` level 3).

lib/tpm2_identity_util.c
jetwhiz: We are definitely interested in adding ECC support, though we don't have an exact timeline on when this would be implemented.  At least for now the priority has been for having RSA support, but ECC support is also a feature we are interested in having for keylime (see https://github.com/keylime/python-keylime/issues/28).

lib/tpm2_options.c
jetwhiz: I agree that using `openssl-backend` is definitely more clear about what is happening than just `no-tpm` -- this long option has been updated to reflect this.  The documentation and integration test has also been updated.

PR #1300, opened 2019-01-31
Add ability to check quotes and output PCR values for quotes

jetwhiz:
Due to a large overlap of changes, this PR contains two new features: 
* Enhance the `tpm2_quote` tool so that it can now output selected PCR hash lists
* Add new `tpm2_checkquote` tool for checking that TPM quotes are valid

The TPM2 quote no longer contains the actual PCR hash values chosen (compared to TPM 1.2), but instead contains only the aggregate hash for these PCR hashes. In many situations it is desirable to have the actual hash values themselves rather than only an aggregate, and implementing a (pcrlist, quote, pcrlist, validate) routine outside a convenient tool such as tpm2_quote just adds complexity and communication overhead.  This PR adds the ability to output the PCR hashes related to a quote after comparing their aggregate hash to the one signed in the quote to ensure they are current. 

The second feature added with this PR directly relates to the new quote functionality described above: checking quotes, including PCR outputs.  The `tpm2_checkquote` tool will validate a quote's signature and will ensure that the PCR list provided matches the aggregate provided in the quote (in case this list got tampered with along the way). 

A large part of this PR involves moving needed routines out of tpm2_pcrlist.c and into the _pcr_ library (since they are now shared code). The `is_pcr_select_bit_set` function was moved into the _tpm2_util_ library. Some functions were renamed to be compliant with the library naming standards (e.g., "pcr_XX" prefix). 

New functionality was added to: 
* openssl library: 
  * `tpm2_openssl_halgid_from_tpmhalg` which converts TPM hash IDs to openssl hash IDs
  * `tpm2_openssl_hash_compute_data` hashes a byte buffer into a TPM2B_DIGEST
  * `tpm2_openssl_hash_pcr_banks` computes the aggregate hash of all selected PCR banks
  * `tpm2_openssl_get_public_RSA_from_pem` a lower-level version of _load_public_RSA_from_pem_ which just returns the RSA object instead of loading it into the TPM (and a related version for ECC).
* util library:
  * `tpm2_util_get_digest_from_quote` pulls the aggregate PCR hash out of a TPM quote
  * `tpm2_util_verify_digests` which verifies that two digests match each other
* pcr library:
  * `pcr_print_pcr_struct` prints out the PCR hashes in YAML format (for tool output)

Also updated are man pages and integration tests for quote and checkquote.


lib/pcr.c
```diff
@@ -173,3 +287,122 @@ bool pcr_parse_list(const char *str, size_t len, TPMS_PCR_SELECTION *pcrSel) {
 
     return true;
 }
+
+
+
```
williamcroberts: ETOOMANY newlines

lib/tpm2_openssl.c
```diff
@@ -424,7 +565,7 @@ static bool load_public_RSA_from_key(RSA *k, TPM2B_PUBLIC *pub) {
     return BN_bn2bin(e, (unsigned char *)&rdetail->exponent);
 }
 
-static bool load_public_RSA_from_pem(FILE *f, const char *path, TPM2B_PUBLIC *pub) {
+RSA* tpm2_openssl_get_public_RSA_from_pem(FILE *f, const char *path) {
```
williamcroberts: `RSA *tpm2_` not `RSA* tpm2_`

lib/tpm2_openssl.h
```diff
@@ -30,9 +30,13 @@
 
 #include <tss2/tss2_sys.h>
 
+#include <openssl/ec.h>
 #include <openssl/err.h>
 #include <openssl/hmac.h>
 
+#include "pcr.h"
+
+
```
williamcroberts: too many newlines

lib/tpm2_openssl.h
```diff
@@ -53,6 +57,15 @@
...
 typedef unsigned char *(*digester)(const unsigned char *d, size_t n, unsigned char *md);
 
+/**
+ * Get an openssl hash algorithm ID from a tpm hashing algorithm ID.
+ * @param algorithm
+ *  The tpm algorithm to get the corresponding openssl version of.
+ * @return
+ *  The openssl hash algorithm id.
+ */
+UINT16 tpm2_openssl_halgid_from_tpmhalg(TPMI_ALG_HASH algorithm);
```
williamcroberts: is it really a uint16, usually openssl is int heavy

lib/tpm2_openssl.c
```diff
@@ -39,14 +39,33 @@
...
 #include <tss2/tss2_sys.h>
 
 #include "files.h"
 #include "log.h"
 #include "tpm2_alg_util.h"
+#include "tpm2_kdfa.h"
 #include "tpm2_openssl.h"
 #include "tpm2_util.h"
 
+UINT16 tpm2_openssl_halgid_from_tpmhalg(TPMI_ALG_HASH algorithm) {
```
williamcroberts: you're returning ints AFAICT

tools/tpm2_checkquote.c
```diff
@@ -0,0 +1,411 @@
...
+            { "format",       required_argument, NULL, 'f' },
+            { "sig",          required_argument, NULL, 's' },
+            { "pcrs",         required_argument, NULL, 'p' },
+            { "pubkey",       required_argument, NULL, 'c' },
+            { "qualify-data",         required_argument, NULL, 'q' },
+    };
+
+
+    *opts = tpm2_options_new("G:m:f:s:t:c:p:q:", ARRAY_LEN(topts), topts,
+                             on_option, NULL, TPM2_OPTIONS_NO_SAPI);
```
williamcroberts: tools that don't need a SAPI go to tools/misc

lib/tpm2_openssl.h
jetwhiz: Yes I think that's correct, the RSA_verify tool that this is used for is expecting an int.  This has been updated.

tools/tpm2_checkquote.c
jetwhiz: Tool has now been moved to tools/misc!

jetwhiz:
> Looks reasonable few nits. Move the new tool to tools/misc since it doesn't use a sapi.
> 
> Fix the memory issues reported by ASAN.

Hi @williamcroberts, thank you for reviewing!  The memory issue has now been fixed and the code cleaned up as requested.  The tool has also been moved to tools/misc now.  It looks like the CI is having some weird issue related to setting up the simulator and resource manager, however: 

````
WARNING:tcti:src/util/io.c:248:socket_connect() Failed to connect to host ::1, port 33118: errno 99: Cannot assign requested address 
````

Is this something we can fix on our end?

**Comment by martinezjavier** [comment]
@jetwhiz when fixing issues pointed out in your pull-requests, please don't add a fixup commit but instead fix it in the commits that have the issues and force push your branch. I usually do something like the following:

```
$ git rebase -i origin/master
$ git commit --amend
$ git rebase --continue
$ git push keylime feat-chkquote -f
```

Since the old version is not preserved, I usually also add message with a changelog, something like:

Changes since v1:
* Fix memory leak introduced in pcr library's pcr_get_banks func
* OpenSSL library's tpm2_openssl_halgid_from_tpmhalg now returns int
* tpm2_checkquote moved to tools/misc (does not rely on SAPI)
* Code cleanup and remove commented out code

**Comment by jetwhiz** [comment]
Hi @martinezjavier -- I have squashed the commits for both feature PRs.  It also looks like the strange error with the CI is now gone after re-building this commit.

File: man/tpm2_checkquote.1.md [hunk 7]
```diff
@@ -0,0 +1,93 @@
...
+provided, verify that the qualifying data and PCR values match those in the quote.
+
+# OPTIONS
+
+  * **-c**, **--key-context**=_KEY\_CONTEXT\_OBJECT_:
+
+    Context object for the key context used for the operation. Either a file
+    or a handle number. See section "Context Object Format".
+
+  * **-G**, **--halg**=_HASH\_ALGORITHM_:
```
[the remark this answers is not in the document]
martinezjavier: @jetwhiz I see that `tpm2_quote` does use `-G` indeed. But I think that we should change it too to `-g` to be consistent with the other tools, what do you think @williamcroberts @joshuagl ?

joshuagl:
> @williamcroberts @joshuagl should we link to the [Deveoper Certificate or Origin document](https://developercertificate.org/l) in CONTRIBUTING.md?

That seems prudent. We should probably add the [Developer Certificate of Origin](https://probot.github.io/apps/dco/) GitHub app to detect this early.
 
> Also, should we add a section documenting that issues bring up on review should be squashed in the original commits and force push your branch? Or is that a common Github workflow so it doesn't need to be documented?

I honestly don't know whether it's common GitHub workflow or not, but it's not too much effort to spell out the expected workflow in the CONTRIBUTING.md

man/tpm2_checkquote.1.md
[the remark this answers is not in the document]
joshuagl: `-G` and `-g` are both used as short options for `--halg`, it's tricky to ensure all tools are using consistent short options so what we're currently aiming for is tools of the same TPM command group to use consistent short options (see [scripts/utils/analyze.py](https://github.com/tpm2-software/tpm2-tools/blob/master/scripts/utils/analyze.py)).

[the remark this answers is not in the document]
martinezjavier: @joshuagl got it, makes sense.

jetwhiz:
Thanks for pointing that out @joshuagl and @martinezjavier -- a new commit has been pushed that is now signed-off.

man/tpm2_checkquote.1.md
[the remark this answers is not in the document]
williamcroberts: @joshuagl @martinezjavier the author is making tpm2_quote match tpm2_certify, which is what is proper. However, the set of commands is just both tools and neither use lowercase `-g`, so for 4.0 we should flip these usages of -G to -g to be as consistent as possible. See #1315

martinezjavier:
> @joshuagl I don't think I appreciate what you're saying here. Can you enlighten me, perhaps offline.

@williamcroberts I think he meant that we may spend a paragraph in CONTRIBUTING.md mentioning the expected workflow and best practices for contribution, i.e:

* Don't mix bug fixes and new features in the same commit, split the changes so each commit just changes one thing.
* Add a 
* Always send a pull-request first against the master branch, and once is merged send a new pull-request with the backports for the stable branches if applicable.
* If there's feedback for your changes, don't add fixup commits but instead fix the commits and force push a new branch.

I asked if there are normal Github workflow or if we should have some docs, Joshua thinks that's worth to write it down. I can propose something tomorrow.

man/tpm2_checkquote.1.md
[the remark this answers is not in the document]
martinezjavier: > @joshuagl @martinezjavier the author is making tpm2_quote match tpm2_certify, which is what is proper. However, the set of commands is just both tools and neither use lowercase `-g`, so for 4.0 we should flip these usages of -G to -g to be as consistent as possible. See #1315

@williamcroberts agreed.

PR #1310, opened 2019-02-09
tpm2_createek EK template and EK nonce

jhand2:
Add functionality to tpm2_createek to read EK template and EK nonce to override `inPublic->publicArea` when present.

Relevant TCG spec regrading EK template and EK nonce (pages 11 and 12): https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_Credential_Profile_EK_V2.1_R13.pdf

Additions:

- refactors code from `tpm2_nvread` to create a function in `tpm2_nv_util.h` to read data from an arbitrary nv index.
- Adds `-t` option to tpm2_createek to use ek template and ek nonce instead of default EK parameters. Only supports ecc and rsa, as these are the only algorithms defined in the EK Credential Profile spec. If the `-t` option is used and there is no ek template, execution with fail (although I think this behavior should be open to discussion in this PR).
- Adds a step to the `createek.sh` integration test to test use of the EK template and EK nonce


tools/tpm2_createek.c
```diff
@@ -266,9 +346,10 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
         { "file",                 required_argument, NULL, 'p' },
         { "format",               required_argument, NULL, 'f' },
         { "context",              required_argument, NULL, 'c' },
+        { "template",             required_argument, NULL, 't' },
```
jhand2: Need to figure out how to add documentation in the manpage for this option

PR #1312, opened 2019-02-11
tpm2_create: add TPM CreateLoaded support

williamcroberts:
Allow tpm2_create to take the -c/--context output option to
produce a context file similar to tpm2_load()

Internally, this will yeild a call to ESYS_CreateLoaded over
ESYS_Create thus producing the loaded ctx object.

Fixes: #832

### DIFF
--- a/man/tpm2_create.1.md
+++ b/man/tpm2_create.1.md
@@ -72,6 +72,13 @@ These options for creating the tpm entity:
   * **-r**, **--privfile**=_OUTPUT\_PRIVATE\_FILE_:
     The output file which contains the sensitive portion of the object, optional.
 
+  * **-o**, **--out-context**=_OUTPUT\_CONTEXT\_FILE_:
+    The output file which contains the key context, optional. The key context is analogous to the context
+    file produced by tpm2_load(1), however is generated via a TPM CreateLoaded command. This this option
+    can be used to avoid the normal *tpm2_create* and *tpm2_load* command sequences and do it all in one
+    command, atomically.
+
+
 [common options](common/options.md)
 
 [common tcti options](common/tcti.md)
@@ -90,10 +97,19 @@ These options for creating the tpm entity:
 
 # EXAMPLES
 
+Create an object whose parent is provided via parent.ctx:
 ```
-tpm2_create -C 0x81010001 -P abc123 -K def456 -g sha256 -G keyedhash-I data.File
-tpm2_create -C file:parent.context -P abc123 -p def456 -g sha256 -G keyedhash -I data.File
-tpm2_create -C 0x81010001 -P 123abc -K 456def -X -g sha256 -G keyedhash -I data.File
+tpm2_create -C parent.ctx -u obj.pub obj.priv
+```
+
+Create an object and seal data to it:
+```
+tpm2_create -C parent.ctx  -K def456 -G keyedhash -I seal.dat -u obj.pub -r obj.priv
+```
+
+Create an rsa2048 object and load it into the TPM:
+```
+tpm2_create -C primary.ctx -G rsa2048 -u obj.pub -r obj.priv -o obj.ctx
 ```
 
 # RETURNS
--- a/tools/tpm2_create.c
+++ b/tools/tpm2_create.c
@@ -39,6 +39,7 @@
 #include <string.h>
 
 #include <tss2/tss2_esys.h>
+#include <tss2/tss2_mu.h>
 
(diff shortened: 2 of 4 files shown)


tools/tpm2_create.c
```diff
@@ -119,39 +119,95 @@ static bool create(ESYS_CONTEXT *ectx) {
...
+            ESYS_TR parentHandle,
+            ESYS_TR shandle1,
+            ESYS_TR shandle2,
+            ESYS_TR shandle3,
+            const TPM2B_SENSITIVE_CREATE *inSensitive,
+            const TPM2B_TEMPLATE *inPublic,
+            ESYS_TR *objectHandle,
+            TPM2B_PRIVATE **outPrivate,
+            TPM2B_PUBLIC **outPublic);
+         */
```
williamcroberts: whoops, delete me

tools/tpm2_create.c
```diff
@@ -220,9 +266,10 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
       { "pubfile",              required_argument, NULL, 'u' },
       { "privfile",             required_argument, NULL, 'r' },
       { "context-parent",       required_argument, NULL, 'C' },
+      { "context",              required_argument, NULL, 'c' },
```
joshuagl: `-c`/`--context` is by convention a child context for a command, would this be better as `-o`/`--out-context` like `tpm2_load` ?

tools/tpm2_create.c
```diff
@@ -119,39 +119,82 @@ static bool create(ESYS_CONTEXT *ectx) {
...
+        if(rval != TPM2_RC_SUCCESS) {
+            LOG_PERR(Esys_Create, rval);
+            return false;
+        }
+        free(creationData);
+        free(creationHash);
+        free(creationTicket);
     }
 
+
```
joshuagl: Spurious whitespace?

tools/tpm2_create.c
```diff
@@ -119,39 +119,82 @@ static bool create(ESYS_CONTEXT *ectx) {
...
-    rval = Esys_Create(ectx, ctx.context_object.tr_handle,
-            shandle1, ESYS_TR_NONE, ESYS_TR_NONE,
-            &ctx.in_sensitive, &ctx.in_public, &outsideInfo, &creationPCR,
-            &outPrivate, &outPublic, &creationData, &creationHash,
-            &creationTicket);
-    if(rval != TPM2_RC_SUCCESS) {
-        LOG_PERR(Esys_Create, rval);
-        ret = false;
-        goto out;
+    ESYS_TR object_handle;
```
joshuagl: Initialise this to ESYS_TR_NONE to placate scan-build?

tools/tpm2_create.c
williamcroberts: Yeah I thought about this as well, initially  I wasn't happy with -o option on tpm2_load but considering you can load a child or parent, it makes more sense.

williamcroberts: Actually, -c is really "this object context". For instance tpm2_readpublic, where as -C is this objects parent.

williamcroberts: Ill keep it same as tpm2_load and we can figure out the proper ones and be consistent when we go through all the options for release.

man/tpm2_create.1.md
```diff
@@ -90,10 +97,19 @@ These options for creating the tpm entity:
...
+```
+tpm2_create -C parent.ctx  -K def456 -G keyedhash -I seal.dat -u obj.pub -r obj.priv
+```
+
+Create an rsa2048 object and load it into the TPM:
 ```
-tpm2_create -C 0x81010001 -P abc123 -K def456 -g sha256 -G keyedhash-I data.File
-tpm2_create -C file:parent.context -P abc123 -p def456 -g sha256 -G keyedhash -I data.File
-tpm2_create -C 0x81010001 -P 123abc -K 456def -X -g sha256 -G keyedhash -I data.File
+tpm2_create -C primary.ctx -G rsa2048 -u obj.pub -r obj.priv -c obj.ctx
```
joshuagl: Should be -o obj.ctx now

williamcroberts: Good catch... these always seem to miss being updated

williamcroberts:
> Hmm, tpm2_readpublic call failed in create.sh integration test.

Helps if you change the case statement to 'o' from 'c' as well.

PR #1318, opened 2019-02-14
INSTALL.MD: Add missing dependency to autoconf-archive

webmeister:
This was already present for Fedora, but was missing from the description for Ubuntu.


webmeister:
> But your commit is missing your Signed-off-by tag.

I wanted to do it quickly and used GitHub's web interface, so I missed that. Now I can't find an amend button in the web interface and need to fix it the usual way again...

PR #1319, opened 2019-02-14
tpm2_capability: Fix infinite loop and out-of-bounds memory access

### LINKED ISSUE #1311: tpm2_getcap -c properties-fixed segfaults
When I run `tpm2_getcap -c properties-variable` everything works, when I run `tpm2_getcap -c properties-variable` fixed the tool dies with a segault.

Running it in gdb points it to:
```
Thread 1 "tpm2_getcap" received signal SIGSEGV, Segmentation fault.
0x004041a4 in tpm2_capability_get (ectx=0x7efff08f, ectx@entry=0x438400, capability=6, property=256, count=2130702480, capability_data=0x7efff0e8, capability_data@entry=0x7efff0e0) at lib/tpm2_capability.c:134
134                     updated->data.algorithms.algProperties[i] =
```
According to strace it seems to be a bit stuck in a loop:
[getcap.strace.txt](https://github.com/tpm2-software/tpm2-tools/files/2852970/getcap.strace.txt)

Tested with a physical tpm

webmeister:
TPM2_GetCapability may return its data in one of several formats, depending
on the capability parameter. All formats are part of a union, but they may
have a different internal structure, so specific union members need to be
accessed to avoid accessing data in the wrong format.

If moreData is set, the next request needs to use a higher value for
property in order to prevent infinite loops. Ideally, the highest value
returned by the TPM +1 is used.

While reworking the code, also avoid unnecessary memory allocations and
just reuse the structure already allocated by Esys_GetCapability.

Fixes #1311.

Signed-off-by: Alexander Steffen <Alexander.Steffen@infineon.com>

### DIFF
--- a/lib/tpm2_capability.c
+++ b/lib/tpm2_capability.c
@@ -40,50 +40,23 @@
 
 #include "tpm2_capability.h"
 
-static size_t tpm2_get_property_data_size(TPM2_CAP capability) {
-
-    size_t size;
-    TPMS_CAPABILITY_DATA dummy;
-
-    switch (capability) {
-        case TPM2_CAP_ALGS:
-            size = sizeof(dummy.data.algorithms.algProperties[0]);
-            break;
-        case TPM2_CAP_HANDLES:
-            size = sizeof(dummy.data.handles.handle[0]);
-            break;
-        case TPM2_CAP_COMMANDS:
-            size = sizeof(dummy.data.command.commandAttributes[0]);
-            break;
-        case TPM2_CAP_PP_COMMANDS:
-            size = sizeof(dummy.data.ppCommands.commandCodes[0]);
-            break;
-        case TPM2_CAP_AUDIT_COMMANDS:
-            size = sizeof(dummy.data.auditCommands.commandCodes[0]);
-            break;
-        case TPM2_CAP_PCRS:
-            size = sizeof(dummy.data.assignedPCR.pcrSelections[0]);
-            break;
-        case TPM2_CAP_TPM_PROPERTIES:
-            size = sizeof(dummy.data.tpmProperties.tpmProperty[0]);
-            break;
-        case TPM2_CAP_PCR_PROPERTIES:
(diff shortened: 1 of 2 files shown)


webmeister:
I'm trying to reproduce the error from the one failing test case. Is there any way to get more information out of the Travis run? It prints tons of logs while compiling, but then during the tests I only see a single line `FAIL: test/integration/tests/verifysignature.sh`. test-suite.log does not contain much information either, just `Done. Your build exited with 1.`

joshuagl:
This was an intermittent failure we sometimes see in CI, I restarted the build and it was green.

The line above "Done. Your build exited with 1." should be "$ cat build/test-suite.log", if you click on that line it will expand to show the full contents of the test-suite.log file.

For example, try in https://travis-ci.org/tpm2-software/tpm2-tools/jobs/493210274

webmeister:
> The line above "Done. Your build exited with 1." should be "$ cat build/test-suite.log", if you click on that line it will expand to show the full contents of the test-suite.log file.

Right, I hadn't noticed the hidden line numbers. That seems more helpful, thanks :)

lib/tpm2_capability.c
```diff
@@ -40,118 +40,123 @@
...
+            case TPM2_CAP_ALGS:
+                APPEND_CAPABILITY_INFORMATION(algorithms, algProperties, .alg, TPM2_MAX_CAP_ALGS);
+                break;
+            case TPM2_CAP_HANDLES:
+                APPEND_CAPABILITY_INFORMATION(handles, handle, , TPM2_MAX_CAP_HANDLES);
+                break;
+            case TPM2_CAP_COMMANDS:
+                APPEND_CAPABILITY_INFORMATION(command, commandAttributes, , TPM2_MAX_CAP_CC);
+                /* workaround because tpm2-tss does not implement attribute commandIndex for TPMA_CC */
+                property &= TPMA_CC_COMMANDINDEX_MASK;
```
joshuagl: Is this a bug in tpm2-tss? If so can we file it there and reference in the comments so that we can remove this workaround at some point?

PR #1320, opened 2019-02-14
tpm2_errata: Do not request unnecessary properties

webmeister:
Passing TPM2_PT_FIXED as count to tpm2_capability_get is semantically
wrong, since it is not a count of anything, but an index into the list of
properties. The intention might have been to just request all the
properties, but this is not necessary since process is only interested in a
small subset of all the properties that can be requested more efficiently.

Signed-off-by: Alexander Steffen <Alexander.Steffen@infineon.com>


joshuagl:
Thanks for this

Issue #1326, opened 2019-02-23
tpm2_policysecret fails when no auth passed


xyzhao2018:
The example shown in https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_policysecret.1.md has bug that no owner auth passed in the following command and result in the error.  
$ tpm2_policysecret -S session.ctx -c $TPM_RH_OWNER -o secret.policy --tcti=mssim
WARNING:esys:src/tss2-esys/api/Esys_PolicySecret.c:379:Esys_PolicySecret_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_PolicySecret.c:147:Esys_PolicySecret() Esys Finish ErrorCode (0x00000982)
ERROR: Esys_PolicySecret(0x982) - tpm:session(1):inconsistent attributes
ERROR: Could not build policysecret
ERROR: Unable to run tpm2_policysecret

joshuagl:
I think the bug here is not so much in the manual page, but that the tool expects an auth value (even when it might be true that no auth value is set).

@idesai would you agree?

idesai:
@joshuagl I tried it at my end without the password and it works just fine. Fixed the man page as well. Thanks.

idesai:
One caveat to my above statement, i found it working with and without the password with an earlier version of the tool: tpm2_policysecret version="3.0.2-731-gf2182cf" tctis="dynamic" tcti-default=tabrmd dlclose=enabled
@joshuagl @williamcroberts @AndreasFuchsSIT do you think this bug may have been introduced after esys integration, in that, there may be a difference in how the password/sensitive part of the command buffer may have been populated? There is also another (related bug I think) with unsealing tool.

joshuagl:
This does indeed look like a bug introduced during the ESAPI port. Thanks for investigating @idesai

PR #1332, opened 2019-03-03
Add test for passwordless reference auth object 

idesai:
The second commit will cause policy secret test to fail on current master, however, that is intentionally so. This because, after the esys port, the policy secret tool fails for passwordless reference auth objects as reported in issue #1326.

### DIFF
--- a/test/integration/tests/tcti/abrmd/policysecret.sh
+++ b/test/integration/tests/tcti/abrmd/policysecret.sh
@@ -83,7 +83,24 @@ rm $session_ctx
 test "$unsealed" == "$SEALED_SECRET"
 
 if [ $? != 0 ]; then
-  echo "failed policysecret integration test"
+  echo "Failed policysecret integration test. Auth ref object password was set"
+  exit 1
+fi
+
+#Test the policy with auth reference object password not set
+unsealed=""
+tpm2_changeauth -O ownerauth
+
+tpm2_startauthsession -a -S $session_ctx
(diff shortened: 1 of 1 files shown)


joshuagl:
If we add the failing test to `XFAIL_TESTS` we should be able to merge this without it causing the CI build to fail, we can then remove the `XFAIL_TESTS` entry once the underlying issue is fixed.

Something like:
`XFAIL_TESTS = test/integration/tests/policysecret.sh`

idesai:
@joshuagl where should it be added - here: https://github.com/tpm2-software/tpm2-tools/blob/735e51d488b744146047319a6df261209e966a52/Makefile.am#L190 ? Because that did not work.

joshuagl:
You need to do it later, after `ALL_SYSTEM_TESTS` has been added to the `TESTS` variable.

```
diff --git a/Makefile.am b/Makefile.am
index 444e0e0..d568ab4 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -204,6 +204,7 @@ check_PROGRAMS = \
     test/unit/test_tpm2_util
 
 TESTS += $(ALL_SYSTEM_TESTS)
+XFAIL_TESTS = test/integration/tests/tcti/abrmd/policysecret.sh
 
 test_unit_test_string_bytes_CFLAGS   = $(AM_CFLAGS) $(CMOCKA_CFLAGS)
 test_unit_test_string_bytes_LDADD    = $(CMOCKA_LIBS) $(LIB_COMMON) $(LDADD)
```

PR #1330, opened 2019-03-02
tools/tpm2_readpublic: Add option -n for saving object name

leeehui:
man/tpm2_readpublic.1.md: add description for option -n

Signed-off-by: Hui Li <leehui_hrbeu71102@163.com>

PR #1334, opened 2019-03-03
Rename short option of --in-file to -i

diabonas:
The CI build in #1330 fails due to an option conflict of [`-f`/`--in-file`](https://github.com/tpm2-software/tpm2-tools/blob/c66e4f069b04dcb13caed1f6408e4d199e6cf925/tools/tpm2_activatecredential.c#L268) in `tpm2_activatecredential` and [`-f`/`--format`](https://github.com/tpm2-software/tpm2-tools/blob/c66e4f069b04dcb13caed1f6408e4d199e6cf925/tools/tpm2_readpublic.c#L139) in `tpm2_readpublic`. This conflict already existed previous to the mentioned PR, but wasn't detected correctly because it was the last entry in the options array.

As per the discussion in this PR, rename the short option of `--in-file` consistently to `-i` to avoid this conflict. This needs to be done in `tpm2_create`, `tpm2_encryptdecrypt` and `tpm2_rsadecrypt` as well: these tools previously used `-I`, which is not in line with using capital letters only for parent objects.

As `tpm2_encryptdecrypt` already has a (unique) option `-i`/`--iv`, rename that one to `-t` to be able to use `-i`/`--input-file` consistently. This option was only introduced on master and the short option doesn't seem to be used anyway (the tests use the long option instead).

The option parser is fixed in a separate PR, see #1335.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ### next
   * tpm2_activatecredential: -f becomes -i
   * tpm2_create: -I becomes -i
+  * tpm2_encryptdecrypt: -I becomes -i
   * tpm2_create: supports TPM command CreateLoaded.
   * Add support for reading authorisation passwords from a file
   * tpm2_createek: now saves a context file for the generated primary's
@@ -48,7 +49,7 @@
   * tpm2_createak: -E becomes -e
   * tpm2_certify: -k becomes -p
   * tpm2_hash: -g changes to -G
-  * tpm2_encryptdecrypt: Support IVs via -i and algorithm modes via -G.
+  * tpm2_encryptdecrypt: Support IVs via -t and algorithm modes via -G.
   * tpm2_hmac: drop -g, just use the algorithm associated with the object.
   * tpm2_getmanufec: -g changes to -G
   * tpm2_createek: -g changes to -G
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
   * tpm2_activatecredential: -f becomes -i
   * tpm2_create: -I becomes -i
   * tpm2_encryptdecrypt: -I becomes -i
+  * tpm2_rsadecrypt: -I becomes -i
   * tpm2_create: supports TPM command CreateLoaded.
   * Add support for reading authorisation passwords from a file
   * tpm2_createek: now saves a context file for the generated primary's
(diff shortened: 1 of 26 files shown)


scripts/utils/analyze.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, tools, full=False):
                 ],
                 "tools": [],
                 "conflict": None,
-                "ignore": set(['r', 'privfile', 'u', 'pubfile', 's', 'secret', 'g', 'halg', 'n', 'name'])
+                "ignore": set(['r', 'privfile', 'u', 'pubfile', 's', 'secret', 'g', 'halg', 'n', 'name', 'f', 'format'])
```
diabonas: As `tpm2_readpublic` is the only tool in the `object` group that uses `-f`/`--format`, the option analyser complains. Ignoring this should be fine, as the short option name conflict with `tpm2_activatecredential` is resolved and the name is consistent with [`tpm2_verifysignature`](https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_verifysignature.1.md).

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Changelog
 ### next
+  * tpm2_activatecredential: -f becomes -I
```
williamcroberts: One of the goals for 4.0 is consistent naming of options as best as possible, focusing on tools in the same group class. I'd like to see capital letters for parent object arguments and lowercase for object arguments.

Well this does address an inconsistency it seems, can you perhaps propose something else?

diabonas: We could name the short option `-i` of course, which would follow the parent/object logic better. In this case, this should be done in `tpm2_create` and `tpm2_rsadecrypt` as well. The only remaining tool to use `--input-file` is `tpm2_encryptdecrypt`, but that one already has a short option `-i` for the initialisation vector `--iv`. What do you think?

williamcroberts: @diabonas seems reasonable.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog
 ### next
+  * tpm2_activatecredential: -f becomes -i
+  * tpm2_create: -I becomes -i
+  * tpm2_encryptdecrypt: -I becomes -i
+  * tpm2_rsadecrypt: -I becomes -i
```
williamcroberts: 4 tools, 4 patches please.

diabonas: Done!

williamcroberts:
Thank you, for some reason github's merge button stopped working and I pushed this directly to master manually.

PR #1335, opened 2019-03-03
Fix option analyser

diabonas:
#1330 discovered a bug in the [option analyser](https://github.com/tpm2-software/tpm2-tools/blob/master/scripts/utils/analyze.py): the last entry of an option array is not parsed if it does not end in a comma.

Unfortunately, the problem is deeper than that: if I understand correctly, the script is supposed to find options that are used by only one tool within the current group. Instead, the [current implementation](https://github.com/tpm2-software/tpm2-tools/blob/c66e4f069b04dcb13caed1f6408e4d199e6cf925/scripts/utils/analyze.py#L219) gathers a list of sets of options for each tool in the current group, then takes the symmetric difference of all sets in this list. This results in the [options that are used by an odd number of tools](https://en.wikipedia.org/wiki/Symmetric_difference#n-ary_symmetric_difference), which includes the unique options, but also gives some false positives. The new implementation in this PR instead collects all options of the tools within a group in a single list and returns the elements that occur exactly once.

Furthermore, the regex used to parse the options from the source code cannot handle option names containing a `-`. Replacing `\w` by `[^"]` (and removing the need for a trailing comma to fix the original issue in #1330) solves this problem, but uncovers a lot of previously undetected option conflicts. Most of these are probably harmless, but there are some that need to be addressed, e.g. `--key-alg` instead of the normally used `--kalg` in [`tpm2_loadexternal`](https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_loadexternal.1.md). As I am not very familiar with the option naming scheme, I would appreciate someone with more expertise looking into these conflicts listed in #1336.

Due to the newly found option conflicts, the CI build currently fails for this PR. The conflicts need to be handled before a merge by either renaming the affected options or adding them to the appropriate ignore list if they are harmless.

### DIFF
--- a/scripts/utils/analyze.py
+++ b/scripts/utils/analyze.py
@@ -270,8 +270,7 @@ class Parser(object):
     '''Parses C files for long option style option blocks'''
 
     regx = re.compile(
-        r'^\s*{\s*"(\w+)"\s*,\s*(?:required_argument|no_argument)\s*,\s*\w+\s*,\s*\'(\w)\'\s*},+$',
-        re.MULTILINE)
+        r'{\s*"([^"]+)"\s*,\s*(?:required_argument|no_argument)\s*,\s*\w+\s*,\s*\'(\w)\'\s*}')
 
     def __init__(self, path=os.getcwd()):
         self._path = path
--- a/scripts/utils/analyze.py
+++ b/scripts/utils/analyze.py
@@ -80,7 +80,7 @@ class ToolConflictor(object):
                 ],
                 "tools": [],
                 "conflict": None,
-                "ignore": set(['r', 'privfile', 'u', 'pubfile', 's', 'secret', 'g', 'halg', 'n', 'name', 'f', 'format'])
+                "ignore": set(['s', 'secret', 'f', 'format'])
             },
             {
                 "gname": "duplication",
@@ -116,14 +116,14 @@ class ToolConflictor(object):
                 "tools-in-group": ["tpm2_certify", "tpm2_quote"],
                 "tools": [],
                 "conflict": None,
-                "ignore": set(['g', 'halg', 'm', 'message', 's', 'signature', 'p', 'pcrs'])
+                "ignore": set(['g', 'halg', 'm', 'message', 'signature', 'pcrs'])
             },
             {
                 "gname": "signing",
                 "tools-in-group": ["tpm2_verifysignature", "tpm2_sign"],
                 "tools": [],
                 "conflict": None,
-                "ignore": set(['f', 'format', 's', 'sig'])
+                "ignore": set(['s', 'sig'])
             },
             {
                 "gname": "integrity",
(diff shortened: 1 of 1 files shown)


diabonas:
The newly reported conflicts are:
```
group: session:
	tools: ['tpm2_startauthsession', 'tpm2_policyrestart']
	conflicts: set(['a', 'auth-policy-session', 'policy-digest-alg', 'g'])
group: object:
	tools: ['tpm2_makecredential', 'tpm2_load', 'tpm2_unseal', 'tpm2_readpublic', 'tpm2_changeauth', 'tpm2_activatecredential', 'tpm2_createprimary', 'tpm2_loadexternal', 'tpm2_create']
	conflicts: set(['new-owner-passwd', 'item-context', 'key-parent-context', 'set-list', 'out-context-name', 'format', 'new-handle-passwd', 'F', 'I', 'current-lockout-passwd', 'O', 'current-handle-passwd', 'pcr-input-file', 'auth-hierarchy', 'key-alg', 'current-endorsement-passwd', 'current-owner-passwd', 'auth-object', 'l', 'enc-key', 'new-endorsement-passwd', 'auth-endorse', 'new-lockout-passwd'])
group: asymmetric:
	tools: ['tpm2_rsaencrypt', 'tpm2_rsadecrypt']
	conflicts: set(['I', 'auth-key', 'in-file', 'p'])
group: symmetric:
	tools: ['tpm2_hmac', 'tpm2_encryptdecrypt', 'tpm2_hash']
	conflicts: set(['C', 'P', 'I', 'c', 'p', 'in-file'])
group: attestation:
	tools: ['tpm2_certify', 'tpm2_quote']
	conflicts: set(['attest-file', 'obj-context', 'auth-ak', 'sel-list', 'auth-key', 'ak-context', 'a', 'G', 'L', 'sig-file', 'id-list', 'key-context', 'c', 'sig-hash-algorithm', 'auth-object', 'l', 'q', 'qualify-data'])
group: signing:
	tools: ['tpm2_verifysignature', 'tpm2_sign']
	conflicts: set(['auth-key', 'p'])
group: integrity:
	tools: ['tpm2_pcrreset', 'tpm2_pcrevent', 'tpm2_pcrlist', 'tpm2_pcrextend']
	conflicts: set(['auth-pcr', 'sel-list', 'i', 'L', 'o', 'P', 'pcr-index', 'out-file'])
group: ea:
	tools: ['tpm2_policyor', 'tpm2_createpolicy', 'tpm2_policycommandcode', 'tpm2_policyauthorize', 'tpm2_policysecret', 'tpm2_policypassword', 'tpm2_policypcr']
	conflicts: set(['a', 'g', 'policy-digest-alg', 'policy-pcr', 'P', 'input-policy-file', 'auth-policy-session'])
group: context:
	tools: ['tpm2_evictcontrol', 'tpm2_flushcontext']
	conflicts: set(['transient-object', 'loaded-session', 'saved-session', 'P', 's', 't', 'auth-hierarchy', 'l'])
group: nv:
	tools: ['tpm2_nvlist', 'tpm2_nvreadlock', 'tpm2_nvwrite', 'tpm2_nvread', 'tpm2_nvdefine', 'tpm2_nvrelease']
	conflicts: set(['f', 'policy-file', 'auth-index', 'p', 'out-file'])
```

williamcroberts:
Amazing catch and fantastic issue tracker for this. Thank you. These options have been a big thorn in my side heading to 4.0 and correcting this huge deficiency in the tool will help dramatically.

joshuagl:
Excellent work here!

I'd be inclined to land this after temporarily changing CI so that the script failures don't cause CI to fail. Treat them as warnings and work through fixing them piecemeal on master.

williamcroberts:
@joshuagl that's actually a good idea, just have it treat failures as warnings for now with a tracker on 4.0 milestone to revert that behavior. You want to land these today?

joshuagl:
Going to rebase and merge this after PR #1379 and having filed #1379 to revert that once #1336 is done.

Issue #876, opened 2018-01-30
Implement stirrandom


williamcroberts:
Implement a tool to use the stirrandom command.

PR #1337, opened 2019-03-04
tools/tpm2_stirrandom: Add command to inject additional input in TPM 

lestums:
Fixes: #876

    Added command to makefile
    Added manual entry for tpm2_stirrandom
    Added command relying on ESAPI
    Added integration test for coverage

### DIFF
--- a/man/tpm2_stirrandom.1.md
+++ b/man/tpm2_stirrandom.1.md
@@ -1,0 +1,73 @@
+% tpm2_stirrandom(1) tpm2-tools | General Commands Manual
+%
+% MARCH 2019
+
+# NAME
+
+**tpm2_stirrandom**(1) - Add "additional information" into TPM RNG state.
+
+# SYNOPSIS
+
+**tpm2_stirrandom** [*OPTIONS*] _INPUT\_FILE_
+
+# DESCRIPTION
+
+**tpm2_stirrandom**(1) Inject "additional information" as bytes into TPM entropy Protected Capability pool.
+
+"Additional information" can be extracted from _INPUT\_FILE_ or being read from stdin
+if _INPUT\_FILE_ is not specified.
+
+Up to 128 bytes can be injected at once through standard input to **tpm2_stirrandom**(1).
+
+If _INPUT\_FILE_ is larger than 128 bytes, **tpm2_stirrandom**(1) will fail.
+
+Adding data through **tpm2_stirrandom**(1) will trigger a reseeding of TPM
+DRBG Protected Capability. It is used when performing any sensitive action
+on a shielded location such as loading a persistent key or acting on a
+Protected Capability like updating TPM firmware.
+
+# OPTIONS
+
+This command has no option
+
+[common options](common/options.md)
+
+[common tcti options](common/tcti.md)
+
+# EXAMPLES
+
+Inject from stdin using echo:
+
+```
+echo -n "myrandomdata" | tpm2_stirrandom
+```
+
+Inject 64 bytes from stdin using a file
+
+```
+dd if=/dev/urandom bs=1 count=64 > myrandom.bin
+tpm2_stirrandom < ./myrandom.bin
+```
+
+Inject bytes from a file and reading up to 128 bytes
+
+```
+dd if=/dev/urandom of=./myrandom.bin bs=1 count=42
+tpm2_stirrandom ./myrandom.bin
+```
+
+# NOTES
+
+Please be aware that even if the "additional information" added
+by **tpm2_stirrandom**(1) can be entropy gathered from other DRBG
+sources, the TPM has no way of determining if the value has any entropy or not.
+As a consequence, it will just be considered as "additional input".
+
+The "additional input" is as defined in NIST SP800-90A (See 
+https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-90.pdf)
+
+# RETURNS
+
+0 on success or 1 on failure.
+
+[footer](common/footer.md)
--- a/scripts/utils/analyze.py
+++ b/scripts/utils/analyze.py
@@ -105,7 +105,7 @@ class ToolConflictor(object):
             },
             {
                 "gname": "random",
-                "tools-in-group": ["tpm2_getrandom"],
+                "tools-in-group": ["tpm2_getrandom", "tpm2_stirrandom"],
                 "tools": [],
                 "conflict": None,
                 "ignore": set()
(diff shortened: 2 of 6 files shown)


tools/tpm2_stirrandom.c
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    return true;
+}
+
+bool tpm2_tool_onstart(tpm2_options **opts) {
+
+    static struct option topts[] = {
+        { "infile", required_argument,  NULL, 'I' },
+        { "size",  required_argument,  NULL, 's' }
```
williamcroberts: Both of these seem weird. First off, the input should be an argument not an option, like getrandom. Secondly, since we can't fseek() the size on stdin, have it default to stdin, but you wont be able to use the helper on line 117. Just read it into a buffer until some max size. stirandom itself will be bounded on some max TPM size, so just fix it to that.

lestums: I also found this dual (and mutual exclusive) option handling quite weird. I was actually seeking criticisms on that :)

I assume i can migrate that to a 'getrandom' approach and have 'stirrandom' tool either read from a file or from stdin up to 128 bytes.

Thanks

man/tpm2_stirrandom.1.md
```diff
@@ -0,0 +1,77 @@
...
+%
+% MARCH 2019
+
+# NAME
+
+**tpm2_stirrandom**(1) - Add "additional information" into TPM RNG state.
+
+# SYNOPSIS
+
+**tpm2_stirrandom** [*OPTIONS*]
```
williamcroberts: You need the argument in here too

tools/tpm2_stirrandom.c
```diff
@@ -0,0 +1,149 @@
...
+        if (!result) {
+            LOG_ERR("Could not convert size to number, got: \"%s\"",
+                    value);
+            return false;
+        }
+        if(ctx.size_to_read > MAX_SIZE_TO_READ){
+            LOG_WARN("Requested size to read is too long. Set size to 128");
+            ctx.size_to_read = MAX_SIZE_TO_READ;
+        }
+        ctx.flags.s = 1;
```
williamcroberts: Both of these options should go away... the interface should be:

Use stdin (no argument via pipe)
```
dd if=/dev/urandom bs=1 count=32 | tpm_stirandom
```

Use stdin (no argument via file)
```
dd if=/dev/urandom bs=1 count=256 of=rand.dat
tpm_stirandom rand.dat < rand.dat
```

use a single file (an argument)
```
dd if=/dev/urandom bs=1 count=192 of=rand.dat
tpm_stirandom rand.dat
```

Fails because too big
```
dd if=/dev/urandom bs=1 count=257 of=rand.dat
tpm_stirandom rand.dat
```

BOTH calls should be bounded by the buffer size of a `TPM2B_SENSITIVE_DATA`  type.

lestums: This is what i had in mind ! Thank you for highlighting :)

tools/tpm2_stirrandom.c
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    ctx.in_file = argv[0];
+
+    return true;
+}
+
+bool tpm2_tool_onstart(tpm2_options **opts) {
+
+    memset(&ctx, 0, sizeof(tpm_stirrandom_ctx));
```
williamcroberts: static declaration means it's guaranteed to be 0's.

tools/tpm2_stirrandom.c
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+        LOG_INFO("Reading from stdin");
+        file = stdin;
+    }
+
+    ctx.in_data.size = fread(&ctx.in_data.buffer, 1, ctx.in_data.size, file);
+    if (ferror(file)) {
+        LOG_ERR("Error while reading data");
+        goto err;
+    }
```
williamcroberts: I think you can replace all of this with a call to `files_load_bytes_from_buffer_or_file_or_stdin` no?

```C
UINT16 size = MAX_SIZE_TO_READ;
bool res = files_load_bytes_from_buffer_or_file_or_stdin(NULL,
ctx._in_file, &size, ctx.in_data.buffer);
if (!res) {
  error;
}
```

tools/tpm2_stirrandom.c
```diff
@@ -0,0 +1,149 @@
...
+    if(ctx.in_data.size == 0){
+        LOG_ERR("Data size to send is zero");
+        goto err;
+    }
+
+    LOG_INFO("Submitting %d bytes to TPM", ctx.in_data.size);
+
+    return true;
+err:
+    if(ctx.in_file)
```
williamcroberts: brackets per coding standard

tools/tpm2_stirrandom.c
```diff
@@ -0,0 +1,149 @@
...
+    UNUSED(flags);
+
+    if(!load_sensitive()){
+        return 1;
+    }
+
+    return do_stirrandom(ectx) != true;
+}
+
+void tpm2_tool_onexit(void) {
```
williamcroberts: remove this if you don't need to do anything

scripts/utils/analyze.py
```diff
@@ -105,10 +105,10 @@ def __init__(self, tools, full=False):
             },
             {
                 "gname": "random",
-                "tools-in-group": ["tpm2_getrandom"],
+                "tools-in-group": ["tpm2_getrandom", "tpm2_stirrandom"],
                 "tools": [],
                 "conflict": None,
-                "ignore": set()
+                "ignore": set(['I', 'infile'])
```
williamcroberts: You shouldn't this part of the patch right? Or is this a side-effect of https://github.com/tpm2-software/tpm2-tools/pull/1335?

lestums: No i just missed it when i squashed / rebased my commits... I will fix that

tools/tpm2_stirrandom.c
lestums: Agree. Just a (unnecessary in that case) habit.

tools/tpm2_stirrandom.c
lestums: Ok with that. For it to work i just need to set size to `MAX_SIZE_TO_READ + 1` as `files_load_bytes_from_buffer_or_file_or_stdin` actually check for buffer size being larger than read data

tools/tpm2_stirrandom.c
lestums: Good catch !

Is it wanted to have CI perform a linting step ? I may open an issue on that and build it using a tool like astyle or clang-format.

williamcroberts: Yeah that would be great

tools/tpm2_stirrandom.c
williamcroberts: Good points in here, maybe it's best to leave it as is.

PR #1346, opened 2019-03-07
nv_define/relase: Default values for hierarchy "-a"

AndreasFuchsSIT:
I noticed that the tpm2_nvdefine and tpm2_nvrelease commands do not work without `-a`.
In the former case, it was a bug with the default value.
In the latter case, no default value was set.
Fixed and added a test for default the value for the hierarchy.

### DIFF
--- a/tools/tpm2_nvdefine.c
+++ b/tools/tpm2_nvdefine.c
@@ -71,7 +71,7 @@ struct tpm_nvdefine_ctx {
 static tpm_nvdefine_ctx ctx = {
     .auth= {
         .session_data = TPMS_AUTH_COMMAND_INIT(TPM2_RS_PW),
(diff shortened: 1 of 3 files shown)


tools/tpm2_nvdefine.c
```diff
@@ -71,7 +71,7 @@ struct tpm_nvdefine_ctx {
 static tpm_nvdefine_ctx ctx = {
     .auth= {
         .session_data = TPMS_AUTH_COMMAND_INIT(TPM2_RS_PW),
-        .hierarchy = ESYS_TR_RH_OWNER
+        .hierarchy = TPM2_RH_OWNER
```
joshuagl: Great catch, thanks!

PR #1357, opened 2019-03-11
tools: Add -u flag for to specify unique data for CreatePrimary

bnevis-i:
This change allows creating a primary object that follows
the guidance of https://trustedcomputinggroup.org/wp-content
/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf

The below should be all on one line:

tpm2_createprimary -a o -G rsa2048:aes128cfb -g sha256
  -o prim.ctx
  -A 'restricted|decrypt|fixedtpm|fixedparent|
      sensitivedataorigin|userwithauth|noda' -u unique.dat

where unique.dat is: 0x00 0x01 (0x00 * 256)

Note that all numbers (as above) must be in little-endian
format. The marhsaling logic will for TPM2B_PUBLIC
will convert to big-endian.

Signed-off-by: Bryon Nevis <bryon.nevis@intel.com>

### DIFF
--- a/lib/tpm2_alg_util.c
+++ b/lib/tpm2_alg_util.c
@@ -880,9 +880,12 @@ bool tpm2_alg_util_public_init(char *alg_details, char *name_halg, char *attrs, 
         }
     }
 
-    /* load a policy from a path if present */
+    /* load the unique portion of TPMT_PUBLIC from a path if present */
     if (unique_file) {
         UINT16 unique_size = sizeof(public->publicArea.unique);
+        /* loaded size may be <= unique_size; user is responsible
+         * for ensuring that that this buffer is formatted as a
+         * TPMU_PUBLIC_ID union. unique_size is max size of the union */
         bool res = files_load_bytes_from_path(unique_file,
                     (UINT8*)&public->publicArea.unique, &unique_size);
         if (!res) {
--- a/lib/tpm2_alg_util.c
+++ b/lib/tpm2_alg_util.c
@@ -865,8 +865,8 @@ bool get_signature_scheme(ESYS_CONTEXT *context,
     return true;
 }
 
-bool tpm2_alg_util_public_init(char *alg_details, char *name_halg, char *attrs, char *auth_policy, TPMA_OBJECT def_attrs,
-        TPM2B_PUBLIC *public) {
+bool tpm2_alg_util_public_init(char *alg_details, char *name_halg, char *attrs, char *auth_policy, char *unique_file,
+        TPMA_OBJECT def_attrs, TPM2B_PUBLIC *public) {
 
     memset(public, 0, sizeof(*public));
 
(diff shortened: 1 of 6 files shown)


lib/tpm2_alg_util.c
```diff
@@ -880,6 +880,16 @@ bool tpm2_alg_util_public_init(char *alg_details, char *name_halg, char *attrs,
         }
     }
 
+    /* load a policy from a path if present */
```
williamcroberts: comment is wrong

lib/tpm2_alg_util.c
```diff
@@ -880,6 +880,16 @@ bool tpm2_alg_util_public_init(char *alg_details, char *name_halg, char *attrs,
...
+    if (unique_file) {
+        UINT16 unique_size = sizeof(public->publicArea.unique);
+        bool res = files_load_bytes_from_path(unique_file,
+                    (UINT8*)&public->publicArea.unique, &unique_size);
+        if (!res) {
+            return false;
+        }
```
williamcroberts: this can return true on a smaller than `sizeof(public->publicArea.unique);` read, is that ok?

tools/tpm2_createprimary.c
```diff
@@ -123,6 +124,12 @@ static bool on_option(char key, char *value) {
             return false;
         }
         break;
+    case 'u':
+        ctx.unique_file = value;
+        if (ctx.unique_file == NULL || ctx.unique_file[0] == '\0') {
+            return false;
+        }
+        break;
```
williamcroberts: You need to update the manpage

tools/tpm2_createprimary.c
```diff
@@ -65,6 +65,7 @@ struct tpm_createprimary_ctx {
     } auth;
     tpm2_hierarchy_pdata objdata;
     char *context_file;
+    char* unique_file;
```
williamcroberts: `char *unique_file`

lib/tpm2_alg_util.c
bnevis-i: Corrected.

tools/tpm2_createprimary.c
bnevis-i: Updated.

lib/tpm2_alg_util.c
bnevis-i: Yes it is. It is a union field. Added inline comment.

tools/tpm2_createprimary.c
bnevis-i: Corrected.

bnevis-i:
Integration test added.

rowanmoul:
I know this PR is long merged, but I figured I could ask here instead of opening an issue.
What is the purpose of appending 0x0001 to the beginning of the unique data? I'm sure it's just something obvious that I'm not seeing, but I can't seem to figure it out despite looking over the code and reading through TCG documents.
All I _can_ determine is that I get a "buffer_size: 4096 with offset: 46 are insufficient for object of size 51919" error if I don't have it. The object size there is an example, as the number varies depending on data that I pass in (not the amount of data which is always 32 random bytes, but the value of the data). The object size seems far too large, which leads me to believe that something is reading past the end of the file, although I didn't see any obvious issues in the code that reads the files.

It seems like more explanation is needed on the man page about the required format of this file, as the description given doesn't allow the successful creation of a key except by pure chance or prior knowledge. I had to look at the tests.

> An optional file input that contains the unique field of TPMT_PUBLIC in little-endian format.

PR #1358, opened 2019-03-11
make the tpm2_sign test for rsapss accept both digest and max salt le…

zt-initech:
…ngth, document the issue in comments

Signed-off-by: Zeev Tarantov <zeev@initech.co.il>


zt-initech:
I'll test it on trusty (Ubuntu 14.04) in a VM, maybe I'll figure out why it failed in Travis CI.

PR #1366, opened 2019-03-19
Makefile.am: include man pages on “make dist” with the dist_man prefix.

dilyanpalauzov:
When PANDOC is absent, “make dist” fails.

When PANDOC is present, the man pages are generated and man pages are included by “make dist”, as these are prefixed with dist_.

There is no point in putting EXTRA_DIST in AM_CONDITIONAL, as the logic of conditionals is irrelevant when “make dist” calculates which files to include.


williamcroberts:
@dilyanpalauzov your autotools knowledge surpasses me by quite a bit; thank you.

Issue #1193, opened 2018-10-05
Add exit codes


electrickite:
When using tools in scripts or non-interactive scenarios, it would be very helpful to return more detailed exit codes. Currently the majority (all?) tools return 0 on success and 1 on any type of failure. 

In order to implement branching logic or error messages based on the type of failure, as far as I know the only way to determine the specific error is to capture stderr and grep the output for error codes. This works to a point, but requires additional commands, processes, and is quite brittle.

Would it be possible to convert TPM error codes into a numeric exit code and return that number after a failure? Alternately, a mapping table of common error codes to numeric exit codes could be created.

joshuagl:
This sounds like a nice idea. If we're going to make this change we should do it before our 4.x release as it'll effectively be a part of the ABI (and we're trying to stabilise and freeze interfaces ahead of 4.x).

electrickite:
@williamcroberts Yes, a full mapping would likely not be possible.

One example from a recent project: The script attempts to unseal a TPM object (or access an NVRAM area) using a PCR policy. If the unseal fails, the tool responds with an exit code of 1. We might guess the reason was a PCR value change, but it could be any number of things from a TCTI communication error to a malformed PCR bank specification from the user to a missing object context or bad object handle.

stderr will contain the RC code, of course, but now we are parsing command output for known error strings or matching a line text against a regex, both pretty brittle.

Another alternative would be to add an option to provide _only_ the RC code on stderr (or stdout?) with the accompanying text. This would allow for less brittle output parsing.

celesteking:
It would also be wonderful if you could add automatic decoding of returned codes. Currently, I have to use `tpm2_rc_decode` constantly in order to understand what's up. Via a special option or an ENV var, doesn't matter. No output when command succeeds and a decoded output if there's an error.

williamcroberts:
> It would also be wonderful if you could add automatic decoding of returned codes. Currently, I have to use `tpm2_rc_decode` constantly in order to understand what's up. Via a special option or an ENV var, doesn't matter. No output when command succeeds and a decoded output if there's an error.

That's done, all the output code has been converted to PERROR macro:
```C
    TSS2_RC rval = Esys_GetRandom(ectx,
                                  ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE,
                                  ctx.num_of_bytes, &random_bytes);
    if (rval != TPM2_RC_SUCCESS) {
[...]

PR #1372, opened 2019-03-24
tools/tpm2_testparms: Add a tool to check if an algorithm suite is supported

lestums:
No implementation issue was opened but this tool might be handy in some situation (like migration or tests)

- Added command to makefile
- Added manual entry for tpm2_selftest
- Added command relying on ESAPI
- Added integration test for coverage

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -114,6 +114,7 @@ bin_PROGRAMS = \
     tools/tpm2_startauthsession \
     tools/tpm2_startup \
     tools/tpm2_stirrandom \
+    tools/tpm2_testparms \
     tools/tpm2_unseal \
     tools/tpm2_verifysignature
 
@@ -185,6 +186,7 @@ tools_tpm2_policycommandcode_SOURCES = tools/tpm2_policycommandcode.c $(TOOL_SRC
 tools_tpm2_stirrandom_SOURCES = tools/tpm2_stirrandom.c $(TOOL_SRC)
 tools_tpm2_policyduplicationselect_SOURCES = tools/tpm2_policyduplicationselect.c $(TOOL_SRC)
 tools_tpm2_policylocality_SOURCES = tools/tpm2_policylocality.c $(TOOL_SRC)
+tools_tpm2_testparms_SOURCES = tools/tpm2_testparms.c $(TOOL_SRC)
 
 if UNIT
 TESTS = $(check_PROGRAMS)
@@ -348,6 +350,7 @@ if HAVE_MAN_PAGES
     man/man1/tpm2_startauthsession.1 \
     man/man1/tpm2_startup.1 \
     man/man1/tpm2_stirrandom.1 \
+    man/man1/tpm2_testparms.1 \
     man/man1/tpm2_unseal.1 \
     man/man1/tpm2_verifysignature.1
 endif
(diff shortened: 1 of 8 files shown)


test/integration/tests/testparms.sh
```diff
@@ -0,0 +1,94 @@
...
+done
+
+# Test that null algorithm raise an error
+if ! tpm2_testparms "null" 2>&1 1>/dev/null | grep -q "unsupported algorithm specification : null"; then
+    echo "tpm2_testparms with 'null' algorithm didn't fail"
+    exit 1
+else
+    true
+fi
+
```
williamcroberts: please add a negative test.

Makefile.am
```diff
@@ -114,6 +114,7 @@ bin_PROGRAMS = \
     tools/tpm2_startauthsession \
     tools/tpm2_startup \
     tools/tpm2_stirrandom \
+    tools/tpm2_testparms \
```
williamcroberts: Wait don't we want this as tpm2_testparams?

lestums: Depends on what we want for the people using the tools : do we value more the matching with TPM2.0 API names or the end-user (whom may not be a TPM expert) comfort by providing meaningful names ?

tools/tpm2_testparms.c
```diff
@@ -0,0 +1,96 @@
...
+};
+
+static tpm_testparms_ctx ctx;
+
+static bool tpm_testparms(ESYS_CONTEXT *ectx) {
+    TSS2_RC rval = Esys_TestParms(ectx, ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE, &(ctx.inputalg));
+    if (rval != TSS2_RC_SUCCESS) {
+        LOG_ERR("TPM TestParms failed !");
+        LOG_PERR(Esys_TestParms, rval);
+        return false;
```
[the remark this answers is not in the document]
lestums: Ack, will do.

Actually refrained from doing that as then it would be the only tool in the current stack to have a tri-state status which concerns TPM communication errors (`tpm2_flushcontext` use 2 as "error converting sapi to esapi handle" and `tpm2_getcap` as "unspecified selector")

This situation raise again the need for a consistent exit code behaviour for tools (as noted in #1193)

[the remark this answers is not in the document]
williamcroberts: I think right now we have 0 for success and 1 for error, so perhaps swap the list for positions 1 and 2 above. That's fine with me.

Makefile.am
williamcroberts: The lack of a is really messing with me. Since it's `Esys_TestParms` and I want tool to api mapping where possible, leave as is.

tools/tpm2_testparms.c
```diff
@@ -0,0 +1,130 @@
...
+                    LOG_ERR("Specified asymmetric algorithm is unsupported");
+                    break;
+                case TPM2_RC_MODE:
+                    LOG_ERR("Specified symmetric mode unsupported");
+                    break;
+                case TPM2_RC_VALUE:
+                default:
+                    LOG_ERR("Unsupported algorithm specification");
+                    break;
+            }
```
williamcroberts: nice error messages

PR #1385, opened 2019-03-29
tools/tpm2_incrementalselftest: Add tool to test specific algorithms

lestums:
- Added command to makefile
- Added manual entry for tpm2_selftest
- Added command relying on ESAPI
- Added integration test for coverage


tools/tpm2_incrementalselftest.c
```diff
@@ -0,0 +1,141 @@
...
+        return true;
+    }
+
+    tpm2_tool_output("Following algorithms remain to be tested ");
+    if(ctx.isempty){
+        tpm2_tool_output("(none scheduled)");
+    } else {
+        tpm2_tool_output("(some may be scheduled)");
+    }
+    tpm2_tool_output(" : \n");
```
williamcroberts: What's the actual output of this whole tool supposed to be?

What algorithms have been tested and whats scheduled to be tested?

We would probably want this output to be YAML compliant and machine parsable.

lestums: The output is a flat list of algorithms to be tested and algorithms that still need testing. 

However there is no way to distinguish between those currently under testing, those planned to be tested and those remaining but not scheduled.

Ok for the YAML structure

williamcroberts:
Is the flow of this to call this, and then call tpm2_gettestresult? Maybe this is one case where they shouldn't be separate tools?

lestums:
The workflow and the results between this command and `tpm2_gettestresult` is distinct. The only situation when those two commands might yield related results is in the case when `tpm2_selftest` has not been called yet and some algorithms are left to be tested. However this situation is unlikely to happen for discrete TPMs on Linux as `tpm2_selftest` (`linux/drivers/char/tpm/tpm2-cmd.c:1030`) is called after startup

Furthermore, `tpm2_gettestresult` is one of the two TPM command callable when the chip is currently in failure mode so IMO this command should be kept simple and callable in any case.

tools/tpm2_incrementalselftest.c
williamcroberts: When you update this PR for YAML compliant output, use the YAML helpers in helpers.sh to verify the output.

PR #1386, opened 2019-03-29
Multiple fixes on PCR / Algorithms banks in lib/pcr.c

lestums:
A bunch of fixes related to hardcoded values for PCR selections and bank algorithms.

### DIFF
--- a/lib/pcr.c
+++ b/lib/pcr.c
@@ -33,6 +33,7 @@
 #include <string.h>
 
 #include <tss2/tss2_sys.h>
+#include <inttypes.h>
 #include <stdbool.h>
 
 #include "pcr.h"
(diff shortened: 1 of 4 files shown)


lib/pcr.c
```diff
@@ -305,6 +305,18 @@ bool pcr_get_banks(ESYS_CONTEXT *esys_context, TPMS_CAPABILITY_DATA *capability_
     *capability_data = *capdata_ret;
 
     unsigned i;
+
+    // If the TPM support more bank algorithm that we currently
+    // able to manage, throw an error
+    if (capability_data->data.assignedPCR.count > sizeof(algs->alg)) {
+        LOG_ERR("Current implementation do not support more than %ld banks "
+                "(%d banks supported by TPM)",
```
williamcroberts: format specifiers wrong, should be %zu and (%zu or PRIu32).

Grammar, "Current implementation does not support more than %zu banks, got: "PRIu32"";

lib/pcr.c
```diff
@@ -305,6 +305,18 @@ bool pcr_get_banks(ESYS_CONTEXT *esys_context, TPMS_CAPABILITY_DATA *capability_
     *capability_data = *capdata_ret;
 
     unsigned i;
+
+    // If the TPM support more bank algorithm that we currently
+    // able to manage, throw an error
+    if (capability_data->data.assignedPCR.count > sizeof(algs->alg)) {
+        LOG_ERR("Current implementation does not support more than %zu banks, "
+                "got %32u banks supported by TPM",
```
williamcroberts: Nope, you just put an arbitrary 32 character width specifier on printf. Please use inttypes.h PRIu32.

Issue #1368, opened 2019-03-19
Optional requirement in configure.ac


dilyanpalauzov:
Configure.ac contains:
```
AC_MSG_WARN([System tests require the python module $1, some tests will fail!])
…
AC_MSG_WARN([Required executable bash not found, system tests require a bash shell!])
AC_MSG_WARN([Required executable python not found, some system tests will fail!])
AC_MSG_WARN([Required executable xxd not found, some system tests will fail!])
```
A related program is either optional or required.  Since ./configure continues when a desired program is missing, then the program is optional.

Reword the above warnings to “MNO not found, some system tests will fail”

williamcroberts:
This should be fail hard when not found

dilyanpalauzov:
The line in configure.ac: `enable_unit="$enable_unit (tpm2_abrmd: $tpm2_abrmd, tpm_server: $tpm_server, bash: $BASH_SHELL, python: $PYTHON, xxd: $XXD)"` has now no more added value.

PR #1387, opened 2019-03-29
configure.ac: Force failure when binary is missing with "enable-tests" on

lestums:
Fixes : #1368

When building with "enable-unit", throw an error if a tool is missing.

Otherwise just inform the user that some features / tests might not
work properly

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,6 +80,38 @@ AC_DEFUN([AC_PYTHON_MODULE],
   fi
 ])
 
+AC_CHECK_PROG([PYTHON],[python],[yes])
+AS_IF(
+    [test "x${PYTHON}" = x"yes"],
+    [],
+    [AS_IF(
+        [test "x$enable_unit" = x"yes"],
+        [AC_MSG_ERROR([Required executable bash not found, system tests require a python interpreter!])],
+        [AC_MSG_WARN([Python interpreter not found. System tests and some utilities will fail])]
+)])
+
+AC_CHECK_PROG([BASH_SHELL],[bash],[yes])
+AS_IF(
+    [test "x${BASH_SHELL}" = x"yes"],
+    [],
+    [AS_IF(
+        [test "x$enable_unit" = x"yes"],
+        [AC_MSG_ERROR([Required executable bash not found, system tests require a bash shell!])],
+        [AC_MSG_WARN([Bash shell not found. System test scripts will fail if executed])]
+)])
+
+AC_PYTHON_MODULE([yaml])
+
+AC_CHECK_PROG([XXD],[xxd],[yes])
+AS_IF(
+    [test "x${XXD}" = x"yes"],
+    [],
+    [AS_IF(
+        [test "x$enable_unit" = x"yes"],
+        [AC_MSG_ERROR([Required executable xxd not found, system tests require xxd!])],
+        [AC_MSG_WARN([xxd not found. Some system test scripts will fail if executed])]
+)])
+
 AC_DEFUN([unit_test_checks],[
 
     PKG_CHECK_MODULES([CMOCKA],[cmocka],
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
The idea behind how configure.ac+Makefile.am are currently written, is to execute tests only if ./configure was called with --enable-unit .  And configure shall only then emit warnings/errors about suboptimal testing environment, when it is supposed to run the tests.

E.g. “./configure && make check” does not run any tests, therefore there shall be no warning “Bash shell not found. System test scripts will fail if executed”.

lestums:
Semantically agree. 

However one can still execute tests without performing "./configure --enable-unit && make check"

So my aim was to inform the user that he won't be able to complain by raising a warning if those tools are actually missing while not activating unit tests

dilyanpalauzov:
If --enable-unit was not provided, then `make test/unit/test_files` prints: ”test/unit/test_files.c:39:10: fatal error: files.h: No such file or directory”, so the tests cannot be compiled with make.

In my opinion, there shall be no --enable-unit feature.  At the end of ./configure a summary shall be printed to the user, if tests are going to be executed on “make check”, that is cmocka, tpm2-abrmd, tpm_server are present.  If tests are not going to be executed, the summary shall provide a reason.  If tests are going to be executed, then there shall be some warnings about missing bash/python/xxd.

Or to tweak it a little bit further, provided somebody has a need for this: if ./configure is called without options and cmocka is missing, no tests are run on “make check”.  If ./configure is called without options and cmocka is present, tests are run on make check. If “./configure --enable-unit” is called and cmocka is missing, ./configure emits an error.

lestums:
By test i meant that most suites are written in Bash and have no hard dependency on toolchain so one could still ./configure and go "hey lets run this neat script to test my feature"

Second one seems a bit odd.

I buy the first one :)

dilyanpalauzov:
tpm-tools contains tests/unit/* and tools/integration/*. On “./configure --enable-unit && make check” tests from both directories are run, on “./configure && make check” nothing is run.

How about running the bash-tests on “make check” even after a pure “./configure” call?  Then the warnings about missing bash will be apropriate.  Or going further, disabling the tests depending on bash, whenever bash is missing (not adding SYSTEM_TESTS to TESTS).  Currently there is a discrepancy, that unit tests are run, whenever cmocka is available, but bash tests are run independent of bash presence (provided cmocka is present).

williamcroberts:
We just need to fix the configure file to be consistent...with all the other projects.

--enable-unit
--enable-integration

When these configure options are enabled, they should each do their respective checking of the system.

No one should be running the test scripts outside of the make check system unless they are power users, and then don't need any autoconf hand-holding.

dilyanpalauzov:
Unless there is a need to fix during ./configure which tests to run (integration ⇔ unit), at “make check”, with code like the one below it can be determined later which group of tests to run.
```
check-integration:
       $(MAKE) TESTS="$(TESTS_INTEGRATION)" check

check-unit:
       $(MAKE) TESTS="$(TESTS_UNIT)" check
```

For tpm2-tss this approach will not work, as the integration tests run only when mssim is complied and the fuzzing tests can be executed, only when mssim is disabled.

configure.ac
```diff
@@ -80,6 +80,38 @@ AC_DEFUN([AC_PYTHON_MODULE],
...
+AC_CHECK_PROG([XXD],[xxd],[yes])
+AS_IF(
+    [test "x${XXD}" = x"yes"],
+    [],
+    [AS_IF(
+        [test "x$enable_unit" = x"yes"],
+        [AC_MSG_ERROR([Required executable xxd not found, system tests require xxd!])],
+        [AC_MSG_WARN([xxd not found. Some system test scripts will fail if executed])]
+)])
+
```
williamcroberts: When configuring a build without testing, their is no dependency on python yaml, xxd, etc. So AFAICT this falsely adds dependencies to the configure step.

williamcroberts:
I don't see cluttering up the configure outcome with warnings that only pertain to tests when configuring non-testing builds. Having possible warning output looks bad from a quality standpoint, and makes it easier for other warnings that actually matter to go unnoticed.

For users that want to run the test scripts outside of the `make check` environment, they would be power users and are on their own.

I don't see this patch series being aligned with these goals. I appreciate the effort. I'm going to close this PR.

PR #1394, opened 2019-04-03
test/integration: Several fixes to helpers.sh file

lestums:
- Fixed syntax on `start_up`
- Fixed multiple coding style warnings (shellcheck)
- Added shebang
- Fixed tabs issues
- Added a check in `shut_down` to prevent `kill` to fail if the process has already been killed


lestums:
Would be cool indeed. Still working on coding style linting for CI so i could add `shellcheck` checks in the same time.

Issue #1405, opened 2019-04-08
tpm2_getrandom issues


udf2457:
There seem to be a few oddities with tpm2_getrandom ?

`tpm2_getrandom 2` = As expected
`tpm2_getrandom 20` = As expected
`tpm2_getrandom 2000` = Insufficient output

`tpm2_getrandom 2 -o test.out` = As expected
`tpm2_getrandom 20 -o test.out` = As expected
`tpm2_getrandom 2000 -o test.out` = Only 32 bytes

`tpm2_getrandom 32` = As expected
`tpm2_getrandom 33` = Still 32 ?

udf2457:
Just saw your update. Thanks for this and yes my preferred route would be better exception handling (since the inability to do something is an exception and normally its nice to tell the user rather than silently fail - or in the case of tpm2 silently default to something else).

PR #1407, opened 2019-04-09
tpm2_getrandom: bound on max hash size

williamcroberts:
Per 16.1 of:
  - https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf

It is recommended that TPMs bound entropy available to max hash size in
bytes and always make that available. TPMs do not issue an error when
size requested comes up short. Thus add checks for these two conditions
and allow a --force flag to override this behavior. Document and test.

$ tpm2_getrandom 2000
ERROR: TPM getrandom is bounded by max hash size, which is: 48
Please lower your request (preferred) and try again or use --force (advanced)
ERROR: Unable to run ./tools/tpm2_getrandom
echo $?
1

Fixes: #1405

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## Changelog
 ### next
+  * tpm2_getrandom: bound input request on max hash size per spec, allow -f to override this.
   * tpm_gettestresult: new tool for getting test results.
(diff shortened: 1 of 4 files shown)


williamcroberts:
@udf2457

PR #1409, opened 2019-04-11
Move tools/tpm2_tool.[ch] in lib/libcommon.a

dilyanpalauzov:
… so that Automake’s default _SOURCES can be used.


dilyanpalauzov:
What exactly causes circular symbol dependency (e.g. which symbol)?

williamcroberts:
IIRC $(TOOL_SRC) has main libcrt depends on main.

dilyanpalauzov:
I see.

Issue #1398, opened 2019-04-04
-G to -g switches


williamcroberts:
@lestums started work on #1355 to change -G to -g/-halg usages where it makes sense. Make this tracker to ensure it gets finished and tag it for 4.0 release as it changes CLI params.

- [X] tools/tpm2_hash.c
- [X] tools/tpm2_quote.c
- [X] tools/tpm2_sign.c
- [ ] tools/tpm2_import.c
- [ ] tools/tpm2_getmanufec.c
- [ ] tools/tpm2_createek.c
- [ ] tools/tpm2_create.c
- [ ] tools/tpm2_createprimary.c
- [ ] tools/tpm2_listpersistent.c
- [ ] tools/tpm2_createak.c
- [ ] tools/tpm2_encryptdecrypt.c
- [ ] tools/tpm2_loadexternal.c
- [X] tools/tpm2_verifysignature.c

PR #1410, opened 2019-04-12
Manpages style makeover and option reallocation

lestums:
A fair amount of modifications, clarifications and options modified to match with current implementation and have comprehensive yet concise man pages.

Also tackled some legacy behaviors, removed dead or unused options, cleaned up outdated examples, etc...

The work done in this PR is in a way related to work started in #1398 but on a global scale for every option category.


test/integration/tests/changeauth.sh
```diff
@@ -51,13 +51,13 @@ new_lockPasswd=newpswd
 
 tpm2_clear
 
-tpm2_changeauth -o $ownerPasswd -e $endorsePasswd -l $lockPasswd
+tpm2_changeauth -w $ownerPasswd -e $endorsePasswd -l $lockPasswd
```
williamcroberts: I was on the fence about this, but this is in the section "Hierarchy commands" which contains tpm2_createprimary that should be using the -o for output. At least sanity within the group.

test/integration/tests/nv.sh
```diff
@@ -64,11 +64,11 @@ cleanup "no-shut-down"
 tpm2_clear
 
 #Test default values for the hierarchy "-a" parameter
-tpm2_nvdefine -Q -x $nv_test_index -s 32 -t "ownerread|policywrite|ownerwrite"
+tpm2_nvdefine -Q -x $nv_test_index -s 32 -A "ownerread|policywrite|ownerwrite"
```
williamcroberts: we need to get this as a non-capital argument, is "-a" used?

Here is a list of the current ones:
```
$ grep -rn "case '.'" | cut -d"'" -f2-2 | sort | uniq
?
a
A
c
C
D
e
E
f
F
g
G
h
H
i
k
K
l
L
m
n
N
o
O
p
P
q
Q
r
s
S
t
T
u
U
v
V
x
X
Z
```
I could consider -y for attributes, but I was hoping one day to be able to specify the public portion of an object via YAML. So I wouldn't want to use that.

Perhaps -w "like with these attributes" thoughts?

williamcroberts: Looks like you want to swap -o for -w in the next patch, so maybe -b

tools/tpm2_pcrevent.c
```diff
@@ -257,7 +257,7 @@ static bool init(void) {
 
     ctx.input = ctx.input ? ctx.input : stdin;
 
-    if ((ctx.auth.session || ctx.flags.P) && !ctx.flags.i) {
+    if ((ctx.auth.session || ctx.flags.P) && !ctx.flags.x) {
         LOG_ERR("Must specify a PCR index via -i with the -%c option.",
```
williamcroberts: -x

williamcroberts:
We also need patches to CHANGELOG.md in here, so we can track all these options changes and then post merge ill go update the tracker tracking the -G/-g switch.

test/integration/tests/changeauth.sh
lestums: I was really puzzled on which toggle to use and `-o` was like 90% of the time used for output files. `-w` was free and close to the owner semantic so i went for it.

Not only it sanitize the "hierarchy" group but also allow a coherent usage of `-o` across all the tools.

test/integration/tests/nv.sh
lestums: `-a` is mostly used for specifying hierarchy :

```
>$ grep -R "'a'[[:space:]]*}" | sed "s%[[:space:]]*%%g"

tools/tpm2_startauthsession.c:{"auth-policy-session",no_argument,NULL,'a'},
tools/tpm2_createpolicy.c:{"auth-policy-session",no_argument,NULL,'a'},
tools/tpm2_nvread.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_evictcontrol.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_nvincrement.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_nvwrite.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_hash.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_createprimary.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_loadexternal.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_nvreadlock.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_nvdefine.c:{"hierarchy",required_argument,NULL,'a'},
tools/tpm2_nvrelease.c:{"hierarchy",required_argument,NULL,'a'},
```

My gears are currently grinded on auth-policy-session (`-p` instead ?).

My goal is to sanctuarize key options : `-o` for output, `-i` for input, etc...

I don't have any objections on swapping `-A` to `-b`. I actually missed the "capitalized is parent" policy on that one

williamcroberts: > My gears are currently grinded on auth-policy-session (-p instead ?).
I'd like to keep -p|-P for authvalue consistent across all tools. For tpm2_startauthsession, -a is for policy over trial type of session. perhaps just long option it --policy and even add --trial? I'd say the same approach for tpm2_createpolicy.

ETOOMANYOPTIONS this gets complicated/monotonous quickly.

lestums: The current behaviour on whether if we start a policy or a trial session rely only on the "-a" toggle.

For the sake of simplicity, it would be easier to state that default behaviour is `SE_TRIAL` and the `--policy` allow a `SE_POLICY` session

lestums:
Added changelog entries and fixed raised issues in review

Waiting for second session of reviews. I'll also do a meticulous review to ensure i didnt break anything.

In that way, i added a test to check toggles implementation (i noticed a couple quirks and mismodification along the way that could have been avoided with that check)

tools/tpm2_nvdefine.c
```diff
@@ -195,14 +195,13 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
         { "index",                  required_argument,  NULL,   'x' },
         { "hierarchy",              required_argument,  NULL,   'a' },
         { "size",                   required_argument,  NULL,   's' },
-        { "attributes",             required_argument,  NULL,   't' },
+        { "index-attributes",       required_argument,  NULL,   'b' },
```
williamcroberts: I think we just want attributes. I think <specifier>-attributes is to much, and I plan on doing a round of long-option name sanity after we get the short options squared away. I wouldn't bother changing this at the moment.

man/tpm2_nvdefine.1.md
```diff
@@ -17,36 +17,43 @@
...
       * **`<num>`** where a raw number can be used.
 
   * **-s**, **--size**=_SIZE_:
-    specifies the size of data area in bytes. Defaults to MAX_NV_INDEX_SIZE
+
+    Specifies the size of data area in bytes. Defaults to **MAX_NV_INDEX_SIZE**
     which is typically 2048.
 
-  * **-t**, **--attributes**=_ATTRIBUTES_
+  * **-b**, **--index-attributes**=_ATTRIBUTES_
```
williamcroberts: don't change the long option name at the moment. Well hit these later.

lestums:
Done on the long naming for `tpm2_nvdefine`. Also rebased on master.

PR #1421, opened 2019-04-17
tpm2_duplicate

GlovePuppet:
This still needs to be tested with tpm2_import. Is it possible to do that with the current test harness? Will the sim import a key it duplicated previously?


williamcroberts:
Two things:
First thing is, please use a rebase workflow not a merge workflow. I'm not sure what the best git foo is to complete that task, but I would just start a new branch with the HEAD on an up-to-date origin/master and cherry pick your 9 patches over and force push. git pull --rebase while you work and have no public dependencies is the cleanest fashion.

Second:
To answer your question, I would imagine it would. I don't think it does that deep of an inspection of objects.

PR #1424, opened 2019-04-18
man: Minor fixes in text

dnoliver:


dnoliver:
it aint much but it's honest work :)

PR #1429, opened 2019-04-19
Man Fixes for createek, createak and quote

dnoliver:
commit https://github.com/tpm2-software/tpm2-tools/commit/7f09ad2eabc3755a7719cc9bdd85b03401df1886: if the template do not have a blank line at the end, the next section is taken as part of the section described in the template. i.e.:

Note the `#EXAMPLES` string that should be a title:

```
Signature Format Specifiers
       Format  selection for the signature output file.  tss (the default) will output a binary blob according to the TPM 2.0 specification and any potential compiler padding.  The op‐
       tion plain will output the plain signature data as defined by the used cryptographic algorithm.  # EXAMPLES

              tpm2_quote -C 0x81010002 -P abc123 -L sha1 -l 16,17,18

              tpm2_quote -C ak.context -P "str:abc123" -L sha1 -l 16,17,18

              tpm2_quote -C 0x81010002 -L sha1 -l 16,17,18

```

commit https://github.com/tpm2-software/tpm2-tools/commit/c8c64a5afa2b2ae3159440acfb1826ae6c98772a and https://github.com/tpm2-software/tpm2-tools/commit/c8c64a5afa2b2ae3159440acfb1826ae6c98772a fixes the `-g` option with `-G`

commit https://github.com/tpm2-software/tpm2-tools/commit/a90189fb6437ab2ecb92909e5eb2f1b8506d6abd fixes the example. I am not sure if the one I am providing is a better example, but at least works for me using the simulator. 

Using the original example, I got the following error:

```
[root@localhost tpm2-tools]$ tpm2_quote -C 0x81010002 -L sha1 -l 16,17,18 --tcti=mssim
ERROR: Could not parse pcr selections, got: "sha1"
```

commit https://github.com/tpm2-software/tpm2-tools/pull/1429/commits/def07014682909fcf360d122b8da640777883fe7 add tpm2_print to the man pages. Tested it with `./configure && make && sudo make install && man tpm2_print`. I hope that it was not left out on purpose @lestums :)


dnoliver:
Found several problems with following the man in `tpm2_checkquote` as well. 

This does not work:
```
tpm2_checkquote -c akpub.pem -m quote.out -s sig.out -p pcrs.out -G sha256 -q abc123
```

Had to change to this, after reading a little bit of the code:

```
tpm2_checkquote -u akpub.pem -m quote.out -s sig.out -F pcrs.out -g sha256 -q abc123
```

But I cannot run it with the simulator:

```
[root@localhost attestation]$ tpm2_checkquote -u akpub.pem -m quote.out -s sig.out -F pcrs.out -g sha256 -q abc123 --tcti=mssim
ERROR: tpm2_checkquote: tool doesn't support the TCTI option
```

Not sure which is the intention of the `tpm2_checkquote`, though :D

PR #1436, opened 2019-04-24
tpm2_import extensions

GlovePuppet:
This PR extends the tpm2_import tool enable it to import TPM objects created by the tpm2_duplicate tool. See the included TCs for example usage

Signed-off-by: GlovePuppet <touched.by.his.noodley.appendage@gmail.com>

### DIFF
--- a/test/integration/tests/import_tpm.sh
+++ b/test/integration/tests/import_tpm.sh
@@ -1,0 +1,141 @@
+#!/bin/bash
+
+TPM_CC_DUPLICATE=0x14B
+
+source helpers.sh
+
+cleanup() {
+rm -f primary.ctx \
+          new_parent.prv new_parent.pub new_parent.ctx \
+          ipolicy.dat dpolicy.dat session.dat \
+          key.prv key.pub key.ctx \
+          dup.prv dup.pub dup.seed \
+          key2.prv key2.pub key2.ctx \
+          sym_key_in.bin \
+          dup.ctx
+
+    if [ "$1" != "no-shut-down" ]; then
+          shut_down
+    fi
+}
+trap cleanup EXIT
+
+start_up
+
+create_policy() {
+    tpm2_startauthsession -Q -S session.dat
+    tpm2_policycommandcode -Q -S session.dat -o $1 $2
+    tpm2_flushcontext -Q -S session.dat
+    rm session.dat
+}
+
+start_session() {
+    tpm2_startauthsession -Q --policy-session -S session.dat
+    tpm2_policycommandcode -Q -S session.dat -o $1 $2
+}
+
+end_session() {
+    tpm2_flushcontext -Q -S session.dat
+    rm session.dat
+}
+
+create_load_new_parent() {
+    # Create new parent
+    tpm2_create -Q -C primary.ctx -g sha256 -G rsa -r new_parent.prv -u new_parent.pub  -b "restricted|sensitivedataorigin|decrypt|userwithauth"
+    # Load new parent key, only the public part
+    tpm2_loadexternal -Q -a o -u new_parent.pub -o new_parent.ctx
+}
+
+load_new_parent() {
+    # Load new parent key, public & private parts
+    tpm2_load -Q -C primary.ctx -r new_parent.prv -u new_parent.pub -o new_parent.ctx
+}
+
+create_load_duplicatee() {
+    # Create the key we want to duplicate
+    create_policy dpolicy.dat $TPM_CC_DUPLICATE
+    tpm2_create -Q -C primary.ctx -g sha256 -G $1 -p foo -r key.prv -u key.pub -L dpolicy.dat -b "sensitivedataorigin|decrypt|userwithauth"
+    # Load the key
+    tpm2_load -Q -C primary.ctx -r key.prv -u key.pub -o key.ctx
+    # Extract the public part for import later
+    tpm2_readpublic -Q -c key.ctx -o dup.pub
+}
+
+do_duplication() {
+    start_session dpolicy.dat $TPM_CC_DUPLICATE
+    if [ "$2" = "aes" ]
+    then
+        tpm2_duplicate -Q -C new_parent.ctx -c key.ctx -g aes -o sym.key -p "session:session.dat" -r dup.dup -s dup.seed
+    else
+        tpm2_duplicate -Q -C new_parent.ctx -c key.ctx -g null -p "session:session.dat" -r dup.dup -s dup.seed
+    fi
+    end_session
+}
+
+do_import_load() {
+    if [ "$2" = "aes" ]
+    then
+        tpm2_import -Q -C new_parent.ctx -k sym.key -u dup.pub -i dup.dup -r dup.prv -s dup.seed  -L dpolicy.dat
+    else
+        tpm2_import -Q -C new_parent.ctx -u dup.pub -i dup.dup -r dup.prv -s dup.seed  -L dpolicy.dat
+    fi
+    tpm2_load -Q -C new_parent.ctx -r dup.prv -u dup.pub -o dup.ctx
+}
+
+test() {
+    # New parent ...
+    create_load_new_parent
+    # Key to be duplicated
+    create_load_duplicatee $1
+    # Duplicate the key
+    do_duplication $2
+    # Remove, we're done with it
+    rm new_parent.ctx
+    # Load the full thing this time
+    load_new_parent
+    # Import & load the duplicate
+    do_import_load $2
+}
+
+# Part 1 : Duplicate 3 varieties of key (aes, rsa or ecc) 
+# and protect them using sym_alg null or aes, verify they
+# can be imported & loaded
+for dup_key_type in aes rsa ecc; do
+    for sym_key_type in aes null; do
+        tpm2_createprimary -Q -a o -g sha256 -G rsa -o primary.ctx
+        test $dup_key_type $sym_key_type
+        cleanup "no-shut-down"
+    done
+done
+
+# Part 2 : 
+# Create a rsa key (Kd)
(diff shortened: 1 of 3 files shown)


test/integration/tests/import_tpm.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Load new parent key, public & private parts
+tpm2_load -Q -C primary.ctx -r new_parent.prv -u new_parent.pub -o new_parent.ctx
+
+# Import the duplicated key
+start_session ipolicy.dat $TPM_CC_IMPORT
+tpm2_import -V -C new_parent.ctx -g null -u dup.pub -d dup.dup -r dup.prv -s dup.seed -P "session:session.dat" -L dpolicy.dat
+end_session
+
+# Load imported key, public & private parts - HOW?
```
williamcroberts: Did you still have this question? -u and -r commands are for public and private portions respectively.

tools/tpm2_import.c
```diff
@@ -312,18 +317,30 @@ static bool on_option(char key, char *value) {
...
         ctx.attrs = value;
         break;
     case 'g':
         ctx.name_alg = value;
         break;
+    case 's':
+        ctx.input_seed_file = value;
+        break;
+    case 'd':
+        ctx.input_duplicate_file = value;
```
williamcroberts: can you re-use -i for this and auto-detect the format?

test/integration/tests/import_tpm.sh
GlovePuppet: Sorry, the comment is a bit cryptic. When I run the script as-is the final load fails. 

WARNING:esys:src/tss2-esys/api/Esys_Load.c:325:Esys_Load_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_Load.c:118:Esys_Load() Esys Finish ErrorCode (0x0000012f) 
ERROR: Eys_Load(0x12F) - tpm:error(2.0): authValue or authPolicy is not available for selected entity

I don't fully understand this but I think I need to satisfy the policy baked into new_parent?

tools/tpm2_import.c
GlovePuppet: will do, I have been trying to keep my fingers off the original code as much as possible.

test/integration/tests/import_tpm.sh
```diff
@@ -0,0 +1,76 @@
...
+end_session
+
+rm new_parent.ctx
+
+# Load new parent key, public & private parts
+tpm2_load -Q -C primary.ctx -r new_parent.prv -u new_parent.pub -o new_parent.ctx
+
+# Import the duplicated key
+start_session ipolicy.dat $TPM_CC_IMPORT
+tpm2_import -V -C new_parent.ctx -g null -u dup.pub -d dup.dup -r dup.prv -s dup.seed -P "session:session.dat" -L dpolicy.dat
```
williamcroberts: Two things here:
1. I don't want to override -g to be the name algorithm or the crypto algorithm.
2. make sure this works with AES128

GlovePuppet: 1.a Sure, how about --symmetric_alg and no short form? 
1.b If the algo is null then you don't need -i (was -d)
1.C I believe null & AES128CFB (same as duplicate)

tools/tpm2_import.c
```diff
@@ -366,30 +386,50 @@ static int check_options(void) {
...
+            case TPM2_ALG_NULL:
+                break;
+            default:
+                if(ctx.input_enc_key_file == NULL) {
+                    LOG_ERR("Expected a symmetric encryption key file to be specified via \"-k\","
+                            " missing option.");
+                    rc = -1;
+                }
+                break;
+        }
```
williamcroberts: this hunk is ultra confusing because of the overloaded -g. If you can re-use -i for the input dup file and -G to say, hey were duplicating it, it should become simpler to read.

tools/tpm2_import.c
```diff
@@ -566,3 +597,138 @@ int tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
...
+                public.publicArea.authPolicy.buffer, &public.publicArea.authPolicy.size);
+    if (!result) {
+        return 1;
+    }
+
+    result = do_import(
+            ectx,
+            parent_object_context.tr_handle,
+            &encrypted_seed, 
+            ctx.input_enc_key_file == NULL ? NULL : &enc_key,
```
williamcroberts: Just initialize enc_key size to 0, then just check in the routine if size is 0, nothing to do versus if NULL nothing to do.

GlovePuppet: My understanding here is that if symmetricAlg is null then encryptionKey should be NULL too. I think then I would have to put the check 'enc_key.size == 0 ? NULL : &enc_key' inside do_import(). 

Is that correct?

tools/tpm2_import.c
GlovePuppet: Will revisit this

GlovePuppet: Agreed, it does make things a lot simpler _but_ this needs to change - the  tpm2_alg_util_flags_* are wrong

    case 'G':
        ctx.key_type = tpm2_alg_util_from_optarg(value,
                tpm2_alg_util_flags_asymmetric
                |tpm2_alg_util_flags_symmetric);
        if (ctx.key_type == TPM2_ALG_ERROR) {
            LOG_ERR("Unsupported key type");
            return false;
        }

GlovePuppet:
I plan to write a few TCs to test the encryption.

Do you have any insights on why the final tpm2_load is not working? I feel there's some policy issue at the root of it

tools/tpm2_import.c
williamcroberts: yep

test/integration/tests/import_tpm.sh
williamcroberts: > 1.a Sure, how about --symmetric_alg and no short form?

i'm fine with that.

> 1.b If the algo is null then you don't need -k

Can we infer the crypto algorithm from -k option or via the dup.dup file coming in?

> 1.C I believe null & AES128CFB (same as duplicate)
> 
> Drop -d (seed) and re-use -i instead, ok.

I thought -s was the seed, no?

Just be very deliberate about command line options, cleanliness and figuring out things on behalf of the user where possible. I want to avoid having a million options for tools, especially for the common case. Guess and check is a valid approach as well.

GlovePuppet: 1.a  -G / --symmetric_alg

1.b I guess if -k is specified we could assume aes else null. That would get rid of the need for --symmetric_alg. I can do it this way if you wish

1.c Yes, -s is seed (my bad). The duplicate was specified by -d but is now specified -i

GlovePuppet:
> Just initialize enc_key size to 0, then just check in the routine if size is 0, nothing to do versus if NULL nothing to do.

Done

> this hunk is ultra confusing because of the overloaded -g. ...

I have simplified the cmd line interface, the definitive list is:

New
-s Seed (in)
-k Sym Alg key file (in) - The absence or presence of this implies the Sym Alg to use (null or aes respectively)
-L Policy (in)

Reused
-i Duplicate (in)
-u Public Key (in)
-r Private Key (out)

-G is unused by tpm mode, mandatory in openssl mode 

Enhanced TCs to duplicate/import keys of type aes, rsa & ecc using sym algs null & aes

TBDs

- Enhance manpages
- Figure out why tpm2_load doesn't work
- Complete TCs

Questions:
Are there any notes about using the debugger with the tpm sim?

Duplicate allows a key to be duplicated with a null parent, the TCG doc specifically mentions this for duplicate but no mention is made for import. @lestums Asked for this feature in tpm2_duplicate, do we know if importing with a null parent is permitted? (I can't get it to work but that doesn't mean much)

GlovePuppet:
I think I can conclude the importing an object and specifying a null parent doesn't & shouldn't work. 

The command descriptor types array for import has none of the ADD_FLAGs found in the duplicate and rewrap descriptors. This makes a null parent invalid for tpm2_import and it fails with 
ERROR: Esys_Import(0x184) - tpm:handle(1):value is out of range or is not correct for the context

Testing against a HW TPM gives the same result

I wonder what the point of duplicate with a null parent is?

GlovePuppet:
Anything else I need to attend to?

williamcroberts:
> Anything else I need to attend to?

Let me look through, thanks.

williamcroberts:
Last item to work on:
1. your commit message
2. the first message/comment block that shows up for the PR

Make these detailed and reflect what and why you did things.

GlovePuppet:
I hope this is satisfactory, thanks for your help again

williamcroberts:
Still have whitespace errors:
```
Applying: tools\tpm2_import
.git/rebase-apply/patch:17: trailing whitespace.
It requires that the parent key object be a RSA key. Can also import a TPM managed 
.git/rebase-apply/patch:37: trailing whitespace.
    Optional. Specifies the file containing the symmetric algorithm key that was used for the 
.git/rebase-apply/patch:38: trailing whitespace.
    inner wrapper. If the file is specified the tool assumes the algorithm is AES 128 in CFB mode 
.git/rebase-apply/patch:46: trailing whitespace.
    When importing a duplicated object this option specifies the file containing the 
.git/rebase-apply/patch:53: trailing whitespace.
    When importing a duplicated object this option specifies the file containing the 
warning: squelched 6 whitespace errors
warning: 11 lines add whitespace errors.
```

Next time check your patch with `git diff --check HEAD^` Fix, git commit --amend and force push.

I took care of these and merged it out of the PR workflow.

williamcroberts:
merged: https://github.com/tpm2-software/tpm2-tools/commit/f254d5efd40f42f06b2c0e9f3947cb4277f71af2

PR #1430, opened 2019-04-22
Wire up support for ak auth password in tpm2_quote tool

jetwhiz:
It looks like the code to take in an AK password existed in tpm2_quote, but wasn't actually wired in to the tool (so it was basically ignored).  

This PR wires in the AK password so it can be used if an AK password is set.

PR #1390, opened 2019-03-31
3.x issue 1389

maage:
This should fix #1389

PR #1437, opened 2019-04-25
[3.X] Fix Travis CI build

diabonas:
The new Dockerfile-based container has `tpm_sever` on the path (in `/usr/local/bin`) instead of in `/ibmtpm974/src`. This should hopefully fix the build failures observed by #1390 and #1430.

### DIFF
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -52,7 +52,7 @@ sleep 5
 
 echo "starting tpm server"
 # start the tpm server
-/ibmtpm974/src/tpm_server &
+tpm_server &
 echo "tpm server started"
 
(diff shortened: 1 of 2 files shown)


diabonas:
One error less in the build log, but tpm2-abrmd still fails to start with the error
> ** (tpm2-abrmd:23883): CRITICAL **: Failed to acquire DBus name com.intel.tss2.Tabrmd. UID 0 must be allowed to "own" this name. Check DBus config and check that this is running as user tss or root.

diabonas:
After fixing the tpm2-abrmd configuration, it works now.

PR #1439, opened 2019-04-26
 Add platform password setup through changeauth

Xeverus:
Hi.

I'd like to propose a patch which allows to setup/change/remove password for platform hierarchy through tpm2_changeauth tool.

### DIFF
--- a/man/tpm2_changeauth.1.md
+++ b/man/tpm2_changeauth.1.md
@@ -4,7 +4,7 @@
 
 # NAME
 
-**tpm2_changeauth**(1) - Configuring authorization for TPM OWNER/ENDORSEMENT/LOCKOUT hierarchies,
+**tpm2_changeauth**(1) - Configuring authorization for TPM OWNER/PLATFORM/ENDORSEMENT/LOCKOUT hierarchies,
 NV Handles and Transient & Persistent object handles.
 
 # SYNOPSIS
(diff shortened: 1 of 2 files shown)


man/tpm2_changeauth.1.md
```diff
@@ -44,6 +48,10 @@ see section "Authorization Formatting".
 
     The current authorization value for the owner hierarchy .
 
+  * **-S**, **--current-platform-passwd**=_CURRENT\_PLATFORM\_AUTH_:
+
+    The current authorization value for the platform hierarchy .
+
```
williamcroberts: I don't want to take -s|-S because were working on fixing the tools options and while this does fit with the existing and adds a useful feature doesn't move us completely in the right direction.

Id like to see almost all of these options go away, and the tool to just be something like:
```
tpm2_changeauth <--parent|-C> <--object|-c> <--old|-o> <--new|-p>
```
Rather than a command line option for each thing. Would you be willing to take this on?

Xeverus: I can try. But I'm not sure what you'd like to see. Can you elaborate a bit more your vision?

williamcroberts: Actually I went through today to see what state we were in for session support in the tools. It looks like it's supported, you should be able to use "session:<file>" as your password and it should respect the session. See 53a6e1a6465a5ff777

So the other change I am seeing here, is removing all of these command line options... just do:
```
tpm2_changeauth <--parent|-C> <--object|-c> <--old|-o> <--new|-p>
```

Where parent:
- persistent handle or transient object file
- object is the object to change, could be nvindex, object or hierarchy type things
- old: old password or session satisfying the auth policy
- new: new password or session containing the new auth policy. Not sure if that last part works

PR #1441, opened 2019-04-26
Backport ability to run tpm2_makecredential off-TPM to 3.X

jetwhiz:
Backport the tpm2_makecredential tool changes that allow off-TPM (openssl) functionality to 3.X

Used some functions from tpm2_import for off-TPM functionality
 - Move needed, overlap code into tpm2_identity_util
Add new global -X/--openssl-backend option for any tools that can operate w/o TPM

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -268,9 +268,8 @@ static char* parse_socket_tcti(void) {
 
 static tcti_conf tcti_get_config(const char *optstr) {
 
-    tcti_conf conf = {
-        .name = NULL
-    };
+    /* set up the default configuration */
+    tcti_conf conf = { 0 };
 
     /* no tcti config supplied, get it from env */
     if (!optstr) {
@@ -294,6 +293,11 @@ static tcti_conf tcti_get_config(const char *optstr) {
             }
         }
     } else {
+        /* handle case of TCTI set as "-T none" */
+        if (!strcmp(optstr, "none")) {
+            return conf;
+        }
+
         parse_env_tcti(optstr, &conf);
     }
 
@@ -400,7 +404,7 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv,
      * grep -rn case\ \'[a-zA-Z]\' | awk '{print $3}' | sed s/\'//g | sed s/\://g | sort | uniq | less
      */
     struct option long_options [] = {
-        { "tcti",          required_argument, NULL, 'T' },
+        { "tcti",          optional_argument, NULL, 'T' },
         { "help",          no_argument,       NULL, 'h' },
         { "verbose",       no_argument,       NULL, 'V' },
         { "quiet",         no_argument,       NULL, 'Q' },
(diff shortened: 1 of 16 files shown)


lib/tpm2_options.c
```diff
@@ -457,6 +458,12 @@ tpm2_option_code tpm2_handle_options (int argc, char **argv,
         case 'Z':
             flags->enable_errata = 1;
             break;
+        case 'X':
+            if (tool_opts) {
+                flags->no_tpm = 1;
+                tool_opts->flags |= TPM2_OPTIONS_NO_SAPI;
+            }
+            break;
```
williamcroberts: I missed this coming into master. This is broken and we need not to introduce this on 3.X. Please make this a makecredential and whatever other tools actually can make use of it specific change.

See: https://github.com/tpm2-software/tpm2-tools/issues/1442

jetwhiz: I think we can move the option from the main tools list into the makecredential option list: 
https://github.com/tpm2-software/tpm2-tools/blob/9ea358e464c3b2885b926b66c0493c5bd93f2611/lib/tpm2_options.c#L334-L342

But I'm not sure how we can move the logic into makecredential from options.c: 
https://github.com/tpm2-software/tpm2-tools/blob/9ea358e464c3b2885b926b66c0493c5bd93f2611/lib/tpm2_options.c#L408-L413

Since SAPI is set up a few lines later (which requires a TPM): 
https://github.com/tpm2-software/tpm2-tools/blob/9ea358e464c3b2885b926b66c0493c5bd93f2611/lib/tpm2_options.c#L448-L452

There doesn't appear to be any way to modify the `tool_opts->flags` variable from within the tools themselves except for in `tpm2_tool_onrun`, but it is too late by then since SAPI will already have been set up. 

If we want to move the logic into makecredential, the only thing I can think of is to allow `on_option` to take in a pointer to `tool_opts->flags` for all tools.  Do you know of another option, or should I try this and update all of the tpm2 tools `on_option` functions?

williamcroberts: You can modify the flags by setting them when setting your on-option handler via:
tpm2_options_new() @ https://github.com/tpm2-software/tpm2-tools/blob/3.X/lib/tpm2_options.h#L142

williamcroberts: Oh they need to be modified post option handler init. We can just make a routine into options to set that flag.

williamcroberts: I almost wonder, rather than making it an option like -X, we allow an empty `-T` or `-T-` to say hey no TCTI and set a flag that says TCTI can be optional. I like that better.

williamcroberts: Feel free to take the changes on:
https://github.com/tpm2-software/tpm2-tools/pull/1445

and backport them to 3.X. I'd wait for the review period to close to see if anyone spots any issues or loudly objects.

jetwhiz: I like that solution ... I'll play around with the patch on my system.

jetwhiz: @williamcroberts I did something similar to your patch for master.  Let me know what you think.

Issue #1206, opened 2018-10-22
tpm2_createprimary: rsa default incorrect


williamcroberts:
As initially reported in this thread: https://github.com/tpm2-software/tpm2-pkcs11/issues/52

The man page for [tpm2_createprimary](https://github.com/tpm2-software/tpm2-tools/blame/master/man/tpm2_createprimary.1.md#L54) states that the default algorithm is `RSA` when it is really [rsa2048:aes256](https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_createprimary.c#L58).

We should look at:
- [x] updating the manpage/code
    - I think  perhaps defaulting to RSA and using the restricted field in the object attributes to determine if
 it needs sym key details would be better. Then choosing AES<max-keysize> on that case. 
- [ ] the code in [tpm2_alg_util_public_init()](https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_alg_util.c#L858) uses the object attributes to try and find sane defaults. However, these choices to not seem to be the best:
    - The use of AES gets us into issues where we select aes256 and only aes128 is supported. Perhaps use getcap to find the biggest size supported.
    ```
    tpm2_getcap - c properties-fixed
    # parse yaml and query property TPM2_PT_CONTEXT_SYM_SIZE
    ```

williamcroberts:
We have this set to 128. The user can externally choose the best algorithm if we pick the wrong one. Getcap is so slow, I would like to avoid going to the TPM for these things.

PR #1443, opened 2019-04-29
Set default aes key size to 128 bit

DomiNic5787:
Set `DEFAULT_PRIMARY_KEY_ALG` of `tpm2_createprimary` to `rsa` instead of `rsa2048:aes256`.

So default algorithm of `tpm2_createprimary` is also compliant with "TCG PC Client Platform TPM Profile (PTP)", where AES256 is still optional.

### DIFF
--- a/lib/tpm2_alg_util.c
+++ b/lib/tpm2_alg_util.c
@@ -347,16 +347,6 @@ static bool handle_rsa(const char *ext, TPM2B_PUBLIC *public) {
     TPMS_RSA_PARMS *r = &public->publicArea.parameters.rsaDetail;
     r->exponent = 0;
 
-    /*
-     * Deal with normalizing the input strings.
-     *
-     * "rsa --> maps to rsa2048:aes128cbc
-     * "rsa:aes --> maps to rsa2048:aes128cbc
-     * "rsa:null" -- maps to rsa2048:null
-     *
-     * This function is invoked with rsa removed.
-     */
-
     bool is_restricted = !!(public->publicArea.objectAttributes & TPMA_OBJECT_RESTRICTED);
 
     size_t len = strlen(ext);
--- a/lib/tpm2_alg_util.c
+++ b/lib/tpm2_alg_util.c
@@ -146,12 +146,10 @@ static bool handle_aes_raw(const char *ext, TPMT_SYM_DEF_OBJECT *s) {
 
     if (!strncmp(ext, "128", 3)) {
         s->keyBits.aes = 128;
+    } else if (!strncmp(ext, "192", 3)) {
+        s->keyBits.aes = 192;
     } else if (!strncmp(ext, "256", 3)) {
         s->keyBits.aes = 256;
-    } else if (!strncmp(ext, "384", 3)) {
-        s->keyBits.aes = 384;
-    } else if (!strncmp(ext, "512", 3)) {
-        s->keyBits.aes = 512;
     } else {
         return false;
     }
--- a/lib/tpm2_alg_util.c
+++ b/lib/tpm2_alg_util.c
@@ -141,7 +141,7 @@ static bool handle_aes_raw(const char *ext, TPMT_SYM_DEF_OBJECT *s) {
     s->algorithm = TPM2_ALG_AES;
 
     if (*ext == '\0') {
-        ext = "256";
+        ext = "128";
     }
 
     if (!strncmp(ext, "128", 3)) {
--- a/man/common/alg.md
+++ b/man/common/alg.md
@@ -65,9 +65,9 @@ general format for specifying this data:
    This portion of the complex algorithm specifier is required. The remaining scheme and symmetric details
    will default based on the type specified and the type of the object being created.
 
-  * aes - Default AES: aes128cfb
+  * aes - Default AES: aes128
   * aes128`<mode>` - 128 bit AES with optional mode (*ctr*|*ofb*|*cbc*|*cfb*|*ecb*). If mode is not
-      specified, defaults to *cfb*.
+      specified, defaults to *null*.
   * aes256`<mode>` - Same as aes128`<mode>`, except for a 256 bit key size.
   * ecc - Elliptical Curve, defaults to ecc256.
   * ecc192 - 192 bit ECC
--- a/man/common/alg.md
+++ b/man/common/alg.md
@@ -68,6 +68,7 @@ general format for specifying this data:
   * aes - Default AES: aes128
   * aes128`<mode>` - 128 bit AES with optional mode (*ctr*|*ofb*|*cbc*|*cfb*|*ecb*). If mode is not
       specified, defaults to *null*.
+  * aes192`<mode>` - Same as aes128`<mode>`, except for a 192 bit key size.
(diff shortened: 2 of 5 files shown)


tools/tpm2_createprimary.c
```diff
@@ -55,7 +55,7 @@
     |TPMA_OBJECT_FIXEDTPM|TPMA_OBJECT_FIXEDPARENT \
     |TPMA_OBJECT_SENSITIVEDATAORIGIN|TPMA_OBJECT_USERWITHAUTH
 
-#define DEFAULT_PRIMARY_KEY_ALG "rsa2048:aes256"
+#define DEFAULT_PRIMARY_KEY_ALG "rsa"
```
williamcroberts: If you walk through https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_alg_util.c#L346 This would end up being:
`rsa2048:null:aes128cfb`

Also, doesn't that spec have AES(128|256)CFB as mandatory:
https://trustedcomputinggroup.org/wp-content/uploads/TCG_PC_Client_Platform_TPM_Profile_PTP_Specification_Family_2.0_Revision_1.3v22.pdf
 at Table 2?

DomiNic5787: Sorry I don't see the problem that it ends up in `rsa2048:null:aes128cfb`. Can you please explain? If we have a problem here, then we have it in 90% of our tests using `tpm2_createprimary` with rsa key, because they use `-G rsa`.

Regarding PTP Spec:

> In Section 3.2, Table 2, the comment for TPM_ALG_AES incorrectly includes 256-bit keys as
necessary minimum support. This revision of the specification should be interpreted to
require support for only 128-bit AES keys.

see https://trustedcomputinggroup.org/wp-content/uploads/Errata_v1.1_TCG_PC_Client_Platform_TPM_Profile_PTP_2.0_r1.03_v22.pdf, 3.1 Errata 1

williamcroberts: > Sorry I don't see the problem that it ends up in `rsa2048:null:aes128cfb`. Can you please explain? If we have a problem here, then we have it in 90% of our tests using `tpm2_createprimary` with rsa key, because they use `-G rsa`.

`-G rsa` isn't the problem, the problem is explained in the errata below, it defaults to 128, i thought 256 was mandatory.

This also ties into issues:
  - https://github.com/tpm2-software/tpm2-tools/issues/1206
  - i think their is more, I cant find them.

We should probably pick defaults based on what the TPM actually supports and picking the best by running getcap internally versus guessing at keysize.

So the question is, do you actually want to fix the algorithm identifiers the correct way, and choose sizes and algorithms by querying the TPM?

DomiNic5787: I set the key size to the full string, now.

I would prefer to set default algorithms to the strongest, minimal required algorithm.
And if someone wants to use the strongest supported algorithm, one can still do this using getcap externally.

Furthermore I set default aes key size to 128 in general and updated/fixed the manpages

man/common/object-alg.md
```diff
@@ -40,18 +40,18 @@ on an RSA scheme, like RSAES_OAEP.
...
+    size to either 1024, 2048 or 4096 respectively.
 
   * **rsa[1024|2048|4096]:[oaep|rsaes]** -
     Similar to **rsa[1024|2048|4096]** option, but provides the ability
     to control the scheme. The algorithms encryption options will default to:
-    aes256cfb.
+    aes128cfb.
 
-  * **rsa[1024|2048]:[oaep|rsaes]:[aes]**
+  * **rsa[1024|2048|4096]:[oaep|rsaes]:[aes]**
```
williamcroberts: this is a non-related fix, separate patch.

man/common/alg.md
```diff
@@ -65,9 +65,10 @@ general format for specifying this data:
    This portion of the complex algorithm specifier is required. The remaining scheme and symmetric details
    will default based on the type specified and the type of the object being created.
 
-  * aes - Default AES: aes128cfb
+  * aes - Default AES: aes128
   * aes128`<mode>` - 128 bit AES with optional mode (*ctr*|*ofb*|*cbc*|*cfb*|*ecb*). If mode is not
-      specified, defaults to *cfb*.
+      specified, defaults to *null*.
```
williamcroberts: Looks like the code was setting the default to null and you're just correcting that issue in the man page, is that right?

Separate patch.

tools/tpm2_createprimary.c
```diff
@@ -55,7 +55,7 @@
     |TPMA_OBJECT_FIXEDTPM|TPMA_OBJECT_FIXEDPARENT \
     |TPMA_OBJECT_SENSITIVEDATAORIGIN|TPMA_OBJECT_USERWITHAUTH
 
-#define DEFAULT_PRIMARY_KEY_ALG "rsa2048:aes256"
+#define DEFAULT_PRIMARY_KEY_ALG "rsa2048:null:aes128cfb"
```
williamcroberts: this patch looks good.

lib/tpm2_alg_util.c
```diff
@@ -141,17 +141,15 @@ static bool handle_aes_raw(const char *ext, TPMT_SYM_DEF_OBJECT *s) {
...
     if (!strncmp(ext, "128", 3)) {
         s->keyBits.aes = 128;
+    } else if (!strncmp(ext, "192", 3)) {
+        s->keyBits.aes = 192;
     } else if (!strncmp(ext, "256", 3)) {
         s->keyBits.aes = 256;
-    } else if (!strncmp(ext, "384", 3)) {
-        s->keyBits.aes = 384;
-    } else if (!strncmp(ext, "512", 3)) {
-        s->keyBits.aes = 512;
```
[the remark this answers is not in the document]
DomiNic5787: I am convinced that they are not present in any TPM, because they do not exist in any AES specification.
I will provide a separate commit.

lib/tpm2_alg_util.c
```diff
@@ -349,16 +347,6 @@ static bool handle_rsa(const char *ext, TPM2B_PUBLIC *public) {
...
-    /*
-     * Deal with normalizing the input strings.
-     *
-     * "rsa --> maps to rsa2048:aes128cbc
-     * "rsa:aes --> maps to rsa2048:aes128cbc
-     * "rsa:null" -- maps to rsa2048:null
-     *
-     * This function is invoked with rsa removed.
-     */
-
```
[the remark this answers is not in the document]
DomiNic5787: This mapping was wrong and after fixing it, the information is that they all map to rsa2048:null:aes128cfb.
I will provide a separate commit and then we can discuss if this comment is still important.

man/common/alg.md
DomiNic5787: Right.
I will provide a separate commit

man/common/object-alg.md
DomiNic5787: I will provide a separate commit

lib/tpm2_alg_util.c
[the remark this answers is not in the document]
williamcroberts: Yeah just delete it in a separate patch

Issue #1433, opened 2019-04-23
Fix the pandoc version for building the documentation


dilyanpalauzov:
Below is the manual page for tpm2_re_decode, generated by
`pandoc 1.19.2.4 Compiled with pandoc-types 1.17.0.5, texmath 0.9, skylighting 0.1.1.5` and `pandoc 2.7.2 Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.7.7`.

[tpm2_rc_decode.1_pandoc_1.19.2.4.txt](https://github.com/tpm2-software/tpm2-tools/files/3107253/tpm2_rc_decode.1_pandoc_1.19.2.4.txt)
[tpm2_rc_decode.1_pandoc_2.0.txt](https://github.com/tpm2-software/tpm2-tools/files/3107254/tpm2_rc_decode.1_pandoc_2.0.txt)

A diff between the files contains:
```diff
@@ -24,3 +24,3 @@
 .IP \[bu] 2
-\f[B]\-h\f[], \f[B]\-\-help=[man|no\-man]\f[]: Display the tools
+\f[B]\-h\f[], \f[B]\[en]help=[man|no\-man]\f[]: Display the tools
 manpage.
```

On my system `man tpm2_rc_decode` shows consistently a single dash before all the long-options, when I use pandoc 2.0 to generate the documentation, but not when pandoc 1.19.2.4 is used.

dilyanpalauzov:
The 2 version used for the file above was `pandoc 2.0 Compiled with pandoc-types 1.17.2, texmath 0.9.4.4, skylighting 0.4.2`, as contained in the first line of tpm2_rc_decode.1_pandoc_2.0.txt, it was not 2.7.2

williamcroberts:
I just tried to reproduce:

Pandoc Versions (1.17.0.5 to 2.7.2):
Old:
```
pandoc2 2.7.2
Compiled with pandoc-types 1.17.5.4, texmath 0.11.2.2, skylighting 0.7.7
Default user data directory: /home/wcrobert/.local/share/pandoc or /home/wcrobert/.pandoc
Copyright (C) 2006-2019 John MacFarlane
Web:  http://pandoc.org
This is free software; see the source for copying conditions.
There is no warranty, not even for merchantability or fitness
for a particular purpose.
```

New:
```
pandoc2 2.7.2
[...]

PR #1446, opened 2019-04-29
man: fixes for dashes with pandoc > 1.17.2

williamcroberts:
Some point in time, pandoc behavior changed with how to deal
with - or -- dashed when converting from markdown to man. They
got changed from a double-dash to an en-dash.

Per https://github.com/jgm/pandoc/issues/5404

Correct this by escaping the dash.

Fixes: #1433

### DIFF
--- a/man/common/options.md
+++ b/man/common/options.md
@@ -3,7 +3,7 @@
 This collection of options are common to many programs and provide
 information that many users may expect.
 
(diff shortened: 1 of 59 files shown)


PeterHuewe:
Why does it fix #1443?

williamcroberts:
> Why does it fix #1443?
> Apart from that it looks good.

So Close #1433

Issue #1453, opened 2019-05-02
tpm2_activatecredential activatehandle auth


dgptha:
The auth of the activateHandle (following "Part 2" naming) in the current implementation doesn't work.
This is fairly easy to test with `test/integration/tests/activatecredential.sh`: set an auth for the created ak (`-P password` option on the `tpm2_createak` command). With that `tpm2_activatecredential` with and without the `-P password` option doesn't work. `tpm2_activatecredential` systematically shows the following error:
```
WARNING:esys:src/tss2-esys/api/Esys_ActivateCredential.c:344:Esys_ActivateCredential_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_ActivateCredential.c:127:Esys_ActivateCredential() Esys Finish ErrorCode (0x0000098e)
ERROR: Esys_ActivateCredential(0x98E) - tpm:session(1):the authorization HMAC check failed and DA counter incremented
ERROR: Unable to run tpm2_activatecredential                                               
```

From my testing I found that on `tpm2_activatecredential.c` when creating the shandle for the activateHandle the keyHandle was actually being used. Changing the following code:
```
ESYS_TR key_shandle = tpm2_auth_util_get_shandle(ectx,
                        ctx.key_ctx_obj.tr_handle, &ctx.auth,
                        ctx.auth_session);
```
with:
```
ESYS_TR key_shandle = tpm2_auth_util_get_shandle(ectx,
                        ctx.ctx_obj.tr_handle, &ctx.auth,
                        ctx.auth_session);
```
solves the problem. With this change the `tpm2_activatecredential` command with the `-P passwd` option now works.

As a side note some variable renaming in `tpm2_activatecredential.c` code would be helpful to better follow the TCG docs. E.g., the `key_shandle` in the above code could be named `activate_shandle` to better match the TCG docs and don't confuse it with the shandle for the keyHandle.

lukehinds:
@dgptha , do you plan to submit a PR with your suggested fix in `key_shandle`?

idesai:
Fix proposed in https://github.com/tpm2-software/tpm2-tools/pull/1499. Thanks @dgptha.

PR #1456, opened 2019-05-03
Backport checkquote tool and PCR outputs for quote tool to 3.X

jetwhiz:
Backporting of features added to master branch to 3.X: 
* Add new tpm2_checkquote tool for checking quotes
  * Pull shared functionality from tpm2_pcrlist into pcr library
  * is_pcr_select_bit_set moved into tpm2_util library
  * Add new functionality to openssl, pcr and util libraries
* The tpm2_quote tool can now output PCR hash lists

New: 
* Add attestation test to check that the full attestation process works


lib/tpm2_util.c
```diff
@@ -41,6 +41,157 @@
 #include "tpm2_tool.h"
 #include "tpm2_util.h"
 
+#define FILE_PREFIX "file:"
+#define FILE_PREFIX_LEN (sizeof(FILE_PREFIX) - 1)
```
williamcroberts: I don't think you need this hunk

man/tpm2_quote.1.md
```diff
@@ -41,18 +41,25 @@
 
   * **-m**, **--message**:
 
-    message output file, records the quote message that makes up the data that
+    Message output file, records the quote message that makes up the data that
```
williamcroberts: separate patch for these to fixups in tpm2_quote.md.1

jetwhiz:
> I noticed the test_attestation.sh script and the verifyquote tool isn't on master, are you planning on porting that to master?

The verify quote tool should be on master, but we put it under the `tools/misc` directory since it does not use the TPM at all -- [master/tools/misc/tpm2_checkquote.c](https://github.com/tpm2-software/tpm2-tools/blob/master/tools/misc/tpm2_checkquote.c).  

Regarding the attestation test, we have one ready for master but it does not pass right now due to authentication issues on the master branch (as discussed on the mailing list: https://lists.01.org/pipermail/tpm2/2019-February/001046.html).  If you'd, like we can submit this with authentication parts stripped out, and then we can add it back after these issues are taken care of?

williamcroberts:
Looks like this bug is filed for 4.0: https://github.com/tpm2-software/tpm2-tools/issues/1453

man/tpm2_quote.1.md
jetwhiz: Should this be a separate PR or just in its own patch/commit?

williamcroberts: Its own commit on this PR is 100% OK.

lib/tpm2_util.c
jetwhiz: Good catch ... I forgot to strip that out when I ported from master

jetwhiz:
I'll keep an eye on that issue and push the attestation tests when the mentioned fix gets merged.  We're also looking to start tracking master for Keylime as well, so we'll be keeping an eye on that issue regardless.

lib/pcr.h
```diff
@@ -35,8 +35,35 @@
 
 #include <tss2/tss2_sys.h>
 
+typedef struct tpm2_algorithm tpm2_algorithm;
+struct tpm2_algorithm {
+    int count;
+    TPMI_ALG_HASH alg[8]; //XXX Why 8?
```
williamcroberts: You should be backporting from the tip of master so you pick up commits like 1bdaef80a8ad7 and 785c7cbc1927

jetwhiz: I just noticed that [82d987](https://github.com/tpm2-software/tpm2-tools/commit/82d987a8e75ebefec4cbf1c2bd3fda87c2aa5e34) hasn't been backported.  Should I include this commit?

williamcroberts: Yep

jetwhiz: Ok, I cherry picked that commit.  The latest should incorporate everything related to this PR from the tip of master.

PR #1466, opened 2019-05-08
Skip tests depending on XXD, when the latter is missing

dilyanpalauzov:


configure.ac
```diff
@@ -107,7 +107,7 @@ AS_IF([test "x$enable_unit" != xno], [
 
     AC_CHECK_PROG([XXD],[xxd],[yes])
     AS_IF([test "x${XXD}" != xyes],
-          [AC_MSG_WARN([Required executable xxd not found, some system tests will fail!])])
+          [AC_MSG_WARN([Executable xxd not found, some system tests will be skipped])])
```
williamcroberts: No, lets just make this a hard failure. I learned one thing working on Android OS, that is, you want your test suite to be stable and have as few branches as possible. If we can't get what we need to test, we should fail loudly.

dilyanpalauzov: Related to https://github.com/tpm2-software/tpm2-tools/issues/1368.

PR #1470, opened 2019-05-09
Add ccache to speed up travis runs.

DomiNic5787:
see [https://github.com/tpm2-software/tpm2-tss/pull/1403](https://github.com/tpm2-software/tpm2-tss/pull/1403)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ services:
 compiler:
 - gcc
 - clang
+cache: ccache
 env:
   matrix:
   - coverity_scan_run_condition='"$CC" = gcc'


diabonas:
Not sure whether that works here, since the software is built in a Docker container, so the ccache directory will also live there. I guess you would need to add ccache to the Dockerfile and bind mount `$HOME/.ccache` from the Travis container to the container.

williamcroberts:
> Not sure whether that works here, since the software is built in a Docker container, so the ccache directory will also live there. I guess you would need to add ccache to the Dockerfile and bind mount `$HOME/.ccache` from the Travis container to the container.

If it's inside the container won't it still use $HOME/.ccache which would be the containers version, where $HOME is /root?

diabonas:
As far as I can see, the Travis build directory [is mounted to `/workspace/tpm2-tools` in the container](https://github.com/tpm2-software/tpm2-tools/blob/c4367da5d502c0abe66f27ecf190168a63f66472/.ci/travis.run#L51), so everything in the build directory is visible to Travis. Everything else in the container is not retained, so the container `/root/.ccache` is lost when Travis tries to upload the cache after the build (from its own `$HOME/.ccache`). Anyway, the Docker container doesn't know about ccache since Travis only adds [`/usr/lib/ccache` to `$PATH`](https://docs.travis-ci.com/user/caching/#ccache-cache), but this isn't propagated to the container.

williamcroberts:
@diabonas I get what you mean now, I was thinking inside the container. Yeah the ccache would be gone and not visible to travis.

PR #1473, opened 2019-05-09
configure/Makefile/helpers.sh: enable running tests without resource manager

dilyanpalauzov:
* When ./configure --enable-unit=mssim is invoked, run the tests towards tpm_server, skipping tpm2-abrmd
* When ./configure --enable-unit or ./configure --enable-unit=abrmd is called, run the tests over the resource manager
* When ./configure or ./configure --enable-unit=no or ./configure --disable-unit is called, do not run tests on “make check”.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,7 +26,11 @@ To build and install the tpm2-tools software the following software is required:
   * To build the man pages you need [pandoc](https://github.com/jgm/pandoc)
   * To enable the new userspace resource manager, one must get tpm2-tabrmd
     (**recommended**).
-  * For the tests: tpm2-abrmd (must be on $PATH) and tpm_server
+  * When ./configure is invoked with --enable-unit or --enable-unit=abrmd,
+    the tests are run towards a resource manager, tpm2-abrmd, which must be on $PATH.
+  * When ./configure is invoked with --enable-unit=mssim, the tests are run directly
+    towards tpm_server, without resource manager.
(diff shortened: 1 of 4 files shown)


AndreasFuchsSIT:
If --enable-unit=mssim then the AC_CHECK_PROG([tpm_server]) should not issue a failure anymore
https://github.com/tpm2-software/tpm2-tools/blob/c4367da5d502c0abe66f27ecf190168a63f66472/configure.ac#L89-L93

williamcroberts:
No, because then we will have to write tests or skip tests that have too many active objects to be resident inside the TPM. The tests are designed to run against the full stack.

I would consider a hook, so folks can run their own tests within the harness and switch the tcti to what they want.

dilyanpalauzov:
The point is, that a tool that works incorrect (does not call flush or whatsoever) towards tpm_server, can pass the tests, when executed towards tpm2-abrmd, as the latter compensates for errors in the tools.

williamcroberts:
> The point is, that a tool that works incorrect (does not call flush or whatsoever) towards tpm_server, can pass the tests, when executed towards tpm2-abrmd, as the latter compensates for errors in the tools.

No it doesn't. Any RM including the in-kernel one should flush objects when the tool closes its connection. For /dev/tpmrm0 it's when close() is called.

williamcroberts:
The only thing that abrmd does that the in-kernel RM doesn't do is mark session handles as alive between tool invocations and extended sessions via -S won't work, which is why they are in specific TCTI test directory.

PR #1474, opened 2019-05-09
Tools: Fix unused return value of BIO_flush

AndreasFuchsSIT:
This is a compile error on some systems.

I have no idea why the tabrmd test is failing...
tabrmd upstream issue ?

### DIFF
--- a/tools/tpm2_getmanufec.c
+++ b/tools/tpm2_getmanufec.c
@@ -277,7 +277,7 @@ char *Base64Encode(const unsigned char* buffer)
     bio = BIO_push(b64, bio);
     BIO_set_flags(bio, BIO_FLAGS_BASE64_NO_NL);
     BIO_write(bio, buffer, SHA256_DIGEST_LENGTH);
-    BIO_flush(bio);
+    UNUSED(BIO_flush(bio));
     BIO_get_mem_ptr(bio, &bufferPtr);
 
     /* these are not NULL terminated */


diabonas:
I think you were probably building with the release version 2.2.2 of tpm2-tss, there `test/integration/tests/tcti/abrmd/extended-sessions.sh` fails for me as well. Switching to tpm2-tss master resolves this issue, I haven't investigated further why that happens.

AndreasFuchsSIT:
Hmpf... ok, so more bisect... :-(

tools/tpm2_getmanufec.c
```diff
@@ -277,7 +277,7 @@ char *Base64Encode(const unsigned char* buffer)
     bio = BIO_push(b64, bio);
     BIO_set_flags(bio, BIO_FLAGS_BASE64_NO_NL);
     BIO_write(bio, buffer, SHA256_DIGEST_LENGTH);
```
williamcroberts: ignored as well too?
```C
 int    BIO_write(BIO *b, const void *buf, int len);
```

AndreasFuchsSIT:
My compiler did not complain here; so I did not bother... ;-)

But I guess catching FILE_IO_ERRORs is allways a good idea...

williamcroberts:
> Hmpf... ok, so more bisect... :-(

The XFAIL tests are not the extended-session.sh test, its the policysecret and policypassword .sh tests.

tools/tpm2_getmanufec.c
```diff
@@ -277,7 +277,7 @@ char *Base64Encode(const unsigned char* buffer)
...
-    BIO_flush(bio);
+    (void)BIO_flush(bio);
```
williamcroberts: Don't just ignore it, handle it

AndreasFuchsSIT: I did a `grep -R BIO_flush` on openssl itself. 99% of invocations ignore the return code.
According to the man page, we'd have to call `BIO_should_retry` and if true, "try again at a later time".
If `BIO_should_retry` is false though, I wouldn't even know what to do ?!?

So should I really make this a
```
while ((r = BIO_flush(bio)) != 1 && BIO_shoud_retry(bio)) {
    sleep(1);
}
if (r) {
     LOG_ERR("Flush failed");
}
```
Sounds awful... ;-)
IMHO, I'd stick with `(void)`. But I'm no openssl-expert...

williamcroberts: Use our UNUSED(x) macro.

AndreasFuchsSIT: That I can do...

Issue #1016, opened 2018-04-27
Support PCR Policy via password language


williamcroberts:
Get rid of the options for supporting a PCR policy and use a minilanguage in the authorizations passwords language.

Something like:
```
-P pcr:f=<pcr_file>
-P pcr:p=<pcr_spec> 
```
I have something started on this branch, for getting the HMAC auth stuff in place. I think we can rip out what we need and at least get PCR policy support in place.

https://github.com/williamcroberts/tpm2.0-tools/tree/bills-hmac-auth

PR #1479, opened 2019-05-09
pcr policy: drop -L and -F options and move to policy lang

williamcroberts:
Drop the -F and -L options on nvread, nvwrite, nvincrement and unseal and support via a pcr authorization mini-language. By default, all tools will now support satisfying simple PCR policies via authorization strings.

Fixes: #1016


martinezjavier:
@williamcroberts I see you already merged this. But for what is worth I agree that it makes it easier to understand and simpler.

Issue #1262, opened 2018-12-20
move to config.h


williamcroberts:
Use a config header vs command line arguments so we don't overflow the command line length.

PR #1486, opened 2019-05-12
configure: switch to autoheader

williamcroberts:
Default configure passes all configure options like #defines through the
command line which can result in over-length commandlines.  Its better
to use a config.h

Fixes: #1262

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,10 @@
 AC_INIT([tpm2-tools],
     [m4_esyscmd_s([git describe --tags --always --dirty])])
 AC_CONFIG_MACRO_DIR([m4])
+
+AX_IS_RELEASE([dash-version])
+AX_CHECK_ENABLE_DEBUG([yes])
+
 AC_PROG_CC
 AC_PROG_LN_S
 LT_INIT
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -1,6 +1,10 @@
 AC_INIT([tpm2-tools],
     [m4_esyscmd_s([git describe --tags --always --dirty])])
 AC_CONFIG_MACRO_DIR([m4])
+
+AX_IS_RELEASE([dash-version])
+AX_CHECK_ENABLE_DEBUG([yes])
```
diabonas: When trying to build a debug version, compiling fails with the warning (treated as an error)
```
/usr/include/features.h:382:4: error: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Werror=cpp]
 #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
    ^~~~~~~
```
This is because `AX_CHECK_ENABLE_DEBUG([yes])` [adds `-g -O0` to the compilation flags](https://www.gnu.org/software/autoconf-archive/ax_check_enable_debug.html). Perhaps use  `AX_CHECK_ENABLE_DEBUG([info])` instead, which only adds `-g`?

DomiNic5787: I'm facing the same problem with this version. Shall we create an issue entry?

diabonas: @DomiNic5787 Sounds good :+1: I am still not sure what the best solution is here, maybe instead of `AX_CHECK_ENABLE_DEBUG([info])` use `--disable-hardening` for debug builds?

diabonas: I opened #1495 to use `AX_CHECK_ENABLE_DEBUG([info])` because I think most people will probably want to keep the hardening, but add debugging symbols.

PR #1487, opened 2019-05-13
Remove duplicated output of persistent handle

PeterHuewe:
With commit 9c3e6387 ("tpm2_evictcontrol: support saving ESYS_TR") the
persistent handle is printed twice, with two different names
('persistentHandle' vs 'persistent-handle').
Only one of them is needed, and the latter is closer to other similar
outputs.


williamcroberts:
Whoops thought I deleted it

PR #1493, opened 2019-05-14
Fix policysecret integration test and revert it from XFAIL list

idesai:


idesai:
Should fix issue #1326

PR #1494, opened 2019-05-15
Add platform hierarchy support on 3.X

remuswu1019:
Add platform hierarchy support and change typo in tpm2_takeownership.1.md.

### DIFF
--- a/man/tpm2_takeownership.1.md
+++ b/man/tpm2_takeownership.1.md
@@ -27,6 +27,12 @@ sign.
     Passwords should follow the password formatting standards, see section
     "Password Formatting".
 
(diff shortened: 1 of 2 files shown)


williamcroberts:
We can't change existing option names on stable api branches, I can take the other 2 patches though. Can you drop the rename and resubmit.

remuswu1019:
> We can't change existing option names on stable api branches, I can take the other 2 patches though. Can you drop the rename and resubmit.

I think the commit message use ambiguous word 'change', but it effect man page.
The original man page not match current code, so I fixed man page to fit current source.

Issue #1496, opened 2019-05-15
Unit-test test_tpm2_auth_util.c does not cleanup or remove the foobar file it creates


idesai:

idesai:
Pending PR 1498

idesai:
@williamcroberts we want it to hit the disk and back since files_load_bytes_from_buffer_or_file_or_stdin needs to be exercised as part of testing file IO with tpm2_auth_util:handle_file routine. Isn't it?

idesai:
@williamcroberts, i get it though and so have made changes to the PR.

PR #1498, opened 2019-05-15
test_tpm2_auth_util.c: Remove foobar file created as part of tpm2_auth_util unit test run

idesai:
A foobar file is created but not cleaned at end of tpm2_auth_util unit test. 
This fixes the issue #1496 

### DIFF
--- a/test/unit/test_tpm2_auth_util.c
+++ b/test/unit/test_tpm2_auth_util.c
@@ -100,7 +100,6 @@ const char *mocked_file_data = "sekretpasswrd";
 FILE * __real_fopen(const char *path, const char *mode);
 FILE * __wrap_fopen(const char *path, const char *mode) {
     if (strcmp (path, "test_tpm2_auth_util_foobar")) {
-        printf("REAL CALLED\n");
         return __real_fopen(path, mode);
     }
     return mock_ptr_type(FILE*);
--- a/test/unit/test_tpm2_auth_util.c
+++ b/test/unit/test_tpm2_auth_util.c
@@ -110,7 +110,6 @@ size_t __wrap_fread(void *ptr, size_t size, size_t nmemb, FILE *stream) {
     if (stream != &mocked_file_stream) {
         return __real_fread(ptr,size,nmemb,stream);
     }
-    UNUSED(stream);
     memcpy(ptr, mocked_file_data, size * nmemb);
     return mock_type (size_t);
 }
@@ -120,7 +119,6 @@ long __wrap_ftell(FILE *stream) {
     if (stream != &mocked_file_stream) {
         return __real_ftell(stream);
     }
-    UNUSED(stream);
     return mock_type (long);
 }
 
@@ -129,9 +127,6 @@ int __wrap_fseek(FILE *stream, long offset, int whence) {
     if (stream != &mocked_file_stream) {
         return __real_fseek(stream, offset, whence);
     }
-    UNUSED(stream);
-    UNUSED(offset);
-    UNUSED(whence);
     return 0;
 }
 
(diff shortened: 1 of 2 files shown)


williamcroberts:
We should make file_fopen() that calls real fopen() so we can safely mock fike_fopen() and prevent this from actually hitting the disk.

idesai:
Bookmarking issue https://github.com/tpm2-software/tpm2-tools/issues/1496 #1496

idesai:
> We should make file_fopen() that calls real fopen() so we can safely mock fike_fopen() and prevent this from actually hitting the disk.

mocked file operations directly.

idesai:
Fixes #1496

test/unit/test_tpm2_auth_util.c
```diff
@@ -94,27 +94,77 @@ static void test_tpm2_password_util_from_optarg_str_escaped_hex_prefix(void **st
...
+    return mock_type (size_t);
+}
+
+long __real_ftell(FILE *stream);
+long __wrap_ftell(FILE *stream) {
+    if (stream != &mocked_file_stream) {
+        return __real_ftell(stream);
+    }
+    UNUSED(stream);
+    return mock_type (long);
```
williamcroberts: we don't do spaces between function name and opening parenthesis. This applies here and throughout

test/unit/test_tpm2_auth_util.c
```diff
@@ -94,27 +94,77 @@ static void test_tpm2_password_util_from_optarg_str_escaped_hex_prefix(void **st
...
+FILE mocked_file_stream;
+const char *mocked_file_data = "sekretpasswrd";
+
+FILE * __real_fopen(const char *path, const char *mode);
+FILE * __wrap_fopen(const char *path, const char *mode) {
+    if (strcmp (path, "test_tpm2_auth_util_foobar")) {
+        printf("REAL CALLED\n");
+        return __real_fopen(path, mode);
+    }
+    return mock_ptr_type(FILE*);
```
williamcroberts: Why not pass a union pointer via will_return() that contains:
```C
union {
bool is_fopen;
struct {
    const char *path;
    const char *mode;
} fopen_args;
struct {
} fememopen_args;
    void *buffer;
    size_t size;
    const char *mode;
};
```

Then all you need to do is mock fopen(), the other calls should stay the same and you have valid data content to check against too if we want to introspect the file or buffer.

williamcroberts: Disregard this, I am overall fine with your approach, just clean up the other nits.

test/unit/test_tpm2_auth_util.c
```diff
@@ -94,27 +94,77 @@ static void test_tpm2_password_util_from_optarg_str_escaped_hex_prefix(void **st
     tpm2_session_close(&session);
 }
 
+FILE mocked_file_stream;
+const char *mocked_file_data = "sekretpasswrd";
+
+FILE * __real_fopen(const char *path, const char *mode);
+FILE * __wrap_fopen(const char *path, const char *mode) {
+    if (strcmp (path, "test_tpm2_auth_util_foobar")) {
+        printf("REAL CALLED\n");
```
williamcroberts: rm printf

test/unit/test_tpm2_auth_util.c
```diff
@@ -94,27 +94,77 @@ static void test_tpm2_password_util_from_optarg_str_escaped_hex_prefix(void **st
...
+    }
+    return mock_ptr_type(FILE*);
+}
+
+size_t __real_fread(void *ptr, size_t size, size_t nmemb, FILE *stream);
+size_t __wrap_fread(void *ptr, size_t size, size_t nmemb, FILE *stream) {
+    if (stream != &mocked_file_stream) {
+        return __real_fread(ptr,size,nmemb,stream);
+    }
+    UNUSED(stream);
```
williamcroberts: shouldn't need this UNUSED(), it's used above it.

Issue #1482, opened 2019-05-10
Personal fork dies on CI


AndreasFuchsSIT:
This is more a request for help:

My personal fork dies on https://github.com/tpm2-software/tpm2-tools/blob/7b93d52bf9cd897ca20de9d94a7f8c3e7a6e2ed5/.ci/docker.run#L92 with
```
fatal: ambiguous argument 'origin/master': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```
(see https://travis-ci.org/AndreasFuchsSIT/tpm2-tools/jobs/530778525)

Any idea, why that is or how I can fix that ?

williamcroberts:
no remote called origin or no branch called master.... I wonder what they call your remote branch combo...

Looks like your remote name is called:
https://travis-ci.org/AndreasFuchsSIT/tpm2-tools/jobs/530778525#L420

diabonas:
The problem here is that Travis CI does a [shallow clone of the current branch](https://travis-ci.org/AndreasFuchsSIT/tpm2-tools/jobs/530778525#L419), in your case the `fix_unused_value` branch. Therefore the `master` branch is not available for comparison. For pull requests, Travis clones the base branch first, then fetches and checks out the PR, so there the logic works.

It might be nice to replace `git diff --check origin/master` by `git diff --check origin/$TRAVIS_BRANCH`, which [points either to the current branch or in case of a pull request, the base branch targeted by the PR](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables). Otherwise the current logic would also fail if it was used on a different branch in the main repository (e.g. on a future `4.X` branch).

In case of feature branches in a personal repository, `git diff --check origin/$TRAVIS_BRANCH` would be a no-op, but it at least wouldn't fail. I think it's quite hard to determine with Git where the current branch forked off from the "upstream" branch, so I think this is the best we can do.

AndreasFuchsSIT:
Thanks for figuring this out...

However, this would mean that people don't receive whitespace errors on travis runs in fork's branches but only once they issue a PR. But yeah, it's already better than the current situation...

diabonas:
[...]

PR #1502, opened 2019-05-16
Fix whitespace errors

diabonas:
Following the discussion in #1482, fix all whitespace errors reported by
```
git diff --check $(git hash-object -t tree /dev/null)
```
We have to be a bit careful with Markdown files because [two trailing spaces at the end of a line are considered as a line break](https://meta.stackexchange.com/questions/40976/what-is-the-reason-for-the-top-secret-two-space-newline-markdown-weirdness), but luckily this feature is not used here.


williamcroberts:
For md looks like you can use a backslash

Issue #1147, opened 2018-08-13
tpm2_clearlock - bad option name on -L and rename


williamcroberts:
The hierarchy for clearcontrol command can be either LOCKOUT or PLATFORM, but the parameter for a password is done via:
```
        { "auth-lockout",   required_argument, NULL, 'L' },
```

This should probably be -P

williamcroberts:
The tool should also be called tpm2_clearcontrol, not clearlock.

williamcroberts:
This -L can also be changed in tpm2_clear

idesai:
> The tool should also be called tpm2_clearcontrol, not clearlock.

The tool is purported to block those with knowledge owner hierarchy authorization from using it to clear the TPM. 

I agree clearlock does not convey that in anyway, rather sounds like it would clear a lockout :)

On a real TPM the owner hierarchy losing it's ability to clear the tpm is a big deal especially since the only way to resolve the situation is to get access to the platform hierarchy which is potentially more difficult on a real platform unless platform firmware has provisions for it.

I feel very tempted to break this tool into two explicit operations:
1. tpm2_disable_clear : this can take owner auth and platform auth
2. tpm2_enable_clear : this can only take platform auth

idesai:
@williamcroberts any thoughts on this ?

williamcroberts:
> > The tool should also be called tpm2_clearcontrol, not clearlock.
> 
> The tool is purported to block those with knowledge owner hierarchy authorization from using it to clear the TPM.
> 
> I agree clearlock does not convey that in anyway, rather sounds like it would clear a lockout :)
> 
> On a real TPM the owner hierarchy losing it's ability to clear the tpm is a big deal especially since the only way to resolve the situation is to get access to the platform hierarchy which is potentially more difficult on a real platform unless platform firmware has provisions for it.
> 
> I feel very tempted to break this tool into two explicit operations:
> 
> 1. tpm2_disable_clear : this can take owner auth and platform auth
> 2. tpm2_enable_clear : this can only take platform auth

No we want the tool name to be close to the TPM command so people can go from spec knowledge to tools easily. The command is clearcontrol, thus tpm2_clearcontrol.

And then make the option names reasonable. We should be specifying hierarchy as a option input:
--hierarchy=<o|l> with it defaulting to platform or lockout (whichever is reasonable) and convert the command line option via a call to `tpm2_hierarchy_from_optarg()` and then a required *argument* of 0|1|set|clear

[...]

PR #1509, opened 2019-05-17
Change clear control tool options to accurately reflect spec language

idesai:
Fixes: #1147

### DIFF
--- a/tools/tpm2_clearcontrol.c
+++ b/tools/tpm2_clearcontrol.c
@@ -2,6 +2,7 @@
 
 #include <stdbool.h>
 #include <stdlib.h>
+#include <string.h>
 
 #include "log.h"
 #include "tpm2_auth_util.h"
@@ -11,30 +12,29 @@
 
 typedef struct clearcontrol_ctx clearcontrol_ctx;
 struct clearcontrol_ctx {
-    bool clear;
-    bool platform;
-
+    TPMI_YES_NO disable_clear;
+    ESYS_TR rh;
     struct {
         char *auth_str;
         tpm2_session *session;
     } auth;
 };
 
-static clearcontrol_ctx ctx;
+static clearcontrol_ctx ctx = {
+    .rh = ESYS_TR_RH_PLATFORM,
+    .disable_clear = 0,
+};
 
 static bool clearcontrol(ESYS_CONTEXT *ectx) {
 
-    ESYS_TR rh = ctx.platform ? ESYS_TR_RH_PLATFORM : ESYS_TR_RH_LOCKOUT;
-    TPMI_YES_NO disable = ctx.clear ? 0 : 1;
+    LOG_INFO ("Sending TPM2_ClearControl(%s) disableClear command with auth handle %s",
+            ctx.disable_clear ? "SET" : "CLEAR",
+            ctx.rh == ESYS_TR_RH_PLATFORM ? "TPM2_RH_PLATFORM" : "TPM2_RH_LOCKOUT");
 
-    LOG_INFO ("Sending TPM2_ClearControl(%s) command on %s",
-            ctx.clear ? "CLEAR" : "SET",
-            ctx.platform ? "TPM2_RH_PLATFORM" : "TPM2_RH_LOCKOUT");
+    ESYS_TR shandle = tpm2_auth_util_get_shandle(ectx, ctx.rh, ctx.auth.session);
 
-    ESYS_TR shandle = tpm2_auth_util_get_shandle(ectx, rh, ctx.auth.session);
-
-    TSS2_RC rval = Esys_ClearControl(ectx, rh,
-                shandle, ESYS_TR_NONE, ESYS_TR_NONE, disable);
+    TSS2_RC rval = Esys_ClearControl(ectx, ctx.rh,
+                shandle, ESYS_TR_NONE, ESYS_TR_NONE, ctx.disable_clear);
     if (rval != TPM2_RC_SUCCESS && rval != TPM2_RC_INITIALIZE) {
         LOG_PERR(Esys_ClearControl, rval);
         return false;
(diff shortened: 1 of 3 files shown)


idesai:
Addresses issue #1147

tools/tpm2_clearcontrol.c
```diff
@@ -24,17 +27,14 @@ static clearcontrol_ctx ctx;
```
williamcroberts: Just initialize the handle to a default.
```C
static clearcontrol_ctx ctx = {
    .rh = ESYS_TR_RH_PLATFORM,
};
```

williamcroberts:
> Addresses issue #1147
No no, do:
Fixes: #1147
As the first comment in the thread, see:
https://help.github.com/en/articles/closing-issues-using-keywords

tools/tpm2_clearcontrol.c
```diff
@@ -44,16 +44,63 @@ static bool clearcontrol(ESYS_CONTEXT *ectx) {
     return true;
 }
 
+static bool set_clearcontrol_auth_handle(const char *value) {
```
williamcroberts: No, use: tpm2_hierarchy_from_optarg();

tools/tpm2_clearcontrol.c
williamcroberts: Just initialize the handle to a default.
```C
static clearcontrol_ctx ctx = {
    .rh = ESYS_TR_RH_PLATFORM,
};
```

tools/tpm2_clearcontrol.c
```diff
@@ -44,16 +44,63 @@ static bool clearcontrol(ESYS_CONTEXT *ectx) {
...
+}
+
 static bool on_option(char key, char *value) {
 
+    bool result;
     switch (key) {
+    case 's':
+        ctx.disable_clear = 1;
+        ctx.set = true;
+        break;
```
williamcroberts: arguments, not options in the call to tpm2_options_new() set the on_arg handler and the parse the arguments like so:
 - argument count in argc must be equal to 1
 - the argument can be one of:
  - 0|c|clear
  - 1|s|set

tools/tpm2_clearcontrol.c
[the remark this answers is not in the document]
idesai: @williamcroberts tpm2_hierarchy_from_optarg does not return ESYS_TR.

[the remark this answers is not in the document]
idesai: also it cannot return lockout handle and rightly so since it is not a hierarchy. See #1510

idesai:
> > Addresses issue #1147
> 
> No no, do:
> Fixes: #1147
> As the first comment in the thread, see:
> https://help.github.com/en/articles/closing-issues-using-keywords

Did not realize there was some automagic there, thanks for the pointer.

tools/tpm2_clearcontrol.c
```diff
@@ -11,30 +12,29 @@
...
 
 static bool clearcontrol(ESYS_CONTEXT *ectx) {
 
-    ESYS_TR rh = ctx.platform ? ESYS_TR_RH_PLATFORM : ESYS_TR_RH_LOCKOUT;
-    TPMI_YES_NO disable = ctx.clear ? 0 : 1;
-
-    LOG_INFO ("Sending TPM2_ClearControl(%s) command on %s",
-            ctx.clear ? "CLEAR" : "SET",
-            ctx.platform ? "TPM2_RH_PLATFORM" : "TPM2_RH_LOCKOUT");
+    LOG_WARN ("Sending TPM2_ClearControl(%s) disableClear command with auth handle %s",
```
williamcroberts: LOG_INFO, why is this warning now?

tools/tpm2_clearcontrol.c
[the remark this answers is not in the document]
williamcroberts: Add a flag to tpm2_hierarchy_flags to return the ESAPI handle constant. Probably change TPMI_RH_PROVISION to something generic. Everything is a UINT32 so it will fit in there.

For example modify each segment to do something like this:
```C
    bool is_o = !strcmp(value, "o");
    if (is_o) {
        if (!(flags & TPM2_HIERARCHY_FLAGS_O)) {
            LOG_ERR("Owner hierarchy not supported by this command.");
            return false;
        }
        *hierarchy = !(flags & TPM2_HIERARCHY_FLAGS_ESAPI) ? TPM2_RH_OWNER : ESYS_TR_RH_OWNER;
        return true;
    }
```

tools/tpm2_clearcontrol.c
idesai: @williamcroberts, for what the tool does - Disable owner capability to clear the TPM - I thought that it should at least warn the user of this choice. So future failed attempts to clear the TPM with owner auth will have a context.

tools/tpm2_clearcontrol.c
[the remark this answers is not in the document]
idesai: @williamcroberts since lockout is a permanent handle and not a hierarchy it would not belong in the routine: tpm2_hierarchy_from_optarg. 

Additionally, The issue i am alluding to in #1510 is that https://github.com/tpm2-software/tpm2-tools/blob/5b96ac3983407fa273b4455a091a399ee15f6b48/lib/tpm2_hierarchy.c#L78 does not do all the checks necessary to convert a hex number to one of hierarchy handles owner, platform, endorsement or null.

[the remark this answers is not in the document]
williamcroberts: The goal here is to have consistent behavior between tools including things like error messages. So unfortunately not using a common routine for changing persistent handles can cause differences in behavior. Since this is just for platform or lockout, I guess its fine for now.

tools/tpm2_clearcontrol.c
williamcroberts: Your warning doesn't convey this. Your warning just says, "hey ive run this command with these arguments". If your trying to convey spec behavior, then say that LOG_WARN("Since your clearing the TPM future clears may fail");

I have no idea exactly what your trying to say here. I would just discard it, but if their is some higher level important information to convey, then do that. Think of a good code comment, don't tell me what the code does (I can read that or im executing the command I know that), tell me why it matters.

idesai: Reverted it to LOG_INFO

PR #1512, opened 2019-05-20
 build: link against tss2-mu

### LINKED ISSUE #1481: Requiring tpm2-tss >=2.3.0
https://github.com/tpm2-software/tpm2-tools/issues/1477#issuecomment-491050505 suggests, that the tpm2_unseal tool relies on the tpm2-tss to clean non-applicable session attributes for decrypt-/encrypt-sessions.
This was introduced to tpm2-tss as enhancement for the 2.3.0-dev series.
If tpm2-tools relies on this feature, it should PKG_CHECK for tpm2-tss >= 2.3.0

diabonas:
Currently the linking of [`libcommon`](https://github.com/tpm2-software/tpm2-tools/blob/54ef8cb2f5ed1f26d9f04852456a0f437600f92a/Makefile.am#L96) relies on the fact that the upstream tpm2-tss pkg-config file explicitly requires tss2-mu. Since this is wrong and will be changed upstream, but we need access to the marshalling functions, explicitly link against tss2-mu. See https://github.com/tpm2-software/tpm2-tss/pull/1417 for a more detailed discussion.

This PR additionally fixes #1481.


PeterHuewe:
@idesai Please review while Bill is out.

PR #1513, opened 2019-05-20
Fix eauth activate and add attestation test (prevent regression)

jetwhiz:
Together with fix from https://github.com/tpm2-software/tpm2-tools/pull/1499 this now matches the docs at [tss2-esys/api/Esys_ActivateCredential.c](https://github.com/tpm2-software/tpm2-tss/blob/master/src/tss2-esys/api/Esys_ActivateCredential.c#L18) (so comment was removed). 

Added attestation test (already on 3.X branch) to test out the full attestation process (which also tests that the included fix works correctly).

### DIFF
--- a/test/integration/tests/attestation.sh
+++ b/test/integration/tests/attestation.sh
@@ -1,0 +1,103 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+
+source helpers.sh
+
+handle_ek=0x81010009
+context_ak=ak.ctx
+handle_nv=0x1500018
+handle_hier=0x40000001
+ek_alg=rsa
+ak_alg=rsa
+digestAlg=sha256
+signAlg=rsassa
+ownerpw=ownerpass
+endorsepw=endorsepass
+ekpw=ekpass
+akpw=akpass
+rand_pcr_value=6ea40aa7267bb71251c1de1c3605a3df759b86b22fa9f62aa298d4197cd88a38
+debug_pcr=16
+debug_pcr_list=15,16,22
+
+file_input_data=secret.data
+file_input_key=nv.data
+output_ek_pub_pem=ekpub.pem
+output_ek_pub=ek.pub
+output_ak_pub_pem=akpub.pem
+output_ak_pub=ak.pub
+output_ak_pub_name=ak.name
+output_mkcredential=mkcred.out
+output_actcredential=actcred.out
+output_quote=quote.out
+output_quotesig=quotesig.out
+output_quotepcr=quotepcr.out
+
+cleanup() {
+  rm -f $file_input_data $file_input_key $output_ek_pub $output_ek_pub_pem $output_ak_pub \
+        $output_ak_pub_pem $output_ak_pub_name $output_mkcredential \
+        $output_actcredential $output_quote $output_quotesig $output_quotepcr $context_ak rand.out
+
+  tpm2_pcrreset -Q $debug_pcr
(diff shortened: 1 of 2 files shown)


test/integration/tests/attestation.sh
```diff
@@ -0,0 +1,100 @@
...
+getrandom 20
+tpm2_quote -Q -C $context_ak -L $digestAlg:15,16,22 -q $loaded_randomness -m $output_quote -s $output_quotesig -p $output_quotepcr -g $digestAlg -P "$akpw"
+
+
+# Verify quote
+tpm2_checkquote -Q -u $output_ak_pub_pem -m $output_quote -s $output_quotesig -F $output_quotepcr -g $digestAlg -q $loaded_randomness
+
+
+# Save U key from verifier
+tpm2_nvdefine -Q -x $handle_nv -a $handle_hier -s 32 -b "ownerread|policywrite|ownerwrite" -p "indexpass" -P "$ownerpw"
```
idesai: @jetwhiz can you specify a policy since you specify the policywrite attribute or drop the attribute if the policy attribute does not matter?

jetwhiz: That's a good point!  We've chosen to use an index password for auth instead of a pcr policy, so I have removed this attribute from the test.

idesai:
@jetwhiz is make check TESTS=test/integration/tests/attestation.sh running successfully at your end? I am suspecting your line endings in the attestation.sh may be having superfluous carriage returns.

test/integration/tests/attestation.sh
```diff
@@ -0,0 +1,100 @@
...
+file_size=`stat --printf="%s" $output_ak_pub_name`
+loaded_key_name=`cat $output_ak_pub_name | xxd -p -c $file_size`
+tpm2_makecredential -Q -T none -e $output_ek_pub -s $file_input_data -n $loaded_key_name -o $output_mkcredential
+tpm2_activatecredential -Q -c $context_ak -C $handle_ek -i $output_mkcredential -o $output_actcredential -P "$akpw" -E "$endorsepw"
+diff $file_input_data $output_actcredential
+
+
+# Quoting
+tpm2_pcrreset -Q 16
+tpm2_pcrextend -Q 16:sha256=6ea40aa7267bb71251c1de1c3605a3df759b86b22fa9f62aa298d4197cd88a38
```
idesai: Avoiding magic values will make the test more clear
sha256=$rand_pcr_value?

test/integration/tests/attestation.sh
```diff
@@ -0,0 +1,100 @@
...
+# Validate keys (registrar)
+file_size=`stat --printf="%s" $output_ak_pub_name`
+loaded_key_name=`cat $output_ak_pub_name | xxd -p -c $file_size`
+tpm2_makecredential -Q -T none -e $output_ek_pub -s $file_input_data -n $loaded_key_name -o $output_mkcredential
+tpm2_activatecredential -Q -c $context_ak -C $handle_ek -i $output_mkcredential -o $output_actcredential -P "$akpw" -E "$endorsepw"
+diff $file_input_data $output_actcredential
+
+
+# Quoting
+tpm2_pcrreset -Q 16
```
idesai: $debug_pcr=16?

jetwhiz:
> @jetwhiz is make check TESTS=test/integration/tests/attestation.sh running successfully at your end? I am suspecting your line endings in the attestation.sh may be having superfluous carriage returns.

Good catch!  That's what I get for pulling this file off of a Windows machine

test/integration/tests/attestation.sh
jetwhiz: I agree, I've moved these into variables now!

PR #1515, opened 2019-05-23
Disable dependency tracking.

idesai:
Since we only do one time builds with travis it is not useful to have dependency
tracking information. This should speed up travis builds.

### DIFF
--- a/.ci/coverity.run
+++ b/.ci/coverity.run
@@ -17,7 +17,7 @@ fi
 
 echo "Performing build with Coverity Scan"
 rm -fr $TRAVIS_BUILD_DIR/cov-int
-./bootstrap && ./configure && make clean
+./bootstrap && ./configure --disable-dependency-tracking && make clean
 cov-build --dir $TRAVIS_BUILD_DIR/cov-int make -j $(nproc)
 
 echo "Collecting Coverity data for submission"
--- a/.ci/docker.run
+++ b/.ci/docker.run
@@ -54,7 +54,7 @@ fi
 # Do a make distcheck in the root, clear it and than
 # cd to the variant directory.
 if [ "$CC" == "gcc" ]; then
-    ./configure
+    ./configure --disable-dependency-tracking
     make distcheck
     make distclean
 fi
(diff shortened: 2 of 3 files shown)


flihp:
I didn't know this flag existed. Interesting. Your commit message says this "should" speed up the CI builds. Would help your case if you posted some stats.

AndreasFuchsSIT:
Looking at the travis-builds over at https://github.com/idesai/tpm2.0-tools/commits/disable_dependency_tracking build-time got worse on the longer run, but faster on the short one...
https://travis-ci.org/idesai/tpm2.0-tools/builds/535427046?utm_source=github_status&utm_medium=notification : 11:46+11:39
https://travis-ci.org/idesai/tpm2.0-tools/builds/536051929?utm_source=github_status&utm_medium=notification : 11:58+11:21

tstruk:
What seems to speedup the TSS CI run quite a bit is ccache.
See: https://docs.travis-ci.com/user/caching/
https://github.com/tpm2-software/tpm2-tss/commit/5b84fb31de3684bb6b247c249f357fb69a684e0a

flihp:
@AndreasFuchsSIT : I looked at the stats from the CI runs but IMHO the numbers aren't conclusive (not enough data to go on). My question was intended to get @idesai to post his analysis that I would expect has better data from multiple local runs on his development system.

williamcroberts:
> What seems to speedup the TSS CI run quite a bit is ccache.
> See: https://docs.travis-ci.com/user/caching/
> [tpm2-software/tpm2-tss@5b84fb3](https://github.com/tpm2-software/tpm2-tss/commit/5b84fb31de3684bb6b247c249f357fb69a684e0a)

Yeah we need the CCACHE changes to land but that PR seemed to go stale when it wasn't as simple to get it to work in docker. See #1470

idesai:
> @AndreasFuchsSIT : I looked at the stats from the CI runs but IMHO the numbers aren't conclusive (not enough data to go on). My question was intended to get @idesai to post his analysis that I would expect has better data from multiple local runs on his development system.

> I didn't know this flag existed. Interesting. Your commit message says this "should" speed up the CI builds. Would help your case if you posted some stats.

@flihp I came across this flag while reading the auto tools manual and I thought it applied to us since we do kick off one time builds in CI. And the language in the manual seemed to suggest that it should reduce the time on one time builds since auto tools would not attempt to generate dependency information during compilation. Interestingly, the numbers are not very conclusive. If the numbers are getting worse like @AndreasFuchsSIT  indicated, then perhaps it may be good idea to not add this flag.   Also not sure what effect it has on the build other than what is mentioned in the auto tools manual, in that, CI would have another build profile as compared to platform builds where dependency tracking is necessary since those are mostly not one time builds.

PR #1522, opened 2019-05-31
Encryptdecrypt: Add functionality to accomodate input data greater than TPM2_MAX_DIGEST_BUFFER or 1KB

idesai:


tools/tpm2_encryptdecrypt.c
```diff
@@ -81,28 +78,55 @@ static bool encrypt_decrypt(ESYS_CONTEXT *ectx, TPM2B_IV *iv_in) {
...
+            return false;
+        }
+
+        if (!append) {
+            result = files_save_bytes_to_file(ctx.out_file_path,
+                out_data->buffer, out_data->size);
         } else {
-            LOG_PERR(Esys_EncryptDecrypt, rval);
+            result = files_append_bytes_to_file(ctx.out_file_path,
+                out_data->buffer, out_data->size);
```
williamcroberts: weird way to do this. Open the file in rw mode and keep the FILE * open the whole time and just write it out.

idesai: using files_write_bytes routine instead.

tools/tpm2_encryptdecrypt.c
```diff
@@ -78,44 +73,74 @@ static tool_rc encrypt_decrypt(ESYS_CONTEXT *ectx, TPM2B_IV *iv_in) {
...
     rc = tool_rc_success;
 
 out:
-    free(out_data);
-    free(iv_out);
+    if (out_data) {
+        free(out_data);
+    }
+    if (iv_out) {
+        free(iv_out);
```
williamcroberts: free doesn't need to be protected by conditionals. free(NULL) is valid.

lib/files.c
```diff
@@ -686,9 +686,25 @@ bool files_read_header(FILE *out, uint32_t *version) {
...
     const char *path, UINT16 *size, BYTE *buf) {
-    bool res;
-    UINT16 original_size = *size;
+    /*
+     * Read from stdin where in size is not fixed or known
+     * Size however cannot be bigger than UINT16_MAX
+     */
+    int c;
+    if (!input_buffer && !path && !*size) {
+        while ((c = getchar()) != EOF) {
```
williamcroberts: Wait, can't we read bigger than 1 byte at a time? like set the buffer size on read to TPM2B_MAX_BUFFER_SIZE (or whatever it is) and then read until you get a short read. If the short read returns more than 0 bytes, encrypt it, else dont.

williamcroberts: You also can't do this here, because you can't read all the data before hand... it would simpler to:
1. get a FILE * to the input source (file or stdin)
while read > 0 do
    2. read max buffer
    3. run encrypt
    4. output to output file
    5. update any IV's/chaining between blocks (not sure if we have to do this)

done

idesai: read > 0..... what does read mean here?
with getchar am reading a byte at a time to be able to determine end of stdin stream of unknown size

williamcroberts: Yeah, terribly slow. So do fread with some buffer size, fread may return:
RETURN VALUE
       On  success,  fread()  and  fwrite()  return the number of items read or written.  This number equals the number of bytes transferred only when size is 1.  If an
       error occurs, or the end of the file is reached, the return value is a short item count (or zero).

fread() does not distinguish between end-of-file and error, and callers must use feof(3) and ferror(3) to determine which occurred.

PR #1526, opened 2019-06-03
Fix test incrementalselftest.sh

DomiNic5787:


DomiNic5787:
@williamcroberts
When this test runs against a "fresh" TPM, where not all algorithms have been tested, the else statement is never covered.
#1529 should improve this

Issue #1532, opened 2019-06-04
incrementalselftest.sh fails due to missing CMAC algorithm support


diabonas:
With the current version 1332 of the [TPM simulator](https://sourceforge.net/projects/ibmswtpm2/), the reworked integration test [`incrementalselftest.sh`](https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/tests/incrementalselftest.sh) fails with
```
ERROR: Got invalid or unsupported algorithm: "unknown3f"
```
The reason for this is that this version of the simulator supports the CMAC algorithm, which `tpm2_getcap` doesn't know and reports as
```
$ tpm2_getcap -T mssim -c algorithms
[...]
unknown3f:
  value:      0x3F
  asymmetric: 0
  symmetric:  1
  hash:       0
  object:     0
  reserved:   0x0
  signing:    1
  encrypting: 0
  method:     0
[...]
```
Support for this algorithm needs to be added to [`tpm2_alg_util.c`](https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_alg_util.c) and possibly other places, and it needs to be removed from [`eccmethods` and `rsamethods`](https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/tests/incrementalselftest.sh#L55-L56) in `incrementalselftest.sh`.

Support for CMAC also requires upstream changes in tpm2-tss, see https://github.com/tpm2-software/tpm2-tss/issues/1436.

PR #1533, opened 2019-06-05
 tpm2_alg_util: add CMAC algorithm

diabonas:
Add CMAC to the list of known algorithms. This PR requires https://github.com/tpm2-software/tpm2-tss/pull/1437 to be merged first so that `TPM2_ALG_CMAC` is defined. The new algorithm is also accidentally included in `eccmethods` and `rsamethods` in `incrementalselftest.sh`, rework the conditions to avoid that.

These are the minimal changes to fix the test suite for the most recent TPM simulator version 1332, please advise if there are more places within tpm2-tools where the new algorithm needs to be added.

Fixes: #1532


diabonas:
This should be good to merge now, CI passes since https://github.com/tpm2-software/tpm2-tss/pull/1437 has been merged upstream.

Issue #1535, opened 2019-06-05
Tests might not fail since traps are not inherited to bash functions


joholl:
In `helpers.sh`, there is a bash trap that exits if a return code other than 0 appears:

```bash
#
# Set the default on ERR handler to print the line number
# and failed command.
#
onerror() {
    echo "$BASH_COMMAND on line ${BASH_LINENO[0]} failed: $?"
    exit 1
}
trap onerror ERR
```

However, this trap is **not** inherited to bash functions. See the [bash reference](https://tiswww.case.edu/php/chet/bash/bashref.html#Shell-Functions):

> All other aspects of the shell execution environment are identical between a function and its caller with these exceptions: [...] the ERR trap is not inherited unless the -o errtrace shell option has been enabled. See Bourne Shell Builtins, for the description of the trap builtin.

If the option `-o errtrace` (or `-E`) is passed, the trap will be inherited:

>    If set, any trap on ERR is inherited by shell functions, command substitutions,
>     and commands executed in a subshell environment. The ERR trap is normally not
>     inherited in such cases.

Due to this issue, there are quite some errors (bash commands not returning 0) which are currently missed. For instance, the test `import.sh` passed although the function `run_aes_import_test()` was broken. See #1146.

I suggest adding a `set -E` at the beginning of `helpers.sh`. Unfortunately, as a consequence, some tests will fail and require further work.

williamcroberts:
> I suggest adding a set -E at the beginning of helpers.sh. Unfortunately, as a consequence, some tests will fail and require further work.

That seems reasonable, why would some tests start failing though? Are they broken and passing?

joholl:
Yes, there are several tests failing, some of which are really broken. I'll keep you updated.

joholl:
I pushed the changes I made to a [feature branch](https://github.com/joholl/tpm2-tools/tree/fix_trap_inheritage). I'll eventually create a PR but I want to discuss some issues first.

- `loadexternal.sh` was completely broken and had many problems which are **fixed** now.
- `incrementalselftest.sh` fails at the moment due to an algorithm named `unknown3f` (printed by tpm2_incrementalselftest). This is probably fixed by [this TSS PR](https://github.com/tpm2-software/tpm2-tss/pull/1437)
- In `import.sh`, `tpm2_evictcontrol` is failing. I'm not sure why it is needed here. Any idea?
[...]

PR #1536, opened 2019-06-05
test: aes256 key trunctuated in import.sh

### LINKED ISSUE #1146: tpm2_import - truncating 256 AES keys
https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/tests/import.sh#L54

The test creates a 32 byte key, but truncates it in the OSSL calls and when setting the public attribute values.

joholl:
In the test `import.sh`, the decryption was broken due to two problems:

- `openssl enc` expects the ciphername as a parameter (either `-aes-128-cfb` or `-aes-256-cfb`). Previously we always passed `-aes-128-cfb` (which broke decryption for aes256)
- `xxd` is used to convert binary data into a hex string. However, `xxd` eventually breaks the line. To prevent this, `-c 256` is added to be able to convert to up to 256 chars without line break. Fixes #1146 

Also, a minor indentation issue is fixed.

Unfortunately, this test passed even if it was broken. This is due to the fact that bash traps are not inherited to bash functions. See #1535.

### DIFF
--- a/test/integration/tests/import.sh
+++ b/test/integration/tests/import.sh
@@ -23,7 +23,7 @@ start_up
 
 run_aes_import_test() {
 
-	dd if=/dev/urandom of=sym.key bs=1 count=$2 2>/dev/null
+	dd if=/dev/urandom of=sym.key bs=1 count=$3 2>/dev/null
 
 	#Symmetric Key Import Test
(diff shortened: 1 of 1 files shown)


williamcroberts:
> Unfortunately, this test passed even if it was broken. This is due to the fact that bash traps are not inherited to bash functions. See #1535.

Oh I did not know that, interesting.

PR #1542, opened 2019-06-07
Allow using env TPM2TOOLS_TCTI for tests

DomiNic5787:
Instead of setting `TPM2TOOLS_TCTI="device"` by default, allow using env `TPM2TOOLS_TCTI`.
In addition exit with error, if no `TPM2TOOLS_TCTI` is set.

### DIFF
--- a/test/integration/helpers.sh
+++ b/test/integration/helpers.sh
@@ -267,12 +267,13 @@ function start_up() {
 
         echo "Running tpm2_startup -c"
         tpm2_startup -c
+    elif [ -n "$TPM2_DEVICE" ]; then
+        export TPM2TOOLS_TCTI="device:$TPM2_DEVICE"
+    elif [ -z "$TPM2TOOLS_TCTI" ]; then
+        echo "No TPM2TOOLS_TCTI defined"
+        exit 1
     else
-        if [ -n "$TPM2_DEVICE" ]; then
-            export TPM2TOOLS_TCTI="device:$TPM2_DEVICE"
-        else
-            export TPM2TOOLS_TCTI="device"
-        fi
+        echo "Using TPM2TOOLS_TCTI $TPM2TOOLS_TCTI"
     fi
 
     echo "Running tpm2_clear"


test/integration/helpers.sh
```diff
@@ -267,12 +267,13 @@ function start_up() {
 
         echo "Running tpm2_startup -c"
         tpm2_startup -c
+    elif [ -n "$TPM2_DEVICE" ]; then
```
williamcroberts: I would just mark TPM2TOOLS_TCTI in the Makefile.am as exported in AM_TESTS_ENVIRONMENT and then in the script if TPM2TOOLS_TCTI is empty, default to abrmd. This way at Make someone can  just point to another TCTI if they so choose.

DomiNic5787: In my opinion TPM2TOOLS_TCTI should not be able to set directly during make process, only indirectly with TPM2_ABRMD, TPM2_SIM and TPM2_DEVICE. (service abrmd is missing!).
This allows maximum flexibility, for example if you want to reuse the test environment (helpers.sh) to write your own scripts.
Finally I am not sure, what is the best solution in case of neither TPM2_ABRMD, TPM2_SIM, TPM2_DEVICE nor TPM2TOOLS_TCTI is set:
1. error
2. set something specific, e.g. "device" or "abrmd" (not my favorite)
3. use find_default_tcti logic (see tpm2_options.c)

AndreasFuchsSIT:
@williamcroberts Could you look into this once more ?
Cause for me that would be nice to test some "future tools" (you know which) with physical TPMs but want to make sure I don't use the pTPMs by accident.

test/integration/helpers.sh
williamcroberts: Doesn't this work:
  - https://www.gnu.org/software/automake/manual/html_node/Scripts_002dbased-Testsuites.html

The AM_TESTS_ENVIRONMENT and TESTS_ENVIRONMENT variables can be used to run initialization code and set environment variables for the test scripts. The former variable is developer-reserved, and can be defined in the Makefile.am, while the latter is reserved for the user, which can employ it to extend or override the settings in the former; for this to work portably, however, the contents of a non-empty AM_TESTS_ENVIRONMENT must be terminated by a semicolon.

williamcroberts: > In my opinion TPM2TOOLS_TCTI should not be able to set directly during make process, only indirectly with TPM2_ABRMD, TPM2_SIM and TPM2_DEVICE. (service abrmd is missing!).
> This allows maximum flexibility, for example if you want to reuse the test environment (helpers.sh) to write your own scripts.

Running the test harness outside of make check, while feasible is not going to be the recommended or supported approach. Passing env variables through automake is the proper way, and if you're doing it out of the make file environment, its on the caller to set it up.

williamcroberts: Also their is no need to have different env variables then what the tools support already.

williamcroberts:
I think this is what you want:
https://github.com/tpm2-software/tpm2-tools/pull/1568

- it prevents accidentally running tests with make check against a TPM when you forget that TPM2TOOLS_TCTI is set

- it allows for:
  1. the simulator to be started externally
  2. abrmd to be started externally
  3. full control of tools tcti

The commit has tons of examples.

DomiNic5787:
I'm fine with #1568

Issue #1539, opened 2019-06-05
refactor tpm interface


williamcroberts:
The work on #1193 ended up with us making a tpm2.c file that contains all the ESYS calls and performs all the tool_rc (name may change) conversions. However, it's still a raw interface. If you look at things like tpm2_policy.c, there is a lot of code like this:
```C
tool_rc tpm2_policy_build_policylocality(ESYS_CONTEXT *ectx,
    tpm2_session *session, TPMA_LOCALITY locality) {

    ESYS_TR handle = tpm2_session_get_handle(session);

    return tpm2_policy_locality(ectx, handle,
                    ESYS_TR_NONE, ESYS_TR_NONE, ESYS_TR_NONE, locality);
}
```
With #1045, the auth and object models get collapsed, and thus a tpm object should contain the tpm2_session *. We can use this idiom to refactor the tpm interface to hide:

1. Only expose session handles required, perhaps not all 3.
2. For commands that take objects, pass the object and hide the session params.

Their are probably other refactorings that can be done to simplify and collapse tool code.

PR #1546, opened 2019-06-09
Refactor activate credential

idesai:
1. Decouples activatecredential  hardcoded auth method of using policysecret to satisfy TPM2_RH_Endorsement auth.

2. Better variable names to align with the spec language.

3. Progresses: #1539

### DIFF
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -861,6 +861,53 @@ tool_rc tpm2_tr_set_auth(
     return tool_rc_success;
 }
 
+tool_rc tpm2_activatecredential(
+        ESYS_CONTEXT *esysContext,
+        tpm2_loaded_object *activatehandleobj,
+        tpm2_loaded_object *keyhandleobj,
+        const TPM2B_ID_OBJECT *credentialBlob,
+        const TPM2B_ENCRYPTED_SECRET *secret,
+        TPM2B_DIGEST **certInfo) {
+
+    ESYS_TR keyobj_session_handle = ESYS_TR_NONE;
+    tool_rc rc = tpm2_auth_util_get_shandle(
+                        esysContext,
+                        keyhandleobj->tr_handle,
+                        keyhandleobj->session,
+                        &keyobj_session_handle); //shandle1
+    if (rc != tool_rc_success) {
+        return rc;
+    }
+
+    ESYS_TR activateobj_session_handle = ESYS_TR_NONE;
+    rc = tpm2_auth_util_get_shandle(
+                esysContext,
+                activatehandleobj->tr_handle,
+                activatehandleobj->session,
+                &activateobj_session_handle); //shandle2
+    if (rc != tool_rc_success) {
+        return rc;
+    }
+
+    TSS2_RC rval = Esys_ActivateCredential(
+                    esysContext,
+                    activatehandleobj->tr_handle,
+                    keyhandleobj->tr_handle,
+                    activateobj_session_handle,
+                    keyobj_session_handle,
+                    ESYS_TR_NONE,
+                    credentialBlob,
+                    secret,
+                    certInfo);
+    if (rval != TPM2_RC_SUCCESS) {
+        LOG_PERR(Esys_ActivateCredential, rval);
+        rc = tool_rc_from_tpm(rval);
+        return rc;
+    }
+
+    return tool_rc_success;
+}
+
 tool_rc tpm2_create(
         ESYS_CONTEXT *esysContext,
         tpm2_loaded_object *parent_obj,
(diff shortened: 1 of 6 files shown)


man/tpm2_activatecredential.1.md
```diff
@@ -13,47 +13,49 @@ key.
...
-    _KEY\_CONTEXT\_OBJECT_.
     Passwords should follow the "authorization formatting standards", see
     section "Authorization Formatting".
 
-  * **-E**, **\--auth-endorse**=_ENDORSE\_PASSWORD_:
+  * **-E**, **\--keyhandle-auth**=_AUTH\_VALUE_:
 
-    The endorsement authorization value, optional. Follows the same formatting
-    guidelines as the key authorization option **-P**.
+    Typically endorsement authorization value, optional. Follows the same
```
williamcroberts: Typically? When is it not?

tools/tpm2_activatecredential.c
```diff
@@ -24,30 +25,26 @@
...
-
     struct {
-        char *auth_str;
-        tpm2_session *session;
-        const char *ctx_arg;
-    } key;
+        const char *ctx_path;
+        const char *auth_str;
+        tpm2_loaded_object object;
+    } keyhandle; //Typically EK
```
williamcroberts: this more than just the keyhandle, isn't it the key itself?

tools/tpm2_activatecredential.c
```diff
@@ -24,30 +25,26 @@
...
 
     struct {
-        char *auth_str;
-        tpm2_session *session;
-    } endorse;
+        const char *ctx_path;
+        const char *auth_str;
+        tpm2_loaded_object object;
+    } activatehandle; //Typically AK
```
williamcroberts: again more than a handle, name it based on the object it represents not the item it represents, that's what the struct members are for.

lib/tpm2.c
```diff
@@ -861,6 +861,53 @@ tool_rc tpm2_tr_set_auth(
...
+tool_rc tpm2_activatecredential(
+        ESYS_CONTEXT *esysContext,
+        tpm2_loaded_object *activatehandleobj,
+        tpm2_loaded_object *keyhandleobj,
+        const TPM2B_ID_OBJECT *credentialBlob,
+        const TPM2B_ENCRYPTED_SECRET *secret,
+        TPM2B_DIGEST **certInfo) {
+
+    ESYS_TR keyobj_session_handle = ESYS_TR_NONE;
+    tool_rc tmp_rc = tpm2_auth_util_get_shandle(
```
williamcroberts: if you never have an rc, you don't need a tmp_rc as it's used elsewhere. When you have a tmp_rc it's usually because you only want to assign to rc on failure and goto a label that returns rc. In short, rename this to rc.

tools/tpm2_activatecredential.c
idesai: The idea was to rename the variables verbatim as in the spec. Yes it is the key itself. Should it instead be called "keyhandle_obj" and "activatehandle_obj"?

man/tpm2_activatecredential.1.md
```diff
@@ -13,47 +13,49 @@ key.
 
 # DESCRIPTION
 
-**tpm2_activatecredential**(1) -  Verify that the given content is protected
-with given key handle for given handle, and then decrypt and return the secret,
-if any password option is missing, assume NULL. Currently only support using
-TCG profile compliant EK as the key handle.
+**tpm2_activatecredential**(1) -  This command enables the association of a
+credential with an object in a way that ensures that the TPM has validated the
+parameters of the credentialed object. If any password option is missing, assume
```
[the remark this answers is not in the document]
idesai: Adding this to the description in the md
`In an attestation scheme , this guarantees the registrar that
the attestation key belongs to the TPM with a qualified parent key in the TPM`

man/tpm2_activatecredential.1.md
idesai: Reverted to previous definition

tools/tpm2_activatecredential.c
idesai: activatehandle_obj?

lib/tpm2.c
idesai: fixed

tools/tpm2_activatecredential.c
williamcroberts: If the spec isn't descriptive or doesn't add clarity, then I would stay away from following it.

idesai: SGTM and so then I will rename it to:
activatehandle --> credentialed_obj
keyhandle --> credential_key
I will update the options and man as well.

idesai:
@williamcroberts closing this one and merging the commits here with PR #1548.

PR #1548, opened 2019-06-11
Refactor tools auth interface

idesai:
1. Better variable names to align with the spec language.
2. Progresses: #1539

### DIFF
--- a/tools/tpm2_certify.c
+++ b/tools/tpm2_certify.c
@@ -26,13 +26,13 @@ struct tpm_certify_ctx {
         const char *ctx_path;
         const char *auth_str;
         tpm2_loaded_object object;
-    } object;
+    } certified_key;
 
     struct {
         const char *ctx_path;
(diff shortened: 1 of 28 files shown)


tools/tpm2_certify.c
```diff
@@ -227,13 +227,13 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
 
     /* Load input files */
     tool_rc rc = tpm2_util_object_load_auth(ectx, ctx.certified_key.ctx_path,
-        ctx.certified_key.auth_str, &ctx.certified_key.object);
+        ctx.certified_key.auth_str, &ctx.certified_key.object, false);
```
williamcroberts: why did you change this? You always pass false, so it's not really needed.

williamcroberts:
FYI you need to update the changelog if tool options change..

Issue #1416, opened 2019-04-15
./bootstrap includes files twice in $ALL_SYSTEM_TESTS


dilyanpalauzov:
To include tests in $SYSTEM_TESTS_TCTI_ABRMD ./bootstrap calls `find test/integration/tests/tcti/abrmd -name "*.sh"`.

To include tests in $SYSTEM_TESTS, ./bootstrap calls `find test/integration/tests/ -name "*.sh"`.

./bootstrap defines `ALL_SYSTEM_TESTS = $(SYSTEM_TESTS) $(SYSTEM_TESTS_TCTI_ABRMD)`.

The variable $SYSTEM_TESTS_TCTI_ABRMD isn’t used on any other place, except in the definition for ALL_SYSTEM_TESTS.

`find test/integration/tests/ -name "*.sh"` includes the files, returned by `find test/integration/tests/tcti/abrmd -name "*.sh"`, so the second call, and the variable $SYSTEM_TESTS_TCTI_ABRMD are superfluous.  Passing the non-POSIX `-maxdepth 1` to find on the first call, would have skipped any subdirectories of test/integration/tests/, leaving $SYSTEM_TESTS_TCTI_ABRMD and $SYSTEM_TESTS disjoint sets.

There is also no point to have separate $LIB_C and $LIB_H, that are joined in $LIB_SRC and are otherwise ignored.  A `src_listvar "lib" "*.[ch] "LIB_SRC"` would have done the same job.

I recommend to get rid of ./bootstrap and src_vars.mk, expect that users call `autoreconf` and list all source, headers and test files explictly in Makefile.am.

williamcroberts:
> I recommend to get rid of ./bootstrap and src_vars.mk, expect that users call `autoreconf` and list all source, headers and test files explictly in Makefile.am.

No n the bootstrap yes on individual listings in Makefile.am, that doesn't bother me.

PR #1549, opened 2019-06-12
test: no longer segregate tests by tcti

williamcroberts:
tpm2-tools currently has no plans for official support running the test framework
against a physical TPM and will require abrmd, thus just move the tests
from the tcti abrmd subdirectory into the main tests directory. This
will cause the double inclusion of tcti test scripts to go away.

Fixes: #1416


DomiNic5787:
`bootstrap` needs to be updated as well.

see [https://travis-ci.org/tpm2-software/tpm2-tools/jobs/544634031#L1306](https://travis-ci.org/tpm2-software/tpm2-tools/jobs/544634031#L1306)

PR #1551, opened 2019-06-13
tools: fix dereferencing of null pointer in tpm2_encryptdecrypt

joholl:
After 75b60808, `iv_out` is dereferenced before the error check of `Esys_EncryptDecrypt2()` or `Esys_EncryptDecrypt()`. This lead to a Segmentation Fault if `Esys_EncryptDecrypt()` failed. Consequently, the error check has to be moved up again.


PeterHuewe:
@idesai  can you merge this please?

PR #1554, opened 2019-06-16
Refactor tools auth interface

idesai:
1. Better variable names to align with the spec language.
2. Progresses: #1539

### DIFF
--- a/lib/object.c
+++ b/lib/object.c
@@ -30,11 +30,15 @@ static tool_rc tpm2_util_object_load2(
             const char *objectstr,
             const char *auth,
             bool do_auth,
-            tpm2_loaded_object *outobject) {
+            tpm2_loaded_object *outobject,
+            bool is_restricted_pswd_session) {
+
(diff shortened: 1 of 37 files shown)


lib/object.c
```diff
@@ -30,19 +30,23 @@ static tool_rc tpm2_util_object_load2(
             const char *objectstr,
             const char *auth,
             bool do_auth,
-            tpm2_loaded_object *outobject) {
+            tpm2_loaded_object *outobject,
+            bool is_restricted_pswd_session) {
```
williamcroberts: isn't this TRUE if esys context is NULL? So can't we just determine based on ESYS context?

idesai: Yes and no. Because esys context is never passed as NULL as it is required in do_ctx_file  OR tpm2_util_sys_handle_to_esys_handle  routines, which ever path it takes.

idesai:
> For the most part this looks reasonable. Any public changing interface (ie tool options) need a CHANGELOG.md entry.

Resolved

PR #1560, opened 2019-06-19
Fix for tools printing auth values to stderr on failed auth attempts

idesai:

### DIFF
--- a/tools/tpm2_encryptdecrypt.c
+++ b/tools/tpm2_encryptdecrypt.c
@@ -326,8 +326,7 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
     tool_rc rc = tpm2_util_object_load_auth(ectx, ctx.encryption_key.ctx_path,
         ctx.encryption_key.auth_str, &ctx.encryption_key.object, false);
     if (rc != tool_rc_success) {
-        LOG_ERR("Invalid object key authorization, got\"%s\"",
-            ctx.encryption_key.auth_str);
+        LOG_ERR("Invalid object key authorization");
         return rc;
     }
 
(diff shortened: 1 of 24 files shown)


williamcroberts:
Ahh codecov is failing you because your code doesn't increase code coverage metrics.
See:
  - https://docs.codecov.io/docs/commit-status

Your patch doesn't negatively impact code coverage, so this is very pedantic. I turned it down in:
  - https://github.com/tpm2-software/tpm2-tools/pull/1563

williamcroberts:
I don't know what the difference is between codecov/patch and codecov/project though

tools/tpm2_encryptdecrypt.c
```diff
@@ -326,8 +326,7 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
     tool_rc rc = tpm2_util_object_load_auth(ectx, ctx.encryption_key.ctx_path,
         ctx.encryption_key.auth_str, &ctx.encryption_key.object, false);
     if (rc != tool_rc_success) {
-        LOG_ERR("Invalid object key authorization, got\"%s\"",
-            ctx.encryption_key.auth_str);
+        LOG_ERR("Invalid object key authorization");
```
williamcroberts: we can probably just delete all these LOG lines, since the call to tpm2_util_object_load_auth() should just LOG all the errors. I would just delete them all, and make sure the routine logs it. This will remove duplicate code from the tools and ensure error messages stay consistent.

williamcroberts: Applies here and throughout the tools code.

williamcroberts: I guess though for things that take multiple objects the exact object that got the wrong password isn't known. N/M

Issue #1462, opened 2019-05-07
tpm2_util_object_load: use consistently


williamcroberts:
Their are a llot of ways in which objects can be selected, for instance "o" for owner hierarchy. Modify tpm2_util_object_load() and use consistently to create a tpm context for all objects and use flags to control what is considered valid input. Use this in all the tools to provide a consistent object loading model from various sources.

williamcroberts:
Probably also allow NV indexes like 1, 2, 3, etc rather than raw handles as well. Flags would likely be around:
* nv indexes - numeric
* persistent - ESYS_TR or numeric
* transient - context files only
* hierarchies: (n|null, o|owner, p|platform, e|endorsement)

PR #1564, opened 2019-06-20
tools: Add back the hierarchy filtering to tools

idesai:
Progresses #1462 

Tools that operate only with certain hierarchies could explicitly
specify it either with a string or a hex handle of the hierarchy.
This is required in tools like tpm2_clear tpm2_clearcontrol
tpm2_dictionarylockout tpm2_evictcontrol tpm2_nvdefine
tpm2_nvrelease tpm2_pcrallocate.

### DIFF
--- a/lib/object.c
+++ b/lib/object.c
@@ -9,7 +9,6 @@
 #include "log.h"
 #include "object.h"
 #include "tpm2_auth_util.h"
-#include "tpm2_hierarchy.h"
 
 #define NULL_OBJECT "null"
 #define NULL_OBJECT_LEN (sizeof(NULL_OBJECT) - 1)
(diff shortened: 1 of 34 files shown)


lib/tpm2_hierarchy.h
```diff
@@ -23,6 +23,8 @@ enum tpm2_hierarchy_flags {
     TPM2_HIERARCHY_FLAGS_ALL  = 0x1F
 };
 
+#define TPM2_HANDLES_ALL TPM2_HIERARCHY_FLAGS_ALL|TPM2_HIERARCHY_SUPPRESS
```
williamcroberts: put this in the enum

lib/object.c
```diff
@@ -9,7 +9,6 @@
 #include "log.h"
 #include "object.h"
 #include "tpm2_auth_util.h"
-#include "tpm2_hierarchy.h"
```
williamcroberts: dont you need this?

idesai: Moved it to object.h. Required when defining tpm2_util_object_load_auth.

PR #1565, opened 2019-06-21
option tests

williamcroberts:
Test some of the options and env parameters.


williamcroberts:
--disable-dlclose config option causes break, need to set will_returns for dlclose() based on this

Issue #1547, opened 2019-06-10
PKCS7 padding in encryptdecrypt tool skips padding when input aligns at block length and strip padding on decryption


idesai:
PKCS7 also requires at least ONE padding byte; so if your input length is exactly modulus block length - you need to add block length number of bytes, containing the value block length.

idesai:
- Fix for adding padding to input with exactly modulus block length
- Fix for stripping padding when decrypting data

PR #1567, opened 2019-06-21
tpm2_encryptdecrypt: Update encryptdecrypt tool and man page

idesai:
Progresses #1547
1. Rename cmdline option to pkcs7-pad
2. Separate out a function for pkcs7 padding
3. Use memset instead of for loops to allow compiler to optimize padding
4. Fix: Padding is necessary also for data that is aligned at block length
5. Add tests for checking padding on data that is aligned at block length 
6. Fix: Strip padding data when decrypting
7. Add tests for checking pad data is stripped on input data with aligned block length
8. Add tests for checking pad data is stripped on input data with unaligned block length

### DIFF
--- a/man/tpm2_encryptdecrypt.1.md
+++ b/man/tpm2_encryptdecrypt.1.md
@@ -32,7 +32,7 @@ specified symmetric key.
 
     Perform a decrypt operation. Default is encryption.
 
-  * **-e**, **\--enable-pkcs7-padding**:
+  * **-e**, **\--pad**:
 
     Enable pkcs7 padding for applicable AES encryption modes cfb/cbc/ecb.
     Applicable only to encryption and for input data with last block shorter
--- a/tools/tpm2_encryptdecrypt.c
+++ b/tools/tpm2_encryptdecrypt.c
@@ -31,13 +31,16 @@ struct tpm_encrypt_decrypt_ctx {
     } encryption_key;
 
     TPMI_YES_NO is_decrypt;
-    TPM2B_MAX_BUFFER data;
+
     uint8_t input_data[UINT16_MAX];
     uint16_t input_data_size;
     char *input_path;
+
     char *out_file_path;
+
     uint8_t padded_block_len;
     bool is_padding_option_enabled;
+
     TPMI_ALG_SYM_MODE mode;
     struct {
         char *in;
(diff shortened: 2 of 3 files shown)


tools/tpm2_encryptdecrypt.c
```diff
@@ -59,6 +59,23 @@ static tool_rc readpub(ESYS_CONTEXT *ectx, TPM2B_PUBLIC **public) {
                       public, NULL, NULL);
 }
 
+static void append_pkcs7_padding_data_to_input(uint8_t *pad_data) {
```
williamcroberts: you also need to strip on decrypt operation if -e is specified.

tools/tpm2_encryptdecrypt.c
```diff
@@ -59,6 +59,23 @@ static tool_rc readpub(ESYS_CONTEXT *ectx, TPM2B_PUBLIC **public) {
...
+
+    if (! (ctx.is_padding_option_enabled && !ctx.is_decrypt &&
+          ctx.data.size < TPM2_MAX_DIGEST_BUFFER &&
```
williamcroberts: I don't think this matters:
```C
ctx.data.size < TPM2_MAX_DIGEST_BUFFER
```
Isn't it always less than that?

tools/tpm2_encryptdecrypt.c
```diff
@@ -91,29 +108,15 @@ static tool_rc encrypt_decrypt(ESYS_CONTEXT *ectx, TPM2B_IV *iv_start) {
...
-                ctx.padded_block_len -(ctx.data.size % ctx.padded_block_len);
-
-            for(i = 0; i < pad_data[0]; i++) {
-                ctx.data.buffer[ctx.data.size + i] = pad_data[0];
-            }
-
-            ctx.data.size += pad_data[0];
-        }
+        uint8_t pad_data = 0;
+        append_pkcs7_padding_data_to_input(&pad_data);
```
williamcroberts: why not just update the buffer size:
    void pkcs7_add_pad(TPM2B_MAX_DIGEST_BUFFER *buf) {
        
        uint8_t pad = buffer->size % ctx.padded_block_len;
        if (!pad) {
            return;
        }
       
        memset(&buffer->buffer[buffer->size], pad, pad);
        buffer->size += pad;
    }
```C
if ( mode is encrypt) {
    pkcs7_add_pad(&ctx.data)
}
    


```

williamcroberts: The data buffer probably shouldn't be in global ctx anymore since were reading it block by block from a file source.

man/tpm2_encryptdecrypt.1.md
```diff
@@ -32,7 +32,7 @@ specified symmetric key.
 
     Perform a decrypt operation. Default is encryption.
 
-  * **-e**, **\--enable-pkcs7-padding**:
+  * **-e**, **\--pkcs7-pad**:
```
williamcroberts: `-e or --pad`

On the off-chance we need to do backwards compatible pad additions we can just do --pad ( defaults to pkcs7) --pad="new pad scheme"

tools/tpm2_encryptdecrypt.c
idesai: Yep. That is already added to the to-do list #1547

idesai: Done

Issue #1371, opened 2019-03-23
No warning/error on default suite being selected if specified suite is ill formated


lestums:
If one attempt to specify such algorithm permutation :
`ecc256<insert_something_here>:....` e.g `ecc256funnytexthere:`
the tool will silently convert the suite into :
`ecc256:null:null`

Same behavior can be obtained by using RSA or any algorithm specification that have data between identifier and the colon (or NULL) and where beginning of the spec. is correct (all permutations of ecc/rsa with a valid curve/key size)
On top of that, algorithms defaults changes between restricted and non restricted modes.
`ecc256:null:null` in non restricted
`ecc256:null:aes128cfb` in restricted

Multiple ways to address the issue :
- Modify manpage to specify default set selected if input is invalid (error-prone and quite abrupt from user's point of view)
- Print a warning & modify manpages to state that user will be warned on invalid algorithm suites (user still need to be careful)
- Throw an error (what i prefer)

What do you suggest ?

williamcroberts:
Throw an error

PR #1572, opened 2019-06-25
tpm2_alg_util: fix extended specifier errors

williamcroberts:
If one attempts to specify such algorithm permutation:
  - ecc256<insert_something_here>
For example:
  -  ecc256funnytexthere

The tool will silently convert the algorithm specifier into ecc256:null:null.

The same behavior can be obtained by using RSA or any algorithm specification
that has data between identifier and the colon (or NULL) and where beginning
of the spec. is correct (all permutations of ecc/rsa with a valid curve/key size)
On top of that, algorithms defaults changes between restricted and non restricted modes.
ecc256:null:null in non restricted
ecc256:null:aes128cfb in restricted

Correct this by issuing errors when invalid text strings are presented.

Fixes: #1371

### DIFF
--- a/lib/tpm2_alg_util.c
+++ b/lib/tpm2_alg_util.c
@@ -117,20 +117,18 @@ static void tpm2_alg_util_for_each_alg(alg_iter iterator, void *userdata) {
     }
 }
 
-static alg_parser_rc handle_aes_raw(const char *ext, TPMT_SYM_DEF_OBJECT *s) {
+static alg_parser_rc handle_sym_common(const char *ext, TPMT_SYM_DEF_OBJECT *s) {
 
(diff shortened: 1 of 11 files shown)


williamcroberts:
```sh
for i in "rsa" "xor" "hmac" "ecc" "keyedhash"; do
    tpm2_testparms "${i}"
done

# Test that RSA signing schemes are supported
for i in ${rsamethods}; do
    tpm2_testparms "rsa:${i}"
done

# Test that ECC signing schemes are supported
for i in ${eccmethods}; do
    tpm2_testparms "ecc:${i}"
done
```
This code breaks, internally test params is calling the extended algorithm code, and things like "rsa:rsapps" break.

PR #1574, opened 2019-06-26
tests/*.sh: Fixes a potential issue If bash is not in /bin, the tests will fail.

### LINKED ISSUE #1467: Tests and bash
./configure calls `AC_CHECK_PROG([BASH_SHELL],[bash],[yes])` and verifies, that bash is on $PATH.  But the tests/integration/tests/*.sh files all start with `#!/bin/bash`.

If `bash` is in /usr/bin, but not in /bin, ./configure will not warn, that the tests will fail.

idesai:
Fixes #1467

Note: Any variable set in the environment when GNU Make is started will be
available inside the Makefile automatically. And so simply specifying bash should be sufficient.


Makefile.am
```diff
@@ -260,7 +260,7 @@ AM_TESTS_ENVIRONMENT =	\
 	TPM2_TOOLS_TEST_FIXTURES=$(abs_top_srcdir)/test/integration/fixtures; \
 	export TPM2_TOOLS_TEST_FIXTURES;
 
-SH_LOG_COMPILER = dbus-run-session
+SH_LOG_COMPILER = dbus-run-session bash
```
idesai: Bash could also be specified as $(SHELL)

williamcroberts:
Good commit message, that note is key to understanding this change.

PR #1578, opened 2019-06-28
Interactive password

### LINKED ISSUE #1194: Interactive password prompt
It would be helpful to have the option of entering passwords interactively when needed. This would

  1. Prevent cleartext passwords form being exposed in command line arguments
  2. Ease the use of tpm2-tools in interactive scripts

Perhaps the option to prompt could be built into the minilanguage? Or the tool could detect when a password is required and prompt if one is not supplied?

idesai:
Fixes #1194 
1. Adds method to specify password from stdin either on a prompt or through process substitution.
2. Adds method to hide password characters when reading from stdin prompt.
3. Update authorization manual to add information on stdin password input mechanism.

### DIFF
--- a/lib/tpm2_auth_util.c
+++ b/lib/tpm2_auth_util.c
@@ -5,6 +5,8 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
+#include <termios.h>
+#include <unistd.h>
 
 #include <tss2/tss2_esys.h>
 #include <tss2/tss2_mu.h>
@@ -249,12 +251,55 @@ out:
     return rc;
 }
 
+static tool_rc console_display_control(bool display) {
+
+    struct termios console;
+    if (tcgetattr(STDIN_FILENO, &console)) {
+        return tool_rc_general_error;
+    }
+
+    if (display) {
+        console.c_lflag |= ECHO;
+    }
+
+    if (!display) {
+        console.c_lflag &= ~((tcflag_t) ECHO);
+    }
+
+    if (tcsetattr(STDIN_FILENO, TCSANOW, &console)) {
+        return tool_rc_general_error;
+    }
+
+    return tool_rc_success;
+}
+
 static tool_rc handle_stdin_password(ESYS_CONTEXT *ectx, tpm2_session **session) {
 
-    tool_rc rc = handle_file(ectx, "file:-", session);
+    bool process_substituion = false;
+    tool_rc rc = console_display_control(false);
+    if (rc != tool_rc_success) {
+       /*
+        * This means password was specified with process substitution method
+        * Example: tpm2_create -p - <<< $testpass
+        */
+        process_substituion = true;
+    }
+
+    if (!process_substituion) {
+        /*
+         * Drop to a password prompt to read password from stdin
+         */
+        printf("Scanning stdin for password:");
+    }
+
+    rc = handle_file(ectx, "file:-", session);
     if (rc != tool_rc_success) {
         LOG_ERR("stdin password error.");
         return rc;
+    }
+
+    if (!process_substituion) {
+        return console_display_control(true);
     }
 
     return tool_rc_success;
--- a/lib/tpm2_auth_util.c
+++ b/lib/tpm2_auth_util.c
@@ -249,6 +249,17 @@ out:
     return rc;
 }
 
+static tool_rc handle_stdin_password(ESYS_CONTEXT *ectx, tpm2_session **session) {
+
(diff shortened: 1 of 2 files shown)


lib/tpm2_auth_util.c
```diff
@@ -272,6 +329,12 @@ tool_rc tpm2_auth_util_from_optarg(ESYS_CONTEXT *ectx, const char *password,
         return handle_file(ectx, password, session);
     }
 
+    /* Is stdin"-" */
+    bool is_stdin = ( (!strncmp(password, "-", 1)) && (strlen(password) == 1) );
```
williamcroberts: Wow... +1 for confusing me, this is just a call to strcmp

man/common/authorizations.md
```diff
@@ -9,21 +9,38 @@ Authorization for use of an object in TPM2.0 can come in 3 different forms:
...
+  * str:<password> Used to indicate it is a raw string. Useful for escaping a
+                   password that starts with the "hex:" prefix.
+  * hex:<password> Used when specifying a password in hex string format.
+  * file:<file>    Used when specifying a password stored in a file. Useful to
+                   prevent leaking the password to UNIX utilities (such as ps).
+  * "-"            Used when specifying a password from stdin. There are 2 ways
+                   to specify the password. A prompt and by the process
+                   substitution method. Note that this is is similar to using
+                   file:-, however this option additionally hides password
+                   characters being echoed back to screen when using prompt.
```
williamcroberts: Huh, I don't understand, if we just fread(stdin, ...), who would be echoing them back, I don't understand why we need termios?

lib/tpm2_auth_util.c
```diff
@@ -5,6 +5,8 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
```
williamcroberts: this block is for std headers, after this we would be system specific libraries (like unistd.h) followed by 3rd party headers followed by internal header files.

Why do you need termios? What benefits does it give us over fread(stdin)?

williamcroberts: Oh I see what you mean:
https://stackoverflow.com/questions/6856635/hide-password-input-on-terminal

man/common/authorizations.md
williamcroberts: Can't you just use getpass from gnulibc?

```
man getpass
NAME
       getpass - get a password

SYNOPSIS
       #include <unistd.h>

       char *getpass(const char *prompt);
```

williamcroberts:
```sh
../../lib/tpm2_auth_util.c: In function 'handle_stdin_password':
../../lib/tpm2_auth_util.c:276:70: error: unused parameter 'password' [-Werror=unused-parameter]
 static tool_rc handle_stdin_password(ESYS_CONTEXT *ectx, const char *password,
```

man/common/authorizations.md
idesai: > Huh, I don't understand, if we just fread(stdin, ...), who would be echoing them back, I don't understand why we need termios?

This is the case where the user is keying in the pswd and auth_util reads it as stdin. In this case user key presses are echoed back.

idesai: > Can't you just use getpass from gnulibc?
> 
> ```
> man getpass
> NAME
>        getpass - get a password
> 
> SYNOPSIS
>        #include <unistd.h>
> 
>        char *getpass(const char *prompt);
> ```

Yes we could, however we will have to add something other than '-' to the auth language. Right now we are telling apart stdin from prompt using termios instead of getpass

williamcroberts: File:- use studio, just - use getpass

williamcroberts:
Something like this seems to be good:
```C
  bool x = isatty(fileno(stdin));
  if (x) {
    char *pass = getpass("enter your password:\n");
    printf("getpass: %s\n", pass);
  } else {
    char buf[1024];
    ssize_t bytes = fread(buf, sizeof(buf), 1, stdin);
    printf("fread %.*s\n", (int)sizeof(buf), buf);
  }
```

It ties it us to GNUC which we kind of already have a soft dependency on, but it avoids using glibc's termios which I think will be less portable. Windows seems to have: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-isatty?view=vs-2019

williamcroberts:
I'll take this over and push an update.

man/common/authorizations.md
williamcroberts: Oh I am an idiot, getpass is deprecated and they want us to go to termios per the manpage...

williamcroberts:
moved to #1589

PR #1580, opened 2019-07-01
tools/tpm2_hierarchycontrol: add new tool and man page

remuswu1019:
tools/tpm2_hierarchycontrol: add tool and its man page
lib: add supporting function for newly added tools

### DIFF
--- a/man/tpm2_hierarchycontrol.1.md
+++ b/man/tpm2_hierarchycontrol.1.md
@@ -1,0 +1,76 @@
+% tpm2_hierarchycontrol(1) tpm2-tools | General Commands Manual
+%
+% July 2019
+
+# NAME
+
+**tpm2_hierarchycontrol**(1) - Enable and disable use of a hierarchy and its
+associated NV storage.
+
+# SYNOPSIS
+
+**tpm2_hierarchycontrol** [*OPTIONS*] _VARIABLE_ _OPERATION_
+
+# DESCRIPTION
+
+**tpm2_hierarchycontrol**(1) - Allows user change phEnable, phEnableNV, shEnable
+and ehEnable when the proper authorization is provided. Authorization should be
+one out of owner hierarchy auth, endorsement hierarchy auth and platform
+hierarchy auth. **As an argument the tool takes the _VARIABLE_ as
+_TPMA\_STARTUP_CLEAR_ bit and \_OPERATION\_ as string clear|set to clear or set
+the _VARIABLE_ bit.**
+
+Note: If password option is missing, assume NULL.
+
+# OPTIONS
+
+  * **-C**, **\--hierarchy**=_AUTH\_HANDLE_:
+
+    Specifies the handle used to authorize. Defaults to the "platform" hierarchy.
+    Supported options are:
+      * **o** for **TPM_RH_OWNER**
+      * **p** for **TPM_RH_PLATFORM**
+      * **`<num>`** where a raw number can be used.
+
+  * **-P**, **\--hierarchy-auth**=_HIERARCHY\_\_AUTH\_VALUE_:
+
+    Specifies the authorization value for the hierarchy. Authorization values
+    should follow the "authorization formatting standards", see section
+    "Authorization Formatting".
+
+[common options](common/options.md)
+
+[common tcti options](common/tcti.md)
+
+[authorization formatting](common/authorizations.md)
+
+# EXAMPLES
+
+## Set phEnableNV with platform hierarchy and its authorization
+```
+tpm2_hierarchycontrol -C p phEnableNV set -P pass
+```
+
+## clear phEnableNV with platform hierarchy
+```
+tpm2_hierarchycontrol -C p phEnableNV clear
+```
+
+## Set shEnable with platform hierarchy
+```
+tpm2_hierarchycontrol -C p shEnable set
+```
+
+## Set shEnable with owner hierarchy
+```
+tpm2_hierarchycontrol -C o shEnable set
+```
+
+## Check current TPMA_STARTUP_CLEAR Bits
+```
+tpm2_getcap properties-variable
+```
+
+[returns](common/returns.md)
+
+[footer](common/footer.md)
--- a/test/integration/tests/hierarchycontrol.sh
+++ b/test/integration/tests/hierarchycontrol.sh
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+source helpers.sh
+
+cleanup() {
+
+  if [ "$1" != "no-shut-down" ]; then
+      shut_down
+  fi
+}
+trap cleanup EXIT
+
+start_up
+
+cleanup "no-shut-down"
+
+# Storage hierarchy
+tpm2_hierarchycontrol -C p shEnable set
+tpm2_hierarchycontrol -C p shEnable clear
+tpm2_hierarchycontrol -C p shEnable set
+tpm2_hierarchycontrol -C o shEnable clear
+
+# Endorsement hierarchy
+tpm2_hierarchycontrol -C p ehEnable set
+tpm2_hierarchycontrol -C p ehEnable clear
+tpm2_hierarchycontrol -C p ehEnable set
+tpm2_hierarchycontrol -C e ehEnable clear
+
+# Platform NV
+tpm2_hierarchycontrol -C p phEnableNV set
+tpm2_hierarchycontrol -C p phEnableNV clear
+tpm2_hierarchycontrol -C p phEnableNV set
+
+# Platform hierarchy
+tpm2_hierarchycontrol -C p phEnable clear
+
+# 0 the handler
+trap - ERR
+
+# ERROR: phEnable may not be 1 using this command
+tpm2_hierarchycontrol -C p phEnable set
+
+# EROOR: Only platform hierarchy handle can be specified for 1
+tpm2_hierarchycontrol -C o shEnable set
+tpm2_hierarchycontrol -C o ehEnable set
+tpm2_hierarchycontrol -C o phEnable set
+tpm2_hierarchycontrol -C o phEnableNV set
+tpm2_hierarchycontrol -C e shEnable set
+tpm2_hierarchycontrol -C e ehEnable set
+tpm2_hierarchycontrol -C e phEnable set
+tpm2_hierarchycontrol -C e phEnableNV set
+
+# ERROR: Permanent handle lockout not supported by this command
+tpm2_hierarchycontrol -C l shEnable set
+tpm2_hierarchycontrol -C l ehEnable set
+tpm2_hierarchycontrol -C l phEnable set
+tpm2_hierarchycontrol -C l phEnableNV set
+tpm2_hierarchycontrol -C l shEnable clear
+tpm2_hierarchycontrol -C l ehEnable clear
+tpm2_hierarchycontrol -C l phEnable clear
+tpm2_hierarchycontrol -C l phEnableNV clear
+
+# ERROR: Only platform and its authorization can be specified for 0
+tpm2_hierarchycontrol -C o ehEnable clear
+tpm2_hierarchycontrol -C o phEnable clear
+tpm2_hierarchycontrol -C o phEnableNV clear
+tpm2_hierarchycontrol -C e shEnable clear
+tpm2_hierarchycontrol -C e phEnable clear
+tpm2_hierarchycontrol -C e phEnableNV clear
+
+exit 0
--- a/tools/tpm2_hierarchycontrol.c
+++ b/tools/tpm2_hierarchycontrol.c
@@ -1,0 +1,170 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+#include <string.h>
+
+#include <log.h>
+#include <tpm2.h>
+#include "tpm2_tool.h"
+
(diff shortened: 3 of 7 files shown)


williamcroberts:
you need a test

man/tpm2_hierarchycontrol.1.md
```diff
@@ -0,0 +1,77 @@
...
+  * **-C**, **\--context-parent**=_PARENT\_CONTEXT\_OBJECT_:
+
+    Context object loaded object's parent. Either a file or a handle number.
+    See section "Context Object Format".
+
+  * **-P**, **\--auth-parent**=_PARENT\_KEY\_AUTH_:
+
+    Optional authorization value to use the parent object specified by **-C**.
+    Authorization values should follow the "authorization formatting standards",
+    see section "Authorization Formatting".
```
williamcroberts: There is no parent object for this command, so it's just -c and -p and its the hierarchy auth itself

remuswu1019: okay, could you suggest the short option for 'property'?

test/integration/tests/hierarchycontrol.sh
```diff
@@ -0,0 +1,66 @@
...
+tpm2_hierarchycontrol -c l -a phEnable clear 2>&1 1>/dev/null | grep "ERROR"
+tpm2_hierarchycontrol -c l -a phEnableNV clear 2>&1 1>/dev/null | grep "ERROR"
+
+# ERROR: Only platform and its authorization can be specified for CLEAR
+tpm2_hierarchycontrol -c o -a ehEnable clear 2>&1 1>/dev/null | grep "ERROR"
+tpm2_hierarchycontrol -c o -a phEnable clear 2>&1 1>/dev/null | grep "ERROR"
+tpm2_hierarchycontrol -c o -a phEnableNV clear 2>&1 1>/dev/null | grep "ERROR"
+tpm2_hierarchycontrol -c e -a shEnable clear 2>&1 1>/dev/null | grep "ERROR"
+tpm2_hierarchycontrol -c e -a phEnable clear 2>&1 1>/dev/null | grep "ERROR"
+tpm2_hierarchycontrol -c e -a phEnableNV clear 2>&1 1>/dev/null | grep "ERROR"
```
williamcroberts: shouldn't these error cases return non-zero to the shell and cause the error trap to run? I don't see the trap handler being cleared for negative testing.

tools/tpm2_hierarchycontrol.c
```diff
@@ -0,0 +1,174 @@
...
+
+    return true;
+}
+
+bool tpm2_tool_onstart(tpm2_options **opts) {
+
+    const struct option topts[] = {
+        { "context",    required_argument, NULL, 'c' },
+        { "auth",       required_argument, NULL, 'p' },
+        { "attribute",  required_argument, NULL, 'a' },
```
williamcroberts: this is called the "enable" in the spec. Attribute is used for object attributes and I don't want that common usage to confuse people here. How about we do -e|--enable

test/integration/tests/hierarchycontrol.sh
williamcroberts: Oh it's probably because grep is succeeding, just do (off top of my head so likely close but wrong):

```sh
# functionality tests here
trap - err 
# negative tests here
```

williamcroberts:
@remuswu1019 do yo have plans to make the last few changes?

tools/tpm2_hierarchycontrol.c
remuswu1019: I have fix this.

joholl:
`toggle-options.sh` fails with `/workspace/tpm2-tools/tools/tpm2_hierarchycontrol.c: missing option -c/--context in /workspace/tpm2-tools/man/tpm2_hierarchycontrol.1.md`

That's because in the code, the option for the context is `-c/--context`, whereas in the `man` page it is `-c/--context-object`.

[Code](https://github.com/tpm2-software/tpm2-tools/commit/5c4c6a4e100f8960739f98cc9a5cb09487301f6f#diff-0793cd8aec37ed09207d3a10b0bc106fR91):
```c
    const struct option topts[] = {
        { "context",    required_argument, NULL, 'c' },
```

[Man page](https://github.com/tpm2-software/tpm2-tools/commit/5c4c6a4e100f8960739f98cc9a5cb09487301f6f#diff-24631e49d3464e11efad35d2367fb3dfR25):
```
  * **-c**, **\--context-object**=_CONTEXT\_OBJECT_:

    Context object for the loaded object. Either a file or a handle number.
    See section "Context Object Format".
```

remuswu1019:
> `toggle-options.sh` fails with `/workspace/tpm2-tools/tools/tpm2_hierarchycontrol.c: missing option -c/--context in /workspace/tpm2-tools/man/tpm2_hierarchycontrol.1.md`
> 
> That's because in the code, the option for the context is `-c/--context`, whereas in the `man` page it is `-c/--context-object`.
> 
> [Code](https://github.com/tpm2-software/tpm2-tools/commit/5c4c6a4e100f8960739f98cc9a5cb09487301f6f#diff-0793cd8aec37ed09207d3a10b0bc106fR91):
> 
> ```c
>     const struct option topts[] = {
>         { "context",    required_argument, NULL, 'c' },
> ```
> 
> [Man page](https://github.com/tpm2-software/tpm2-tools/commit/5c4c6a4e100f8960739f98cc9a5cb09487301f6f#diff-24631e49d3464e11efad35d2367fb3dfR25):
> 
> ```
>   * **-c**, **\--context-object**=_CONTEXT\_OBJECT_:
> 
>     Context object for the loaded object. Either a file or a handle number.
>     See section "Context Object Format".
> ```

Thanks for the reminder.

test/integration/tests/hierarchycontrol.sh
remuswu1019: I have modified test method.
Would you help to check it again?

man/tpm2_hierarchycontrol.1.md
williamcroberts: Options need to change:
  - -p should be -P and --auth-key should be --auth-hierarchy.
  - -c should be -C and --context-object should be --hiearchy.

test/integration/tests/hierarchycontrol.sh
```diff
@@ -0,0 +1,76 @@
...
+trap cleanup EXIT
+
+start_up
+
+cleanup "no-shut-down"
+
+tpm2_clear
+
+# Clear the handler
+trap - ERR
```
williamcroberts: reverse the order of the tests so you just do normal testing up front, then clear the trap handler, then do negative testing.

man/tpm2_hierarchycontrol.1.md
```diff
@@ -0,0 +1,77 @@
...
+    Context object for the loaded object. Either a file or a handle number.
+    See section "Context Object Format".
+
+  * **-p**, **\--auth-key**=_KEY\_AUTH_:
+
+    Optional authorization value to use the key specified by **-c**.
+    Authorization values should follow the "authorization formatting standards",
+    see section "Authorization Formatting".
+
+  * **-e**, **\--enable**=_TPMA\_STARTUP\_CLEAR\_BITS:
```
williamcroberts: kill this option and make it argument.
So
```
tpm2_hierarchycontrol -c e -e phEnable set
```
becomes:
```sh
tpm2_hierarchycontrol -c e phEnable set
```

williamcroberts: I would also accept:
```
phEnable=set|1
phEnable=clear|0
```
Sample Code:
```C
#include <string.h>

int main(int argc, char *argv[]) {

  char *split = strchr(argv[1], '=');
  if (!split) {
    return 1;
  }

  char *key=argv[1];
  *split = '\0';
  char *value = split + 1;

  printf("%s=%s\n", key, value);

  return 0;
}
```

tools/tpm2_hierarchycontrol.c
```diff
@@ -0,0 +1,174 @@
...
+
+    return true;
+}
+
+bool tpm2_tool_onstart(tpm2_options **opts) {
+
+    const struct option topts[] = {
+        { "context-object",	required_argument, NULL, 'c' },
+        { "auth-key",		required_argument, NULL, 'p' },
+        { "enable",		required_argument, NULL, 'e' },
```
williamcroberts: delete

tools/tpm2_hierarchycontrol.c
```diff
@@ -0,0 +1,174 @@
...
+        break;
+    }
+
+    return true;
+}
+
+bool tpm2_tool_onstart(tpm2_options **opts) {
+
+    const struct option topts[] = {
+        { "context-object",	required_argument, NULL, 'c' },
```
williamcroberts: C --hierarchy

tools/tpm2_hierarchycontrol.c
```diff
@@ -0,0 +1,174 @@
...
+        { "auth-key",		required_argument, NULL, 'p' },
```
williamcroberts: -P --auth-hiearchy

man/tpm2_hierarchycontrol.1.md
remuswu1019: > Options need to change:
> 
> * -p should be -P and --auth-key should be --auth-hierarchy.
> * -c should be -C and --context-object should be --hiearchy.

Okay, I will commit again with this change.

tools/tpm2_hierarchycontrol.c
remuswu1019: What did mean "delete"?
Delete these 4 lines or modify them to correct option?

test/integration/tests/hierarchycontrol.sh
remuswu1019: Thank you for this comment, I will change the testing flow.

tools/tpm2_hierarchycontrol.c
joholl: I think @williamcroberts is referring to his comment above

> kill this option and make it argument.
> So
> ```sh
> tpm2_hierarchycontrol -c e -e phEnable set
> ```
> becomes:
> ```sh
> tpm2_hierarchycontrol -c e phEnable set
> ```

As a consequence there would not be an option `-e`/`--enable` any more.

williamcroberts: Yeah just delete the -e line and move it into the tools arguments. Sorry for the thrash on the review, we are scrutinizing our options as we move into the 4.0 release with vigor, as we are stuck with those choices. As we move forward, we will stabilize the CLI.

man/tpm2_hierarchycontrol.1.md
remuswu1019: I will use first one.

tools/tpm2_hierarchycontrol.c
remuswu1019: No problem, please review patches again, Thanks.

man/tpm2_hierarchycontrol.1.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+## Set shEnable with owner hierarchy
+```
+tpm2_hierarchycontrol -C o shEnable set
+```
+
+## Check current TPMA_STARTUP_CLEAR Bits
+```
+tpm2_getcap -c properties-variable
```
williamcroberts: tpm2_getcap -c is now an argument.

man/tpm2_hierarchycontrol.1.md
```diff
@@ -0,0 +1,75 @@
...
+% July 2019
+
+# NAME
+
+**tpm2_hierarchycontrol**(1) - Enable and disable use of a hierarchy and its
+associated NV storage.
+
+# SYNOPSIS
+
+**tpm2_hierarchycontrol** [*OPTIONS*] _OPERATION_
```
williamcroberts:
```
_VARIABLE_ _OPERATION_
```

man/tpm2_hierarchycontrol.1.md
```diff
@@ -0,0 +1,75 @@
...
+
+# DESCRIPTION
+
+**tpm2_hierarchycontrol**(1) - Allows user change phEnable, phEnableNV,
+shEnable and ehEnable when the proper authorization is provided. Authorization
+ should be one out of owner hierarchy auth, endorsement hierarchy auth and
+ platform hierarchy auth. **As an argument the tool takes the \_OPERATION\_ as
+ string clear|set to clear or set the TPMA_STARTUP_CLEAR bits.**
```
williamcroberts: this needs to be updated too, for it taking two arguments.

test/integration/tests/hierarchycontrol.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# ERROR: Only platform and its authorization can be specified for 0
+tpm2_hierarchycontrol -C o ehEnable clear
+tpm2_hierarchycontrol -C o phEnable clear
+tpm2_hierarchycontrol -C o phEnableNV clear
+tpm2_hierarchycontrol -C e shEnable clear
+tpm2_hierarchycontrol -C e phEnable clear
+tpm2_hierarchycontrol -C e phEnableNV clear
+
+exit 0
```
williamcroberts: Tests look great, I like how the tool evolved!

williamcroberts:
The build failed on import, occasionally we get transient issues, not 100% sure why, I restarted it. Just update your man page to reflect the change in arguments.

Issue #880, opened 2018-02-01
Check that handle values passed to the tools are between a supported range


martinezjavier:
The TPM2 Handle space is split for the different TPM2 resources. For example, NV Index Areas start at 0x01000000, Transient Objects at 0x80000000 and Persistent Objects at 0x81000000. This means that each set of commands are only supposed to be used within the Handle range of the resources that they manage.

So for example TPM2_Unseal should only be used with handles from the Transient Objects range and TPM2_NV_Write for handles from the Persistent Objects range.

Right now any hex value can be used as a handle and even commands that will fails are sent to the TPM. This can be easily checked in the tools to avoid that.

williamcroberts:
ESAPI port handles:
- transient objects
- persistent objects

However NV space is left, and also NV space would be nice if you could just do 1,2,3,4 and it just adds the 0x01000000 to it.

idesai:
@williamcroberts, esapi doesn't check the range for NV; but does check for transient and persistent? Sounds like this should be checked in esapi.
However, there is more to parsing the handle value than it's relevance to a specific tool. What about reserved key handles and NV Indices. There is a TCG registry of these and the handle ranges **vary depending on which TCG TPM workgroup** we are talking about. I am not sure the tpm simulator restricts you on writing these ranges today -- I do know that Intel(r) PTT or Intel's integrated TPM does prevent you from writing NV Indices reserved for EK resulting in TPM error. This being one example. The full registry is here https://trustedcomputinggroup.org/wp-content/uploads/RegistryOfReservedTPM2HandlesAndLocalities_v1p1_pub.pdf

williamcroberts:
Your over thinking this, its more like the top byte of the handle should be 0x1:
```C
#define TPM2_HT_NV_INDEX       ((TPM2_HT) 0x01) /* NV Index  assigned by the caller */
#define TPM2_HR_SHIFT             ((TPM2_HC) 24)
#define TPM2_HR_NV_INDEX          ((TPM2_HC) (TPM2_HT_NV_INDEX << TPM2_HR_SHIFT))
```

However, this begs the question, handles like 0, 1, 2 ... N (within range) should work too. I don't think when people want to write to index 0 that they have to write 0x01000000 (or something like that). So we probably want to say, if the top byte is set in the handle it must be 0x01, else if the top byte is all 0 set it to 1. This way things like 0 as input (pcr index 0), just internally turn into the right handle.

PR #1588, opened 2019-07-09
lib/tpm2_hierarchy: Add filtering for NV Indices

idesai:
Progresses #1462
Progresses #880

1. Add filter for checking handle is in valid NV index range
2. Add option to specify NV indices simply as an offset. Eg. 1 for NV index @ 01000001
3. Add unit tests for the added functionality.

### DIFF
--- a/lib/tpm2_hierarchy.c
+++ b/lib/tpm2_hierarchy.c
@@ -29,6 +29,84 @@
  * @return
  *  True on success, False otherwise.
  */
+
+static bool filter_hierarchy_handles(TPMI_RH_PROVISION hierarchy,
+    tpm2_hierarchy_flags flags) {
+
+    switch(hierarchy) {
+        case TPM2_RH_OWNER:
+            if ( !(flags & TPM2_HIERARCHY_FLAGS_O) ) {
+                LOG_ERR("Unexpected handle - TPM2_RH_OWNER");
+                return false;
+            }
+            break;
+        case TPM2_RH_PLATFORM:
+            if ( !(flags & TPM2_HIERARCHY_FLAGS_P) ) {
+                LOG_ERR("Unexpected handle - TPM2_RH_PLATFORM");
+                return false;
+            }
+            break;
+        case TPM2_RH_ENDORSEMENT:
+            if ( !(flags & TPM2_HIERARCHY_FLAGS_E) ) {
+                LOG_ERR("Unexpected handle - TPM2_RH_ENDORSEMENT");
+                return false;
+            }
+            break;
+        case TPM2_RH_NULL:
+            if ( !(flags & TPM2_HIERARCHY_FLAGS_N) ) {
+                LOG_ERR("Unexpected handle - TPM2_RH_NULL");
+                return false;
+            }
+            break;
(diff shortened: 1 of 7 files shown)


lib/tpm2_hierarchy.c
```diff
@@ -29,6 +29,82 @@
  * @return
  *  True on success, False otherwise.
  */
+
+static bool filter_hierarchy_handles(TPMI_RH_PROVISION *hierarchy,
```
williamcroberts: hiearchy is read and never updated, so no point in the pointer here.

lib/tpm2_hierarchy.c
```diff
@@ -29,6 +29,82 @@
...
+
+    return true;
+}
+
+static bool filter_handles(TPMI_RH_PROVISION *hierarchy, tpm2_hierarchy_flags flags) {
+
+    bool result = true;
+    switch(*hierarchy & TPM2_HR_RANGE_MASK) {
+        case 0: //If user simply specifies an index offset
+            *hierarchy += TPM2_HR_NV_INDEX;
```
williamcroberts: Check that NV flag is set.

lib/tpm2_hierarchy.h
```diff
@@ -14,13 +14,16 @@ typedef enum tpm2_hierarchy_flags tpm2_hierarchy_flags;
 
 enum tpm2_hierarchy_flags {
     TPM2_HIERARCHY_FLAGS_NONE = 0,
+
```
williamcroberts: remove

lib/tpm2_hierarchy.h
```diff
@@ -14,13 +14,16 @@ typedef enum tpm2_hierarchy_flags tpm2_hierarchy_flags;
...
     TPM2_HIERARCHY_FLAGS_O    = 1 << 0,
     TPM2_HIERARCHY_FLAGS_P    = 1 << 1,
     TPM2_HIERARCHY_FLAGS_E    = 1 << 2,
     TPM2_HIERARCHY_FLAGS_N    = 1 << 3,
     TPM2_HIERARCHY_FLAGS_L    = 1 << 4,
-    TPM2_HIERARCHY_SUPPRESS   = 1 << 5,
     TPM2_HIERARCHY_FLAGS_ALL  = 0x1F,
+
+    TPM2_HANDLES_FLAGS_NV     = 1 << 5,
+
```
williamcroberts: same here, extra new lines

lib/tpm2_hierarchy.c
```diff
@@ -81,32 +157,8 @@ bool tpm2_hierarchy_from_optarg(const char *value,
      * If the caller specifies the expected valid hierarchies, either as string,
      * or hex handles, they are additionally filtered here.
      */
-    if (!(flags & TPM2_HIERARCHY_FLAGS_O) && *hierarchy == TPM2_RH_OWNER) {
```
williamcroberts: aren't the NV and HIERARCHY flag ranges now mutually exclusive, you should probably check that. With that said, we can assume the caller knows what they are doing since it's an internal routine.

Issue #1042, opened 2018-05-17
audit manpages to tools


williamcroberts:

joshuagl:
- ensure manpage formatting is consistent
  - consistent whitespace use (blank line after option name before description?)
  - use of quotation marks around section names
  - consistent formatting of options i.e. code formatting or bold formatting?

joholl:
From #1557: don't forget to update the wiki pages.

williamcroberts:
General:
 - [ ] remove dates from all man pages

- tpm2_activatecredential:
  - [ ] -E becomes -e
  - [ ] -P becomes -p to match the lowercase object -c option.
  - [ ] - full end to end example

- tpm2_certify:
  - [ ] - -C and -c should be reversed since the signing key has authority over the certifying key.
  - [ ] -P and -p flipped with -C and -c.
  - [ ] -s --sig-file to --signature
  - [ ] --out-attest-file to --attestation
  - [ ] update examples

- tpm2_checkqoute:
  - [ ] is pcrs_from_file() routine a helper, if so use the helper. Perhaps coalesce with other tools that use -F.
  - [ ] stdout outputs don't make sense, remove them.

- tpm2_clear:
  - [ ] - -L becomes an argument
  - [ ] -p becomes -c and uses the load object interface directly with the user value. Keep default the same.

- tpm2_clearcontrol:
  - a becomes -c because thats the object being worked on.

- tpm2_create:
 - [ ] - example incomplete needs a createprimary call.

PR #1600, opened 2019-07-21
Rename options

idesai:
Progresses #1039 and #1042
1. Tools requiring hierarchy specified should use option "hierarchy" --> auth-hierarchy goes away
2. Tools requiring single authentication should use option "auth" alone --> auth-hierarchy, hierarchy-auth, object-auth etc. goes away
3. Tools requiring multiple auths should follow "object-auth" to tell apart auths--> auth-hierarchy becomes hierarchy-auth, auth-key becomes key-auth, etc.
4. halg becomes hash-algorithm
5. kalg,algorithm becomes key-algorithm
6. sig becomes signature
7. All options drop the term file. Eg. in-file becomes input
8. All tools using hierarchy input should specify it  as C
9. Endorsement hierarchy will use known abbreviations and becomes eh
10. set-list, sel-list becomes pcr-list with short option becoming "l"
11. alg becomes algorithm in general
12. Tools with single option taking type hierarchy or NV or other tpm handle types should call it "object-context" and use short option "C"
13. qualify-data becomes qualification-data
14. pcr-input short option becomes "f" and long options becomes pcr

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 ## Changelog
 ### next
+  * tpm2_quote: short and long options changed
+    auth-ak is now ak-auth
+    id-list is now pcr-index
+    sel-list is now pcr-list
+    qualify-data is now qualification-data
+    pcrs is now pcr
+    halg is now hash-signature
+    -l is now -i
+    -L is now -l
+    -p is now -f
+    -f is now -F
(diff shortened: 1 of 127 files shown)


williamcroberts:
@idesai:
- good summary on github
- good break down of individual commits. Commit granularity was great.
- passes CI

Excellent change.

PR #1608, opened 2019-07-25
Fix tpm2_encryptdecrypt pcks7 padding scheme

### LINKED ISSUE #1547: PKCS7 padding in encryptdecrypt tool skips padding when input aligns at block length and strip padding on decryption

idesai:
Fixes #1547 

In order to make the tpm2_encryptdecrypt interoperable with openssl, below fixes have been implemented:
1. Do not apply pkcs7 padding to all aes modes and in particular to aes cfb mode
2. The input iv for aes ecb mode should be set to null
3. Padded block length was wrongly set to key size; instead it should have been fixed to AES block length of 16 bytes or 128 bits.

### DIFF
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -1372,7 +1372,7 @@ tool_rc tpm2_encryptdecrypt(
                 iv_out);
     }
 
-    if (tpm2_error_get(rval) == TPM2_RC_COMMAND_CODE) {
+    if (rval != TPM2_RC_SUCCESS) {
         if (*version == 2) {
             LOG_PERR(Esys_EncryptDecrypt2, rval);
         } else {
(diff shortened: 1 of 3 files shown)


williamcroberts:
My fault, ill fix this up.
```
/workspace/tpm2-tools
man/tpm2_loadexternal.1.md:15: trailing whitespace.
+object's seed. The command allows loading of a public area or both a public 
```

williamcroberts:
> tpm2_loadexternal

Weird I am not seeing that in my local tree....

williamcroberts:
Where are my tests? Their should be one encrypting with ossl and decrypting with tpm. You can import the shared symmetric key into the TPM for testing.

idesai:
> Where are my tests? Their should be one encrypting with ossl and decrypting with tpm. You can import the shared symmetric key into the TPM for testing.

@williamcroberts, tests coming up :)

idesai:
> Needs test, try to test both pad, no pad, and the modes that don't do padding.

Will add a test for no padding as well.

PR #1609, opened 2019-07-25
test: remove external hex conversion of random bytes

joholl:
For quoting, 20 random bytes are needed. However, the conversion of these random bytes into a hex stream is done two times, first with `--hex` (introduced with 460afc31) and then by calling `xxd`. This results in 80 characters (representing 40 bytes) which is not what we want.

With `tpm2_getrandom`'s internal conversion into a hex stream, the call of `xxd` should be ommitted.

### DIFF
--- a/test/integration/tests/attestation.sh
+++ b/test/integration/tests/attestation.sh
@@ -35,7 +35,7 @@ cleanup() {
   rm -f $file_input_data $file_input_key $output_ek_pub $output_ek_pub_pem $output_ak_pub \
         $output_ak_pub_pem $output_ak_pub_name $output_mkcredential \
         $output_actcredential $output_quote $output_quotesig $output_quotepcr $context_ak \
-        rand.out session.ctx
+        session.ctx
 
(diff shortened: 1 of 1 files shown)


test/integration/tests/attestation.sh
```diff
@@ -91,7 +85,7 @@ diff $file_input_data $output_actcredential
 tpm2_pcrreset -Q $debug_pcr
 tpm2_pcrextend -Q $debug_pcr:sha256=$rand_pcr_value
 tpm2_pcrlist -Q
-getrandom 20
+loaded_randomness=$(tpm2_getrandom --hex 20)
```
williamcroberts: Good catch. I set binary as the default output, but that looks silly. I chose that based on what I though `openssl dgst` was doing, but I was dead wrong. I'm going to change the default to hex and also add a binary option.

williamcroberts: Wait shouldn't this be a fix to getrandom wrapper?

PR #1615, opened 2019-07-25
tcti: Replace tpm2_tcti_ldr with libtss2-tctildr.

flihp:
This commit removes the tpm2_tcti_ldr module replacing it with the new
tctildr library from the tpm2-tss repo. This is a "drop in" replacement,
leaving the existing `--tcti` command line interface unchanged.


flihp:
This new library went into tpm2-tss in https://github.com/tpm2-software/tpm2-tss/pull/1470. The goal is to have this in the next / 2.3 release. I figure adding support downstream is the next step. Not sure how you want to fit this into the 4.0 release schedule but it works when you want it.

williamcroberts:
-539 +67, perfect

PR #1634, opened 2019-08-01
 tpm2_nvreadlock: use index auth if no --hierarchy is specified

diabonas:
Consider the following example:
```
tpm2_createpolicy --policy-pcr --pcr-list sha256:0,1,2,3 --policy policy.data
tpm2_nvdefine --size 5 --attributes 'policyread|ownerwrite|read_stclear' --policy policy.data 0x1500018
echo test | tpm2_nvwrite --hierarchy o --input - 0x1500018
tpm2_nvread --auth pcr:sha256:0,1,2,3 0x1500018
tpm2_nvreadlock --auth pcr:sha256:0,1,2,3 0x1500018
```
The last command fails with
```
WARNING:esys:src/tss2-esys/api/Esys_NV_ReadLock.c:303:Esys_NV_ReadLock_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_NV_ReadLock.c:109:Esys_NV_ReadLock() Esys Finish ErrorCode (0x0000012f) 
ERROR: Failed to lock NVRAM area at index 0x1500018
ERROR: Esys_NV_ReadLock(0x12F) - tpm:error(2.0): authValue or authPolicy is not available for selected entity
ERROR: Unable to run tpm2_nvreadlock
```
In contrast to `tpm2_nvread`, you need to explicitly specify the NV index in `--hierarchy` for the command to work:
```
tpm2_nvreadlock --hierarchy 0x1500018 --auth pcr:sha256:0,1,2,3 0x1500018
```
This pull request brings the behaviour of `tpm2_nvreadlock` in line with `tpm2_nvread`, `tpm2_nvwrite` and `tpm2_nvincrement` so that above example works as expected. The code is directly taken from these tools. Additionally the man pages of the NV tools are updated to have a unified, up-to-date description for `-C`/`--hierarchy`.

### DIFF
--- a/tools/tpm2_nvreadlock.c
+++ b/tools/tpm2_nvreadlock.c
@@ -17,13 +17,16 @@ struct tpm_nvreadlock_ctx {
     TPM2_HANDLE nv_index;
 };
 
-static tpm_nvreadlock_ctx ctx = {
-    .auth_hierarchy.ctx_path = "owner",
-};
+static tpm_nvreadlock_ctx ctx;
 
 
 static bool on_arg(int argc, char **argv) {
-
+    /* If the user doesn't specify an authorization hierarchy use the index
+    * passed to -x/--index for the authorization index.
+    */
+    if (!ctx.auth_hierarchy.ctx_path) {
+        ctx.auth_hierarchy.ctx_path = argv[0];
+    }
     return on_arg_nv_index(argc, argv, &ctx.nv_index);
 }
 
--- a/man/tpm2_nvincrement.1.md
+++ b/man/tpm2_nvincrement.1.md
@@ -16,10 +16,9 @@ handle range "TPM2_HR_NV_INDEX".
 
 # OPTIONS
 
-  * **-C**, **\--hierarchy**=_AUTH_:
+  * **-C**, **\--hierarchy**=_AUTH_HANDLE_:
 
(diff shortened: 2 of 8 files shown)


tools/tpm2_nvreadlock.c
```diff
@@ -14,16 +14,22 @@ struct tpm_nvreadlock_ctx {
         tpm2_loaded_object object;
     } auth_hierarchy;
 
+    bool is_auth_hierarchy_specified;
```
williamcroberts: no need for this bool, just check if path is NULL

tools/tpm2_nvreadlock.c
```diff
@@ -14,16 +14,22 @@ struct tpm_nvreadlock_ctx {
...
+    .is_auth_hierarchy_specified = false,
 };
 
 
 static bool on_arg(int argc, char **argv) {
-
+    /* If the user doesn't specify an authorization hierarchy use the index
+    * passed to -x/--index for the authorization index.
+    */
+    if (!ctx.is_auth_hierarchy_specified) {
```
williamcroberts: if ctx.path is NULL then...

tools/tpm2_nvreadlock.c
diabonas: Done, I also changed this in `tpm2_nvincrement`, `tpm2_nvread` and `tpm2_nvwrite` to keep the code consistent everywhere.

PR #1636, opened 2019-08-01
tpm2_policyor.1.md: Update example

### LINKED ISSUE #1628: man/tpm2_policyor: example issues
Their are some questions that arrose when I looked into the policyor manpage.

 - [x] why do we fake a pcr extension on sha1 but build policy against bank sha256?
 - [x] why don't we show the pcr extension occurring and the other branch succeeding? Can we 
       have this?

Please address these concerns.

idesai:
Fixes #1628


williamcroberts:
need signoff (git commit --amend -s) and rebase conflict resolution

Issue #1389, opened 2019-03-31
tpm2_policy_pcr_build: pcr_values wrong type


maage:
https://github.com/tpm2-software/tpm2-tools/blob/fdc0fb34c49bdeaca63c511a0fb6f246b2b2aa32/lib/tpm2_policy.c#L90

From: https://bugzilla.redhat.com/show_bug.cgi?id=1694462

My version is;

tpm2-tools-3.1.4-1.fc30.x86_64

But error code seems to be the same.

```
[root@localhost ~]# tpm2_pcrlist -L sha256:0,1,2,3,4,5,6,7,8,9 -o dig
sha256 :
  0  : dc47a38bd654f863cce3ce1db9da513ba91f763d0f1344a18b441efc8c4ce535
  1  : 0824fb47bacb30e202416aa8edca75f8490f47e65f43e9f7ca0c5a579300f09e
  2  : cca4fa22e5754c1d6757efb91a699b54a5e95227c18ab6a3f2b03f58b1e486c1
  3  : 3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969
  4  : dceeaf355ad7f887d2813f3a2e00fe57bff6ecf8bb58ca9d9e4390df85195192
  5  : 2d774735843911e1e2c735fb075aa11784c070ca82cf954882ce6c6fec4a3041
  6  : 3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969
  7  : 3ec8b3bf34e2c20d9470e053aa1a5fc6bad72e51af74f8a6f011ed929c572120
  8  : 232f1ebad23fe79915c931f5f7fed548304151d409564988703b0bd58101f735
  9  : ade01d3d21e6fdb68fd86cbeaefc48c7a0359535361a9d8c98e397b761ba21fb
[root@localhost ~]# tpm2_createpolicy -P -L sha256:0,1,2,3,4,5,6,7,8,9 -F dig -f /dev/null
*** stack smashing detected ***: <unknown> terminated
Aborted (core dumped)
```

Stack trace is about the same as main issue.

```
(gdb) up
#5  0x000055580a8f31de in tpm2_policy_pcr_build (sapi_context=0x55580b089dc0, policy_session=0x55580b08ae10, 
    pcr_selections=<optimized out>, raw_pcrs_file=0x7ffdf261466d "/tmp/tmp.d98wPpcRy6/pcr.digest")
[...]

PR #1638, opened 2019-08-01
Add ability to use multiple banks in the PCR mini language

### LINKED ISSUE #1635: Cannot specify multiple banks in the PCR mini language
Consider the following tool output:
```
$ tpm2_nvread --auth pcr:sha256:0,1,2,3+sha1:0,1,2,3 0x1234567
ERROR: Could not open file: "sha1:0,1,2,3" error: No such file or directory
ERROR: Could not retrieve raw_pcrs_file size
ERROR: Invalid handle authorization
ERROR: Unable to run tpm2_nvread
```
According to the documentation, the syntax for `--auth` should be correct:
> ## PCR Authorizations
>
> You can satisfy a PCR policy using the "pcr:" prefix and the PCR minilanguage. The PCR
minilanguage is as follows:
> `<pcr-spec>+<raw-pcr-file>`
>
> The PCR spec is documented in in the section "PCR bank specifiers".
>
> The `raw-pcr-file` is an **optional** the output of the raw PCR contents as returned by *tpm2_pcrread(1)*.

> # PCR Bank Specifiers
>
> PCR Bank Selection lists follow the below specification:
>
> ```
> <BANK>:<PCR>[,<PCR>] or <BANK>:all
> ```
>
> multiple banks may be separated by '+'.

However the problem is that `+` is used both to separate multiple banks as well as to separate the optional `raw-pcr-file` from the PCR specification. This results in everything after the first `+` being interpreted as `raw-pcr-file`.

Probably the easiest fix is to use a different separator for `raw-pcr-file` in [`handle_pcr()`](https://github.com/tpm2-software/tpm2-tools/blob/923ae0e364d766354bdcefaeaed9efe4a23408be/lib/tpm2_auth_util.c#L178).

diabonas:
Fix #1635 by changing the separator for `raw-pcr-file` from `+` to `=`. The equals sign was chosen as the new separator to have a syntax somewhat similar to `tpm2_pcrextend`.

While fixing this issue, I came across a different, unrelated problem with using multiple banks: the feature to use a `raw-pcr-file` of expected PCR values is broken due to a [broken loop counter in `tpm2_policy.c`](https://github.com/tpm2-software/tpm2-tools/blob/d381875156eb9352b1a1ef4cc7d0c903cba878b6/lib/tpm2_policy.c#L56-L59), resulting in errors like
```
$ tpm2_unseal -p pcr:sha256:0+sha1:0=file
WARNING:esys:src/tss2-esys/api/Esys_PolicyPCR.c:288:Esys_PolicyPCR_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_PolicyPCR.c:100:Esys_PolicyPCR() Esys Finish ErrorCode (0x000001c4)
ERROR on line: "53" in file: "lib/log.h": Esys_PolicyPCR(0x1C4) - tpm:parameter(1):value is out of range or is not correct for the context
ERROR on line: "67" in file: "tools/tpm2_unseal.c": Invalid item handle authorization
ERROR on line: "145" in file: "tools/tpm2_tool.c": Unable to run tpm2_unseal
```
A fix for this has already been proposed by @maage in #1389 and #1390, I cherry-picked the second commit from there.

To make sure PCR specifications with multiple banks get some more test coverage in the future, I replaced the previously used PCR specification `sha1:0,1,2,3` by `sha256:0,1,2,3+sha1:0,1,2,3` in the integration tests.

### DIFF
--- a/lib/tpm2_auth_util.c
+++ b/lib/tpm2_auth_util.c
@@ -175,7 +175,7 @@ static tool_rc handle_pcr(ESYS_CONTEXT *ectx, const char *policy, tpm2_session *
 
     const char *pcr_str = dup;
     const char *raw_path = NULL;
-    char *split = strchr(dup, '+');
+    char *split = strchr(dup, '=');
     if (split) {
         *split = '\0';
         raw_path = split + 1;
--- a/lib/tpm2_policy.c
+++ b/lib/tpm2_policy.c
@@ -54,9 +54,9 @@ static bool evaluate_populate_pcr_digests(TPML_PCR_SELECTION *pcr_selections,
          */
         UINT8 k;
         for (k = 0; k < total_indices_for_this_alg; k++) {
-            pcr_values->digests[dgst_cnt + k].size = dgst_size;
+            pcr_values->digests[dgst_cnt].size = dgst_size;
+            dgst_cnt++;
         }
-        dgst_cnt++;
     }
 
     //Check if the input pcrs file size is the same size as the pcr selection setlist
--- a/man/common/authorizations.md
+++ b/man/common/authorizations.md
@@ -90,7 +90,7 @@ session:session.ctx+hex:11223344
 
 You can satisfy a PCR policy using the "pcr:" prefix and the PCR minilanguage. The PCR
 minilanguage is as follows:
-`<pcr-spec>+<raw-pcr-file>`
+`<pcr-spec>=<raw-pcr-file>`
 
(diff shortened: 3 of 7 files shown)


williamcroberts:
We should probably add specific tests into the unit tests as well.

williamcroberts:
I'm going to take this as is, as it fixes an issue and has tests. For bonus points submit some unit tests. I filed issue #1641 for that.

PR #1643, opened 2019-08-03
Changes to tool tpm2_getmanufec

idesai:
- Short and long options changed
- ~cmdline option to specify ek public is now arg~
- Removed ek key creation and management logic.
- ~Change ekcert-portal from arg to cmdline option~
- Remove dead short option
- Renamed the tool to tpm2_getekcertificate
- Moved the tool to misc

Additionally,

    1. If the TPM manufacturer is the IBM simulator, error out since the
       simulator endorsement keys aren't certified by IBM.
    2. If the TPM manufacturer is Intel aka the the TPM2 device is PTT,
       also if the tpmGeneratedEPS bit is set it implies that the soc
       or pch has a firmware that has mitigations for Intel security
       advisory SA-00086. And so another utility must be used to retrieve
       the endorsement key certificate. More information on the advisory:
       https://www.intel.com/content/www/us/en/security-center/advisory/intel-sa-00086.html
       The alternative utility and the instructions can be found here:
       https://github.com/intel/INTEL-SA-00086-Linux-Recovery-Tools.

Progresses #1042

### DIFF
--- a/man/tpm2_getekcertificate.1.md
+++ b/man/tpm2_getekcertificate.1.md
@@ -14,53 +14,11 @@ server.
 
 **tpm2_getekcertificate**(1) - Retrieve the Endorsement key Certificate for
 the TPM endorsement key from the TPM manufacturer's endorsement certificate hosting
-server.
+server. The argument _URL_ specifies the address for the ek certificate portal.
 
 # OPTIONS
 
-  * **-P**, **\--eh-auth**=_ENDORSE\_AUTH_:
-
-    Specifies current endorsement authorization.
-    Authorizations should follow the "authorization formatting standards", see
-    section "Authorization Formatting".
-
-  * **-p**, **\--ek-auth**=_EK\_AUTH_
-
-    Specifies the EK authorization when created.
-    Same formatting as the endorse authorization value or **-e** option.
-
-  * **-w**, **\--owner-auth**=_OWNER\_AUTH_
-
-    Specifies the current owner authorization.
-    Same formatting as the endorse authorization value or **-e** option.
-
-  * **-H**, **\--persistent-handle**=_HANDLE_:
-
-    Specifies the handle used to make EK  persistent.
-    If a value of **-** is passed the tool will find a vacant persistent handle
-    to use and print out the automatically selected handle.
-
-  * **-G**, **\--key-algorithm**=_ALGORITHM_:
-
-    Specifies the algorithm type of EK.
-    See section "Supported Public Object Algorithms" for a list of supported
-    object algorithms. See section "Algorithm Specifiers" on how to specify
-    an algorithm argument.
-
-  * **-o**, **\--output**=_FILE_:
-
-    Specifies the file used to save the public portion of EK.
-
-  * **-N**, **\--non-persistent**:
-
-    Specifies to readout the EK public without making it persistent.
-
-  * **-O**, **\--offline**=_FILE_:
-
-    Specifies the file that contains an EK retrieved from offline
-    platform that needs to be provisioned.
-
-  * **-E**, **\--ec-cert**=_EC\_CERT\_FILE_:
+  * **-E**, **\--ec-cert**=_EK\_CERTIFICATE\_FILE_:
 
     Specifies the file used to save the Endorsement key certificate retrieved from
     the TPM manufacturer provisioning server. Defaults to stdout if not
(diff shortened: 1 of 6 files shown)


idesai:
@williamcroberts Makefile.am changes to move the tool to misc in commit dddd6c3 is causing the compiled tool to show in both tools and tools/misc folder. Any thoughts?

man/tpm2_getekcertificate.1.md
```diff
@@ -0,0 +1,63 @@
...
+# NOTES
+
+When the verbose option is specified, additional curl debugging information is
+provided by setting the curl mode verbose, see
+<https://curl.haxx.se/libcurl/c/CURLOPT_VERBOSE.html> for more information.
+
+# EXAMPLES
+
+```
+tpm2_getekcertificate -X -o ECcert.bin ek.bin -a https://tpm.manufacturer.com/ekcertserver/
```
williamcroberts: full working example

man/tpm2_getekcertificate.1.md
```diff
@@ -0,0 +1,63 @@
...
+
+# NAME
+
+**tpm2_getekcertificate**(1) - Retrieve the Endorsement key Certificate for the TPM
+endorsement key from the TPM manufacturer's endorsement certificate hosting
+server.
+
+# SYNOPSIS
+
+**tpm2_getekcertificate** [*OPTIONS*] _EK\_PUBLIC\_FILE_
```
williamcroberts: I am not 100% sure what the _EK\_PUBLIC\_FILE_ is as it's not documented nor in the examples on how obtain it. Im assuming its the public portion of the object. Make the argument the URL and make the ek an argument, so it just follows the normal -u convention. If it's supposed to be ek context, then use -c.

williamcroberts:
> @williamcroberts Makefile.am changes to move the tool to misc in commit [dddd6c3](https://github.com/tpm2-software/tpm2-tools/commit/dddd6c39688bdb68ed4889b8837a8a023d604cfa) is causing the compiled tool to show in both tools and tools/misc folder. Any thoughts?

Probably didn't clean you tree before moving and recompiling so it's old. If you look at the timestamp is it the same as the other tool?

man/tpm2_getekcertificate.1.md
idesai: > > @williamcroberts Makefile.am changes to move the tool to misc in commit [dddd6c3](https://github.com/tpm2-software/tpm2-tools/commit/dddd6c39688bdb68ed4889b8837a8a023d604cfa) is causing the compiled tool to show in both tools and tools/misc folder. Any thoughts?
> 
> Probably didn't clean you tree before moving and recompiling so it's old. If you look at the timestamp is it the same as the other tool?

I did do a clean and distclean prior to compiling again. The timestamps match too. The only difference I can tell from the other misc/tools is that the getmanufec( or now getekcertificate) has CFLAGS and LDADD that I am also moving to tools_misc_tpm2_getekcertificate

man/tpm2_getekcertificate.1.md
idesai: This will be an interesting one. In that, this would never work on a simulator. Nevertheless, I get your point. The example at least need to show how you arrive at the EK public required as an input to be given to the tool.

man/tpm2_getekcertificate.1.md
idesai: > I am not 100% sure what the _EK_PUBLIC_FILE_ is as it's not documented nor in the examples on how obtain it. Im assuming its the public portion of the object. Make the argument the URL and make the ek an argument, so it just follows the normal -u convention. If it's supposed to be ek context, then use -c.

Yes it exclusively requires 2 inputs -
1. The public portion of the endorsement key and 
2. The portal address that host the certificate for the public key.

I will update the example section to show how to retrieve the EK_PUBLIC_FILE. Should this then still follow the -u convention. I was thinking from a factory line perspective where you have one internet facing server to retrieve multiple ek certificates. If the factory used a script the only variable would be the ek public of the various platforms which can be passed as an argument.

man/tpm2_getekcertificate.1.md
idesai: Updated example section

man/tpm2_getekcertificate.1.md
idesai: Updated example section

williamcroberts: flip so the URL is not an option, the pub is via -u

williamcroberts:
rebased on #1653

PR #1645, opened 2019-08-05
manpage audit

williamcroberts:
* tpm2_sign:
     - Drop -m and -d option arguments and make them sole argument.
     - Remove the -m option and make -d toggle if input is a digest.
  * tpm2_encryptdecrypt: drop -i input option and make argument.
  * tpm2_rsadecrypt: drop -i input option and make argument.
  * tpm2_rsaencrypt: make output binary and either stdout or file based on -o.
  * tpm2_rsadecrypt: add support for label via -l.
  * tpm2_rsaencrypt: add support for label via -l.

### DIFF
--- a/man/tpm2_encryptdecrypt.1.md
+++ b/man/tpm2_encryptdecrypt.1.md
@@ -6,12 +6,13 @@
 
 # SYNOPSIS
 
-**tpm2_encryptdecrypt** [*OPTIONS*]
+**tpm2_encryptdecrypt** [*OPTIONS*] _FILE_
 
 # DESCRIPTION
 
 **tpm2_encryptdecrypt**(1) - Performs symmetric encryption or decryption with a
-specified symmetric key.
+specified symmetric key on the contents of _FILE_.
+If _FILE_ is not specified, defaults to *stdin*.
 
 # OPTIONS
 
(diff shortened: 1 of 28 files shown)


man/tpm2_encryptdecrypt.1.md
```diff
@@ -74,11 +70,20 @@ specified symmetric key.
...
-tpm2_encryptdecrypt -c key.dat -p abc123 -i <filePath> -o <filePath>
-tpm2_encryptdecrypt -c 0x81010001 -p 123abca  -i <filePath> -o <filePath>
-tpm2_encryptdecrypt -c 0x81010001 --enable-pkcs7-padding -i <filePath> -o <filePath>
+
+# Encrypt and Decrypt some data
+```bash
+echo "my secret" > secret.dat
+tpm2_encryptdecrypt -c key.ctx -o secret.enc secret.dat
+tpm2_encryptdecrypt -d -c key.ctx -o secret.dec secret.enc
+cat secret.dec
```
idesai: Do we want to compare the results or simply print the output to stdout in the example section.

man/tpm2_rsaencrypt.1.md
```diff
@@ -49,8 +48,24 @@ The key referenced by key-context is **required** to be:
...
+## Encrypt using RSA
+```bash
+echo "my message" > msg.dat
+tpm2_rsaencrypt -c key.ctx -o msg.enc msg.dat
+```
+
+## Decrypt using RSA
+```bash
+tpm2_rsadecrypt -c key.ctx -o msg.ptext msg.enc
+cat msg.ptext
```
idesai: Print to stdout or compare ?

williamcroberts:
> example section - proposing that we should compare the result instead of printing to stdout. But ignore if it is consistent with other examples.

I think it's more clear when the example has some visual feedback of what's going on. It's also consistent with what I have been doing. Either way is fine for now, we can always update example sections as we see fit. I would err on showing the user whats going on with cat/echo type things.

PR #1648, opened 2019-08-06
selftest and bash examples for manpage audit

williamcroberts:
- audit the selftest set of tools
- make sure EXAMPLES have ```bash decorator


idesai:
Good one. whitespace on verifysignature markdown on line 23 needs fixed.

PR #1668, opened 2019-08-13
tpm2_policyauthorize: Fix incorrect text in error logging

### LINKED ISSUE #1658: tpm2_policyauthorize: weird options
-c is mentioned in is_checkinput_options_ok() but their is no -c option. Also, perhaps input should be an argument.

idesai:
Fixes #1658 

Fixed option descriptions in the is_check_input_options_ok routines
Changed the global variable from verifying_pubkey_path to verifying_pubkey_name_path.

### DIFF
--- a/tools/tpm2_policyauthorize.c
+++ b/tools/tpm2_policyauthorize.c
@@ -16,7 +16,7 @@ struct tpm2_policyauthorize_ctx {
    //File path for the policy qualifier data
    const char *qualifier_data_path;
    //File path for the verifying public key name
-   const char *verifying_pubkey_path;
+   const char *verifying_pubkey_name_path;
    //File path for the verification ticket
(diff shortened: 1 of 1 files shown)


tools/tpm2_policyauthorize.c
```diff
@@ -16,7 +16,7 @@ struct tpm2_policyauthorize_ctx {
    //File path for the policy qualifier data
    const char *qualifier_data_path;
    //File path for the verifying public key name
-   const char *verifying_pubkey_path;
+   const char *verifying_pubkey_name_path;
```
williamcroberts: You should have split this up in 2 patches, I won't nack it this time, but I will next time. 1 patch for the variable rename and 1 patch for the actual logging error fix.

PR #1682, opened 2019-08-21
DO NOT MERGE - demonstrate undetected error in unit tests

joholl:
Warnings (i.e. errors due to -Werror) can be introduced to unit tests and will not be detected by travis. This is a demonstration.


williamcroberts:
Fixed in: https://github.com/tpm2-software/tpm2-tools/pull/1684 I'll pull this after 4.0

Issue #1689, opened 2019-08-23
"tpm2_getcap commands" is giving truncated names for commands with underscores


mdempsky:
"tpm2_getcap commands" is printing truncated names like "nv", "pcr", or "ec" for commands like NV_Read, PCR_Extend, or EC_Ephemeral, respectively.

It looks like this is just because the table in tpm2_getcap.c is broken:

```
$ grep '{ TPM2_CC_.*_' tools/tpm2_getcap.c 
        { TPM2_CC_NV_UndefineSpaceSpecial, "nv" },
        { TPM2_CC_NV_UndefineSpace, "nv" },
        { TPM2_CC_NV_DefineSpace, "nv" },
        { TPM2_CC_PCR_Allocate, "pcr" },
        { TPM2_CC_PCR_SetAuthPolicy, "pcr" },
        { TPM2_CC_PP_Commands, "pp" },
        { TPM2_CC_NV_GlobalWriteLock, "nv" },
        { TPM2_CC_NV_Increment, "nv" },
        { TPM2_CC_NV_SetBits, "nv" },
        { TPM2_CC_NV_Extend, "nv" },
        { TPM2_CC_NV_Write, "nv" },
        { TPM2_CC_NV_WriteLock, "nv" },
        { TPM2_CC_NV_ChangeAuth, "nv" },
        { TPM2_CC_PCR_Event, "pcr" },
        { TPM2_CC_PCR_Reset, "pcr" },
        { TPM2_CC_NV_Read, "nv" },
        { TPM2_CC_NV_ReadLock, "nv" },
        { TPM2_CC_ECDH_ZGen, "ecdh" },
        { TPM2_CC_RSA_Decrypt, "rsa" },
        { TPM2_CC_HMAC_Start, "hmac" },
        { TPM2_CC_ECDH_KeyGen, "ecdh" },
        { TPM2_CC_NV_ReadPublic, "nv" },
        { TPM2_CC_RSA_Encrypt, "rsa" },
        { TPM2_CC_ECC_Parameters, "ecc" },
        { TPM2_CC_PCR_Read, "pcr" },
        { TPM2_CC_PCR_Extend, "pcr" },
        { TPM2_CC_PCR_SetAuthValue, "pcr" },
        { TPM2_CC_NV_Certify, "nv" },
        { TPM2_CC_ZGen_2Phase, "zgen" },
        { TPM2_CC_EC_Ephemeral, "ec" },
[...]

PR #1692, opened 2019-08-23
tpm2_getcap: fix invalid command string name output

williamcroberts:
The tpm2_getcap command had a map for converting command code
vlaue to textual strings. This map was invalid and also hard
to read. Also, a command code map already existed in
tpm2_cc_util.c, so rather than duplicating this map, coalesce
the map to that file.

Update the map in tpm2_cc_util.c so the command values match
what was being defined in the TSS common overviews document
and coordinating tss2_tpm2_types.h file. Add some additional
unit tests to validate the map and update integration tests.

Fixes: #1689


lib/tpm2_cc_util.c
```diff
@@ -3,6 +3,7 @@
 #include <string.h>
 
 #include "log.h"
+#include "tpm2_util.h"
```
williamcroberts: not needed

Issue #1694, opened 2019-08-23
tpm2_changeauth fails to change NV Index auth


idesai:
There are 2 issues:
[ ] In the tpm2_changeauth the routine object_needs_parent was not looking for the object handle to be an absolute value transient or persistent. Instead was looking for a non zero value and returning the boolean. This caused the tool to interpret that NV index handle range requires a parent.
[ ] The Esys call to nv change auth partially succeeds, in that, while the auth change occurs successfully there is an issue with the response buffer handling in esys. For this bug to close in its entirety it will require the esys bug be also closed as well https://github.com/tpm2-software/tpm2-tss/issues/1498

PR #1696, opened 2019-08-23
Fix change auth bug

idesai:
Partially solves #1694.

    The routine object_needs_parent was not looking for
    the object handle to be an absolute value transient
    or persistent. Instead was looking for a non zero
    value and returning the boolean. This caused the tool
    to interpret that NV index handle range requires a
    parent.


idesai:
> Partially solves #1694.
> 
> ```
> The routine object_needs_parent was not looking for
> the object handle to be an absolute value transient
> or persistent. Instead was looking for a non zero
> value and returning the boolean. This caused the tool
> to interpret that NV index handle range requires a
> parent.
> ```

The second issue needs to be fixed in esys https://github.com/tpm2-software/tpm2-tss/issues/1498

tools/tpm2_changeauth.c
```diff
@@ -112,8 +112,13 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
 
 static inline bool object_needs_parent(tpm2_loaded_object *obj) {
 
-    return (obj->handle & TPM2_HR_TRANSIENT)
-            || (obj->handle & TPM2_HR_PERSISTENT);
+    switch (obj->handle & TPM2_HR_RANGE_MASK) {
```
williamcroberts: Don't do this with a switch....your adding a bunch of jumps for something that could be optimized to just bit wise logic and likely all in the register file. I don't think the compiler will be able to optimize a switch as well. Just add the mask:

```C
TPM2_HC h = obj->handle & TPM2_HR_RANGE_MASK;
return (h & TPM2_HR_TRANSIENT)
            || (h & TPM2_HR_PERSISTENT);
```

Issue #1649, opened 2019-08-06
fix ecdsa signature


williamcroberts:
This bug originally stemmed from aPKCS11 bug:
https://github.com/tpm2-software/tpm2-pkcs11/issues/277

However, PKCS11 uses a different ECDSA signature format, it just concatenates R+S directly. For the tools, and interoperability with ossl, the ASN1 encoded version should be used. The code is currently doing that, but in a poor way. Below you'll see code suggested by OSSL. It's not tested and likely needs massaging, however it's a good point to start to refactor our signature code in the tools.

mdempsky:
It's probably worth clarifying that the relevant part of tpm2-software/tpm2-pkcs11#277 to tpm2-tools is [OpenSSL's recommended way to generate an ASN.1-encoded ECDSA signature](https://github.com/tpm2-software/tpm2-pkcs11/issues/277#issuecomment-518740965), which can replace the "hack" code in tpm2-tools/lib/tpm2_convert.c's extract_ecdsa:

https://github.com/tpm2-software/tpm2-tools/blob/758e4f980f2168a7a274b522c0ec3a4e0d5265b3/lib/tpm2_convert.c#L407-L416

However, the overall issue there (that PKCS#11 CKM_ECDSA requires raw R/S signatures without ASN.1-encoding) doesn't seem applicable to tpm2-tools, as it's not a PKCS#11 provider.

williamcroberts:
Got it thanks for the clarification. I see that their are two formats for the ECDSA signature, R+S and ASN1.

williamcroberts:
Actually, our code could be improved here to use what the OSSL folks suggested:

```C
ECDSA_SIG *create_ecdsa_sig(unsigned char *r, size_t rlen, unsigned char *s,
size_t slen)
{
    ECDSA_SIG *sig = ECDSA_SIG_new();
    if (sig == NULL)
        return NULL;
   sig->r = BN_bin2bn(r, rlen, NULL);
   sig->s = BN_bin2bn(s, slen, NULL);
   if (sig->r == NULL || sig->s == NULL) {
        ECDSA_SIG_free(sig);
        return NULL;
    }
    return sig;
}
```

PR #1702, opened 2019-08-26
tpm2_convert: fix ecdsa signature generation

williamcroberts:
No need to hand jam ASN1 structure if you just populate
an ECDSA_SIG object and work on that by setting the R
and S fields.

Fixes: #1649


williamcroberts:
I guess it doesn't like shorthand if(p) then do statements

Issue #1641, opened 2019-08-01
unit tests for pcr minilangauge


williamcroberts:
In reference to #1635 we should have some additional testing for pcr minilangauge

PR #1707, opened 2019-08-27
 test_tpm2_auth_util: add unit tests for PCR mini language 

diabonas:
These tests assert that `pcr:<pcr-spec>=<raw-pcr-file>` is split correctly, the tests whether `pcr_parse_selections` correctly parses the PCR specification and `tpm2_policy_build_pcr` builds the right policy from that are in [`test_pcr`](https://github.com/tpm2-software/tpm2-tools/blob/master/test/unit/test_pcr.c) and [`test_tpm2_policy`](https://github.com/tpm2-software/tpm2-tools/blob/master/test/unit/test_tpm2_policy.c), resp.

These two other unit tests could also use some love, though:
- test_pcr only whether using friendly names produces the same result as using raw algorithm specifiers, it could test whether the `TPML_PCR_SELECTION` structure is actually correct.
- test_tpm2_policy builds an example policy with only one bank, this didn't catch the second problem fixed in #1638. Adding an example with two banks would be helpful, but requires changes to the currently quite inflexible test structure.

Fixes: #1641


diabonas:
@williamcroberts I forgot to free the return value in the unit tests, fixed now.

PR #1720, opened 2019-08-28
Kill openssl 1.0.x

williamcroberts:


williamcroberts:
Lets wait until end of year

nullr0ute:
Any reason why you wouldn't leave this open to track it so it's not lost?

PR #1729, opened 2019-09-06
Modify tests to make them run on FreeBSD

tstruk:


test/integration/helpers.sh
```diff
@@ -138,7 +138,7 @@ tpm2_test_original_cwd=""
 tpm2_test_cwd=""
 function switch_to_test_dir() {
     tpm2_test_original_cwd=`pwd`;
-    tpm2_test_cwd=$(mktemp --directory --tmpdir=/tmp tpm2_test_XXXXXX)
+    tpm2_test_cwd=$(mktemp -d /tmp/tpm2_test_XXXXXX)
```
diabonas: While you're at it, we should respect the [POSIX `$TMPDIR` environment variable](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03) if it is set:
```
tpm2_test_cwd=$(mktemp -d "${TMPDIR:-/tmp}/tpm2_test_XXXXXX")

File: configure.ac [hunk 2]
```diff
@@ -114,7 +114,12 @@ AS_IF([test "x$enable_unit" != xno], [
...
     AS_IF([test $XXD = missing],
           [AC_MSG_ERROR([Required executable xxd not found, some system tests will fail!])])
-    enable_unit="$enable_unit (tpm2_abrmd: $tpm2_abrmd, tpm_server: $tpm_server, bash: $BASH_SHELL, python: $PYTHON, xxd: $XXD)"
+
+    AC_CHECK_PROG([NETSTAT], [netstat], found, missing)
+    AS_IF([test $NETSTAT = missing],
+          [AC_MSG_ERROR([Required executable netstat not found, some system tests will fail!])])
+
+    enable_unit="$enable_unit (tpm2_abrmd: $tpm2_abrmd, tpm_server: $tpm_server, bash: $BASH_SHELL, python: $PYTHON, xxd: $XXD, netstat: $NETSTAT)"
 ])
```
diabonas: netstat is deprecated in favour of ss (from the iproute2 package) [according to its man page](http://net-tools.sourceforge.net/man/netstat.8.html), so  we should use that instead.

test/integration/helpers.sh
```diff
@@ -153,6 +153,15 @@ tpm2_abrmd_pid=""
 tpm2_tcti_opts=""
 tpm2tools_tcti=""
 
+sock_tool="unknow"
+OS=$(uname)
+
+if [ "$OS" == "Linux" ]; then
+    sock_tool="netstat -lntp"
```
diabonas: As per the comment above, replace by
```
sock_tool="ss -lntp4"
```
The `-4` argument adds more robustness in case the simulator only manages to bind to the IPv6 interface (which doesn't help since the TCTI library only uses IPv4), see https://github.com/tpm2-software/tpm2-tss/pull/1234 for more information.

test/integration/helpers.sh
```diff
@@ -153,6 +153,15 @@ tpm2_abrmd_pid=""
 tpm2_tcti_opts=""
 tpm2tools_tcti=""
 
+sock_tool="unknow"
```
diabonas: Typo "unknow*n*"

test/integration/helpers.sh
```diff
@@ -153,6 +153,15 @@ tpm2_abrmd_pid=""
...
+OS=$(uname)
+
+if [ "$OS" == "Linux" ]; then
+    sock_tool="netstat -lntp"
+elif [ "$OS" == "FreeBSD" ]; then
+    sock_tool="sockstat -l"
```
diabonas: FreeBSD might have the same problem with the simulator occasionally binding only to IPv6, so using `sockstat -l4` would be more robust here as well (and is supported according to the [FreeBSD man page](https://www.freebsd.org/cgi/man.cgi?query=sockstat&manpath=FreeBSD+12.0-RELEASE+and+Ports)).

test/integration/helpers.sh
```diff
@@ -171,19 +180,18 @@ function start_sim() {
...
         echo "Attempting to start simulator on port: $tpm2_sim_port"
         $TPM2_SIM -port $tpm2_sim_port &
         tpm2_sim_pid=$!
         sleep 1
 
-        # Do not rely on whether netstat is present or not and directly fetch
-        # data in relevent /proc file
-        tpm2_sim_port_inode="$(awk -v port="$(printf ':%04X' "$tpm2_sim_port")" '$2 ~ port { print $10 }' /proc/net/tcp)"
-        tpm2_mssim_tcti_expected_port_inode="$(awk -v port="$(printf ':%04X' "$tpm2_mssim_tcti_expected_port")" '$2 ~ port { print $10 }' /proc/net/tcp)"
+        ${sock_tool} 2>/dev/null | grep ${TPM2_SIM} | grep ${tpm2_sim_port}
```
diabonas: I think it would be more robust to check whether that is actually the process with PID `$tpm2_sim_pid` and not any process named `$TPM2_SIM="tpm_server"`, otherwise there might be race conditions between different tests attempting to start their simulator on the same ports. Does the output of `sockstat` include the PID?

test/integration/helpers.sh
```diff
@@ -171,19 +180,18 @@ function start_sim() {
...
+        tpm2_sim_port_rc=$?
+        ${sock_tool} 2>/dev/null | grep ${TPM2_SIM} | grep ${tpm2_sim_cmd_port}
```
diabonas: Same problem here, that only checks whether any simulator has managed to bind to the port, not if it is the one we started.

tstruk:
@diabonas PR updated. Thanks for the feedback.

tstruk:
Thank you @diabonas

PR #1731, opened 2019-09-09
tpm2_identity_util: replace "encrypt_seed" with "share_secret"

### LINKED ISSUE #1730: tpm2_identity_util_encrypt_seed_with_public_key API is clumsy
tpm2_identity_util_encrypt_seed_with_public_key implements the "Secret Sharing" mechanism for RSA (TPM, Part 1, Section B.10) and ECC (Ibid., Section C.6).

However, the usage of the `seed` parameter differs depending on whether the public key is an RSA or ECC key. In particular, when sharing a secret with RSA, the *caller* has to provide random bytes (e.g., via RAND_bytes); but when sharing a secret with ECC, the *callee* (i.e., tpm2_identity_util_encrypt_seed_with_public_key) actually overwrites the seed with an ECDH-computed one (making the RAND_bytes call redundant).

I'd suggest changing the API to make the callee always responsible for initializing the seed parameter, and moving the RAND_bytes call into the RSA-specific secret sharing code. This would make the API more like a typical generic [key-encapsulation API](http://bench.cr.yp.to/call-kem.html).

mdempsky:
This commit changes the
tpm2_identity_util_encrypt_seed_with_public_key in two ways:

1. It changes the callee to always be responsible for generating the
seed, independent of the public key algorithm. Previously, the caller
had to generate a random seed for RSA keys, which was unnecessary for
ECC keys.

2. It renames "encrypt_seed" to "share_secret" to be more consistent
with the name that the TPM 2.0 Library Specification uses for this
mechanism.

Fixes #1730.

Signed-off-by: Matthew Dempsky <matthew@dempsky.org>

### DIFF
--- a/lib/tpm2_identity_util.c
+++ b/lib/tpm2_identity_util.c
@@ -6,6 +6,8 @@
 
 #include <tss2/tss2_mu.h>
 
+#include <openssl/rand.h>
+
 #include "log.h"
 #include "tpm2_alg_util.h"
 #include "tpm2_identity_util.h"
@@ -96,7 +98,7 @@ static TPM2_KEY_BITS get_pub_asym_key_bits(TPM2B_PUBLIC *public) {
     return 0;
 }
 
-static bool encrypt_seed_with_tpm2_rsa_public_key(TPM2B_DIGEST *protection_seed,
+static bool share_secret_with_tpm2_rsa_public_key(TPM2B_DIGEST *protection_seed,
         TPM2B_PUBLIC *parent_pub, unsigned char *label, int label_len,
         TPM2B_ENCRYPTED_SECRET *encrypted_protection_seed) {
     bool rval = false;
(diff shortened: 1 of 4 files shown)


mdempsky:
Looks like the GCC failure is unrelated:

```
../test/integration/tests/sign.sh: line 113: "1"<<16 + "0"<<8 + "2": syntax error: operand expected (error token is ""1"<<16 + "0"<<8 + "2"")
../test/integration/tests/sign.sh: line 113: "1"<<16 + "1"<<8 + "1": syntax error: operand expected (error token is ""1"<<16 + "1"<<8 + "1"")
../test/integration/tests/sign.sh: line 164: [: : integer expression expected
```

lib/tpm2_identity_util.c
```diff
@@ -116,6 +118,12 @@ static bool encrypt_seed_with_tpm2_rsa_public_key(TPM2B_DIGEST *protection_seed,
 
     TPMI_ALG_HASH parent_name_alg = parent_pub->publicArea.nameAlg;
 
+    /*
+     * RSA Secret Sharing uses a randomly generated seed (Part 1, B.10.3).
+     */
+    protection_seed->size = tpm2_alg_util_get_hash_size(parent_name_alg);
+    RAND_bytes(protection_seed->buffer, protection_seed->size);
```
williamcroberts: check the return code, 1 is success.

mdempsky: Done.

lib/tpm2_identity_util.c
```diff
@@ -116,6 +118,15 @@ static bool encrypt_seed_with_tpm2_rsa_public_key(TPM2B_DIGEST *protection_seed,
 
     TPMI_ALG_HASH parent_name_alg = parent_pub->publicArea.nameAlg;
 
+    /*
+     * RSA Secret Sharing uses a randomly generated seed (Part 1, B.10.3).
+     */
+    protection_seed->size = tpm2_alg_util_get_hash_size(parent_name_alg);
+    if (RAND_bytes(protection_seed->buffer, protection_seed->size) <= 0) {
```
idesai: we try not to hide conditionals. please assign the function output to an rc variable

mdempsky: Done.

PR #1732, opened 2019-09-09
Make tests more portable

tstruk:
Some portability changes required to run the integration tests on FreeBSD


test/integration/tests/pcrextend.sh
```diff
@@ -14,8 +14,7 @@ declare -A alg_hashes=(
 
 digests=""
 # test a single algorithm based on what is supported
-for alg in `tpm2_getcap pcrs | sed -r -e '1d' -e s/'\s+-\s+(\w+):.*'/'\1'/g`; do
-  alg=`echo $alg | cut -d\( -f 1-1`;
+for alg in `tpm2_getcap pcrs | awk {'print $2'} | cut -d: -f 1`; do
```
idesai: @tstruk for purpose of portability, should the cut be an awk instead?

tstruk: @idesai updated to use awk

PR #1739, opened 2019-09-16
expose some useful utility functions in a shared library

### LINKED ISSUE #1738: Consider providing a shared library of helper functions for building other TPM tools?
For our project we have built a program which performs a little sequence of TPM operations using some hardcoded constants and policy for our particular configuration. We reused a lot of the excellent code you have built for tpm2-tools. Essentially we have our own custom, product-specific TPM tool.

Currently we're building our program by just grabbing libcommon.a out of the tpm2-tools build process and then linking that into our program, but it's a bit messy.

Would you consider providing a shared library, say `libtpm2tools.so`, containing some of the reusable helper functions that are useful for building TPM-related tools externally?

danc86:
Fixes #1738

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,7 +14,7 @@ include src_vars.mk
 ACLOCAL_AMFLAGS = -I m4 --install
 
 INCLUDE_DIRS = -I$(top_srcdir)/tools -I$(top_srcdir)/lib
-LIB_COMMON := lib/libcommon.a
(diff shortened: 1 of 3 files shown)


williamcroberts:
I think this is going to need some discussion. We have talked about providing a stable API helper library, but I think exposing the the tools internal library as is, isn't the way to go. I never really intended for that stuff to be a stable API.

PR #1743, opened 2019-09-18
Check for shasum

williamcroberts:


configure.ac
```diff
@@ -93,33 +93,38 @@ AC_DEFUN([AC_PYTHON_MODULE],
...
           [AC_MSG_ERROR([Required executable python not found, some system tests will fail!])])
 
     AC_PYTHON_MODULE([yaml])
 
-    AC_CHECK_PROG([XXD], [xxd], found, missing)
+    AC_CHECK_PROG([XXD], [xxd], yes, no)
     AS_IF([test $XXD = missing],
           [AC_MSG_ERROR([Required executable xxd not found, some system tests will fail!])])
 
-    AC_CHECK_PROG([SS], [ss], found, missing)
```
tstruk: It is used:
https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/helpers.sh#L160

williamcroberts:
> Should we also check for mktemp?

yeah ill add that in and drop the SS patch.

PR #1746, opened 2019-09-20
Support saving TPMT_PUBLIC area with tpm2_readpublic

asac:


asac:
it seems testsuite is broken ... i cant see how my pull request could have broken that. are we sure current master tests cleanly?

diabonas:
You need to document the new option `--public` in the [man page](https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_nvreadpublic.1.md), that is why [CI is failing](https://travis-ci.org/tpm2-software/tpm2-tools/jobs/587408855#L4308).

tools/tpm2_readpublic.c
```diff
@@ -57,6 +58,13 @@ static tool_rc read_public_and_save(ESYS_CONTEXT *ectx) {
         }
     }
 
+    if (ctx.public_path)
```
williamcroberts: The `TPM2B_PUBLIC` is saved via the call to `tpm2_convert_pubkey_save()` with the format of `pubkey_format_tss`, which contains the `TPMT_PUBLIC` data, whats the reason that doesn't work for you?

idesai: +1 to what Bill said. You can specify this in the command line option with the option: 
--format tss

asac: i have split in two commits. first gives a bit of rational in the commit message.

Point is that TPM2B_PUBLIC is the TSS format. And that's great if you want to feed it back into TPM tools.

That said, we face the challenge that we need to calculate the "Name" we got from a tpm2_certify from a public area dump someone exported with tpm2-tools.

According to "Architecture Spec", a "Name" is built like: nameAlg || hash(TPMT_PUBLIC), so the TPM2B_PUBLIC won't work without further processing. Having a TPMT_PUBLIC export (which --public is about) would be convenient and also feels like it fits the spirit of this command (e.g. called readpublic).

For the name algorithm, check chapter 16 for persistent objects here: https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p55_Part1_Architecture_pubrev.pdf

Let me know your thoughts.

williamcroberts: > i have split in two commits. first gives a bit of rational in the commit message.
> 
> Point is that TPM2B_PUBLIC is the TSS format. And that's great if you want to feed it back into TPM tools.

The default format is a TPM2B_PUBLIC which is a size followed by a TPMT_PUBLIC, so you can just go past the size field and get the TPMT_PUBLIC structure. 

You could just use DD, a bit "hacky" but it would work.
```sh
tpm2_readpublic -c primary.ctx -o tpm2b_public.dat

# skip the UINT16 size field.
dd if=tpm2b_public.dat of=tpmt_public.dat skip=2 bs=1

cat tpmt_public.dat | openssl sha256
(stdin)= 914a1abc3f6b71166e4b8d3a02ee338524bb5b82a7c9e1a59f28dc97f7f2fd0c
```
Add the `00` and the algorithm `0b` and you get the full name:
`000b914a1abc3f6b71166e4b8d3a02ee338524bb5b82a7c9e1a59f28dc97f7f2fd0c`

> 
> That said, we face the challenge that we need to calculate the "Name" we got from a tpm2_certify from a public area dump someone exported with tpm2-tools.

If you need the name, it's provided by the YAML output dumped to stdout.

```sh
$ ./tools/tpm2_readpublic -c object.ctx
name: 000b914a1abc3f6b71166e4b8d3a02ee338524bb5b82a7c9e1a59f28dc97f7f2fd0c
qualified name: 000bdb4f165a0db93bc1cccbc5266f69ceaafe1bf96bcab11ca73d6f155a15a66591
name-alg:
  value: sha256
  raw: 0xb
```
> 
> According to "Architecture Spec", a "Name" is built like: nameAlg || hash(TPMT_PUBLIC), so the TPM2B_PUBLIC won't work without further processing. Having a TPMT_PUBLIC export (which --public is about) would be convenient and also feels like it fits the spirit of this command (e.g. called readpublic).

If you want the TPMT_PUBLIC I am not opposed to doing it, just opposed to your implementation. Their is already a -o for output and -f for format. So it would be adding a
format selector, something like tss-no-size so it saves the first structure embedded in the TPM2B_, in this case the TPMT.

Adding another option isn't the right approach.

> 
> For the name algorithm, check chapter 16 for persistent objects here: https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p55_Part1_Architecture_pubrev.pdf
> 
> Let me know your thoughts.

williamcroberts: Actually for the format string name, just use tpmt if you choose to add a format specified.

asac:
thanks for the feedback. i agree that tpmt is good as format string. will
update pr tomorrow using that (currently travelling).

asac:
@williamcroberts updated this PR following your input. Now we can do tpm2_readpublic -c HANDLE -f tpmt -o /tmp/public.tpmt and sha256sum of that file will match the suffix of the 'name' ... e.g.

```
tpm2_readpublic -c 0x81010001 -f tpmt -o /tmp/out.tpmt
name: 000bb7e1fe396dc93e5a14099717d5658c994123b8198d30d51b695e8fdafcd70872
[...]
# cat /tmp/out.tpmt | sha256sum 
b7e1fe396dc93e5a14099717d5658c994123b8198d30d51b695e8fdafcd70872  -
```
Let me know ....

asac:
hmm ... seems that travis is failing on fedora gcc in loadexternal test... can't see how this could be related .... Is that test usually stable?

idesai:
> hmm ... seems that travis is failing on fedora gcc in loadexternal test... can't see how this could be related .... Is that test usually stable?

@asac not sure why that happens, have kicked off the build again.

idesai:
> > hmm ... seems that travis is failing on fedora gcc in loadexternal test... can't see how this could be related .... Is that test usually stable?
> 
> @asac not sure why that happens, have kicked off the build again.

That went alright. Thanks for the PR.

lib/files.c
```diff
@@ -719,6 +719,9 @@ tool_rc files_save_ESYS_TR(ESYS_CONTEXT *ectx, ESYS_TR handle, const char *path)
 SAVE_TYPE(TPM2B_PUBLIC, public)
 LOAD_TYPE(TPM2B_PUBLIC, public)
 
+SAVE_TYPE(TPMT_PUBLIC, public_area)
```
idesai: @asac tpmt_public vs public_area, what are your thoughts for the resulting function naming files_save_public_area vs files_save_tpmt_public? 

I was thinking files_save_public and files_save_public_area seem to imply the same thing.

williamcroberts: Call it template

lib/files.h
```diff
@@ -141,6 +141,18 @@ tool_rc files_save_ESYS_TR(ESYS_CONTEXT *ectx, ESYS_TR handle, const char *path)
  */
 bool files_save_public(TPM2B_PUBLIC *public, const char *path);
 
+
+/**
+ * Serializes a TPMT_PUBLIC to the file path provided.
+ * @param public
```
williamcroberts: rename the param from public to template. the TPMT_PUBLIC is the public template details. Since we're using public usually for the TPM2B_PUBLIC, this will help prevent confusion when glancing at the code.

lib/files.h
```diff
@@ -141,6 +141,18 @@ tool_rc files_save_ESYS_TR(ESYS_CONTEXT *ectx, ESYS_TR handle, const char *path)
...
+ *  The TPMT_PUBLIC to save to disk.
+ * @param path
+ *  The path to save to.
+ * @return
+ *  true on success, false on error.
+ */
+bool files_save_public_area(TPMT_PUBLIC *public, const char *path);
```
williamcroberts: same here, rename this to: files_save_public_template()

williamcroberts:
merged on: https://github.com/tpm2-software/tpm2-tools/pull/1767

PR #1761, opened 2019-09-26
lib/openssl: pad ECC key params

### LINKED ISSUE #1759: Add padding for ECC keys
See: https://github.com/tpm2-software/tpm2-tss/issues/1518

joholl:
When reading an ECC key from a PEM file, one of it's parameters (private scalar, public coordinates x and y) can start with leading zero bytes. The `BN_num_bytes()` returns the bytes required to store such a parameter, e.g. for a `prime256v1` curve 31 bytes instead of 32 bytes if the parameter starts with `0x00`.

Additionally, `BN_bn2bin()` does not pad the parameter to its expected size which can lead to problems in the TPM. Hence, the key sizes need to be calculated properly and the ECC params padded.

Fixes #1759

### DIFF
--- a/lib/tpm2_openssl.c
+++ b/lib/tpm2_openssl.c
@@ -13,6 +13,30 @@
 #include "tpm2_alg_util.h"
 #include "tpm2_openssl.h"
 
+/* compatibility function for OpenSSL versions < 1.1.0 */
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+static int BN_bn2binpad(const BIGNUM *a, unsigned char *to, int tolen) {
+    int r;
+    int topad;
+    int islen;
+
+    islen = BN_num_bytes(a);
+
+    if (tolen < islen)
+        return -1;
+
+    topad = tolen - islen;
+
+    memset(to, 0x00, topad);
+    r = BN_bn2bin(a, to + topad);
+    if (r == 0) {
+        return -1;
+    }
+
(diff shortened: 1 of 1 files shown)


joholl:
There are other calls of `BN_num_bytes()` and `BN_bn2bin()` for RSA and AES which this commit does not deal with. We should tackle that, as well.

lib/tpm2_openssl.c
```diff
@@ -708,27 +708,27 @@ static bool load_public_ECC_from_key(EC_KEY *k, TPM2B_PUBLIC *pub) {
...
+    unsigned y_size = (EC_GROUP_get_degree(group) + 7) / 8;
     if (y_size > sizeof(Y->buffer)) {
         LOG_ERR("X coordinate is too big. Got %u expected less than or equal to"
                 " %zu", y_size, sizeof(Y->buffer));
         goto out;
     }
 
-    X->size = BN_bn2bin(x, X->buffer);
+    X->size = BN_bn2binpad(x, X->buffer, x_size);
     if (X->size != x_size) {
```
williamcroberts: This function isn't available on openssl < 1.1. you'll have to add a compat routine until we drop 1.0.2 support.

joholl: Thank you. I added a static compat function with `#if OPENSSL_VERSION_NUMBER < 0x10100000L`.

PR #1762, opened 2019-09-26
Policysigned - Add new policy tool

idesai:


idesai:
> make sure your updating the CHANGELOG for each tool you add

Thanks @williamcroberts. Will do.

idesai:
> You don't use the nonceTPM output, so why create the option?

It is to be used in the policysigned and policysecret

williamcroberts:
> > You don't use the nonceTPM output, so why create the option?
> 
> It is to be used in the policysigned and policysecret

I didn't see any test that uses it, where is it?

idesai:
> > > You don't use the nonceTPM output, so why create the option?
> > 
> > 
> > It is to be used in the policysigned and policysecret
> 
> I didn't see any test that uses it, where is it?

Ah I see the confusion :), In the works. I will start using the draft PR feature.

williamcroberts:
> > > > You don't use the nonceTPM output, so why create the option?
> > > 
> > > 
> > > It is to be used in the policysigned and policysecret
> > 
> > 
> > I didn't see any test that uses it, where is it?
> 
> Ah I see the confusion :), In the works. I will start using the draft PR feature.

Yeah when you introduce a feature, you need to introduce a consumer at the same time.

PR #1766, opened 2019-10-02
tpm2_startauthsession: Add option to specify session nonce size

idesai:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### next
 
+* tpm2_startauthsession:
+  - Added an option, -s or --nonce-size, to specify the maximum session nonce
+  size for the session started with tpm2_startauthsession.
+
 * misc:
   - support "tpmt" as a public key output format that only saves the TPMT structure.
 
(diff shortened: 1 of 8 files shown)


idesai:
> Why do I need this, ESAPI selects a nonce if not specified. Do we need to expose this to users? The less we expose the simpler the tool is.

This decides the size of nonceTPM which is an input to the policysigned tool.

idesai:
> > Why do I need this, ESAPI selects a nonce if not specified. Do we need to expose this to users? The less we expose the simpler the tool is.
> 
> This decides the size of nonceTPM which is an input to the policysigned tool.

It is an optional argument used in conjunction with policysigned policysecret etc.

idesai:
> Why do I need this, ESAPI selects a nonce if not specified. Do we need to expose this to users? The less we expose the simpler the tool is.

@williamcroberts, on another thought, all i need is the nonceTPM. I can still let ESAPI decide the nonce size. I will drop the commits for specifying the nonce size and keep the one to return the nonceTPM.

PR #1776, opened 2019-10-08
Map TPMI_RH_HIERARCHY type to Esys_TR type

tstruk:
To continue discussion started in https://github.com/tpm2-software/tpm2-tss/pull/1531 the tools can handle the API change by some internal remapping like this.

### DIFF
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -8,6 +8,7 @@
 #include "tpm2.h"
 #include "tpm2_auth_util.h"
 #include "tpm2_tool.h"
+#include "config.h"
 
 #define TPM2_ERROR_TSS2_RC_ERROR_MASK 0xFFFF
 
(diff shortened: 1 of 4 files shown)


lib/tpm2.c
```diff
@@ -591,13 +592,36 @@ tool_rc tpm2_evictcontrol(ESYS_CONTEXT *esys_context,
...
+        case TPM2_RH_ENDORSEMENT:
+            return ESYS_TR_RH_ENDORSEMENT;
+        case ESYS_TR_RH_PLATFORM:
+            return ESYS_TR_RH_PLATFORM;
+        default:
+            return TSS2_ESYS_RC_BAD_VALUE;
+    }
+#elif defined(ESYS_2_3)
+    return hierarchy;
+#else
```
idesai: @tstruk does this branch imply, versions lesser than 2.3? Does it mean also that versions prior to 2.3 did not support ESAPI?

lib/tpm2.h
```diff
@@ -340,4 +340,6 @@ tool_rc tpm2_policy_nv_written(ESYS_CONTEXT *esys_context,
         ESYS_TR policy_session, ESYS_TR shandle1, ESYS_TR shandle2,
         ESYS_TR shandle3, TPMI_YES_NO written_set);
 
+uint32_t fix_esys_hierarchy(TPMI_RH_HIERARCHY hierarchy);
```
idesai: @tstruk is this going to be an errata? if so, should this function be renamed as such with an errata number?

lib/tpm2.c
tstruk: It is for tools >= 4.0, which require ESAPI >= 2.3.0.
This is inline with the configure check.

lib/tpm2.h
tstruk: My understanding is that the errata mechanism is targeted for TPM spec issues that are fixed by errata. The ESAPI issue that this change fixes is not TPM spec related, but it is the implementation issue, so no errata number exists for it.

idesai: > My understanding is that the errata mechanism is targeted for TPM spec issues that are fixed by errata. The ESAPI issue that this change fixes is not TPM spec related, but it is the implementation issue, so no errata number exists for it.

I see. LGTM.

lib/tpm2.c
```diff
@@ -591,13 +592,36 @@ tool_rc tpm2_evictcontrol(ESYS_CONTEXT *esys_context,
...
+{
+#if defined(ESYS_3_0)
+    switch (hierarchy) {
+        case TPM2_RH_NULL:
+            return ESYS_TR_RH_NULL;
+        case TPM2_RH_OWNER:
+            return ESYS_TR_RH_OWNER;
+        case TPM2_RH_ENDORSEMENT:
+            return ESYS_TR_RH_ENDORSEMENT;
+        case ESYS_TR_RH_PLATFORM:
```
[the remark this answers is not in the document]
tstruk: Yes.

lib/tpm2.h
idesai: @tstruk, I think we should leave a comment explaining the necessity for this function and reference the tpm2-tss issue#

tstruk: Agree. Will send an update soon.

lib/tpm2.c
```diff
@@ -591,13 +592,36 @@ tool_rc tpm2_evictcontrol(ESYS_CONTEXT *esys_context,
     return tool_rc_success;
 }
 
+uint32_t fix_esys_hierarchy(TPMI_RH_HIERARCHY hierarchy)
```
williamcroberts: Why a uint32_t and not one of the other typedefs?

tstruk: Because it can return either TPMI_RH_HIERARCHY or ESYS_TR, which are both typedefs of uint32_t anyway.

PR #1780, opened 2019-10-11
Portable tests

joholl:
Some of the commands fail since certain requirements are not met, e.g. ECC curves or algorithms are not supported etc. In this case the tests in question should skip such algorithms or have the test return `77` (SKIP) altogether.

To make the tests more portable, I introduce following changes:

- Introduce  new tool `tpm2_getname` which returns the name of a TPM object. This tool is used in tests later. Any feedback is welcome!
- Use `tpm2_getname` in tests as an alternative to `tpm2_encryptdecrypt` if `tpm2_encryptdecrypt` is not supported
- Add `try` function to `helpers.sh` which is used to handle expected errors (e.g. algorithm not supported). In this case, sub-tests are skipped.
- Skip tests `toggle_options` if required files are not present (e.g. directory `man`)
- Replace `tpm2_encryptdecrypt` by `tpm2_sign` in test `abrmd_policyauthvalue`

Note:
I introduced a new tool `tpm_getname`. This tool might not be necessary to archieve what I wanted, but I did not find a way to determine an object's name using the tools. Please review this new tool in particular. Thanks!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -511,6 +511,14 @@ AM_TESTS_ENVIRONMENT =	\
 	SRCDIR=$(abs_srcdir); export SRCDIR; BUILDDIR=$(abs_builddir); export BUILDDIR; \
 	export TPM2_TOOLS_TEST_FIXTURES;
 
+if PERSISTENT
+    AM_TESTS_ENVIRONMENT +=	\
+        TPM2TOOLS_TEST_PERSISTENT=true; export TPM2TOOLS_TEST_PERSISTENT;
+else
+    AM_TESTS_ENVIRONMENT +=	\
+        TPM2TOOLS_TEST_PERSISTENT=false; export TPM2TOOLS_TEST_PERSISTENT;
+endif
+
 SH_LOG_COMPILER = dbus-run-session bash
 AM_SH_LOG_FLAGS = --
 endif
--- a/test/integration/tests/abrmd_policypassword.sh
+++ b/test/integration/tests/abrmd_policypassword.sh
@@ -10,13 +10,12 @@ key_ctx=key.ctx
 key_pub=key.pub
 key_priv=key.priv
 plain_txt=plain.txt
-encrypted_txt=enc.txt
-decrypted_txt=dec.txt
+signature_txt=signature.txt
 testpswd=testpswd
 
 cleanup() {
     rm -f $policypassword $session_ctx $o_policy_digest $primary_key_ctx \
-    $key_ctx $key_pub $key_priv $plain_txt $encrypted_txt $decrypted_txt
+    $key_ctx $key_pub $key_priv $plain_txt $signature_txt
 
     tpm2_flushcontext $session_ctx 2>/dev/null || true
 
@@ -39,19 +38,18 @@ rm $session_ctx
 
 tpm2_createprimary -C o -c $primary_key_ctx
 
-tpm2_create -g sha256 -G aes -u $key_pub -r $key_priv -C $primary_key_ctx \
+tpm2_create -g sha256 -G ecc -u $key_pub -r $key_priv -C $primary_key_ctx \
 -L $policypassword -p $testpswd
 
 tpm2_load -C $primary_key_ctx -u $key_pub -r $key_priv -c $key_ctx
-tpm2_encryptdecrypt -c $key_ctx -o $encrypted_txt -p $testpswd $plain_txt
+tpm2_sign -c $key_ctx -p $testpswd -o $signature_txt $plain_txt
+tpm2_verifysignature -c key.ctx -m $plain_txt -s $signature_txt
 
 tpm2_startauthsession --policy-session -S $session_ctx
 tpm2_policypassword -S $session_ctx -L $policypassword
-tpm2_encryptdecrypt -c $key_ctx -o $decrypted_txt -d \
--p session:$session_ctx+$testpswd $encrypted_txt
+tpm2_sign -c $key_ctx -p session:$session_ctx+$testpswd -o $signature_txt $plain_txt
+tpm2_verifysignature -c key.ctx -m $plain_txt -s $signature_txt
 tpm2_flushcontext $session_ctx
 rm $session_ctx
-
-diff $plain_txt $decrypted_txt
 
 exit 0
--- a/test/integration/tests/create.sh
+++ b/test/integration/tests/create.sh
@@ -49,28 +49,37 @@ test "$policy_orig" == "$policy_new"
 #
 # Test the extended format specifiers
 #
-tpm2_create -Q -C context.out -g sha256 -G aes256cbc -u key.pub -r key.priv
+# aes128cfb (mandatory for PCClient TPMs)
+tpm2_create -Q -C context.out -g sha256 -G aes128cfb -u key.pub -r key.priv
 tpm2_load -Q -C context.out -u key.pub -r key.priv -c key1.ctx
 tpm2_readpublic -c key1.ctx > out.yaml
 keybits=$(yaml_get_kv out.yaml "sym-keybits")
 mode=$(yaml_get_kv out.yaml "sym-mode" "value")
-test "$keybits" -eq "256"
-test "$mode" == "cbc"
-
-tpm2_create -Q -C context.out -g sha256 -G aes128ofb -u key.pub -r key.priv
-tpm2_load -Q -C context.out -u key.pub -r key.priv -c key2.ctx
-tpm2_readpublic -c key2.ctx > out.yaml
-keybits=$(yaml_get_kv out.yaml "sym-keybits")
-mode=$(yaml_get_kv out.yaml "sym-mode" "value")
 test "$keybits" -eq "128"
-test "$mode" == "ofb"
+test "$mode" == "cfb"
+
+# aes256ofb (if supported)
+if is_alg_supported aes256ofb; then
+  mode="$(populate_alg_modes $strongest_aes | head -n1)" # e.g. aes128ecb
+  tpm2_create -Q -C context.out -g sha256 -G aes256ofb -u key.pub -r key.priv
+  tpm2_load -Q -C context.out -u key.pub -r key.priv -c key2.ctx
+  tpm2_readpublic -c key2.ctx > out.yaml
+  keybits=$(yaml_get_kv out.yaml "sym-keybits")
+  mode=$(yaml_get_kv out.yaml "sym-mode" "value")
+  test "$keybits" -eq "256"
+  test "$mode" == "ofb"
+fi
+
+exit 0
 
 #
 # Test scheme support
 #
 
 for alg in "rsa1024:rsaes" "ecc384:ecdaa4-sha256"; do
-  tpm2_create -Q -C context.out -g sha256 -G "$alg" -u key.pub -r key.priv
+  if is_alg_supported $alg; then
+    tpm2_create -Q -C context.out -g sha256 -G "$alg" -u key.pub -r key.priv
+  fi
 done
 
 # Test createloaded support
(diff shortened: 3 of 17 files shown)


idesai:
> Some of the commands fail since certain requirements are not met, e.g. ECC curves or algorithms are not supported etc. In this case the tests in question should skip such algorithms or have the test return `77` (SKIP) altogether.
> 
> To make the tests more portable, I introduce following changes:
> 
> * Introduce  new tool `tpm2_getname` which returns the name of a TPM object. This tool is used in tests later. Any feedback is welcome!
> * Use `tpm2_getname` in tests as an alternative to `tpm2_encryptdecrypt` if `tpm2_encryptdecrypt` is not supported
> * Add `try` function to `helpers.sh` which is used to handle expected errors (e.g. algorithm not supported). In this case, sub-tests are skipped.
> * Skip tests `toggle_options` if required files are not present (e.g. directory `man`)
> * Replace `tpm2_encryptdecrypt` by `tpm2_sign` in test `abrmd_policyauthvalue`
> 
> Note:
> I introduced a new tool `tpm_getname`. This tool might not be necessary to archieve what I wanted, but I did not find a way to determine an object's name using the tools. Please review this new tool in particular. Thanks!

Questions - 
1. How is name an alternative to encryptdecrypt ?
2. tpm2_readpublic, tpm2_loadexternal, tpm2_load give you the object names including transient objects. 
3. Names of permanent entities like NV indices, PCRs, hierarchies, etc. is the handle value itself.
4. Actually, I like toggle_options to complain if a man page is not supplied with the tool. It is kind of its purpose, isnt it?
5. What is the need for tpm2_encryptdecrypt being replaced with tpm2_sign? I am suspecting you are changing tests so they run on a real TPM. I don't think we should go down that path. Everything on tpm2-software repositories is run against the simulator and I'd like to keep the tools consistent as well.

joholl:
Actually, I'm not sure if getname is a proper way to test if load() succeeded (alternatively to encryptdecrypt). Do you have any idea?

Isn't the purpose of toggle options to test the consistency of the code/man pages? If the man pages are not present, IMHO this test should be skipped (the man pages could be consistent, we cannot tell).

You are right, as the title of this PR implies, I want to enable skipping certain tests if features are not supported on a hardware/pcclient-compliant TPM. As you saw in my PR, I did my best to not remove any test code for the simulator. Note that I'm not asking you to support/advertise the tests being able to run on a hardware TPM, though. IMO it does not hurt if the tests are able to skip testing unsupported functionality.

williamcroberts:
> Some of the commands fail since certain requirements are not met, e.g. ECC curves or algorithms are not supported etc. In this case the tests in question should skip such algorithms or have the test return `77` (SKIP) altogether.

The simulator has a defined set of what it supports... all this stuff is really for running against hardware, which I don't want to support. I don't want to worry about the test suite breaking your TPM. 

> 
> To make the tests more portable, I introduce following changes:
> 
> * Introduce  new tool `tpm2_getname` which returns the name of a TPM object. This tool is used in tests later. Any feedback is welcome!
> * Use `tpm2_getname` in tests as an alternative to `tpm2_encryptdecrypt` if `tpm2_encryptdecrypt` is not supported
> * Add `try` function to `helpers.sh` which is used to handle expected errors (e.g. algorithm not supported). In this case, sub-tests are skipped.
> * Skip tests `toggle_options` if required files are not present (e.g. directory `man`)
> * Replace `tpm2_encryptdecrypt` by `tpm2_sign` in test `abrmd_policyauthvalue`
> 
> Note:
> I introduced a new tool `tpm_getname`. This tool might not be necessary to archieve what I wanted, but I did not find a way to determine an object's name using the tools. Please review this new tool in 
particular. Thanks!

tpm2_readpublic and tpm2_create and tpm2_createprimary all output the name, is that not good enough?

williamcroberts:
> Actually, I'm not sure if getname is a proper way to test if load() succeeded (alternatively to encryptdecrypt). Do you have any idea?

If tpm2_load fails the command will exit with a non-zero status. What exactly do you want to verify, that the key is indeed created by the TPM and not a MITM?

> 
> Isn't the purpose of toggle options to test the consistency of the code/man pages? If the man pages are not present, IMHO this test should be skipped (the man pages could be consistent, we cannot tell).

IIRC, It tests against the markdown files, not against built or installed manpages.

> 
> You are right, as the title of this PR implies, I want to enable skipping certain tests if features are not supported on a hardware/pcclient-compliant TPM. As you saw in my PR, I did my best to not remove any test code for the simulator. Note that I'm not asking you to support/advertise the tests being able to run on a hardware TPM, though. IMO it does not hurt if the tests are able to skip testing unsupported functionality.

What if something changes with the simulator or the CI and we start skipping tests? The CI would still pass. We would need to constantly verify nothing was skipped. I'm really leery of hardware testing with the test harness.

williamcroberts:
So whats our general consensus on this, im leaning towards a no on this. @idesai thoughts? @joholl how badly do you need this?

PeterHuewe:
In my opinion, it would be beneficial to have the ability to skip tests if the underlying device test does not support a certain command. 
This is really not a dramatic change and would help us, and thus the project a lot.

I definitely understand your take on we don't want to run or support tests on hardware, this is respected and understood. If I have a tpm with the function set of the simulator, the tests could already break it, so the skipping does not change much here.

However, to be quite honest, we spend a lot of time and effort on verifying whether the tss and the tools work without issues on our TPMs, so nothing will break in the field.

In the future the tss and the tools will be used against real tpms, and the bug reports will end up here.
This is something we want to prevent.

It would make our life way easier, so we can run the testsuites unmodified on our tpms here (which we have no problem if they break, we have unlimited supply :) and uncover **real** problems as soon as they come up. (We also sometimes find issues with the reference code found in the simulator).

Just think about it as extensive additional testing, which comes with a very small price tag.

For the CI, I think it can be configured that it will mark skipped tests.

AndreasFuchsSIT:
IMHO, using EncryptDecrypt is not the nicest anyways. It just makes it somewhat harder if you want to test an actual TPM without benefit.
Thus, I think, HMAC or UnSeal would be better candidates for testing...

Makefile.am
```diff
@@ -47,6 +47,7 @@ bin_PROGRAMS = \
     tools/tpm2_dictionarylockout \
     tools/tpm2_duplicate \
     tools/tpm2_getcap \
+    tools/tpm2_getname \
```
williamcroberts: tpm2_readpublic doesn't work?

test/integration/tests/abrmd_policypassword.sh
```diff
@@ -32,6 +35,12 @@ cleanup "no-shutdown"
 
 echo "plaintext" > $plain_txt
 
+
+encryptdecrypt_supp=false
+if [ ! -z "$(tpm2_getcap commands | grep 'encryptdecrypt:')" ]; then
+  encryptdecrypt_supp=true
+fi
```
williamcroberts: Use a different command. Change the key type to rsa and do a sign and verify operation.

test/integration/tests/create.sh
```diff
@@ -27,9 +27,20 @@ tpm2_createprimary -Q -C o -g sha1 -G rsa -c context.out
 # values.
 for gAlg in `populate_hash_algs`; do
     for GAlg in rsa keyedhash ecc aes; do
-        echo "tpm2_create -Q -C context.out -g $gAlg -G $GAlg -u key.pub \
+        echo "tpm2_create -Q -C context.out -g $gAlg -G $GAlg -u key.pub
         -r key.priv"
-        tpm2_create -Q -C context.out -g $gAlg -G $GAlg -u key.pub -r key.priv
+
+        # Some TPMs might not be able to create aes256 keys (error 0x000002c4)
```
williamcroberts: use getcap and look up the size of the encrypted buffer. There is a field in the output that actually relates.

test/integration/tests/toggle_options.sh
```diff
@@ -11,10 +11,22 @@ if [ "$OS" == "FreeBSD" ]; then
     exit 0
 fi
 
-srcdir="$(readlink -e "$(dirname "$0")")"
-toolsdir="$(readlink -e "${srcdir}"/../../../tools)"
-mandir="$(readlink -e "${srcdir}"/../../../man)"
+srcdir="$(readlink -m "$(dirname "$0")")"
```
williamcroberts: Why would any of these be missing, you're running the test scripts on the host? Can't you just modify TESTS variable to include what you want to run?

joholl:
Thank you for your feedback @williamcroberts, I'll rework this PR as soon as I can.

joholl:
I have done a complete rework of this PR. I did my best to keep it simple.

### New Functions `is_alg_supported` and `is_cmd_supported `

I added two new functions `is_alg_supported `and `is_alg_supported`  to helpers.sh. Most importantly: for the **simulator** (both mssim and swtpm), this function will **error if the alg/cmd is unsupported**. This means testing cannot silently break.

`EncryptDecrypt` was replaced by `SignVerify` (asymmetric) and `ReadPublic` (symmetric). The latter only applies to non-simulator TPMs that do not support `EncryptDecrypt`.

### Add cleanup

In some cases an NV index could remain. Some other tests did not reset the passwords set via `tpm2_changeauth`. Added additional cleanup.

### Test fore relative change of NV counter

The NV counter was incremented a bunch of times and compared to absolute numbers. Changed to relative numbers, i.e. "is the index now `a+1`?"

### Add configure option --disable-persistent

Some tests might leave the TPM in a dirty state since we are not able to reset. These tests can now be skipped via passing `--disable-persistent` or setting `TPM2TOOLS_TEST_PERSISTENT=false`. Only if `--disable-persistent` was given or if `TPM2TOOLS_TEST_PERSISTENT` is set to exactly `false` (and nothing else), the tests will be skipped. By default, they will be executed.

So far, only the test `hierarchycontrol.sh` is affected.

### Add further utility functions

For your convenience, I added utility functions to `helpers.sh` which behave like `populate_hash_algs`. They use `tpm2_testparms` and do not work for hashes.

- `populate_alg_lengths`: call e.g. with `aes` and get a LF-separated list:   
```
aes128
aes256
```
- `weakest_alg`:  call e.g. with `aes` and get `aes128`
- `strongest_alg`: call e.g. with `aes` and get `aes256`
- `populate_alg_modes`: call e.g. with `aes256` and get a LF-separated list:   
```
aes256ctr
aes256ofb
aes256cbc
aes256cfb
aes256ecb
```

These do not work solely for `aes`, but also for other non-hash algorithms. If you do *not* want these functions added, I can remove this commit.

joholl:
Travis fails to the missing tpm2-abrmd tag `2.3.0-rc0`. It's [2.3.0](https://github.com/tpm2-software/tpm2-abrmd/releases/tag/2.3.0) now.

Makefile.am
joholl: Used tpm2_readpublic for non-simulator TPMs which do not support Tpm2_encryptdecrypt now.

test/integration/tests/abrmd_policypassword.sh
joholl: Done.

test/integration/tests/create.sh
joholl: I added a function is_alg_supported based on tpm2_testparms instead. It can be used for different types of algorithms such as aes and rsa.

test/integration/tests/toggle_options.sh
joholl: You're right, in a normally cloned repo, they should be present. Removed this change.

joholl:
@williamcroberts ping

joholl:
Sure thing. I'll get to it as soon as I can.

joholl:
Rebased and pushed

idesai:
@williamcroberts, @joholl thanks! I will stage this for 4.3 release.

Issue #1782, opened 2019-10-15
Checkquote no longer outputs proper YAML


jetwhiz:
Note that the commit here breaks YAML output (particularly when outputting PCR data): https://github.com/tpm2-software/tpm2-tools/commit/350fae1f410e95b5f9a21059e7c6d1e279857801#diff-6e12f2d1da3a30287442b0cbb76b67eaL87

Now the signature buffer is dumped out to STDOUT without the necessary YAML key field (removed in the commit above).  

https://github.com/tpm2-software/tpm2-tools/blob/ef3bd3aa2c190a8d2d4b38ea4f1442dcc994cc69/tools/misc/tpm2_checkquote.c#L80

williamcroberts:
whoop sorry a little overzealous with that cleanup

williamcroberts:
Fixed in #1784 and I am working on a 4.0.1 release candidate with this fix.

jetwhiz:
Thanks, @williamcroberts !

PR #1783, opened 2019-10-15
Add YAML key for sigBuffer to fix checkquote

jetwhiz:
If PCRs provided then they will print with 'pcr' key
  Not having the 'sigBuffer' key here breaks YAML formatting

One potential way to address issue https://github.com/tpm2-software/tpm2-tools/issues/1782


williamcroberts:
See #1784

PR #1799, opened 2019-10-22
tpm2_policycountertimer: Add tool to enable command TPM2_PolicyCountertimer

### LINKED ISSUE #858: Implement policycountertimer
Implement a tool to use the policycountertimer command.

idesai:
Fixes #858 

### DIFF
--- a/man/tpm2_policycountertimer.1.md
+++ b/man/tpm2_policycountertimer.1.md
@@ -1,0 +1,142 @@
+% tpm2_policycountertimer(1) tpm2-tools | General Commands Manual
+
+# NAME
+
+**tpm2_policycountertimer**(1) - Enables policy authorization by evaluating the
+comparison operation on the TPM parameters time, clock, reset count, restart
+count and TPM clock safe flag.
+
+# SYNOPSIS
+
+**tpm2_policycountertimer** [*OPTIONS*] [*ARGUMENT*]
+
+# DESCRIPTION
+
+**tpm2_policycountertimer**(1) - Enables policy authorization by evaluating the
+comparison operation on the TPM parameters time, clock, reset count, restart
+count and TPM clock safe flag. If time/clock, it is input as milliseconds value.
+The parameter and the value is given as a command line argument as below:
+```
+tpm2_policycountertimer -S session.ctx safe
+tpm2_policycountertimer -S session.ctx clock=<N ms>
+tpm2_policycountertimer -S session.ctx time=<N ms>
+tpm2_policycountertimer -S session.ctx resets=<N>
+tpm2_policycountertimer -S session.ctx restarts=<N>
+```
+By default comparison tests for equality and also by default it tests for time.
+
+# OPTIONS
+
+  * **-L**, **\--policy**=_FILE_:
+
+    File to save the policy digest.
+
+  * **-S**, **\--session**=_FILE_:
+
+    The policy session file generated via the **-S** option to
+    **tpm2_startauthsession** or saved off of a previous tool run.
+
+  * **--eq**
+
+    if value of current time in the TPM = value of specified input time.
+
+  * **--neq**
+
+   if value of current time in the TPM != value of specified input time.
+
+  * **--sgt**
+
+   if signed value of current time in the TPM > signed value of specified input
+   time.
+
+  * **--ugt**
+
+   if unsigned value of current time in the TPM > unsigned value of specified
+   input time.
+
+  * **--slt**
+
+   if signed value of current time in the TPM < signed value of specified
+   input time.
+
+  * **--ult**
+
+   if unsigned value of current time in the TPM < unsigned value of specified
+   input time.
+
+  * **--sge**
+
+   if signed value of current time in the TPM >= signed value of specified
+   input time.
+
+  * **--uge**
+
+   if unsigned value of current time in the TPM >= unsigned value of specified
+   input time.
+
+  * **--sle**
+
+   if signed value of current time in the TPM <= unsigned value of specified
+   input time.
+
+  * **--ule**
+
+   if unsigned value of current time in the TPM <= unsigned value of specified
+   input time.
+
+  * **--bs**
+
+    if all bits set in value of current time in the TPM are set in value of
+    specified input time.
+
+  * **--bc**
+
+    if all bits set in value of current time in the TPM are clear in value of
+    specified input time.
+
+## References
+
+[common options](common/options.md) collection of common options that provide
+information many users may expect.
+
+[common tcti options](common/tcti.md) collection of options used to configure
+the various known TCTI modules.
+
+# EXAMPLES
+
+Create a sealing object with an authorization policy that evaluates only for
+first minute of TPM restart.
+
+## Create the policy and the sealing object
+
+```bash
+tpm2_startauthsession -S session.ctx
+
+tpm2_policycountertimer -S session.ctx -L policy.countertimer --ult 60000
+
+tpm2_flushcontext session.ctx
+
+tpm2_createprimary -C o -c prim.ctx -Q
+
+echo "SUPERSECRET" | \
+tpm2_create -Q -u key.pub -r key.priv -i- -C prim.ctx \
+-L policy.countertimer -a "fixedtpm|fixedparent" -c key.ctx
+```
+
+## Unsealing should work in the first minute after TPM restart
+
+```bash
+tpm2_startauthsession -S session.ctx --policy-session
+
+tpm2_policycountertimer -S session.ctx --ult 60000
+
+tpm2_unseal -c key.ctx -p session:session.ctx
+
+tpm2_flushcontext session.ctx
+```
+
+[returns](common/returns.md)
+
+[limitations](common/policy-limitations.md)
+
+[footer](common/footer.md)
--- a/test/integration/tests/abrmd_policycountertimer.sh
+++ b/test/integration/tests/abrmd_policycountertimer.sh
@@ -1,0 +1,87 @@
+# SPDX-License-Identifier: BSD-3-Clause
+
+source helpers.sh
+
+cleanup() {
+
+  rm -f session.ctx prim.ctx key.pub key.priv key.ctx policy.countertimer.minute
+
+  if [ "$1" != "no-shut-down" ]; then
+     shut_down
+  fi
+}
+
+trap cleanup EXIT
+
+start_up
+
+cleanup "no-shut-down"
+
+tpm2_clear
+
+#
+# Create a sealing object with a policy that evaluates for first minute after
+# TPM restart. NOTE the time is 60000 milliseconds.
+#
+tpm2_startauthsession -S session.ctx
+
+tpm2_policycountertimer -S session.ctx -L policy.countertimer.minute --ult \
+60000
+
+tpm2_flushcontext session.ctx
+
+tpm2_createprimary -C o -c prim.ctx -Q
+
+echo "SUPERSECRET" | \
+tpm2_create -Q -u key.pub -r key.priv -i- -C prim.ctx \
+-L policy.countertimer.minute -a "fixedtpm|fixedparent" -c key.ctx
+
+#
+# ASSUMING 1 minute hasn't elapsed since clear, Try unseal in the first minute
+# -- Should pass
+#
+tpm2_startauthsession -S session.ctx --policy-session
+
+tpm2_policycountertimer -S session.ctx -L policy.countertimer.minute --ult \
+60000
+
+tpm2_unseal -c key.ctx -p session:session.ctx
+
+tpm2_flushcontext session.ctx
+
+#
+# Test if a policycountertimer evaluates with the clock
+#
+tpm2_clear
+tpm2_startauthsession -S session.ctx --policy-session
+tpm2_policycountertimer -S session.ctx --ult clock=60000
+tpm2_flushcontext session.ctx
+
+#
+# Test if a policycountertimer evaluates with the TPM clocks safe flag
+# Assuming the safe flag is set since with just started and cleared the TPM
+#
+tpm2_clear
+tpm2_startauthsession -S session.ctx --policy-session
+tpm2_policycountertimer -S session.ctx safe
+tpm2_flushcontext session.ctx
+
+#
+# Test if a policycountertimer evaluates with the TPM reset count
+# Assuming the value is zero since we just cleared the TPM
+#
+tpm2_clear
+tpm2_startauthsession -S session.ctx --policy-session
+tpm2_policycountertimer -S session.ctx resets=0
+tpm2_flushcontext session.ctx
+
+#
+# Test if a policycountertimer evaluates with the TPM restart count
+# Assuming the value is zero since we just cleared the TPM
+#
+tpm2_clear
+tpm2_startauthsession -S session.ctx --policy-session
+tpm2_policycountertimer -S session.ctx restarts=0
+tpm2_flushcontext session.ctx
+
+exit 0
--- a/tools/tpm2_policycountertimer.c
+++ b/tools/tpm2_policycountertimer.c
@@ -1,0 +1,255 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+
+#include <stddef.h>
+#include <stdlib.h>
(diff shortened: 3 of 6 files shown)


tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,198 @@
...
+};
+
+static bool on_arg(int argc, char **argv) {
+
+    if (argc > 1) {
+        LOG_ERR("Specify one argument - The Comparison-Operation");
+        return false;
+    }
+
+        if (!strcmp(argv[0], "eq")) {
```
williamcroberts: indent issues

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,198 @@
...
+    switch (key) {
+    case 'L':
+        ctx.policy_digest_path = value;
+        break;
+    case 'S':
+        ctx.session_path = value;
+        break;
+    case 'i':
+        result = process_input_operand_b(value);
+        goto on_option_out;
```
williamcroberts: uhhhh, just use a break?

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,198 @@
...
+    case 0:
+        result = process_input_offset(value);
+        goto on_option_out;
```
williamcroberts: break

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,198 @@
...
+bool is_input_option_args_valid(void) {
+
+    if (!ctx.session_path) {
+        LOG_ERR("Must specify -S session file.");
+        return false;
+    }
+
+    if (!ctx.operand_b.size) {
+        LOG_ERR("Data to compare is of size 0");
+        return false;
```
williamcroberts: is this really an error? Is it possible to compare against an array of 0 length? Or perhaps this should be a warning?

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /*
+     * Ensure input data at offset is smaller than the TPMS_TIME_INFO size.
+     */
+    if ((ctx.operand_b.size + ctx.offset) > sizeof(TPMS_TIME_INFO)){
```
williamcroberts: why TPMS_TIME_INFO, is this something in the spec? Can you quote the source?

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,198 @@
...
+    char *input_file = strcmp("-", value) ? value : NULL;
+    bool result;
+
+    if (input_file) {
+        result = files_get_file_size_path(value,
+                (long unsigned *) &ctx.operand_b.size);
+    }
+    if (input_file && !result) {
+        return false;
+    }
```
williamcroberts: I dont think you need block 75 to 81, the files_load_bytes_from... routine handles it all

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,198 @@
...
+
+    result = files_load_bytes_from_buffer_or_file_or_stdin(NULL, input_file,
+            (UINT16 *)&ctx.operand_b.size,
+            (unsigned char *)&ctx.operand_b.buffer);
```
williamcroberts: get rid of that ampersand on the buffer thats why its complaining. You shouldn't need to cast these.

tools/tpm2_policycountertimer.c
idesai: The comparison in this command (TPM2_PolicyCounterTimer) is in the TPMS_TIME_INFO structured data.

williamcroberts: Ahh looking at the tests it is more clear now, you create some string in milliseconds and pass through XXD converting it to binary and then copy it into the structure. One, your check is wrong, you want to check against the size of the time portion of the field, which is a U64 right? I don't see the utility in setting the TPMS_TIME_INFO structure like this and requiring an offset.

I would just make time in miliseconds an argument to the tool and make the comparison operator a required option.

Then in the future, if people need to adjust other portions of the TPMs_TIME_INFO, we can set them by name:

```bash
tpm2_policycountertimer --gte 123456
tpm2_policycountertimer --gte time=123456
tpm2_policycountertimer --gte clock=123456 reset_count=2, etc.
```

I think for right now, I would just do milliseconds as an argument and populate time only.

idesai: > Ahh looking at the tests it is more clear now, you create some string in milliseconds and pass through XXD converting it to binary and then copy it into the structure. One, your check is wrong, you want to check against the size of the time portion of the field, which is a U64 right? I don't see the utility in setting the TPMS_TIME_INFO structure like this and requiring an offset.
> 
> I would just make time in miliseconds an argument to the tool and make the comparison operator a required option.
> 
> Then in the future, if people need to adjust other portions of the TPMs_TIME_INFO, we can set them by name:
> 
> ```shell
> tpm2_policycountertimer --gte 123456
> tpm2_policycountertimer --gte time=123456
> tpm2_policycountertimer --gte clock=123456 reset_count=2, etc.
> ```
> 
> I think for right now, I would just do milliseconds as an argument and populate time only.

Although you can only specify one operand at a time as a single input operand buffer and an offset is ultimately landing in the esapi call

williamcroberts: > > I think for right now, I would just do milliseconds as an argument and populate time only.
> 
> Although you can only specify one operand at a time as a single input operand buffer and an offset is ultimately landing in the esapi call

Yeah but we're not trying to expose ESAPI directly, we're trying to create tools that mortals can use

man/tpm2_policycountertimer.1.md
```diff
@@ -0,0 +1,144 @@
...
+first minute of TPM restart.
+
+## Create the policy and the sealing object
+
+```bash
+tpm2_startauthsession -S session.ctx
+
+SIXTYTHOUSANDMILLISECONDS=60000
+tpm2_policycountertimer -S session.ctx -L policy.countertimer --ult \
+"$SIXTYTHOUSANDMILLISECONDS"
```
williamcroberts: don't use variables in the examples... if you need more context add a comment saying what you need to convey.

man/tpm2_policycountertimer.1.md
```diff
@@ -0,0 +1,144 @@
...
+tpm2_create -Q -u key.pub -r key.priv -i- -C prim.ctx \
+-L policy.countertimer -a "fixedtpm|fixedparent" -c key.ctx
+```
+
+## Unsealing should work in the first minute after TPM restart
+
+```bash
+tpm2_startauthsession -S session.ctx --policy-session
+
+tpm2_policycountertimer -S session.ctx --ult "$SIXTYTHOUSANDMILLISECONDS"
```
williamcroberts: See this variable isn't defined here... just drop it and use 60000 directly.

test/integration/tests/abrmd_policycountertimer.sh
```diff
@@ -0,0 +1,88 @@
...
+tpm2_unseal -c key.ctx -p session:session.ctx
+
+tpm2_flushcontext session.ctx
+
+#
+# Test if a policycountertimer evaluates with the clock
+#
+tpm2_clear
+tpm2_startauthsession -S session.ctx --policy-session
+tpm2_policycountertimer -S session.ctx --ult clock $SIXTYTHOUSANDMILLISECONDS
```
williamcroberts: The code looks like it can process multiple things, like clock 60000 restart 2, etc, is this true? I don't see you testing that. Also, I would make these key=value arguments like dd so it's more clear.

```C
for (i=0; i < argc; i++) {
  char *a = argv[i]
  char *split = strchr(a, '=');
  if (!split) {
    // error
  }
  split[0] = '\0';
  const char *value = split + 1;
  const char *key = a;

  if (!value[0]) {
    //error
  }

  // look up key and process value
}
```

tools/tpm2_policycountertimer.c
idesai: To clarify if multiple of these conditions have to be true, then it can be combined in an OR or AND fashion by calling the tool to extend the same session context, just like other policy tools.

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,251 @@
...
+            break;
+    }
+
+    ctx.operand_b.size = size;
+    int rc = tpm2_util_hex_to_byte_structure(string_bytes, &ctx.operand_b.size,
+        ctx.operand_b.buffer);
+    if (rc) {
+        LOG_ERR("Error converting the milliseconds input.");
+        result = false;
+    }
```
williamcroberts: I have almost no idea what you're trying to do here and this code looks very wrong.

You have some data structure you need to fill up with a number starting as a string right? The data structure is either 64 or 32 bits wide? Is this your "to big endian" routine?

Cant you do?
```C
if (is_32_bit) {
  tpm2_util_string_to_uint32(str, value);
  tpm_structure.field = tpm2_util_hton_32(str, value);
} else {
  tpm2_util_string_to_uint64(str, value);
  tpm_structure.field = tpm2_util_hton_64(str, value);
}
```

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,251 @@
...
+        } else if (!strcmp("resets", argv[arg_ctr])) {
+            inline_process_time_info(argc, argv, arg_ctr, 16, sizeof(uint32_t));
+            arg_ctr++;
+        } else if (!strcmp("restarts", argv[arg_ctr])) {
+            inline_process_time_info(argc, argv, arg_ctr, 20, sizeof(uint32_t));
+            arg_ctr++;
+        } else if (!strcmp("safe", argv[arg_ctr])) {
+            ctx.offset = 24;
+            ctx.operand_b.size = 1;
+            ctx.operand_b.buffer[0] = TPM2_YES;
```
williamcroberts: why? Explain this magic in a comment

test/integration/tests/abrmd_policycountertimer.sh
idesai: Policycountertimer is an immediate assertion that will fail instantly in a non-trial policy session.

tools/tpm2_policycountertimer.c
idesai: > I have almost no idea what you're trying to do here and this code looks very wrong.
> 
> You have some data structure you need to fill up with a number starting as a string right? The data structure is either 64 or 32 bits wide? Is this your "to big endian" routine?
> 
> Cant you do?
> 
> ```c
> if (is_32_bit) {
>   tpm2_util_string_to_uint32(str, value);
>   tpm_structure.field = tpm2_util_hton_32(str, value);
> } else {
>   tpm2_util_string_to_uint64(str, value);
>   tpm_structure.field = tpm2_util_hton_64(str, value);
> }
> ```

No. We are not filling a data structure. Rather we take the number starting as a string, convert it to hex and then break it into a byte stream which forms the operand buffer. The operand buffer is not a single specific structure member but could be any of the structure members merely used for comparison. And so an offset is additionally specified in the TPM/ESAPI command. The TPM then consumes the offset and the buffer for its internal comparison of the buffer and a corresponding TPM state/value.

idesai: I wish the TPM command itself was simpler - rather than have a buffer and structure member offset as an index into the structure members, why not have unique inputs for all structure members in which case your proposed routine would have been handy.

tools/tpm2_policycountertimer.c
idesai: essentially these magic numbers are the offsets in the TPMS_TIME_INFO structure. Although I must have a macro definition for the offset values, I agree.

test/integration/tests/abrmd_policycountertimer.sh
williamcroberts: Huh? But can't multiple fields be set?

tools/tpm2_policycountertimer.c
williamcroberts: Libmu have anything?

tools/tpm2_policycountertimer.c
williamcroberts: Populate a tpms_time_info struct and the use libmu to convert it to a byte array

test/integration/tests/abrmd_policycountertimer.sh
idesai: No, they can't be. The last one parsed will replace operand buffer content. This is why I error out if more than two arguments are received. Although, I see your point - I will resort to key=value mechanism and obviate the do{}while.

tools/tpm2_policycountertimer.c
idesai: TPM expects only a single structure-member data in the operand buffer.

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    return true;
+}
+
+#define OFFSET_TPMS_TIME_INFO_TIME 0
+#define OFFSET_TPMS_TIME_INFO_CLOCK 8
+#define OFFSET_TPMS_TIME_INFO_RESETS 16
+#define OFFSET_TPMS_TIME_INFO_RESTARTS 20
+#define OFFSET_TPMS_TIME_INFO_SAFE 24
```
williamcroberts: use offsetof() in stddef.h.

tools/tpm2_policycountertimer.c
williamcroberts: So you're convert logic should be something like:
1. Get the key and value the user wants to set.
2. Convert the value the user provides into either a u64 or u32.
3. serialize that to the TPM2B_OPERAND
4. set offset for the command to the value of offsetof() macro for the key field in the TPMS structure.

For step 2, you're converting it to a hex string and and then populating that hex string in into the TPM2B_OEPRAND buffer. That should be equivalent of converting it to a big endian and populating the buffer. Just call  the host to tpm2_util_hton_32 or tpm2_util_hton_64 routines, and then copy that into the buffer as is. No need to be malloc'ing up.

test/integration/tests/abrmd_policycountertimer.sh
williamcroberts: Ahh I missed that.

williamcroberts:
@idesai here is a patch that does it as I describe
[0001-bills-fixups.patch.txt](https://github.com/tpm2-software/tpm2-tools/files/3778987/0001-bills-fixups.patch.txt)

To apply:
wget https://github.com/tpm2-software/tpm2-tools/files/3778987/0001-bills-fixups.patch.txt
git am 0001-bills-fixups.patch.txt

tools/tpm2_policycountertimer.c
williamcroberts: If you error out and stop the tool this would be LOG_ERR

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,271 @@
...
+    case 9:
+        ctx.operation = TPM2_EO_UNSIGNED_LE;
+        break;
+    case 10:
+        ctx.operation = TPM2_EO_BITSET;
+        break;
+    case 11:
+        ctx.operation = TPM2_EO_BITCLEAR;
+        break;
+    default:
```
williamcroberts: If you specify multiple options, the tool works respecting the last option. I would make all the option cases fall through and use the number to index into an array of operations for the assignment and then set a flag that option handked. And check that flag in this handler that it's not set. If it's set issue an error.

tools/tpm2_policycountertimer.c
```diff
@@ -0,0 +1,271 @@
...
+        value = split + 1;
+        key = a;
+    }
+
+    if (!value[0]) {
+      LOG_ERR("Must specify a corresponding value");
+      return false;
+    }
+
+  // look up key and process value
```
williamcroberts: Indent error

williamcroberts:
@idesai here is another patch
[0001-tools-tpm2_policycountertimer-verify-options.patch.txt](https://github.com/tpm2-software/tpm2-tools/files/3780423/0001-tools-tpm2_policycountertimer-verify-options.patch.txt)

Issue #1820, opened 2019-10-28
nvextend man page is wrong


idesai:
nvextend is identical to nvsetbits man page.

perhaps copy-paste-(forgotten to edit for nvextend).

williamcroberts:
Yep copy paste forgot to edit

Issue #1631, opened 2019-07-31
qualifyingdata should be file or byte array


williamcroberts:
When taking in qualifying data like in tpm2_quote, check to see if its a file and use it as binary else try and convert it from hex string

williamcroberts:
Check all the tools

PR #1822, opened 2019-10-29
qualifying data as hex string or binary file and manpage fix

williamcroberts:
- Unify qualifying data like options so the data argument can be a hex string or binary file.
- Fix the tpm2_nvextend manpage.

Fixes: #1631 
Fixes: #1820

### DIFF
--- a/lib/tpm2_policy.c
+++ b/lib/tpm2_policy.c
@@ -156,7 +156,7 @@ tool_rc tpm2_policy_build_pcr(ESYS_CONTEXT *ectx, tpm2_session *policy_session,
 
 tool_rc tpm2_policy_build_policyauthorize(ESYS_CONTEXT *ectx,
         tpm2_session *policy_session, const char *policy_digest_path,
-        const char *policy_qualifier_path,
+        const char *qualifying_data,
         const char *verifying_pubkey_name_path, const char *ticket_path) {
 
     bool result = true;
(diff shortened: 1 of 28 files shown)


lib/tpm2_policy.c
```diff
@@ -265,25 +260,17 @@ tool_rc tpm2_policy_build_policysecret(ESYS_CONTEXT *ectx,
...
-    unsigned long file_size = 0;
-    bool result = true;
-    if (policy_qualifier_path) {
-        result = files_get_file_size_path(policy_qualifier_path, &file_size);
-        if (!result) {
-            return tool_rc_general_error;
-        }
-    }
-
-    TPM2B_NONCE policy_qualifier = { .size = (uint16_t) file_size };
```
williamcroberts: @idesai this code is wrong, and broken. file_size is a long, which can hold a value larger than 16 bits. You could be reading only partial files as you cast it done without checking that it is not larger than UINT16_MAX.

Additionally, this code was over complicated... all you need to do is:
1. set a size to the max size of the buffer
2. call files_load_bytes_from_path(), which will read up to max bytes and error out if the file hasn't been fully consumed as well as update the size parameter to the actual size in bytes read.
3. go on your way using the contents, validate size if your expecting something certain.

tools/tpm2_certifycreation.c
```diff
@@ -227,31 +227,15 @@ static tool_rc process_certifycreation_input(ESYS_CONTEXT *ectx,
...
+        return tool_rc_success;
     }
 
-    unsigned long file_size = 0;
-    result = files_get_file_size_path(ctx.policy_qualifier_path, &file_size);
-    if (!result || file_size == 0) {
-        return tool_rc_general_error;
-    }
-
-    *policy_qualifier = malloc(file_size + sizeof(uint16_t));
```
williamcroberts: @idesai this code is wild.
1. if you need to malloc a TPM2B_* data type, just malloc based on the sizeof of the type. We're not counting the few bytes it wastes.
2. Why did you even malloc it? There was no point in complicating the flow with needing to free and dealing with oom. Just stack allocate and pass a pointer, which is funny, as a lot of the other parameter structures worked like that.

Issue #859, opened 2018-01-30
Implement policycphash


williamcroberts:
Implement a tool to use the policycphash command.

idesai:
- [ ] Add SAPI build and link references to access Esys_GetSysContext
- [ ] Add a function to retrieve sapi context from Esys API call
- [ ] Add a function to calculate cpHash
- [ ] Add tool, test, man page for enabling TPM2_PolicyCpHash

**Add option to output cpHash for tools**
- [ ] tpm2_nvsetbits
- [ ]tpm2_nvextend
- [ ] tpm2_nvincrement

idesai:
cpHash for remaining tools to be added when implementing command and session audit tools.

PR #1826, opened 2019-10-31
Tpm2 policycphash

idesai:
Progresses #859


lib/tpm2.c
```diff
@@ -2079,3 +2142,80 @@ tool_rc tpm2_gettime(ESYS_CONTEXT *ectx,
...
+    }
+
+    //CpBuffer
+    memcpy(to_hash + offset, command_parameters, command_parameters_size);
+
+    //cpHash
+    tool_rc rc = tool_rc_success;
+    bool result = tpm2_openssl_hash_compute_data(halg, to_hash, to_hash_len,
+        cp_hash);
+    free(to_hash);
```
idesai: Should be Esys_Free

williamcroberts:
@idesai can you rebase this?

.ci/download-deps.sh
```diff
@@ -3,8 +3,8 @@
 
 function get_deps() {
 
-	TSS_VERSION="2.3.1"
-	ABRMD_VERSION="2.3.0"
+	TSS_VERSION=master
+	ABRMD_VERSION=master
```
williamcroberts: Do we need these set to master or can we bump them to a stable release for the feature you need?

idesai: @williamcroberts the change for retrieving the sapi handle is in tpm2-tss master. I also see it is in v2.4rc1. Perhaps I can tag this as 2.4 and wait until the rc becomes a full release.

PR #1846, opened 2019-11-13
tpm2_getcap: Fix bitwise operator order for calculating cHandles

chrisccoulson:
'>>' has a higher precedence than '&' which results in the value of cHandles
being computed as `tpma_cc & 0xE`. This just adds some parentheses to correct
the operator order so that cHandles is calculated and displayed correctly.


idesai:
Thanks, @chrisccoulson, had to rebase so closing via #1848

PR #1848, opened 2019-11-15
Fix getcap

idesai:
'>>' has a higher precedence than '&' which results in the value of cHandles
being computed as tpma_cc & 0xE. This just adds some parentheses to correct
the operator order so that cHandles is calculated and displayed correctly.


idesai:
Closing #1846  via #1848

PR #1856, opened 2019-11-21
Match types of counter variable and comparison bound

diabonas:
Fix several small alerts found by LGTM:
- [`TPML_PCR_SELECTION.count`](https://github.com/tpm2-software/tpm2-tss/blob/1c99117f4d102fd465655ed86aca58909a13c10d/include/tss2/tss2_tpm2_types.h#L1046) is `UINT32`, so the counter variable needs to be `UINT32` as well.
- [`TPMS_PCR_SELECTION.sizeofSelect`](https://github.com/tpm2-software/tpm2-tss/blob/1c99117f4d102fd465655ed86aca58909a13c10d/include/tss2/tss2_tpm2_types.h#L951) is `UINT8`, but after multiplication with 8 it might only fit into an `int` (which is guaranteed to be at least 16 bits long) due to integer promotion.

### DIFF
--- a/lib/pcr.c
+++ b/lib/pcr.c
@@ -191,8 +191,8 @@ bool pcr_print_pcr_struct(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
         tpm2_tool_output("  %s:\n", alg_name);
 
         // Loop through all PCRs in this bank
-        UINT8 pcr_id;
-        for (pcr_id = 0; pcr_id < pcr_select->pcrSelections[i].sizeofSelect * 8;
+        unsigned int pcr_id;
+        for (pcr_id = 0; pcr_id < pcr_select->pcrSelections[i].sizeofSelect * 8u;
                 pcr_id++) {
             if (!tpm2_util_is_pcr_select_bit_set(&pcr_select->pcrSelections[i],
                     pcr_id)) {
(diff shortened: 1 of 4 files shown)


lib/pcr.c
```diff
@@ -191,7 +191,7 @@ bool pcr_print_pcr_struct(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
         tpm2_tool_output("  %s:\n", alg_name);
 
         // Loop through all PCRs in this bank
-        UINT8 pcr_id;
+        int pcr_id;
```
williamcroberts: I think the better fix would be to force everything to unsigned. I think the multiplication is going signed since the literal will int in an int, C uses int as it's type and thus we're getting signed results from it. I think changing the 8 to 8u will fix the issue and prevent us from undefined overflow on the counter variable.

diabonas: If I understand the integer promotion rules correctly, using `8u` instead of `8` will still convert the upper bound to `unsigned int`, let me run a quick test with LGTM on this PR. Using `8u` in combination with changing `pcr_id` to `unsigned int` seems reasonable.

diabonas: Yeah, using only `8u` [does not fix these three alerts](https://github.com/tpm2-software/tpm2-tools/pull/1856#issuecomment-558713772). Changing both `8`→`8u` and `UINT8`→`unsigned int` should hopefully bring it back to fixing all four alerts.

diabonas: Indeed, we are [back to normal](https://github.com/tpm2-software/tpm2-tools/pull/1856#issuecomment-558718538).

PR #1863, opened 2019-12-04
tpm2_pcrextend,tpm2_alg_util: fix sha1 examples

dnoliver:
Signed-off-by: Nicolas Oliver <nicolasoliver03@gmail.com>


williamcroberts:
Yep no alias for sha: https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_alg_util.c#L54

PR #1882, opened 2020-01-10
[RFC] tpm2_eventlog: Add tool to parse the TPM2 event log.

flihp:
The format for the event log parsed by this tool is documented in
section 9 of the "TCG PC Client Platform Firmware Profile
Specification" (aka "PFP"). This is the "crypto-agile" format capable of
representing digests of varying size. Some constants / values necessary
for understanding the log contents are spread out over other specs but
the structure is documented in the PFP.

This log is a series of concatinated binary structures describing events
generated / collected by the firmware on platform boot. Each 'event' has
3 primary components: First is a header describing the PCR extended, the
'type' of the event and the number of digests in the event. After this a
series of digest structures follows. The final data in this structure is
the event size and a binary representation of the event. The total size
of the log must be inferred.

This tool takes the log file as it's only parameter (positional). It
parses the log and writes an ascii / text representation to stdout. The
option processing and 'main' function for the tool is located in
'tools/misc/tpm2_eventlog.c'. If the log file parameter is omitted the
tool will fallback to parsing the event log exposed by the kernel via
securityfs (/sys/kernel/security/tpm0/binary_bios_measurements).

The log parsing logic and the text formatting logic are located in the
'lib/tpm2_eventlog' and 'lib/tpm2_eventlog_ascii' modules respectively.
The main driver initiates the log parsing through the 'foreach_event2'
function that takes a reference to the start of the log, its size and a
callback function. This callback is invoked by the parsing logic for
each event structure encountered in the log. It's passed a reference to
the event and the size of the event (validated by the parser before
invoking the callback). The display logic is implemented as callbacks
invoked the the parser.

Additionally this commit adds unit tests and a single integration test.
Unit tests are used to test the parser and display logic. The single
integration test invokes the tool passing it a single binary event
structure taken from a VM booting the OVMF firmware with an emulated
TPM2 backend.


flihp:
I'm just now realizing that when I replaced the file reading code from my original work with the existing mechanisms available in the 'files' module it broke the code that pulls the log from securityfs. The pattern from the files module requires we first stat the file size and then pass the size to the 'read' function. I don't think this will work for the securityfs file.

flihp:
PR updated so `tpm2_eventlog` now requires the path to the eventlog passed as a positional parameter. Additional cleanup made to better align with the coding standard and general sanity.

flihp:
Another thing I hadn't considered till now is that the event log is all in host byte order. So the integration test I've added here will fail on big-endian systems since I took the test data (an event log entry) form an Intel platform (little-endian).

lib/tpm2_eventlog.c
```diff
@@ -0,0 +1,143 @@
+#include <inttypes.h>
+#include <stdlib.h>
+#include <tss2/tss2_tpm2_types.h>
+
+#include "log.h"
+#include "tcg-efi-event.h"
+#include "tpm2_eventlog.h"
+
+/* Functions to parse / process the structures that make up the eventlog. */
+size_t sizeof_alg(UINT16 alg_id) {
```
williamcroberts: This is a duplicate function of tpm2_alg_util_get_hash_size, and it's also missing SM3_256...

lib/tpm2_eventlog.c
```diff
@@ -0,0 +1,143 @@
...
+         size -= event_size) {
+
+        size_t digests_size = 0;
+        if (size < sizeof(*eventhdr)) {
+            LOG_ERR("corrupted log, insufficient size for event header: %zu", size);
+            return false;
+        }
+        event_size = sizeof(*eventhdr);
+
+        if (!foreach_digest2(eventhdr->Digests,
```
williamcroberts: code conventions are to not hide function calls in conditionals unless an assignment is required. In this case do something like:
```C
bool result = foreach_digest2();
if (!result) {
  return false;
}
```

lib/tpm2_eventlog.c
```diff
@@ -0,0 +1,143 @@
...
+        }
+        event_size += sizeof(*event);
+
+        if (size < event_size + event->EventSize) {
+            LOG_ERR("size insufficient for event data");
+            return false;
+        }
+        event_size += event->EventSize;
+
+        if (!callback(eventhdr, event_size, data)) {
```
williamcroberts: same about functions in conditionals here.

lib/tpm2_eventlog_ascii.c
```diff
@@ -0,0 +1,208 @@
...
+#include <stdlib.h>
+
+#include <tss2/tss2_tpm2_types.h>
+
+#include "log.h"
+#include "tcg-efi-event.h"
+#include "tpm2_eventlog.h"
+#include "tpm2_eventlog_ascii.h"
+
+char* get_alg_name(UINT16 alg_id) {
```
williamcroberts: Any reason we cant use tpm2_alg_util_algtostr? Do you need the output to be "TPM2_ALG_SHA1" or is "sha1" good enough? If you need to full name can we smarten up that logic and add a flag to the routine to get the canonical name?

lib/tpm2_eventlog_ascii.c
```diff
@@ -0,0 +1,208 @@
...
+        return "TPM2_ALG_SHA256";
+    case TPM2_ALG_SHA384:
+        return "TPM2_ALG_SHA384";
+    case TPM2_ALG_SHA512:
+        return "TPM2_ALG_SHA512";
+    default:
+        return "UNKNOWN_ALGORITHM";
+    }
+}
+char* eventtype_to_string (UINT32 event_type) {
```
williamcroberts: `const char *`

Also, here and throughout, `<type> *<varname>`. The * should be tight to the variable name not the type.

lib/tpm2_eventlog.c
flihp: Thx. I missed the `tpm2_alg_util` while porting this over.

lib/tpm2_eventlog_ascii.c
flihp: No reason. Didn't see the `tpm2_alg_util` module when porting. Resolved.

lib/tpm2_eventlog_ascii.c
flihp: Good catch. Went with `char const *`.

flihp:
"ascii" was chosen since the goal was to convert the binary log to human readable text on the console and every other prefix I could come up with was too long. The format is of my own "design" (using the term loosely) since TCG doesn't  standardize a human readable text formats. I avoided JSON because I didn't want to encourage folks to use the output from this tool as input into anything else since the TCG either already has or is currently standardizing formats for that purpose.

The tool will output partial results before a formatting error is encountered. Initially this was by design because some UEFI code I wrote mysteriously corrupted the last few bytes or appended a few (I can't remember which) when capturing the eventlog. Chasing that bug down was a distraction from the larger goal so I just compensated in the tool. I'm not partial either way and adding a bit of complexity to prevent any output from corrupted logs shouldn't take too much time.

flihp:
Got a build failure after fixing up the changes requested. Oddly this failure is in a seemingly unrelated test `test/integration/tests/getekcertificate.sh` for only one of the build matrix entries (AMD64 with code coverage enabled) and I can't repro locally. Any ideas?

williamcroberts:
> Got a build failure after fixing up the changes requested. Oddly this failure is in a
> seemingly unrelated test test/integration/tests/getekcertificate.sh for only one
> of the build matrix entries (AMD64 with code coverage enabled) and I can't repro
> locally. Any ideas?

Usually something like a timeoout, just hit the restart button

williamcroberts:
> "ascii" was chosen since the goal was to convert the binary log to human readable
> text on the console and every other prefix I could come up with was too long.
> The format is of my own "design" (using the term loosely) since TCG doesn't
> standardize a human readable text formats. I avoided JSON because I didn't want

Darn me, I said JSON and meant YAML. YAML is human readable in its basic form.

> to encourage folks to use the output from this tool as input into anything else
> since the TCG either already has or is currently standardizing formats for that
> purpose.

Folks then will try grep/awk and then itll be a brittle hell. I standardized on stdout tool
output being YAML, so I don't want to break away from that here.
The tools currently can output malformed YAML on exit != 0, so this short output is a
common issue.  You don't need to use the emitter, as your output doesn't look wildly
deep and complex, so basic tpm2_tool_output() calls with hand jammed YAML output
should be sufficient.

> 
> The tool will output partial results before a formatting error is encountered.
> Initially this was by design because some UEFI code I wrote mysteriously
> corrupted the last few bytes or appended a few (I can't remember which) when
> capturing the eventlog. Chasing that bug down was a distraction from the larger
> goal so I just compensated in the tool. I'm not partial either way and adding a bit
> of complexity to prevent any output from corrupted logs shouldn't take too much
> time.

That’s where the YAML emitter library is nice... I have sample code in the pkcs11 library.
YAML is human readable.

flihp:
> Usually something like a timeoout, just hit the restart button.

I don't have the permissions required to do that.

> I standardized on stdout tool output being YAML, so I don't want to break away from that here.

The only thing I don't like about using either JSON or YAML for this is, AFAIK, the recommended way to represent binary data blobs is base64 encoded strings. For most of the debugging that motivated this work I was comparing hex dumps of binary logs to digests generated by the `shaXsum` tools all of which deal in hex. I don't think this is sufficient to justify deviation from the local standard. Running the output through another layer of translation to dump hex from the base64 isn't too heavy a lift and it would have got me to 95% of what I needed.

> The tools currently can output malformed YAML on exit != 0, so this short output is a
common issue.  You don't need to use the emitter, as your output doesn't look wildly
deep and complex, so basic tpm2_tool_output() calls with hand jammed YAML output
should be sufficient.

Good pointer. I'll take a look.

williamcroberts:
> The only thing I don't like about using either JSON or YAML for this is, AFAIK, the recommended way to represent binary data blobs is base64 encoded strings. For most of the debugging that 

Iv'e never seen that recommendation, nor in this case would I follow it. All the tools dump binary fields as lower case hex strings. Which means using tools like xxd to convert back to binary works.

williamcroberts:
> > Usually something like a timeoout, just hit the restart button.
> 
> I don't have the permissions required to do that.

I didn't realize restarting a build took advanced permissions, funny. You can trigger a build via a PR but you cant re-start it, that seems silly. Likely no way to enforce that it's an actual user associated with the PR rather than some random web-bot tampering with projects.

williamcroberts:
> > > Usually something like a timeoout, just hit the restart button.
> > 
> > 
> > I don't have the permissions required to do that.
> 
> I didn't realize restarting a build took advanced permissions, funny. You can trigger a build via a PR but you cant re-start it, that seems silly. Likely no way to enforce that it's an actual user associated with the PR rather than some random web-bot tampering with projects.

Weird, I couldn't even re-start it at first. I had to sign out of travis and sign back in to get the restart icon to appear.

williamcroberts:
> > The only thing I don't like about using either JSON or YAML for this is, AFAIK, the recommended way to represent binary data blobs is base64 encoded strings. For most of the debugging that
> 
> Iv'e never seen that recommendation, nor in this case would I follow it. All the tools dump binary fields as lower case hex strings. Which means using tools like xxd to convert back to binary works.

Thats for the actual "tag:yaml.org,2002:binary" data type. We just convert to string. Theirs also different forms of YAML output, im not sure of the proper terms here, but one is all tagged up with extra metadata like type and braces and the other doesn't have this. The plain style output is what we're targeting here. I think the plain style is "block" and the marked up version is "flow", but don't quote me.

flihp:
YAML output should be functional. The only thing in the output that `yamllint` complains about is line lengths > 80 characters. Breaking these lines up seems like more work than it's worth IMHO. Commit message and tests are updated.

The only change I couldn't make is using the `tpm2_tool_output` function as you requested. As implemented currently this function cannot be used in code that's covered by a unit test. This limitation is the result of the function depending on global data defined in `tools/tpm2_tool.c`. This module defines the `main` function for the tools and so linking it into a unit test causes duplicate definitions.

I don't expect fixing this would be hard but it's not central to this work. AFAIK the only downside to using `printf` directly is that we lose the ability to validate the eventlog format w/o any output (--quiet). Would be a nice feature but not required.

man/tpm2_eventlog.1.md
```diff
@@ -0,0 +1,39 @@
...
+# OPTIONS
+
+This tool takes no tool specific options.
+
+  * **ARGUMENT** The command line argument is the path to a binary TPM2
+    eventlog.
+
+## References
+
+[common options](common/options.md) collection of common options that provide
```
williamcroberts: This has -Q in it, ill fix up the quiet code and pull this in, you got it far enough

flihp: Thanks.

williamcroberts:
Moved: https://github.com/tpm2-software/tpm2-tools/pull/1898

PR #1906, opened 2020-01-23
tpm2_tool_output: Move tool output variable and macros to new module.

flihp:
Defining the 'output_enabled' global variable in the tpm2_tool module
makes it unusable from any executable that defines it's own 'main'
function (mostly affecting unit tests). e2b7f1b1752f works around this
by defining the variable in the test. This works but requires all unit
tests using `tpm2_tool_output` to do the same. This may be a source of
mysteries for those working on unit tests in the future.

This commit removes this burden from unit test writers by moving
`output_enabled` out of the tpm2_tool module and into one of it's
own: tpm2_tool_output. This causes the symbol to be included in the
local utility library where any executable may link against it. This
commit also adds a new macro 'tpm2_tool_output_disable` to allow the
tpm2_tool module to disable output without having to access the
`output_enabled` variable directly.

### DIFF
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -10,6 +10,7 @@
 #include "tpm2_eventlog.h"
 #include "tpm2_eventlog_yaml.h"
 #include "tpm2_tool.h"
+#include "tpm2_tool_output.h"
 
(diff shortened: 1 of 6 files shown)


flihp:
The formatting in the commit message got a bit borked here on account of my quotes being mismatched. I just fixed this and pushed.

PR #1914, opened 2020-01-30
tools: fix incompatible pointers in tpm2_eventlog

joholl:
In `tpm2_eventlog`, a `size_t *` pointer is passed to a function which expects an `unsigned int long *`. That caused build failure on e.g. Raspberry Pi.

### DIFF
--- a/tools/misc/tpm2_eventlog.c
+++ b/tools/misc/tpm2_eventlog.c
@@ -42,7 +42,7 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
 
     bool ret = false;
     UINT8 *eventlog;
-    size_t size = 0;
+    unsigned long size = 0;
 
     if (filename == NULL) {
         LOG_ERR("Missing required positional parameter, try -h / --help");
@@ -55,7 +55,7 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
 
     eventlog = calloc(1, size);
     if (eventlog == NULL){
-        LOG_ERR("failed to allocate %zd bytes: %s", size, strerror(errno));
+        LOG_ERR("failed to allocate %lu bytes: %s", size, strerror(errno));
         return tool_rc_general_error;
     }
 


tools/misc/tpm2_eventlog.c
```diff
@@ -55,7 +55,7 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
 
     eventlog = calloc(1, size);
     if (eventlog == NULL){
-        LOG_ERR("failed to allocate %zd bytes: %s", size, strerror(errno));
```
flihp: Your commit message states that we're passing the wrong parameter to a function but it looks more like the issue is caused by the use of the 'zd' format specifier. Couldn't we fix this by replacing the 'zd' with a 'zu' instead of changing the type of the 'size' variable?

joholl: Nope, it seems that `size_t` is a `unsigned int` on my RPi.

```
tools/misc/tpm2_eventlog.c:51:46: error: passing argument 2 of ‘files_get_file_size_path’ from incompatible pointer type [-Werror=incompatible-pointer-types]
     ret = files_get_file_size_path(filename, &size);
                                              ^
In file included from tools/misc/tpm2_eventlog.c:8:0:
./lib/files.h:377:6: note: expected ‘long unsigned int *’ but argument is of type ‘size_t * {aka unsigned int *}’
 bool files_get_file_size_path(const char *path, unsigned long *file_size);
      ^~~~~~~~~~~~~~~~~~~~~~~~
```

williamcroberts: > Your commit message states that we're passing the wrong parameter to a function but it looks more like the issue is caused by the use of the 'zd' format specifier. Couldn't we fix this by replacing the 'zd' with a 'zu' instead of changing the type of the 'size' variable?

files_get_file_size_path() takes unsigned long. So he updates the declaration and logging macro.

flihp: Ah I see now. The compiler output was helpful. The issue is with the type of the size parameter for the `files_get_file_size_path` function (which is unsigned long, not size_t).

PR #1922, opened 2020-02-07
Modify INSTALL.md to reference latest version of TPM simulator

moneeb777:


moneeb777:
If this change is OK then we should probably update the wiki also.

idesai:
@moneeb777 I imagine this will make the release process difficult to tame. A fixed version of dependencies helps anchor a specific release for a well-understood behavior. This is particularly important to find and fix bugs.
@williamcroberts your thoughts?

moneeb777:
@idesai I do not remember seeing a dependency on a specific version of the simulator anywhere in the docs so I assumed the latest one should always be preferred. I actually encountered a problem while compiling the simulator because the wiki suggested using ibmtpm974 which has a known bug: https://sourceforge.net/p/ibmswtpm2/discussion/general/thread/a44c4f901f/?limit=25#1d7f/336e

We can also consider that currently, the wiki and INSTALL.md reference two different versions(974 and 1332).

idesai:
> @idesai I do not remember seeing an explicit dependency on a specific version of the simulator in the docs so I assumed the latest one should generally be preferred. I actually encountered a problem while compiling the simulator because the wiki suggested using ibmtpm974 which has a known bug: https://sourceforge.net/p/ibmswtpm2/discussion/general/thread/a44c4f901f/?limit=25#1d7f/336e
> 
> We can also consider that currently, the wiki and INSTALL.md reference two different versions(974 and 1332).

Yes, the wiki definitely needs an update amongst other things.

PR #1944, opened 2020-03-10
test: fix yaml.load warning

yizhao1:
The yaml.load(f) is deprecated since pyyaml 5.1.
Use yaml.load(f, Loader=yaml.BaseLoader) instead of it.
See https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

Fixes warning:
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated,
as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.

Signed-off-by: Yi Zhao <yi.zhao@windriver.com>

### DIFF
--- a/test/system/test_tpm2_activecredential.sh
+++ b/test/system/test_tpm2_activecredential.sh
@@ -61,7 +61,7 @@ from __future__ import print_function
 import yaml
 
 with open('ak.out', 'r') as f:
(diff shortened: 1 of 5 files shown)


williamcroberts:
We're not taking patches on 3.X

PR #1952, opened 2020-03-13
test: only support python3

yizhao1:
The python2 is EOL and in some distributions (e.g. CentOS 8/RHEL 8),
there is no python symbolic link by default which will cause an error
when running the test scripts:
$ ./test_tpm2_activecredential.sh: line 66: python: command not found

See: https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/
So drop python2 support and only keep python3.

Signed-off-by: Yi Zhao <yi.zhao@windriver.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -121,24 +121,25 @@ AS_IF(
     [],
     [AC_MSG_WARN([Required executable bash not found, system tests require a bash shell!])])
 
-AC_CHECK_PROG([PYTHON],[python],[yes])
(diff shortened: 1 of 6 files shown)


williamcroberts:
While python2 is EOL, many distributions that package it are not. This is the same reason we're not killing off OSSL < 1.1 support.

You could, at configure time, figure out which python executable exists, and pass that in the test environment via AM_TESTS_ENVIRONMENT in Makefile.am

williamcroberts:
3.X is end of life....

PR #1969, opened 2020-04-02
 test/integration/helpers.sh: Add cert less option to FAPI test config.

JuergenReppSIT:
For FAPI there will be a PR which will change the provisioning behaviour. An error will be returned if no certificate for the generated EK exists.
To enable FAPI integration tests with the simulator and also with physical
TPMS the cert less option was added to the helper. So no error is signalled
in simulator tests without certificates and the certificate verification is
skipped in tests with physical TPMs.

### DIFF
--- a/test/integration/helpers.sh
+++ b/test/integration/helpers.sh
@@ -416,6 +416,7 @@ cat > $tempdir/fapi_config.json <<EOF
     "system_dir": "$tempdir/${KEYSTORE_SYSTEM}",
     "tcti": "${TPM2TOOLS_TCTI}",
     "system_pcrs" : [],
+    "ek_cert_less": "yes",
     "log_dir" : "$tempdir/${LOG_DIR}",
 }
 EOF


AndreasFuchsSIT:
@idesai I don't know if we really want to wait for the first bugfix-release with this.

Background: We have a bug-fix in tpm2-tss https://github.com/tpm2-software/tpm2-tss/pull/1693 that we need to backport to 4.2.2. The problem is, that this bugfix will trigger a bug in the tpm2-tools test-scripts.

So as soon as we backport there, make-check here will fail.

IMHO we should merge this PR and give it a Quick 2-day RC...

Sorry for cause such problems...

williamcroberts:
> @idesai I don't know if we really want to wait for the first bugfix-release with this.
> 
> Background: We have a bug-fix in tpm2-tss [tpm2-software/tpm2-tss#1693](https://github.com/tpm2-software/tpm2-tss/pull/1693) that we need to backport to 2.4.x. The problem is, that this bugfix will trigger a bug in the tpm2-tools test-scripts.
> 
> So as soon as we backport there, make-check here will fail.
> 
> IMHO we should merge this PR and give it a Quick 2-day RC...
> 
> Sorry for cause such problems...

@idesai I agree with @AndreasFuchsSIT on this, can we get this added to the RC2, and roll RC3 next week?

PR #1991, opened 2020-05-02
Remove PCR Padding

lukehinds:
Previously tpm2_pcrread used to display PCR values with no padding
(aka `0x`). This change reverts back to the original non padded
format.

Signed-off-by: Luke Hinds <lhinds@redhat.com>

### DIFF
--- a/man/tpm2_pcrread.1.md
+++ b/man/tpm2_pcrread.1.md
@@ -42,7 +42,6 @@ sha256 :
 
     The output file to write the PCR values in binary format, optional.
 
-
 [common options](common/options.md)
 
 [common tcti options](common/tcti.md)
--- a/tools/tpm2_pcrread.c
+++ b/tools/tpm2_pcrread.c
@@ -46,7 +46,7 @@ static bool print_pcr_values(void) {
                 return false;
             }
 
-            tpm2_tool_output("  %-2d: 0x", pcr_id);
+            tpm2_tool_output("  %-2d: ", pcr_id);
 
             int k;
             for (k = 0; k < ctx.pcrs.pcr_values[vi].digests[di].size; k++) {


idesai:
@lukehinds thank you for the PR. My thought is that adding the option seems a little excessive to remove the "0x " concatenation. The initial thought when it was put might have been to elicit that the PCR data displayed is a hex byte stream. But I don't see why we can't update the man page with this information. Can you please update the PR with the following changes:
- [ ] Remove the extra "--nopad" "-n" option
- [ ] Replace in place the 0x to simply display the PCR value (you've already done this)
- [ ] Add information in the man page about the PCR value displayed being a stream of hex bytes.

williamcroberts:
This breaks the values in YAML.
```yaml
foo: 0x1234
bar: 1234
```
```python
Python 2.7.17 (default, Apr 15 2020, 17:20:14) 
[GCC 7.5.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml
>>> y = yaml.load(open("a.yaml"))
>>> print(y['bar'])
1234
>>> print(y['foo'])
4660
```
Thus we need to revert this. Whatever you have reading this information, just have it load it up as YAML.

lukehinds:
I honestly never knew about that (TIL) [0], with that in mind I would be ok with a revert, although I still like the option in my first patch of having the flag to display with a hex as default and then just show the hash if used

[0] https://yaml.org/type/int.html

williamcroberts:
@lukehinds it's valid scalar type for YAML. See:
- https://yaml.org/spec/1.0/#id2490752
- https://yaml.org/type/int.html

The issue seems to arise on implementations and how they treat the type based on underlying native data types. Since that is just the YAML without the type detail, you can pull it as you choose.

If you handle it as a string,  you can just check for a leading 0x and remove it with TrimPrefix.
```
func TrimPrefix(s, prefix string) string
TrimPrefix returns s without the provided leading prefix string. If s doesn't start with prefix, s is returned unchanged
```

If you want an actual number representation, you can just unmarshal it into `big.Int`
https://play.golang.org/p/HeMAJ57NhQN

That seems to marshal and unmarshal stuff just fine, all I did was take an example and make it big.Int instead of int. Thats actually the first Go code ive ever touched.

But adding options to tools, for something like that seems unnecessary, I don't want to complicate the tools any further than necessary. Since this is an implementation issue, that is not spec compliant to YAML (per yaml spec Integers represent arbitrarily sized finite mathematical integers), id say theirs no reason to add that workaround in the tool. Especially if other parsers have issues, I don't want to add switches for them all.

lukehinds:
@williamcroberts - that's fair enough. And for any other edge case or lang that does not adhere to the yaml spec, its really not a big deal to have to strip / chomp etc for the first two characters. 

@idesai I am happy for this to be reverted.

PR #1992, opened 2020-05-05
Add FAPI support to all tpm2_* tools

AndreasFuchsTPM:
By introducing a "tss2:" prefix to denote a FAPI object.

Open would be the Question
- Should the FAPI-calling be encapsulated in a different file than object.c ?
- Should we get the TCTI from the FAPI context or do we want to allow ppl to use FAPI-context objects in the regular tcti context ?

### DIFF
--- a/.ci/download-deps.sh
+++ b/.ci/download-deps.sh
@@ -3,7 +3,8 @@
 
 function get_deps() {
 
-	TSS_VERSION=2.4.0
+	TSS_VERSION=master
+#	TSS_VERSION=2.4.0
 	ABRMD_VERSION=2.3.1
 
 	echo "pwd starting: `pwd`"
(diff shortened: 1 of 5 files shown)


williamcroberts:
> By introducing a "tss2:" prefix to denote a FAPI object.
> 
> Open would be the Question
> 
> * Should the FAPI-calling be encapsulated in a different file than object.c ?

Yes I think that. Just handle it in the normal code flows that exist.

> * Should we get the TCTI from the FAPI context or do we want to allow ppl to use FAPI-context objects in the regular tcti context ?

I would say get the FAPI TCTI, but we would then need to make sure that if --tcti is set, we error out or at least say we ignore it in a warning.

AndreasFuchsSIT:
> > * Should we get the TCTI from the FAPI context or do we want to allow ppl to use FAPI-context objects in the regular tcti context ?
> 
> I would say get the FAPI TCTI, but we would then need to make sure that if --tcti is set, we error out or at least say we ignore it in a warning.

I thought about this some more and if we get the FAPI TCTI we run into problems with session usage, since StartAuthSession does not take a FAPI object. So we'd either need to add a flag there as well or just not allow session support when using FAPI objects.
Hmpf... :-(

AndreasFuchsSIT:
I'll close this for now, due to lack of time to work on this and I don't want to keep the PR tab cluttered much longer

williamcroberts:
> I'll close this for now, due to lack of time to work on this and I don't want to keep the PR tab cluttered much longer

Thanks I'm a zero-inboxer, open things in lists give me anxiety :-p

PR #1993, opened 2020-05-06
Fix detection of hardened c flags supported by the compiler

tstellar:
For the clang compiler, adding an unsupported warning option generates
a warning (-Wunknown-warning-option), but not an error.  So all of the
add_hardened_c_flag checks will pass even for unsupported warnings.

However, when the code is actually being compiled, the -Werror flag is
used, which turns the -Wunknown-warning-option warnings into errors and
causes the build to fail.

In order to fix this, we need to be sure to pass -Werror to the
AX_CHECK_COMPILE_FLAG macro in order to correctly detect which
warnings are unsupported.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -192,7 +192,8 @@ AS_IF([test x"$enable_hardening" != x"no"], [
   AC_DEFUN([add_hardened_c_flag], [
     AX_CHECK_COMPILE_FLAG([$1],
       [EXTRA_CFLAGS="$EXTRA_CFLAGS $1"],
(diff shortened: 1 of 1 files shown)


williamcroberts:
fedora30 and opensuse-leap both fail, so we would need to perhaps disable hardening for those two builds. Or just make it so if we can't add it, we don't die, we just warn?

tstellar:
Closed in favor of: https://github.com/tpm2-software/tpm2-tools/pull/2075

PR #1994, opened 2020-05-06
Test: Add test for tpm2-tss master branch

AndreasFuchsTPM:
This PR triggers some bugs in the tools vs the master branch of tpm2-tss.

Could this have to do with the API/ABI break in ESYS regarding the ESYS_TR for TPM2_HANDLE change ?
I thought that was fixed ?


AndreasFuchsSIT:
So I did bisect this with tpm2-tss and the commit that breaks the tools was indeed https://github.com/tpm2-software/tpm2-tss/commit/9d48cc7922ad8dc3496444b04c94439c5bfbec8e `"esys: change the hierarchy type according to the spec"`

Apparently some cases of `fix_esys_hierarchy()` slipped in https://github.com/tpm2-software/tpm2-tools/commit/780800c0be69a49b9097f8eae653cdb0623d2100 `"Map TPMI_RH_HIERARCHY type to Esys_TR type"`.

I'm currently tracing those down and will post a separate PR.

AndreasFuchsSIT:
Once https://github.com/tpm2-software/tpm2-tools/pull/1999 is in master and https://github.com/tpm2-software/tpm2-tss/pull/1719 is merged as well, I'll rebase so CI succeeds and we can merge this one as well.

.travis.yml
```diff
@@ -12,6 +12,7 @@ env:
   - DOCKER_TAG=ubuntu-18.04
   - DOCKER_TAG=fedora-30
   - DOCKER_TAG=opensuse-leap
+  - DOCKER_TAG=ubuntu-18.04 TPM2_TSS_VERSION=master
```
williamcroberts, line 15: this is a really good idea, we can set the minimum stable version and be building against master as we go. Im struggling if we want both gcc and clang builds or just add this as an include build and generate only one build, but I guess we do indeed want to know if something weird with clang or gcc pops up.

williamcroberts (approved):

PR #1996, opened 2020-05-06
Enable CI on FreeBSD

tstruk:
Enable CI builds on FreeBSD


tstruk:
This PR is blocked on this https://github.com/tpm2-software/tpm2-abrmd/pull/711

tstruk:
> ill make a bug to drop the patch apply part so we dont forget

Good idea. Once the change in merged in RM the cirrus CI will start failing so that will be another reminder.

PR #1997, opened 2020-05-06
Revert "Remove PCR Padding"

williamcroberts:
Reverts tpm2-software/tpm2-tools#1991


idesai:
> Reverts #1991

@williamcroberts did #1991 break something?

PR #1999, opened 2020-05-07
Fix ESYS_TR hierarchy transition

AndreasFuchsSIT:
Fix those cases of TPM2_RH_ to ESYS_TR_RH_ translations that were missed in 780800c0be69a49b9097f8eae653cdb0623d2100 as discovered in https://github.com/tpm2-software/tpm2-tools/pull/1994

I hope I did not miss any further cases, but at least tests are now successful.

This should also be backported to 4.1.x and 4.2.x as well.

### DIFF
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -1054,39 +1054,51 @@ tpm2_evictcontrol_skip_esapi_call:
 }
 
 /* This function addresses ESAPI change that changes parameter type from
- * Esys_TR to TPMI_RH_HIERARCHY and breaks backwards compatibility.
+ * Esys_TR to TPMI_RH_HIERARCHY or TPMI_RH_ENABLES and breaks backwards
+ * compatibility.
  * To keep the tools parameters consistent after v4.0 release we need to
  * map the values to appropriate type based on the version of the ESYS API.
  * Note: the mapping is based on the ESYS version recognized at compile time.
  * The TSS change can be found here:
  * https://github.com/tpm2-software/tpm2-tss/pull/1531
  */
-uint32_t fix_esys_hierarchy(TPMI_RH_HIERARCHY hierarchy)
+TSS2_RC fix_esys_hierarchy(uint32_t in, uint32_t *out)
 {
 #if defined(ESYS_3_0)
-    switch (hierarchy) {
+    switch (in) {
         case ESYS_TR_RH_NULL:
+            /* FALLTHRU */
         case ESYS_TR_RH_OWNER:
+            /* FALLTHRU */
         case ESYS_TR_RH_ENDORSEMENT:
+            /* FALLTHRU */
         case ESYS_TR_RH_PLATFORM:
+            /* FALLTHRU */
         case ESYS_TR_RH_PLATFORM_NV:
-            return hierarchy;
+            *out = in;
+            return TSS2_RC_SUCCESS;
         case TPM2_RH_NULL:
-            return ESYS_TR_RH_NULL;
+            *out = ESYS_TR_RH_NULL;
+            return TSS2_RC_SUCCESS;
         case TPM2_RH_OWNER:
-            return ESYS_TR_RH_OWNER;
+            *out = ESYS_TR_RH_OWNER;
+            return TSS2_RC_SUCCESS;
         case TPM2_RH_ENDORSEMENT:
-            return ESYS_TR_RH_ENDORSEMENT;
+            *out = ESYS_TR_RH_ENDORSEMENT;
+            return TSS2_RC_SUCCESS;
         case TPM2_RH_PLATFORM:
-            return ESYS_TR_RH_PLATFORM;
+            *out = ESYS_TR_RH_PLATFORM;
+            return TSS2_RC_SUCCESS;
         case TPM2_RH_PLATFORM_NV:
-            return ESYS_TR_RH_PLATFORM_NV;
+            *out = ESYS_TR_RH_PLATFORM_NV;
+            return TSS2_RC_SUCCESS;
         default:
-            LOG_ERR("An unknown hierarchy handle was passed: 0x%08x", hierarchy);
-            return 0xffffffff;
+            LOG_ERR("An unknown hierarchy handle was passed: 0x%08x", in);
+            return TSS2_ESYS_RC_BAD_VALUE;
     }
 #elif defined(ESYS_2_3)
-    return hierarchy;
+    *out = in;
+    return TSS2_RC_SUCCESS;
 #else
 #error "Need to define either ESYS_3_0 or ESYS_2_3"
 #endif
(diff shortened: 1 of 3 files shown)


lib/tpm2.c
```diff
@@ -1065,6 +1065,12 @@ uint32_t fix_esys_hierarchy(TPMI_RH_HIERARCHY hierarchy)
 {
 #if defined(ESYS_3_0)
     switch (hierarchy) {
+        case ESYS_TR_RH_NULL:
+        case ESYS_TR_RH_OWNER:
+        case ESYS_TR_RH_ENDORSEMENT:
+        case ESYS_TR_RH_PLATFORM:
+        case ESYS_TR_RH_PLATFORM_NV:
```
williamcroberts, line 1072: You need falls-thru comments so the compiler won't complain, CFLAGS `-Wextra` enables `-Wimplicit-fallthrough=3`. Details can be found here:
  - https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

Some versions of gcc/clang are more pedantic than others, but I know occasionally Fedora folks will come back with patches. This may be better for an if/elseif type of thing now.

lib/tpm2.c
```diff
@@ -1073,14 +1079,16 @@ uint32_t fix_esys_hierarchy(TPMI_RH_HIERARCHY hierarchy)
             return ESYS_TR_RH_ENDORSEMENT;
         case TPM2_RH_PLATFORM:
             return ESYS_TR_RH_PLATFORM;
+        case TPM2_RH_PLATFORM_NV:
+            return ESYS_TR_RH_PLATFORM_NV;
         default:
-            return TSS2_ESYS_RC_BAD_VALUE;
+            LOG_ERR("An unknown hierarchy handle was passed: 0x%08x", hierarchy);
+            return 0xffffffff;
```
williamcroberts, line 1086: This will cause something else to fail right? Is this an exceptional case, can someone actually reach this?

lib/tpm2.c
```diff
@@ -2245,9 +2253,9 @@ tool_rc tpm2_hierarchycontrol(ESYS_CONTEXT *esys_context,
         goto tpm2_hierarchycontrol_skip_esapi_call;
     }
 
-
     TSS2_RC rval = Esys_HierarchyControl(esys_context, auth_hierarchy->tr_handle,
-            shandle, ESYS_TR_NONE, ESYS_TR_NONE, enable, state);
+            shandle, ESYS_TR_NONE, ESYS_TR_NONE,
+            fix_esys_hierarchy(enable), state);
```
williamcroberts, line 2258: make these separate calls, and return a proper error. just passing a magic ~0 seems like a bad idea. The error occurs in one place with the LOGE but the failure will be in the actual ESYS calls ... seems odd

tools/tpm2_hierarchycontrol.c
```diff
@@ -138,7 +138,6 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
         LOG_ERR("Invalid authorization");
         return rc;
     }
-
```
williamcroberts, line 141: unrelated from the patch, drop

lib/tpm2.c
AndreasFuchsSIT: I thought `/* FALLTHRU */` was only needed where the `case` statements were not in consecutive lines ?

lib/tpm2.c
AndreasFuchsSIT: Should not be reachable, but was reached before this PR. So this is more for developers than users.
I could make this an `assert(0)` if you prefer.

tools/tpm2_hierarchycontrol.c
AndreasFuchsSIT: ACK

lib/tpm2.c
AndreasFuchsSIT: See above, this is more for debugging purposes.
I did not want to change the function prototype from what it was before.
I guess it could be `TSS2_RC fix_esys_hierarchy(TPM2_HANDLE in, ESYS_TR *out);` and then we'd have actual explicit error handling, but that's going to be some more code...

AndreasFuchsSIT: Wrong mistake here, it would be `TSS2_RC fix_esys_hierarchy(TPM2_HANDLE in, uint32_t *out);` of course...

williamcroberts: Im not tracking, we're not changing any function prototypes I was thinking:
```C

   TSS2_RC rval = fix_esys_hierarchy(enable, &enable)
   if (rval != TSS2_RC_SUCCESS) {
       LOG_ERR();
       return rc/goto;
   }

    rval = Esys_HierarchyControl(esys_context, auth_hierarchy->tr_handle,
            shandle, ESYS_TR_NONE, ESYS_TR_NONE, enable, state);
            shandle, ESYS_TR_NONE, ESYS_TR_NONE,
            enable, state);
```

lib/tpm2.c
williamcroberts: Oh yeah, I didn't realize that. It's only really if theirs some code in the statements....

lib/tpm2.c
williamcroberts: Id rather just make this an error and have the caller handle it

lib/tpm2.c
AndreasFuchsSIT: I know what you thought...

What I meant: I did not create `fix_esys_hierarchy()`. That function was already in there and I did not want to change its signature.

If you prefer that I can do it of course...

williamcroberts:
@AndreasFuchsSIT here is a patch you can apply
[patch.diff.txt](https://github.com/tpm2-software/tpm2-tools/files/4594569/patch.diff.txt)

AndreasFuchsSIT:
Already done...

lib/tpm2.c
```diff
@@ -1054,33 +1054,53 @@ tool_rc tpm2_evictcontrol(ESYS_CONTEXT *esys_context,
...
+    switch (in) {
+        case ESYS_TR_RH_NULL:
+            /* FALLTHRU */
+        case ESYS_TR_RH_OWNER:
+            /* FALLTHRU */
+        case ESYS_TR_RH_ENDORSEMENT:
+            /* FALLTHRU */
+        case ESYS_TR_RH_PLATFORM:
+            /* FALLTHRU */
+        case ESYS_TR_RH_PLATFORM_NV:
```
williamcroberts, line 1077: I think you're right, we don't need fallthru, but they do make it explicit that you didn't intend to do something somewhere.

williamcroberts:
Im backporting this now, when I have it staged on the other branches ill merge.

williamcroberts:
> Fix those cases of TPM2_RH_ to ESYS_TR_RH_ translations that were missed in [780800c](https://github.com/tpm2-software/tpm2-tools/commit/780800c0be69a49b9097f8eae653cdb0623d2100) as discovered in #1994
> 
> I hope I did not miss any further cases, but at least tests are now successful.
> 
> This should also be backported to 4.1.x and 4.2.x as well.

Backport:
4.1.X: #2005
4.2.X: #2006

PR #2004, opened 2020-05-08
tpm2_sign: Fix for signature failures when key is restricted type

idesai:

Ideally this should never have been in the sign tool but in hash tool instead, but this needs to go in for options portability.
~~- [ ] Add hierarchy support~~
Since an auth isnt required, let's stick with the owner hierarchy for the validation ticket and use it internally in the tool.

### DIFF
--- a/man/tpm2_sign.1.md
+++ b/man/tpm2_sign.1.md
@@ -2,7 +2,7 @@
 
 # NAME
 
-**tpm2_sign**(1) - Sign a hash using the TPM.
+**tpm2_sign**(1) - Sign a hash or message using the TPM.
 
 # SYNOPSIS
 
@@ -10,12 +10,19 @@
 
 # DESCRIPTION
 
-**tpm2_sign**(1) - Signs an externally provided message or hash with the
-specified symmetric or asymmetric signing key.
+**tpm2_sign**(1) - Generates signature of specified message or message-digest
+using the specified symmetric or asymmetric signing key.
 
-If the signing key is a restricted signing key, then validation can be provided
-via the **-t** output. The ticket indicates that the TPM performed the hash of
-the message.
+When signing a message, **tpm2_sign** utility first calculates the digest of the
+message similar to the **tpm2_hash** command. It also generates a validation
+ticket under TPM2_RH_NULL or TPM2_RH_OWNER hierarchies respectively for
+unrestricted or the restricted signing keys.
+
+While signing messages is a provision in this tool it is recommended to use the
+**tpm2_hash** tool first and pass the digest and validation ticket.
+
+NOTE: If the signing key is a restricted signing key, then validation and digest
+must be provided via the **-t** input. The ticket indicates that the TPM performed the hash of the message.
 
 # OPTIONS
 
--- a/tools/tpm2_sign.c
+++ b/tools/tpm2_sign.c
@@ -88,50 +88,42 @@ out:
 
 static tool_rc init(ESYS_CONTEXT *ectx) {
 
-    bool option_fail = false;
-
-    if (ctx.cp_hash_path && ctx.output_path) {
-        LOG_ERR("Cannout output signature when calculating cpHash");
-        return tool_rc_option_error;
-    }
-
-    if (!ctx.signing_key.ctx_path) {
-        LOG_ERR("Expected option c");
-        option_fail = true;
-    }
-
-    if (!ctx.flags.o && !ctx.cp_hash_path) {
-        LOG_ERR("Expected option o");
-        option_fail = true;
-    }
-
-    if (option_fail) {
-        return tool_rc_option_error;
-    }
-
-    if (ctx.flags.d && ctx.flags.t) {
-        LOG_WARN("When using a pre-computed digest the validation ticket"
-                " is ignored.");
-    }
-
-    if (ctx.flags.d || !ctx.flags.t) {
-        ctx.validation.tag = TPM2_ST_HASHCHECK;
-        ctx.validation.hierarchy = TPM2_RH_NULL;
-        memset(&ctx.validation.digest, 0, sizeof(ctx.validation.digest));
-    }
-
     /*
-     * Set signature scheme for key type, or validate chosen scheme is allowed for key type.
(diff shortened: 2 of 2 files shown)


idesai:
@tstruk any idea what is the "main" failure in the checks?

tstruk:
> @tstruk any idea what is the "main" failure in the checks?

That's cirrus CI running the build on FreeBSD. Don't worry about this fail. It is fixed on master with this commit: https://github.com/tpm2-software/tpm2-tools/commit/719615f746a9179e77dc4030810ad64e97e7aace

williamcroberts:
> @tstruk any idea what is the "main" failure in the checks?

That's cirrus cloud CI and is building on freeBSD, you should be able to follow the links to build logs if needed.

tools/tpm2_sign.c
```diff
@@ -88,83 +88,93 @@ static tool_rc sign_and_save(ESYS_CONTEXT *ectx) {
...
-        LOG_WARN("When using a pre-computed digest the validation ticket"
-                " is ignored.");
-    }
-
-    if (ctx.flags.d || !ctx.flags.t) {
-        ctx.validation.tag = TPM2_ST_HASHCHECK;
-        ctx.validation.hierarchy = TPM2_RH_NULL;
-        memset(&ctx.validation.digest, 0, sizeof(ctx.validation.digest));
+    if (!ctx.flags.d && ctx.flags.t) {
+        LOG_WARN("Invalid validation ticket since no digest specified.");
```
williamcroberts, line 119: let them know that were ignoring -t still

tools/tpm2_sign.c
```diff
@@ -88,83 +88,93 @@ static tool_rc sign_and_save(ESYS_CONTEXT *ectx) {
...
         if (!input) {
             LOG_ERR("Could not open file \"%s\"", ctx.input_file);
             return tool_rc_general_error;
         }
 
-        rc = tpm2_hash_file(ectx, ctx.halg, TPM2_RH_NULL, input, &ctx.digest,
-                NULL);
+        TPMT_TK_HASHCHECK *temp_validation_ticket;
+        rc = tpm2_hash_file(ectx, ctx.halg, TPM2_RH_OWNER, input, &ctx.digest,
+                &temp_validation_ticket);
```
williamcroberts, line 135: Does this always work, even if the object is in a different hierarchy? It either has to always work, or we leave this advanced feature for discrete tpm2_hash calls and clearly spell this out in the manpage.

tools/tpm2_sign.c
```diff
@@ -88,83 +88,93 @@ static tool_rc sign_and_save(ESYS_CONTEXT *ectx) {
...
-    /* else process it as a pre-computed digest */
+    /*
+     * else process it as a pre-computed digest
+     */
     ctx.digest = malloc(sizeof(TPM2B_DIGEST));
     if (!ctx.digest) {
         LOG_ERR("oom");
         return tool_rc_general_error;
     }
-
```
williamcroberts, line 160: not applicable to this patch series

tools/tpm2_sign.c
idesai: It does always work. Although I really don't like duplicating the hashing code in the signing tool. Perhaps it still needs to be around for compatibility reasons. What are your thoughts on dropping the hashing mechanisms from signing tool and only sign digests with or without validation tickets?

williamcroberts: It needs to be around for compatibility reasons, if we make the default behavior a digest and not a message, it would break all existing callers (5.0 release). I would just leave it, set it as owner hierachy **and document it in the manpage**. So folks know if they want more features they need to call tpm2_hash and then tpm2_sign for full control.

man/tpm2_sign.1.md
```diff
@@ -2,20 +2,29 @@
...
-**tpm2_sign**(1) - Signs an externally provided message or hash with the
-specified symmetric or asymmetric signing key.
+**tpm2_sign**(1) - Generates signature of specified message or message-digest
+using the specified symmetric or asymmetric signing key.
 
-If the signing key is a restricted signing key, then validation can be provided
-via the **-t** output. The ticket indicates that the TPM performed the hash of
-the message.
+When signing a message, **tpm2_sign** utility first calculates the digest of the
+message like in the **tpm2_hash**. It also generates a validation ticket under
```
williamcroberts, line 17: Grammer:
```When signing a message, **tpm2_sign** utility first calculates the digest of the
message like in the **tpm2_hash**. It also generates a validation ticket under..``

"like in the tpm2_hash" does not compute. Try:

```When signing a message, **tpm2_sign** utility first calculates the digest of the
message simialr to the **tpm2_hash(1)** command. It also generates a validation ticket under...```

File: man/tpm2_sign.1.md [hunk 5]
```diff
@@ -2,20 +2,29 @@
...
+TPM2_RH_NULL or TPM2_RH_OWNER hierarchies respectively for un-restricted or the
```
williamcroberts, line 18: unrestricted doesn't need a hyphen

man/tpm2_sign.1.md
```diff
@@ -2,20 +2,29 @@
...
+restricted signing keys.
+
+While signing messages is a provision in this tool it is recommended to use the
```
williamcroberts, line 21: "...is a provision in this tool, ..." needs a comma after tool

man/tpm2_sign.1.md
```diff
@@ -2,20 +2,29 @@
...
+**tpm2_hash** tool first and pass the digest and validation ticket for
+flexibility in choice of message length and hierarchy of choice for generating
```
williamcroberts, line 23: How does it affect message length?

idesai: Ah... i was looking at the lib/tpm2_hash.c where message lengths longer than max buffer were split and hashed in loops. I will drop that comment.

williamcroberts: > Ah... i was looking at the lib/tpm2_hash.c where message lengths longer than max buffer were split and hashed in loops. I will drop that comment.

Im confused, if you were looking at that code, wouldn't that mean that message lengths of any size are supported?

PR #2010, opened 2020-05-08
FAPI: deduplicate variable name

AndreasFuchsTPM:
lgtm reported on a local variable with the same name as a global. Renamed local variable so we don't hide the global anymore.

### DIFF
--- a/tools/fapi/tss2_template.c
+++ b/tools/fapi/tss2_template.c
@@ -470,25 +470,25 @@ char* ask_for_password() {
 #else /* FAPI_3_0 */
     char *pw;
 #endif /* FAPI_3_0 */
-    char *password = NULL;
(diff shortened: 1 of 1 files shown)


tools/fapi/tss2_template.c
```diff
@@ -470,25 +470,25 @@ char* ask_for_password() {
 #else /* FAPI_3_0 */
     char *pw;
 #endif /* FAPI_3_0 */
-    char *password = NULL;
```
williamcroberts, line 473: this is why when I do globals, I make them start with an underscore or put them in a structure with a variable called global or something distinct... ive had these bugs bite me in the A$$

williamcroberts (approved):

PR #2016, opened 2020-05-11
travis: add ubuntu-20.04 and drop ubuntu-16.04

tstruk:


tstruk:
@williamcroberts @pdxjohnny @AndreasFuchsSIT is ubuntu-16.04 docker image used anywhere else in TSS except tpm2-tss and tpm2-tools? If not I will rm it from tpm2-software-container after this is merged.

williamcroberts:
@tstruk seems to be what you said and tpm2-pkcs11:
```bash
$ for d in `find . -maxdepth 1 -type d -name tpm2-\*`; do grep -q ubuntu $d/.travis.yml 2>/dev/null; if [ "$?" -eq 0 ]; then echo $d; fi; done;
./tpm2-pkcs11
./tpm2-tss
./tpm2-tools
```

Yeah once we have no reference to it, kill it.

williamcroberts:
@tstruk can you do the add as one patch and the rm as a separate patch, so if for some reason we can just revert them independently.

PR #2018, opened 2020-05-12
[RFC] tpm2_netlisten command

williamcroberts:
I still have a bunch of cleanup to do, but wanted to solicit feedback. I attached a general diagram below, The idea is that it takes raw tpm command buffers and just forwards them to the TCTI of choice.

Depends on:
https://github.com/tpm2-software/tpm2-tss/pull/1727

[network-tcti.pdf](https://github.com/tpm2-software/tpm2-tools/files/4616049/network-tcti.pdf)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -114,6 +114,7 @@ bin_PROGRAMS += \
     tools/tpm2_load \
     tools/tpm2_loadexternal \
     tools/tpm2_makecredential \
+    tools/tpm2_netlisten \
     tools/tpm2_nvdefine \
     tools/tpm2_nvextend \
     tools/tpm2_nvincrement \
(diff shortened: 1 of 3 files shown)


AndreasFuchsSIT:
Comments over at https://github.com/tpm2-software/tpm2-tss/pull/1727

tstruk:
Shouldn't that be a system service/daemon, probably started/controlled by systemd instead of a command? Who will run this command and when?

williamcroberts:
> Shouldn't that be a system service/daemon, probably started/controlled by systemd instead of a command? Who will run this command and when?

Its just a process, who starts it doesn't matter. You could do it as a systemd service, and initrc service, run on the command line, doesn't matter. But this can be replaced with tcpserver and some minor changes to tpm2_send command, so we don't need this. I just learned tcpserver command exists.

Secondly, @AndreasFuchsSIT has an interesting idea where we just make a tcti that forks a child process and exexs whatever you want, and then opens stdin/stdout to communicate with it and kills the child process on TCTI finalization. So you can just interact with whatever you want to throw in there.

So with something like that, and existing network/ssh toolkits, you should be able to build whatever you need simply.

PR #2035, opened 2020-05-19
TSS2 Tools Backport Fixes

cplappert:
- Several fixes for the tss2 tools which also should be back ported to 4.2.X:
  - Fix some double-free errors
  - Shorthand command -f does not need argument
  - Update tss2_encrypt to the new FAPI interface
  - An additional \n was appended when redirecting to stdout
  - Adapt optional and mandatory parameters for tools in code
  - Extract Fapi version from Fapi_GetInfo
  - Error handling in case of multiple inputs and/or outputs from/to stdin/stdout
  - Fix and update content of man pages 
  - Add parameter types to all man pages 

Signed-off-by: Christian Plappert christian.plappert@sit.fraunhofer.de

### DIFF
--- a/tools/fapi/tss2_template.c
+++ b/tools/fapi/tss2_template.c
@@ -367,7 +367,6 @@ int open_write_and_close(const char* path, bool overwrite, const void *output,
             fprintf (stderr, "write(2) to stdout failed: %m\n");
             return 1;
         }
-        putchar ('\n');
         return 0;
     }
 
--- a/tools/fapi/tss2_template.c
+++ b/tools/fapi/tss2_template.c
@@ -123,25 +123,42 @@ static tpm2_option_code tss2_handle_options (
                 fprintf (stderr, "Not enough memory\n");
                 goto out;
             }
-            char *fapi_version;
-            TSS2_RC ret = Fapi_GetInfo(fctx, &fapi_version);
+            char *info;
+            TSS2_RC ret = Fapi_GetInfo(fctx, &info);
             if (ret != TSS2_RC_SUCCESS) {
                 fprintf (stderr, "Fapi_GetInfo returned %u\n", ret);
                 free(prog_name);
                 goto out;
             }
-            char* target = strndup(fapi_version, 50); // TODO: Extract correct version
-            /* N.B. the returned version is UTF-8 and printing it correct
-             * depends on the locale */
-            Fapi_Free(fapi_version);
-            printf ("%s from " PACKAGE " version " VERSION " using at run time "
-                    "FAPI library version %s.\nCopyright (C) 2019 Fraunhofer SI"
-                    "T.\nLicense BSD-3-Clause.\nMy homepage is https://github.com/tpm2"
-                    "-software/tpm2-tools .  The homepage of the FAPI library i"
-                    "s https://github.com/tpm2-software/tpm2-tss .\n",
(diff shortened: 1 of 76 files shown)


tools/fapi/tss2_template.c
```diff
@@ -123,25 +123,42 @@ static tpm2_option_code tss2_handle_options (
...
+                t = t + strlen("\"version\"");
+                version = (char*) malloc(strlen(t) + 1);
+                if (!version) {
+                    fprintf (stderr, "malloc(2) failed: %m\n");
+                    return 1;
+                }
+                ret = sscanf(t, "%*[^\"]\"%[^\"]%*[*]", version);
+                /* Version string is not larger than 128 characters */
+                if (ret!=1 || strlen(version) > 128 ) {
+                    version = "not found";
```
williamcroberts, line 146: You just orphaned the pointer, you need free(version) before you reseat it.

tools/fapi/tss2_template.c
```diff
@@ -123,25 +123,42 @@ static tpm2_option_code tss2_handle_options (
...
+            char *version = NULL;
+            char *t = strstr(info, "\"version\"");
+            if (t) {
+                t = t + strlen("\"version\"");
+                version = (char*) malloc(strlen(t) + 1);
+                if (!version) {
+                    fprintf (stderr, "malloc(2) failed: %m\n");
+                    return 1;
+                }
+                ret = sscanf(t, "%*[^\"]\"%[^\"]%*[*]", version);
```
williamcroberts, line 143: LGTM is complaining that the malloc doesn't include space for a terminating null character, but as far as I can work out from the most complicated format specifier I've ever layed eyes on, is that it should be fine, as:
- scanf family nul terminates strings
- the malloc is whats left of t past version, and some of it is discarded so it will always fit

I Ran this in a fuzzer:
```C
#include <stdlib.h>
#include <string.h>
#include <stdio.h>
#include <stdint.h>

int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {

    // LibFuzzer doesn't have an option to NUL term, boo!
    char *info = calloc(1, Size + 1);
    if (!info) {
	    fprintf(stderr, "%s", "oom\n");
	    return 0;
    }
    memcpy(info, Data, Size);

    char *version = NULL;
    char *t = strstr(info, "\"version\"");
    if (t) {
        t = t + strlen("\"version\"");
        version = (char*) malloc(strlen(t) + 1);
        if (!version) {
            fprintf (stderr, "malloc(2) failed: %m\n");
            goto out;
        }
        
	int ret = sscanf(t, "%*[^\"]\"%[^\"]%*[*]", version);
        if (ret!=1 || strlen(version) > 128 ) {
	    // NEED TO ADD
            free(version);
            version = "not found";
            t = NULL;
        }
    } else {
        version = "not found";
    }
    
    printf("version=\"%s\"\n", version);

    if (t) {
        free(version);
    }

out:
    free(info);
    return 0;
}
```
```bash
clang -o fuzz -g -O1 -fsanitize=fuzzer,address version-fuzz.c
./fuzz
```
Why not just plop in the m modifier to the sscanf and have it allocate version? Then you can dump the version allocation there and just check sscanf for issues.

cplappert: Cool. Thanks for the additional research.
I will use the m modifier and fix above issue in a new PR.

PR #2044, opened 2020-05-22
tss_*: Fix orphaned pointer and let sscanf allocate memory

cplappert:
This PR addresses the comments in #2035/#2036

- Deallocate the version pointer before reseating again. 
- Improve code by letting sscanf allocate memory itself.


williamcroberts:
Travis is passing, but Cirrus CI is failing complaining that tpm_server doesn't exist.
I re toggled it, not sure why that dependency is not being fulfilled. Well see what happens again.

PR #2074, opened 2020-05-29
Several bugfixes for the fapi tools

cplappert:
Some fixes for the tss2 tools: 
- tss2_*: Fix bash-completion for tss2_pcrextend and tss2_verifysignature
- tss2_*: Add force option to tss2_list
- tss2_*: Consistent force option for all fapi tools
- tss2_*: Do not decode non-TPM errors
- tss2_*: Enhance integration tests to test changes of optional/mandatory parameters
- tss2_*: Add --hex parameter to tss2_getrandom
- tss2_*: Fix autocompletion issue in all fapi tools
- tss2_*: Switch tss2_* to with-"="-style
- tss2_*: Add size parameter to tss2_createseal

### DIFF
--- a/dist/bash-completion/tpm2-tools/tss2_authorizepolicy
+++ b/dist/bash-completion/tpm2-tools/tss2_authorizepolicy
@@ -18,8 +18,8 @@ _tss2_authorizepolicy()
 
     $split && return
 
-    COMPREPLY=( $(compgen -W "-h --help -v --version --policyPath -P
-      --keyPath -p --policyRef -r" -- "$cur") )
+    COMPREPLY=( $(compgen -W "-h --help -v --version --policyPath= -P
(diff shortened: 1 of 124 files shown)


cplappert:
The fapi-encrypt-decrypt test fails with an older TSS version, since it do not support key creation without type. Should this one statement in the test be removed?

cplappert:
> The fapi-encrypt-decrypt test fails with an older TSS version, since it does not support key creation without type. Should this one statement in the test be removed?

Resolved this issue by commenting out the statement in the test with requirement for tpm2-tss>= 2.4.2

cplappert:
> The only change I have is to test the fapi version and conditional that test.

Ok. I implemented this in the current version.

cplappert:
> > The only change I have is to test the fapi version and conditional that test.
> 
> Ok. I implemented this in the current version.

PR #2075, opened 2020-05-29
Don't use unsupported c flags when compiling with clang

tstellar:


tstruk:
> @tstruk I'm trying to extend your fix to include clang on all platforms not just BSD.

@tstellar looks good to me, but @idesai or @williamcroberts need to make the call on this one.

configure.ac
```diff
@@ -266,7 +266,7 @@ add_c_flag([-D_GNU_SOURCE], [AC_MSG_ERROR([Cannot enable -D_GNU_SOURCE])])
 # Best attempt compiler options that are on newer versions of GCC that
 # we can't widely enforce without killing other peoples builds.
 # Works with gcc only. Needs to be disabled on BSD and clang
-AS_IF([test "$HOSTOS" = "Linux"],
+AS_IF([test "$CC" = "gcc"],
```
williamcroberts, line 269: Will this work if CC is set to something like gcc-8?

tstruk: No that's just a string comparison, but the default compiler should be used with something like
update-alternatives --set gcc /usr/bin/gcc-8

williamcroberts: > No that's just a string comparison, but the default compiler should be used with something like
> update-alternatives --set gcc /usr/bin/gcc-8

That's what I thought, but we can't make folks switch their system wide compiler to have to switch it back. That also gets used for building python bindings, kernel modules, etc. In some of those scripts.

It dumps it to the shell, not sure if it's bash so I won't do the wildcard pattern matting in the [[ ]] double brackets, but we could do something like:
```bash
echo $CC | grep gcc
```
Then it won't really care, as long as the filename has gcc it will do it.

diabonas:
Could you test #2098, please, as it appears to be tackling the same issue as this PR and is more robust then hardcoding specific compiler names?

williamcroberts:
Closing this. Looks like between this and #2098 the fix is going in a different direction. Feel free to submit a new PR.

PR #2076, opened 2020-05-30
Make all of the helper functions in each tool static

osresearch:
A few of the program had various helper functions exported; this makes them all static so that they do not conflict with each other when combined in a busybox style executable.

It also identified that `tools/tpm2_policynv.c` and `tools/tpm2_policyauthorizenv.c` do not use their `is_input_option_args_valid()` function, and that a few tools had `tpm2_onexit()` functions where it seems like `tpm2_tool_onexit()` was intended.

### DIFF
--- a/tools/tpm2_createek.c
+++ b/tools/tpm2_createek.c
@@ -419,7 +419,7 @@ out:
     return rc;
 }
 
-void tpm2_onexit(void) {
+void tpm2_tool_onexit(void) {
 
     tpm2_hierarchy_pdata_free(&ctx.objdata);
 }
--- a/tools/tpm2_policyauthorizenv.c
+++ b/tools/tpm2_policyauthorizenv.c
@@ -80,7 +80,8 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
     return *opts != NULL;
 }
 
-bool is_input_option_args_valid(void) {
+#if 0
+static bool is_input_option_args_valid(void) {
 
     if (!ctx.session_path) {
         LOG_ERR("Must specify -S session file.");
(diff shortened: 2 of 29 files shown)


osresearch:
The motivation for making this change is in #2077, although it needs some assistance in figuring out how to integrate with the autoconf generated `Makefile`

tools/tpm2_policynv.c
```diff
@@ -157,7 +157,8 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
     return *opts != NULL;
 }
 
-bool is_input_option_args_valid(void) {
+#if 0
+static bool is_input_option_args_valid(void) {
```
williamcroberts, line 161: probably want to call this in the beginning of `tpm2_tool_onrun`.

tools/tpm2_policyauthorizenv.c
```diff
@@ -80,7 +80,8 @@ bool tpm2_tool_onstart(tpm2_options **opts) {
     return *opts != NULL;
 }
 
-bool is_input_option_args_valid(void) {
+#if 0
+static bool is_input_option_args_valid(void) {
```
williamcroberts, line 84: Probably want to call this in tpm2_tool_onrun.

tools/tpm2_createek.c
```diff
@@ -419,7 +419,7 @@ tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
     return rc;
 }
 
-void tpm2_onexit(void) {
+void tpm2_tool_onexit(void) {
```
williamcroberts, line 422: I am quite surprised we never saw a leak. But yes, this is correct.

osresearch:
I rebased to HEAD and added the parameter checking calls to the two programs.

PR #2077, opened 2020-05-30
Bundle all binaries into a single busybox style executable

osresearch:
It is convenient to have a single executable for ease of installation in initrd or embedded systems; this RFC pull request adds a `bundle` script that uses `objcopy --redefine-sym` to rename all of the `tpm2_tool_onrun()`, `tpm2_tool_onstart()`, `tpm2_tool_onstop()` and `tpm2_tool_onexit()` functions to `${TOOL_NAME}_onrun()` etc.  A modified version of `tpm2_tool.c` looks at the first argument and uses that to determine which handler functions to call.

The current script is a bit of a hack to demonstrate that it works and is feasible; figuring out how to integrate it into the autoconf build might be a challenge. Please offer suggestions for how to improve this patch!

This depends on #2076 to ensure that there are no symbol overlaps between the various `tools/*.o` files. The real diff to examine is in https://github.com/tpm2-software/tpm2-tools/commit/c466c280db00b91d24f3e3fe1a3435859f2f5dec


osresearch:
It looks like the single `tpm2` program idea was briefly discussed back in 2017, although the implementation then was a wrapper that called `exec()` on the other binary: https://github.com/tpm2-software/tpm2-tools/pull/575/commits/6bf38edffd35a88e917fccc6badaad779a41bdb6 

This patch differs in that it bundles all of them into a single executable so that there are fewer programs necessary to install into an `initrd` or other embedded system. The resulting `tpm2` binary is 571KB after stripping (112KB gzip'ed); all 97 of the `tpm2_*` tools and misc are 5.3MB after stripping (1.8MB gzip'ed).

This patch could be extended to work like busybox where `argv[0]` is used to select the command, which would allow all of the subcommands to be replaced with symlinks in the installation.

osresearch:
```
/tmp/cirrus-ci-build/test/integration/helpers.sh: line 286: tpm_server: command not found
Could not start simulator at port: 63545
/tmp/cirrus-ci-build/test/integration/helpers.sh: line 304: kill: (79212) - No such process
Tries left: 0
Maximum attempts reached. Aborting
Shutting down
```
Not sure what happened to `tpm_server` in the CI -- none of the rest of the build should be affected since this only touches the new `bundle/` directory.

bundle/tpm2_bundle.c
```diff
@@ -0,0 +1,219 @@
...
+    esys_teardown(esys_context);
+    Tss2_TctiLdr_Finalize(&tcti_context);
+}
+
+static ESYS_CONTEXT* ctx_init(TSS2_TCTI_CONTEXT *tcti_ctx) {
+
+    TSS2_ABI_VERSION abi_version = SUPPORTED_ABI_VERSION;
+    ESYS_CONTEXT *esys_ctx;
+
+    TSS2_RC rval = Esys_Initialize(&esys_ctx, tcti_ctx, &abi_version);
```
williamcroberts, line 70: You can just pass NULL for the ABI version. I'm assuming this is a copy paste, so we're doing it in tools.

osresearch: Right, that's a direct copy of the `tools/tpm2_tool.c` program: https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_tool.c#L52

osresearch:
I'd prefer to find a way to have the busybox style built by default, in addition to the standalone tools, so that it eventually becomes part of the system packages and doesn't require a custom build to integrate into the initramfs or dracut environments.

The `objcopy` trick to rename the symbols in the `*.o` files is perhaps not as portable as renaming all of the entry points in the various tool source code or using a constructor segment, although the use of `__weak__` symbols is also ready a compiler specific technique, so perhaps that is not a concern.

williamcroberts:
> I'd prefer to find a way to have the busybox style built by default, in addition to the standalone tools, so that it eventually becomes part of the system packages and doesn't require a custom build to integrate into the initramfs or dracut environments.

Start with optional, then we can turn it on by default. Many distro packagers set their own configure, CFLAGS, etc as well. So this necessarily wouldn't prevent them from building with or without it. The reason we don't just do a busybox command-let as the only option and by default, which I wanted to do when I started the project, is size. Most of the embedded IoT crowd just copy the tools they need to the remote target and they were unwilling to budge. Even though the vast majority of the code is in the library that is linked statically.

> 
> The `objcopy` trick to rename the symbols in the `*.o` files is perhaps not as portable as renaming all of the entry points in the various tool source code or using a constructor segment, although the use of `__weak__` symbols is also ready a compiler specific technique, so perhaps that is not a concern.

It's easier if we can just rename, the less magic the better.

osresearch:
I removed almost all of the magic (including the `__attribute__((__weak__))` annotations that were in the `tpm2_tools.h` header), which unfortunately involved touching almost every file to make their methods static.  Now each tool `.c` file has a struct with its methods:
```
typedef struct {
	const char * name;
	tpm2_tool_onstart_t onstart;
	tpm2_tool_onrun_t onrun;
	tpm2_tool_onstop_t onstop;
	tpm2_tool_onexit_t onexit;
} tpm2_tool;
```

The new magic is in a macro that registers the command with `tpm2_tool.c` in a constructor and then `main()` consults the `argv[0]` to determine which one to run: https://github.com/tpm2-software/tpm2-tools/pull/2077/files#diff-d04f495181baf3e61d74bd26450844ac

All of the standalone tools default to running their pre-existing methods; the only one that registers more than a single tool is the `tpm2_tool` program. I wasn't able to figure out how to expand variables in the `Makefile.am`, so this has all of the tool explicitly listed (ideally it would use the `$(bin_PROGRAMS)` variable instead).

What are your thoughts on this approach?

williamcroberts:
Im going to add this to 4.3 release.

osresearch:
The only issue that might come up for embedded applications is that the `tpm2_getekcertificate` tool wants to link against libcurl.  It could still be built as a standalone executable to avoid adding that requirement to the bundled executable.

williamcroberts:
> The only issue that might come up for embedded applications is that the `tpm2_getekcertificate` tool wants to link against libcurl. It could still be built as a standalone executable to avoid adding that requirement to the bundled executable.

I've never seen an embedded device that hasn't had curl in a long time. But we can make a configure switch to disable tools if we get push back.

`--disable-tools='a,b,c,d...'`

That's how *ideally* i'd like it to work, takes a CSV of tools to remove. Then when we add to the sources we just don't add it. We can loop by using $(subst) to swap , (comma) for spaces and then the foreach should work on that space delimited list.

But this is in the off chance we need it. I wouldn't worry about it now, but +1 if you add it.

FYI I am out the rest of the week. Ill be back on Monday, so I can pick up the discussion and merge this as soon as it's ready. Thanks again.

osresearch:
I've added a loop in `install-exec-hook` to symlink all of the tools that go into the `tpm2` program: b803931846324d06cc9457093b2c9408f8bb8272

One complication with making the symlinks in the `install-exec-hook` is that they might not be available during the tests, so I think it is necessary to update all of the test scripts to call `tpm2 foo` instead of `tpm2_foo`. Not positive; haven't figured out how to run the tests manually yet.

williamcroberts:
Commit 0b958e7707446 is applied tree-wide, but we always need to make sure that we keep backwards compat with the older `tpm2_` prefix symlinks. So make sure those still work. I would just keep a few tests with the older tpm2_ and some with the commandlet style. This way we have coverage on the distpatch table.

tools/tpm2_tool.c
```diff
@@ -61,18 +62,101 @@ static ESYS_CONTEXT* ctx_init(TSS2_TCTI_CONTEXT *tcti_ctx) {
...
+}
+
+static const tpm2_tool * tpm2_tool_lookup(int *argc, char *** argv)
+{
+    // no tools linked in?
+    if (!tools)
+        return NULL;
+
+    // only one tool linked in? use it no matter the program name.
+    if (tool_count == 1)
```
williamcroberts, line 102: no, if you run tpm2 just die with  expected commandlet name or something. Consistent with other commandlets like git.

tools/tpm2_tool.c
```diff
@@ -61,18 +62,101 @@ static ESYS_CONTEXT* ctx_init(TSS2_TCTI_CONTEXT *tcti_ctx) {
...
+static const tpm2_tool ** tools;
+static unsigned tool_count;
+
+void tpm2_tool_register(const tpm2_tool * tool)
+{
+    if (!tools)
+        tools = calloc(1024, sizeof(*tools));
+
+    //printf("%s: register %s\n", __func__, tool->name);
+    if (tools)
```
williamcroberts, line 77: brackets on everything.

tools/tpm2_tool.c
```diff
@@ -61,18 +62,101 @@ static ESYS_CONTEXT* ctx_init(TSS2_TCTI_CONTEXT *tcti_ctx) {
...
+        tools[tool_count++] = tool;
```
williamcroberts, line 78: This could overflow if someone added a bazillion tools and didn't update this, make sure bounds checking is done and it complains loudly.

For extra points we could calculate the tools size in the Makefile and pass this in so the list is just the right size and we don't have waste.

tools/tpm2_tool.c
williamcroberts: Looks like return NULL.

tools/tpm2_tool.c
```diff
@@ -61,18 +62,101 @@ static ESYS_CONTEXT* ctx_init(TSS2_TCTI_CONTEXT *tcti_ctx) {
...
+    return NULL;
+}
+
+
 /*
  * This program is a template for TPM2 tools that use the SAPI. It does
  * nothing more than parsing command line options that allow the caller to
  * specify which TCTI to use for the test.
  */
-int main(int argc, char *argv[]) {
```
williamcroberts, line 69: Don't change stuff like this, here and throughout. We do lists of pointers with the array brackets.

williamcroberts:
> One complication with making the symlinks in the `install-exec-hook` is that they are not be available during the tests, so I think it is necessary to update all of the test scripts to call `tpm2 foo` instead of `tpm2_foo`:

Oh duh, ignore my other statement.

tools/tpm2_tool.c
```diff
@@ -61,18 +62,101 @@ static ESYS_CONTEXT* ctx_init(TSS2_TCTI_CONTEXT *tcti_ctx) {
     return esys_ctx;
 }
 
+/*
+ * Build a list of the TPM2 tools linked into this executable
+ */
+static const tpm2_tool ** tools;
```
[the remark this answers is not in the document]
osresearch: That doesn't work since it would need to have a static size (and the `calloc()` removed):
```
tools/tpm2_tool.c:74:15: error: ‘tools’ has an incomplete type ‘const tpm2_tool *[]’ {aka ‘const struct <anonymous> *[]’}
   74 |         tools = calloc(1024, sizeof(*tools));
```

tools/tpm2_tool.c
osresearch: That's left over from when I was still building the tools as separate programs in addition to the bundled one.  That way if you copied `tpm2_pcrread` to `foo`, it would still read the PCRs since that tool had only a single handler in the array.  It can go away now.

tools/tpm2_tool.c
osresearch: This change was necessary because `tpm2_tool_lookup()` modified the command line based on if it was invoked as `tpm2` or as `tpm2_foo`.  With `char * argv[]` it is not possible to do the `argv++` to skip the 0th argument.

williamcroberts:
> -----Original Message-----
> From: Trammell Hudson <notifications@github.com>
> Sent: Monday, June 8, 2020 10:12 AM
> Cc: Roberts, William C <william.c.roberts@intel.com>; Comment
> <comment@noreply.github.com>
> Subject: Re: [tpm2-software/tpm2-tools] RFC: Bundle all binaries into a single
> busybox style executable (#2077)
> 
> @osresearch commented on this pull request.
> 
> ________________________________
> 
> In tools/tpm2_tool.c <https://github.com/tpm2-software/tpm2-
> tools/pull/2077#discussion_r436783713> :
> 
> > @@ -61,18 +62,101 @@ static ESYS_CONTEXT* ctx_init(TSS2_TCTI_CONTEXT
> *tcti_ctx) {
> 
>      return esys_ctx;
> 
>  }
> 
> 
> 
> +/*
> 
> + * Build a list of the TPM2 tools linked into this executable
> 
> + */
> 
> +static const tpm2_tool ** tools;
> 
> 
> That doesn't work since it would need to have a static size (and the calloc()
> removed):

Why even calloc it, just allocate it statically.

tools/tpm2_tool.c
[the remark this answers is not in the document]
williamcroberts: Can you just statically allocate it since the array size is static below, just heap allocated?

osresearch:
Woo! The tests are all passing again!

Setting the size statically requires having `make` compute the length of the `$(tpm2_tools)` list and pass it in via a command line flag.  I've added that, as well as a test that the invocation of the bundled executable in various ways.

williamcroberts:
> Woo! The tests are all passing again!
> 
> Setting the size statically requires having `make` compute the length of the `$(tpm2_tools)` list and pass it in via a command line flag. I've added that, as well as a test that the invocation of the bundled executable in various ways.

Well you could of just set it to 1024 like the calloc, but thats way better.

martinezjavier:
> @martinezjavier are you ok with this? Will this break anything?

I missed this pull-request, but I do like the idea a lot. And I can't think of anything that may break with this change.

tstruk:
Is the motivation for this change to make the initrd installation easier, or is it to change the style to more busybox like? I don't think there is any limitation in initrd with regards to number of files or size, is there? If the motivation is mainly style, we could have a simple shell wrapper "tpm2" that would call the appropriate executable:
```
$ cat /usr/bin/tpm2
#!/bin/sh

PARS=($@)
tpm2_$1 ${PARS[@]:1}
```
which would be called it like this:
```
$ tpm2 getrandom 32 -Tmssim | xxd
00000000: 62fa 60d9 f9a3 659a f102 27ba 4a96 a3e7  b.`...e...'.J...
00000010: 6613 2ded 3b34 411d 134e b091 5756 d369  f.-.;4A..N..WV.i
```

osresearch:
The goal is mostly for size -- I want to be able to include the `tpm2` executable in a Heads or Linuxboot firmware initrd image, which are typically limited to 4-16 MB total since they are stored in the flash on the mainboard.

williamcroberts:
> Is the motivation for this change to make the initrd installation easier, or is it to change the style to more busybox like? I don't think there is any limitation in initrd with regards to number of files or size, is there? If the motivation is mainly style, we could have a simple shell wrapper "tpm2" that would call the appropriate executable:
> 
> ```
> $ cat /usr/bin/tpm2
> #!/bin/sh
> 
> PARS=($@)
> tpm2_$1 ${PARS[@]:1}
> ```
> 
> which would be called it like this:
> 
> ```
> $ tpm2 getrandom 32 -Tmssim | xxd
> 00000000: 62fa 60d9 f9a3 659a f102 27ba 4a96 a3e7  b.`...e...'.J...
> 00000010: 6613 2ded 3b34 411d 134e b091 5756 d369  f.-.;4A..N..WV.i
> ```

Huge size win, anything more than 3 tools and it's a win and we don't have to go to a shared object for the library the tools carry around.

AndreasFuchsSIT:
@williamcroberts @idesai @osresearch Should this be done for tss2_* tools in a similar fashion ?

williamcroberts:
> @williamcroberts @idesai @osresearch Should this be done for tss2_* tools in a similar fashion ?

Yeah that'd be great if you got the bandwidth. I made a bug so we track this on 5.0. See #2105.

Issue #2072, opened 2020-05-28
tpm2_pcrread is not compatible with tpm2_checkquote


williamcroberts:
tpm2_checkquote and tpm2_quote have a different metadata format than tpm2_pcrread. tpm2_quote includes the pcr selection information, tpm2_pcrread does not, just the hashes in order.

We should:
make tpm2_pcrread to be able to output either binary format. We will call them plain, which is the current format, or metadata, which is the quote format. We probably want to version the metadata format.

tpm2_checkquote to be able to read any format, and know what format it is by a PCR selection string. If it's pcr selection string, then the plain tpm2_pcrread is being used. If no pcr slection string then the metadata format is being used.

williamcroberts:
@idesai you asked, "Current output from tpm2_pcrread is used between the tools so that will continue to be the default (plain) output. What should the new option be called to output PCR-selection information?"

Perhaps:

  * -f plain (current and default, just hashes)
  * -f meta (selection specifier and hashes)

I don't like the "meta" option, im open to suggestions and improvements on that.

Then in consuming tools if they:

1. Specify a PCR file without the PCR selection string, its meta.
2. Specify a PCR file with the PCR selection string, its plain.

idesai:
2 fixes need to go in
- [ ] Enable tpm2_checkquote to read in PCR-digest data without TPML_PCR_SELECTION header and add PCR-Selection input option.
- [ ] Enable tpm2_pcrread to output TPML_PCR_SELECTION by adding an option to make this non-default choice.

idesai:
> 2 fixes need to go in
> 
> * [x]  Enable tpm2_checkquote to read in PCR-digest data without TPML_PCR_SELECTION header and add PCR-Selection input option.
> * [ ]  Enable tpm2_pcrread to output TPML_PCR_SELECTION by adding an option to make this non-default choice.

Fix-1 through #2114

idesai:
> > 2 fixes need to go in
> > 
> > * [x]   Enable tpm2_checkquote to read in PCR-digest data without TPML_PCR_SELECTION header and add PCR-Selection input option.
> > * [ ]   Enable tpm2_pcrread to output TPML_PCR_SELECTION by adding an option to make this non-default choice.
> 
> Fix-1 through #2114

We agreed that it is sufficient to add support for PCR data without TPML_PCR_SELECTION in the tpm2_checkquote tool, so Fix-1 should suffice to close this issue.

PR #2093, opened 2020-06-08
tpm2_eventlog: compute PCR values based on the events

osresearch:
This patch computes the PCR values for the sha1 and sha256 hashes based on the events in the log and outputs an additional yaml section with the predicted PCRs to make it easy to compare to the attested quote.

### DIFF
--- a/lib/tpm2_eventlog.c
+++ b/lib/tpm2_eventlog.c
@@ -52,11 +52,28 @@ bool foreach_digest2(tpm2_eventlog_ctx_t * ctx, int pcrId, TCG_DIGEST2 const *di
         }
 
         uint8_t * pcr = NULL;
+        if (pcrId > TPM2_MAX_PCRS) {
+            LOG_ERR("PCR%d > max %d", pcrId, TPM2_MAX_PCRS);
+        } else
         if (alg == TPM2_ALG_SHA1) {
             pcr = ctx->sha1_pcrs[pcrId];
+            ctx->sha1_used |= (1 << pcrId);
         } else
         if (alg == TPM2_ALG_SHA256) {
             pcr = ctx->sha256_pcrs[pcrId];
+            ctx->sha256_used |= (1 << pcrId);
+        } else
+        if (alg == TPM2_ALG_SHA384) {
+            pcr = ctx->sha384_pcrs[pcrId];
+            ctx->sha384_used |= (1 << pcrId);
+        } else
+        if (alg == TPM2_ALG_SHA512) {
+            pcr = ctx->sha512_pcrs[pcrId];
+            ctx->sha512_used |= (1 << pcrId);
+        } else
+        if (alg == TPM2_ALG_SM3_256) {
+            pcr = ctx->sm3_256_pcrs[pcrId];
+            ctx->sm3_256_used |= (1 << pcrId);
         } else {
             LOG_WARN("PCR%d algorithm %d unsupported", pcrId, alg);
         }
--- a/lib/tpm2_eventlog.c
+++ b/lib/tpm2_eventlog.c
@@ -28,7 +28,7 @@ bool digest2_accumulator_callback(TCG_DIGEST2 const *digest, size_t size,
  * hold the digest. The size of the digest is passed to the callback in the
  * 'size' parameter.
  */
-bool foreach_digest2(tpm2_eventlog_ctx_t * ctx, int pcrId, TCG_DIGEST2 const *digest, size_t count, size_t size) {
+bool foreach_digest2(tpm2_eventlog_ctx_t *ctx, unsigned pcr_index, TCG_DIGEST2 const *digest, size_t count, size_t size) {
 
     if (digest == NULL) {
         LOG_ERR("digest cannot be NULL");
@@ -51,35 +51,30 @@ bool foreach_digest2(tpm2_eventlog_ctx_t * ctx, int pcrId, TCG_DIGEST2 const *di
             return false;
         }
 
-        uint8_t * pcr = NULL;
-        if (pcrId > TPM2_MAX_PCRS) {
-            LOG_ERR("PCR%d > max %d", pcrId, TPM2_MAX_PCRS);
-        } else
-        if (alg == TPM2_ALG_SHA1) {
-            pcr = ctx->sha1_pcrs[pcrId];
-            ctx->sha1_used |= (1 << pcrId);
-        } else
-        if (alg == TPM2_ALG_SHA256) {
-            pcr = ctx->sha256_pcrs[pcrId];
-            ctx->sha256_used |= (1 << pcrId);
-        } else
-        if (alg == TPM2_ALG_SHA384) {
-            pcr = ctx->sha384_pcrs[pcrId];
-            ctx->sha384_used |= (1 << pcrId);
-        } else
-        if (alg == TPM2_ALG_SHA512) {
-            pcr = ctx->sha512_pcrs[pcrId];
-            ctx->sha512_used |= (1 << pcrId);
-        } else
-        if (alg == TPM2_ALG_SM3_256) {
-            pcr = ctx->sm3_256_pcrs[pcrId];
-            ctx->sm3_256_used |= (1 << pcrId);
+        uint8_t *pcr = NULL;
+        if (pcr_index > TPM2_MAX_PCRS) {
+            LOG_ERR("PCR%d > max %d", pcr_index, TPM2_MAX_PCRS);
+        } else if (alg == TPM2_ALG_SHA1) {
+            pcr = ctx->sha1_pcrs[pcr_index];
+            ctx->sha1_used |= (1 << pcr_index);
+        } else if (alg == TPM2_ALG_SHA256) {
+            pcr = ctx->sha256_pcrs[pcr_index];
+            ctx->sha256_used |= (1 << pcr_index);
+        } else if (alg == TPM2_ALG_SHA384) {
+            pcr = ctx->sha384_pcrs[pcr_index];
+            ctx->sha384_used |= (1 << pcr_index);
+        } else if (alg == TPM2_ALG_SHA512) {
+            pcr = ctx->sha512_pcrs[pcr_index];
+            ctx->sha512_used |= (1 << pcr_index);
+        } else if (alg == TPM2_ALG_SM3_256) {
+            pcr = ctx->sm3_256_pcrs[pcr_index];
+            ctx->sm3_256_used |= (1 << pcr_index);
         } else {
-            LOG_WARN("PCR%d algorithm %d unsupported", pcrId, alg);
+            LOG_WARN("PCR%d algorithm %d unsupported", pcr_index, alg);
         }
 
         if (pcr && !tpm2_openssl_pcr_extend(alg, pcr, digest->Digest, alg_size)) {
-            LOG_ERR("PCR%d extend failed", pcrId);
+            LOG_ERR("PCR%d extend failed", pcr_index);
             return false;
         }
 
@@ -201,7 +196,7 @@ bool parse_event2(TCG_EVENT_HEADER2 const *eventhdr, size_t buf_size,
     return true;
 }
 
-bool foreach_event2(tpm2_eventlog_ctx_t * ctx, TCG_EVENT_HEADER2 const *eventhdr_start, size_t size) {
+bool foreach_event2(tpm2_eventlog_ctx_t *ctx, TCG_EVENT_HEADER2 const *eventhdr_start, size_t size) {
 
     if (eventhdr_start == NULL) {
         LOG_ERR("invalid parameter");
--- a/lib/tpm2_eventlog.c
+++ b/lib/tpm2_eventlog.c
@@ -2,6 +2,10 @@
 #include <stdlib.h>
 #include <tss2/tss2_tpm2_types.h>
 
+#include <openssl/buffer.h>
+#include <openssl/evp.h>
+#include <openssl/sha.h>
+
 #include "log.h"
 #include "efi_event.h"
 #include "tpm2_alg_util.h"
@@ -27,8 +31,7 @@ bool digest2_accumulator_callback(TCG_DIGEST2 const *digest, size_t size,
  * hold the digest. The size of the digest is passed to the callback in the
  * 'size' parameter.
  */
-bool foreach_digest2(TCG_DIGEST2 const *digest, size_t count, size_t size,
-                     DIGEST2_CALLBACK callback, void *data) {
+bool foreach_digest2(tpm2_eventlog_ctx_t * ctx, int pcrId, TCG_DIGEST2 const *digest, size_t count, size_t size) {
 
     if (digest == NULL) {
         LOG_ERR("digest cannot be NULL");
@@ -48,8 +51,26 @@ bool foreach_digest2(TCG_DIGEST2 const *digest, size_t count, size_t size,
             LOG_ERR("insufficient size for digest buffer");
             return false;
         }
-        if (callback != NULL) {
-            ret = callback(digest, alg_size, data);
+
+        if (digest->AlgorithmId == TPM2_ALG_SHA1) {
+            uint8_t * const pcr = ctx->sha1_pcrs[pcrId];
+            SHA_CTX sha1;
+            SHA1_Init(&sha1);
+            SHA1_Update(&sha1, pcr, alg_size);
+            SHA1_Update(&sha1, digest->Digest, alg_size);
+            SHA1_Final(pcr, &sha1);
+        } else
+        if (digest->AlgorithmId == TPM2_ALG_SHA256) {
+            uint8_t * const pcr = ctx->sha256_pcrs[pcrId];
+            SHA256_CTX sha256;
+            SHA256_Init(&sha256);
+            SHA256_Update(&sha256, pcr, alg_size);
+            SHA256_Update(&sha256, digest->Digest, alg_size);
+            SHA256_Final(pcr, &sha256);
+        }
+
+        if (ctx->digest2_cb != NULL) {
+            ret = ctx->digest2_cb(digest, alg_size, ctx->data);
             if (!ret) {
                 LOG_ERR("callback failed for digest at %p with size %zu", digest, alg_size);
                 break;
@@ -137,9 +158,13 @@ bool parse_event2(TCG_EVENT_HEADER2 const *eventhdr, size_t buf_size,
     }
     *event_size = sizeof(*eventhdr);
 
-    ret = foreach_digest2(eventhdr->Digests, eventhdr->DigestCount,
-                          buf_size - sizeof(*eventhdr),
-                          digest2_accumulator_callback, digests_size);
+    tpm2_eventlog_ctx_t ctx = {
+        .data = digests_size,
+        .digest2_cb = digest2_accumulator_callback,
+    };
+    ret = foreach_digest2(&ctx, eventhdr->PCRIndex,
+                          eventhdr->Digests, eventhdr->DigestCount,
+                          buf_size - sizeof(*eventhdr));
     if (ret != true) {
         return false;
     }
@@ -161,10 +186,7 @@ bool parse_event2(TCG_EVENT_HEADER2 const *eventhdr, size_t buf_size,
     return true;
 }
 
-bool foreach_event2(TCG_EVENT_HEADER2 const *eventhdr_start, size_t size,
-                    EVENT2_CALLBACK event2hdr_cb,
-                    DIGEST2_CALLBACK digest2_cb,
-                    EVENT2DATA_CALLBACK event2_cb, void *data) {
+bool foreach_event2(tpm2_eventlog_ctx_t * ctx, TCG_EVENT_HEADER2 const *eventhdr_start, size_t size) {
 
     if (eventhdr_start == NULL) {
         LOG_ERR("invalid parameter");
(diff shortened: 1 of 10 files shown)


osresearch:
The tests don't know about the new `tpm2_eventlog_ctx_t`, so they can't properly test it right now. Perhaps the helper functions and struct should be moved into a separate internal header so that they are not exposed to callers of `yaml_eventlog()`.

osresearch:
On more consideration, the PCR tracking should be moved into the non-yaml part of the eventlog code and an additional callback for processing the PCRs at the end, so that `checkquote` can use it as well to validate that the eventlog matches the quote.

I've converted this back to a draft while I make those changes.

osresearch:
@williamcroberts what are your thoughts on this minor re-archtecting of the `tpm2_eventlog` library to use a table of function pointers and have it compute the PCRs as part of parsing the eventlog?

The motivation for adding this functionality is that the remote attestation process for some applications involves verifying that the eventlog records the the actual values that were measured into the PCRs. This could be done as a separate program after running `tpm2_checkquote` as well, although it might make sense to have it in the quote checking program.

osresearch:
```
../tools/misc/tpm2_checkquote.c:355:32: warning: The right operand of '<' is a garbage value
        for (unsigned i = 0; i < pcr_select.count; i++) {
                               ^ ~~~~~~~~~~~~~~~~
```

That's not a very nice thing for the compiler to say about `i`. I'm not sure what it even means.

williamcroberts:
> On more consideration, the PCR tracking should be moved into the non-yaml part of the eventlog code and an additional callback for processing the PCRs at the end, so that `checkquote` can use it as well to validate that the eventlog matches the quote.
> 
> I've converted this back to a draft while I make those changes.

I would just add an output switch to dump computed PCR's to a file. This also is slightly related to #2072, in that their are two binary PCR output formats.

osresearch:
I had originally added code to dump the computed PCRs to a file, but then noticed that the output was already yaml, so it seemed more consistent to add another section that matches the same format as `tpm2_pcrread`.

lib/tpm2_eventlog_yaml.c
```diff
@@ -401,14 +401,44 @@ bool yaml_specid_callback(TCG_EVENT const *event, void *data) {
...
+
 bool yaml_eventlog(UINT8 const *eventlog, size_t size) {
 
     size_t count = 0;
+    tpm2_eventlog_ctx_t ctx = {
+        .data = &count,
+        .specid_cb = yaml_specid_callback,
+        .event2hdr_cb = yaml_event2hdr_callback,
+        .digest2_cb = yaml_digest2_callback,
+        .event2_cb = yaml_event2data_callback,
```
williamcroberts, line 433: These seem static except for the test code, so why are we exposing them as fn pointers? Just call the function directly and you can use ld wrap for hijacking during unit tests.

I don't want to add indirection complexity unless its needed, and I see that Phillip already did, your just moving where they get set, but I don't see the value in either of them.

williamcroberts:
> ```
> ../tools/misc/tpm2_checkquote.c:355:32: warning: The right operand of '<' is a garbage value
>         for (unsigned i = 0; i < pcr_select.count; i++) {
>                                ^ ~~~~~~~~~~~~~~~~
> ```
> 
> That's not a very nice thing for the compiler to say about `i`. I'm not sure what it even means.

It might be because your shadowing the global variable ctx as a local parameter.
https://lgtm.com/projects/g/tpm2-software/tpm2-tools/rev/pr-f738c5f5c56fa09cdeab2d1102d91a4cbc2072b9

williamcroberts:
> I had originally added code to dump the computed PCRs to a file, but then noticed that the output was already yaml, so it seemed more consistent to add another section that matches the same format as `tpm2_pcrread`.
> I had originally added code to dump the computed PCRs to a file, but then noticed that the output was already yaml, so it seemed more consistent to add another section that matches the same format as `tpm2_pcrread`.

As long as it's backwards compat. IIUC, -e will dump the PCR values. The YAML maps you should always be able to add without breaking backwards compat as long as the existing data remains untouched in both type and name. Usually everything is a dictionary off of a key, occasionally a list.

lib/tpm2_eventlog_yaml.c
```diff
@@ -307,7 +307,7 @@ bool yaml_event2hdr_callback(TCG_EVENT_HEADER2 const *eventhdr, size_t size,
         return false;
     }
 
-    tpm2_tool_output("- Event[%zu]:\n", *count++);
+    tpm2_tool_output("- Event[%zu]:\n", (*count)++);
```
osresearch, line 310: This is the only non-backwards compatible change, which fixes a bug in the event numbering of `tpm2_eventlog`

osresearch:
The patch did introduce a slight backwards incompatibility to fix a bug in the yaml generation of the existing tool -- the event numbering was broken. And I just added additional tests to ensure that the output of `tpm2_eventlog` is valid yaml;

However, I'd like to propose some additional changes to the output of this tool -- the pcr addition was not valid since the output was an array of hashes of hashes, so I had to adjust the structure slightly. Now it outputs a dictionary with `events` and `pcrs`, and in the `events` section the numbering is corrected (before it was adding a key `Event[X]` with a null value).

There are some other changes that might be worth making for consistency.  Things like:

```
    SpecID:
      - Signature: Spec ID Event03
        platformClass: 0
        specVersionMinor: 0
        specVersionMajor: 2
        specErrata: 0
        uintnSize: 2
        numberOfAlgorithms: 2
        Algorithms:
          - Algorithm[0]:
            algorithmId: sha1
            digestSize: 20
          - Algorithm[1]:
            algorithmId: sha256
            digestSize: 32
        vendorInfoSize: 0
```

The case of these keys doesn't match the rest of the file (all the rest have initial uppercase). For instance, `algorithmId` doesn't match `AlgorithmId` elsewhere. The `Algorithms` section is an array of hashes, where each hash has an entry `Algorithm[0]` or `Algorithm[1]` with an empty key. 

Also, including the `numberOfAlgorithms`, `AlgorithmCount` and `DigestCount` in the various entries is also redundant since the length of the array implicitly encodes the count. The `EventSize`, `UnicodeNameLength` and `VariableDataLength` are also likely redundant.

Hopefully these aren't too breaking for the users of the tool and will make it easier for new users to make use of the eventlog and pcr values.

osresearch:
The test failure looks spurious:
```
FAIL: test/integration/tests/send
=================================
...
ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:394:Esys_StartAuthSession_Finish() Received a non-TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:136:Esys_StartAuthSession() Esys Finish ErrorCode (0x000a0008) 
ERROR: Esys_StartAuthSession(0xA0008) - tcti:Fails to connect to next lower layer
ERROR: Invalid lockout authorization
ERROR: Unable to run tpm2_clear
Shutting down
```

lib/tpm2_eventlog.h
```diff
@@ -17,23 +17,25 @@ typedef bool (*EVENT2DATA_CALLBACK)(TCG_EVENT2 const *event, UINT32 type,
...
 typedef bool (*SPECID_CALLBACK)(TCG_EVENT const *event, void *data);
 
+typedef struct {
+    void * data;
+    SPECID_CALLBACK specid_cb;
+    EVENT2_CALLBACK event2hdr_cb;
+    DIGEST2_CALLBACK digest2_cb;
+    EVENT2DATA_CALLBACK event2_cb;
+    uint8_t sha1_pcrs[TPM2_MAX_PCRS][TPM2_SHA1_DIGEST_SIZE];
+    uint8_t sha256_pcrs[TPM2_MAX_PCRS][TPM2_SHA256_DIGEST_SIZE];
```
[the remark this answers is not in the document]
osresearch: I've updated it for all of the current supported ones.  Something that does dynamic allocation of PCR banks and tracks which are in use could be added, although it would require quite a bit additional code.

lib/tpm2_eventlog_yaml.c
osresearch: I'm not sure how the test code could do the ld wrapping trick to hijack the function pointers since there are many different combinations used in the tests.  The busybox executable also introduces a problem with making these global functions since both the `tpm2_eventlog.c` and `tpm2_checkquote.c` programs want to call into the eventlog parser.

Using a streaming visitor pattern might be the wrong approach for this; perhaps it would be better to parse the event log and return a tree structure of some sort that the caller can then traverse on their own, rather than having dynamic callbacks. It's not like these logs are all that large.

tools/misc/tpm2_checkquote.c
```diff
@@ -304,6 +342,69 @@ static tool_rc init(void) {
         }
     }
 
+    if (ctx.flags.eventlog) {
+        tpm2_eventlog_ctx_t eventlog_ctx = {};
+        if (!eventlog_from_file(&eventlog_ctx, ctx.eventlog_path)) {
```
williamcroberts, line 357: dont hide functions in conditionals, set an intermediate value and check that in the if statement. The only exceptions are if it makes the code easier.

tools/misc/tpm2_checkquote.c
```diff
@@ -304,6 +342,69 @@ static tool_rc init(void) {
...
+                if (vi >= pcrs->count || di >= pcrs->pcr_values[vi].count) {
+                    LOG_ERR("Something wrong, trying to print but nothing more");
+                    eventlog_fail = true;
+                    break;
+                }
+
+                // Compare this digest to the computed value from the eventlog
+                const TPM2B_DIGEST *pcr = &pcrs->pcr_values[vi].digests[di];
+                const uint8_t * pcr_q = pcr->buffer;
+		const uint8_t * pcr_e = NULL;
```
williamcroberts, line 384: indent

lib/tpm2_eventlog_yaml.c
williamcroberts: Ahh yeah, I grepped for parse_eventlog and I now see the differences in uses. This approach is fine.

test/unit/test_tpm2_eventlog.c
```diff
@@ -42,15 +45,50 @@ static void test_foreach_digest2(void **state) {
...
-    assert_true(foreach_digest2(digest, 1, TCG_DIGEST2_SHA1_SIZE, foreach_digest2_test_callback, NULL));
+
+    tpm2_eventlog_ctx_t ctx = { .digest2_cb = foreach_digest2_test_callback };
+    assert_true(foreach_digest2(&ctx, 0, digest, 1, TCG_DIGEST2_SHA1_SIZE));
 }
 static void test_foreach_digest2_cbnull(void **state){
 
     (void)state;
-    uint8_t buf [TCG_DIGEST2_SHA1_SIZE] = { 0, };
+    uint8_t buf [TCG_DIGEST2_SHA1_SIZE] = { };
```
williamcroberts, line 55: Fix this and all other empty initializers and use `{ 0 }`

test/unit/test_tpm2_eventlog.c
```diff
@@ -42,15 +45,50 @@ static void test_foreach_digest2(void **state) {
...
 
-    assert_true(foreach_digest2(digest, 1, TCG_DIGEST2_SHA1_SIZE, NULL, NULL));
+    tpm2_eventlog_ctx_t ctx = {};
+    assert_true(foreach_digest2(&ctx, 0, digest, 1, TCG_DIGEST2_SHA1_SIZE));
+}
+static void test_sha1(void **state){
+
+    (void)state;
+    uint8_t buf [TCG_DIGEST2_SHA1_SIZE] = { 0, };
+    uint8_t sha1sum[] = { 0x31,0x19,0x5d,0x69,0x35,0x16,0x3c,0x79,0xa9,0x67,0x22,0xba,0x7d,0x4b,0x11,0x35,0x24,0x89,0xf4,0x8b };
```
williamcroberts, line 65: format these into symetrical blocks, like:
```C
uint8_t sha1sum[] = { 
    0x31,0x19,0x5d,0x69,0x35,0x16,0x3c,0x79,0xa9,0x67,
    0x22,0xba,0x7d,0x4b,0x11,0x35,0x24,0x89,0xf4,0x8b 
};
```

here and throughout

```

File: test/unit/test_tpm2_eventlog.c [hunk 8]
```diff
@@ -42,15 +45,50 @@ static void test_foreach_digest2(void **state) {
...
+    const int pcrIndex = 3;
+
+    TCG_DIGEST2 * digest = (TCG_DIGEST2*) buf;
```
williamcroberts, line 68: * tight to variable, here and throughout

test/unit/test_tpm2_eventlog.c
```diff
@@ -42,15 +45,50 @@ static void test_foreach_digest2(void **state) {
...
-    assert_true(foreach_digest2(digest, 1, TCG_DIGEST2_SHA1_SIZE, NULL, NULL));
+    tpm2_eventlog_ctx_t ctx = {};
+    assert_true(foreach_digest2(&ctx, 0, digest, 1, TCG_DIGEST2_SHA1_SIZE));
+}
+static void test_sha1(void **state){
+
+    (void)state;
+    uint8_t buf [TCG_DIGEST2_SHA1_SIZE] = { 0, };
+    uint8_t sha1sum[] = { 0x31,0x19,0x5d,0x69,0x35,0x16,0x3c,0x79,0xa9,0x67,0x22,0xba,0x7d,0x4b,0x11,0x35,0x24,0x89,0xf4,0x8b };
+    const int pcrIndex = 3;
```
williamcroberts, line 66: No CamelCasing pcr_index. (I know the code base is not 100% consistent, but lets not add to the issue)

test/unit/test_tpm2_eventlog.c
```diff
@@ -42,15 +45,50 @@ static void test_foreach_digest2(void **state) {
...
+
+    TCG_DIGEST2 * digest = (TCG_DIGEST2*) buf;
+    digest->AlgorithmId = TPM2_ALG_SHA1,
+    memcpy(digest->Digest, "the magic words are:", TPM2_SHA1_DIGEST_SIZE);
+
+    tpm2_eventlog_ctx_t ctx = {};
+    assert_true(foreach_digest2(&ctx, pcrIndex, digest, 1, TCG_DIGEST2_SHA1_SIZE));
+    // memcmp() == 0 if the buffers match
+    assert_false(memcmp(ctx.sha1_pcrs[pcrIndex], sha1sum, sizeof(sha1sum)));
```
williamcroberts, line 75: assert_memory_equal(), here and throughout.

tools/misc/tpm2_checkquote.c
```diff
@@ -216,6 +219,37 @@ static bool pcrs_from_file(const char *pcr_file_path,
...
+        return false;
+    }
+
+    uint8_t * eventlog = calloc(1, size);
+    if (!eventlog) {
+        LOG_ERR("OOM");
+        return false;
+    }
+
+    uint16_t size_tmp = size; // wtf uint16?
```
williamcroberts, line 241: TPM size field is uint16_t, so the lore was that we're always bounded by that. In this case, we're not. So it might be better just to use stdio directly and comment why.

tools/misc/tpm2_checkquote.c
```diff
@@ -216,6 +219,37 @@ static bool pcrs_from_file(const char *pcr_file_path,
...
+    if (!files_load_bytes_from_path(file_path, eventlog, &size_tmp)) {
+        free(eventlog);
+        return false;
+    }
+
+    bool rc = parse_eventlog(ctx, eventlog, size);
```
williamcroberts: You could make this take a file pointer perhaps.... Then we don't need to load the whole thing up at once in memory. If you need a FILE stream from memory, don't use fmemopen. If you look through the logs you'll see it got removed, someone has some system without that. However, you can use the OPENSSL BIO routines, since we require OSSL.

lib/tpm2_eventlog.h
```diff
@@ -17,23 +17,25 @@ typedef bool (*EVENT2DATA_CALLBACK)(TCG_EVENT2 const *event, UINT32 type,
                                     void *data);
 typedef bool (*SPECID_CALLBACK)(TCG_EVENT const *event, void *data);
 
+typedef struct {
+    void * data;
+    SPECID_CALLBACK specid_cb;
+    EVENT2_CALLBACK event2hdr_cb;
+    DIGEST2_CALLBACK digest2_cb;
+    EVENT2DATA_CALLBACK event2_cb;
```
williamcroberts, line 25: Damn you Phil! ... we don't do capital type names. No need to fix in your series.

lib/tpm2_eventlog.h
```diff
@@ -17,23 +17,25 @@ typedef bool (*EVENT2DATA_CALLBACK)(TCG_EVENT2 const *event, UINT32 type,
...
+    uint8_t sha1_pcrs[TPM2_MAX_PCRS][TPM2_SHA1_DIGEST_SIZE];
+    uint8_t sha256_pcrs[TPM2_MAX_PCRS][TPM2_SHA256_DIGEST_SIZE];
+} tpm2_eventlog_ctx_t;
```
williamcroberts, line 36: We also don't do this _t stuff. If it's in the type location for the language it's obviously a type.

dnoliver:
I run into https://github.com/tpm2-software/tpm2-tools/pull/2093#issuecomment-642578549 before, thank you for fixing it!

My problem was that I could not use the output of tpm2_eventlog, convert it to json, and pass it to `jq` (command line json proccessor). The reason for doing that it so make queries to the event log, for example "Give me all the events related to PCR 7" and similar.

Hope this fixes it!

lib/tpm2_eventlog.c
```diff
@@ -48,8 +51,26 @@ bool foreach_digest2(TCG_DIGEST2 const *digest, size_t count, size_t size,
...
+
+        if (digest->AlgorithmId == TPM2_ALG_SHA1) {
+            uint8_t * const pcr = ctx->sha1_pcrs[pcrId];
+            SHA_CTX sha1;
+            SHA1_Init(&sha1);
+            SHA1_Update(&sha1, pcr, alg_size);
+            SHA1_Update(&sha1, digest->Digest, alg_size);
+            SHA1_Final(pcr, &sha1);
+        } else
+        if (digest->AlgorithmId == TPM2_ALG_SHA256) {
```
williamcroberts: ` } else if {`

williamcroberts:
> The patch did introduce a slight backwards incompatibility to fix a bug in the yaml generation of the existing tool -- the event numbering was broken. And I just added additional tests to ensure that the output of `tpm2_eventlog` is valid yaml;
> 
> However, I'd like to propose some additional changes to the output of this tool -- the pcr addition was not valid since the output was an array of hashes of hashes, so I had to adjust the structure slightly. Now it outputs a dictionary with `events` and `pcrs`, and in the `events` section the numbering is corrected (before it was adding a key `Event[X]` with a null value).
> 
> There are some other changes that might be worth making for consistency. Things like:
> 
> ```
>     SpecID:
>       - Signature: Spec ID Event03
>         platformClass: 0
>         specVersionMinor: 0
>         specVersionMajor: 2
>         specErrata: 0
>         uintnSize: 2
>         numberOfAlgorithms: 2
>         Algorithms:
>           - Algorithm[0]:
>             algorithmId: sha1
>             digestSize: 20
>           - Algorithm[1]:
>             algorithmId: sha256
>             digestSize: 32
>         vendorInfoSize: 0
> ```
> 
> The case of these keys doesn't match the rest of the file (all the rest have initial uppercase). For instance, `algorithmId` doesn't match `AlgorithmId` elsewhere. The `Algorithms` section is an array of hashes, where each hash has an entry `Algorithm[0]` or `Algorithm[1]` with an empty key.
> 
> Also, including the `numberOfAlgorithms`, `AlgorithmCount` and `DigestCount` in the various entries is also redundant since the length of the array implicitly encodes the count. The `EventSize`, `UnicodeNameLength` and `VariableDataLength` are also likely redundant.
> 
> Hopefully these aren't too breaking for the users of the tool and will make it easier for new users to make use of the eventlog and pcr values.

Since it wasn't valid YAML before (I guess I didn't review it well enough), and it should have been, I think we're just fixing it. Lets just make sure this in the changelog. Master is headed to 5.0 for the busybox style changes, and I had to make some slightly backwards incompat changes to the default hash algorithm when not specified.

williamcroberts:
> -----Original Message-----
> From: Nicolas Oliver <notifications@github.com>
> Sent: Thursday, June 11, 2020 10:48 AM
> Subject: Re: [tpm2-software/tpm2-tools] tpm2_eventlog: compute PCR values
> based on the events (#2093)
> 
> I run into #2093 (comment) <https://github.com/tpm2-software/tpm2-
> tools/pull/2093#issuecomment-642578549>  before, thank you for fixing it!
> 
> My problem was that I could not use the output of tpm2_eventlog, convert it to
> json, and pass it to jq (command line json proccessor). The reason for doing that
> it so make queries to the event log, for example "Give me all the events related
> to PCR 7" and similar.

Interesting, I always use python because I have been searching for a command line processor,
thanks for mentioning jq.

dnoliver:
There is also a wrapper for YAML, that do the conversion and calls jq for you

https://github.com/kislyuk/yq

I noticed that the JSON output made no sense. Hence the queries made no sense as well.
I had a TODO item to report that problem it, but @osresearch was faster and more effective than me :D

test/unit/test_tpm2_eventlog.c
osresearch: Oh, right. This is one other those areas that I feel the the C standard committee messed up (see also the trailing `,` in JSON).

test/unit/test_tpm2_eventlog.c
osresearch: I was following the structures, which are CamelCased.  Fixed!

lib/tpm2_eventlog.h
osresearch: Renamed to `tpm2_eventlog_context`.

osresearch:
I've made the style guide fixes and rebased to pickup the busybox bundled `tpm2` patches.

Issue #2097, opened 2020-06-09
codebase will not compile with clang


thac0:
After running `bootstrap` and `CC=clang CXX=clang++ ./configure` followed by `make`, it's clear the codebase is currently tuned for `gcc`, and only `gcc`. As such, configure should fail if a sufficiently-versioned `gcc` is not found.

```
$ make
  CC       tools/fapi/tss2_decrypt-tss2_template.o
error: unknown warning option '-Wstringop-overflow=4'; did you mean '-Wshift-overflow'? [-Werror,-Wunknown-warning-option]
error: unknown warning option '-Wstringop-truncation'; did you mean '-Wstring-conversion'? [-Werror,-Wunknown-warning-option]
error: unknown warning option '-Wduplicated-branches' [-Werror,-Wunknown-warning-option]
error: unknown warning option '-Wduplicated-cond' [-Werror,-Wunknown-warning-option]
error: unknown warning option '-Wbool-compare' [-Werror,-Wunknown-warning-option]
make: *** [Makefile:5719: tools/fapi/tss2_decrypt-tss2_template.o] Error 1
```

Snippet from `config.log`:
```
configure:11866: checking whether C compiler accepts -Wstringop-overflow=4
configure:11885: clang -c  -g  -Wstringop-overflow=4  -D_FORTIFY_SOURCE=2 conftest.c >&5
warning: unknown warning option '-Wstringop-overflow=4'; did you mean '-Wshift-overflow'? [-Wunknown-warning-option]
1 warning generated.
configure:11885: $? = 0
configure:11893: result: yes
```

clang version:
```
$ clang -v
clang version 10.0.0
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-pc-linux-gnu/10.1.0
Found candidate GCC installation: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.1.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-pc-linux-gnu/10.1.0
Found candidate GCC installation: /usr/lib64/gcc/x86_64-pc-linux-gnu/10.1.0
Selected GCC installation: /usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/10.1.0
Candidate multilib: .;@m64
Candidate multilib: 32;@m32
Selected multilib: .;@m64
```

Codebase:
```
$ git describe HEAD
[...]

PR #2098, opened 2020-06-09
configure.ac: make -W cflag checks more robust

thac0:
Stumbled upon this while attempting to compile the codebase
(unsuccessfully) with clang. As it happens, the underlying autoconf
macro (AX_CHECK_COMPILE_FLAG) will only fail if the compiler returns
non-zero. After a few moments of exploration, this is not always
sufficient.

In the case of clang, an unrecognized cflag will emit a warning, but
clang will return 0. Therefore, configure will (incorrectly) conclude
the cflag is recognized. See #2097 for the result.

In the case of gcc, if a cflag is recognized but used incorrectly
(example: -Wformat-security by itself), gcc will emit a warning but
return 0.

So, add smarts (EXTRA-FLAGS) to the macro, which results in clang
or gcc exiting with an error should it encounter a cflag it does not
recognize or a cflag used incorrectly.

Signed-off-by: Joe Konno <joe.konno@intel.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -207,7 +207,8 @@ AS_IF([test x"$enable_hardening" != x"no"], [
   AC_DEFUN([add_hardened_c_flag], [
     AX_CHECK_COMPILE_FLAG([$1],
       [EXTRA_CFLAGS="$EXTRA_CFLAGS $1"],
-      [AC_MSG_ERROR([Cannot enable $1, consider configuring with --disable-hardening])]
+      [AC_MSG_ERROR([Cannot enable $1, consider configuring with --disable-hardening])],
+      [-Werror]
     )
   ])
 
@@ -225,12 +226,13 @@ AS_IF([test x"$enable_hardening" != x"no"], [
     )
   ])
 
+  add_hardened_c_flag([-Werror])
   add_hardened_c_flag([-Wall])
   add_hardened_c_flag([-Wextra])
-  add_hardened_c_flag([-Werror])
 
+  add_hardened_c_flag([-Wbool-compare])
   add_hardened_c_flag([-Wformat])
-  add_hardened_c_flag([-Wformat-security])
+  add_hardened_c_flag([-Wformat -Wformat-security])
   add_hardened_c_flag([-Wstack-protector])
   add_hardened_c_flag([-fstack-protector-all])
   add_hardened_c_flag([-Wstrict-overflow=5])
@@ -256,7 +258,8 @@ and submitting patches upstream!])
 AC_DEFUN([add_c_flag], [
   AX_CHECK_COMPILE_FLAG([$1],
     [EXTRA_CFLAGS="$EXTRA_CFLAGS $1"],
-    $2
+    [$2],
+    [-Werror]
   )
 ])
 
@@ -271,7 +274,7 @@ AS_IF([test "$HOSTOS" = "Linux"],
        add_c_flag([-Wstringop-truncation])
        add_c_flag([-Wduplicated-branches])
        add_c_flag([-Wduplicated-cond])
-       add_hardened_c_flag([-Wbool-compare])],[])
+])
 
 # Best attempt, strip unused stuff from the binary to reduce size.
 # Rather than nesting these and making them ugly just use a counter.


configure.ac
```diff
@@ -225,12 +226,12 @@ AS_IF([test x"$enable_hardening" != x"no"], [
...
   ])
 
+  add_hardened_c_flag([-Werror])
   add_hardened_c_flag([-Wall])
   add_hardened_c_flag([-Wextra])
-  add_hardened_c_flag([-Werror])
 
   add_hardened_c_flag([-Wformat])
-  add_hardened_c_flag([-Wformat-security])
+  add_hardened_c_flag([-Wformat -Wformat-security])
```
williamcroberts, line 234: Why this change?

thac0: > Why this change?

Because, if `gcc -Wformat-security` is run by itself, you get:
```
$ gcc -Wformat-security -o foo foo.c
cc1: warning: ‘-Wformat-security’ ignored without ‘-Wformat’ [-Wformat-security]
```
While it is technically 'recognized', it would error out with -Werror flipped:
```
$ gcc -Werror -Wformat-security -o foo foo.c
cc1: error: ‘-Wformat-security’ ignored without ‘-Wformat’ [-Werror=format-security]
```
So, in my mind, checking for this particular flag's existence on its own is invalid.

tstellar:
I filed a very similar pull request in #1993 , but closed it in favor of #2075.  It seems to me that trying to define a set of hardening flags that work for clang might work better, but I am happy with any solution that will fix the clang build with hardening.

configure.ac
```diff
@@ -207,7 +207,8 @@ AS_IF([test x"$enable_hardening" != x"no"], [
   AC_DEFUN([add_hardened_c_flag], [
     AX_CHECK_COMPILE_FLAG([$1],
       [EXTRA_CFLAGS="$EXTRA_CFLAGS $1"],
-      [AC_MSG_ERROR([Cannot enable $1, consider configuring with --disable-hardening])]
+      [AC_MSG_ERROR([Cannot enable $1, consider configuring with --disable-hardening])],
```
williamcroberts, line 210: Something else is still broken:
```
configure: WARNING: Compiling with --disable-hardening is dangerous!
you should consider fixing the configure script compiler flags
and submitting patches upstream!
checking whether C compiler accepts -D_GNU_SOURCE... yes
checking whether C compiler accepts -Wstringop-overflow=4... no
checking whether C compiler accepts -Wstringop-truncation... no
checking whether C compiler accepts -Wduplicated-branches... no
checking whether C compiler accepts -Wduplicated-cond... no
checking whether C compiler accepts -Wbool-compare... no
configure: error: Cannot enable -Wbool-compare, consider configuring with --disable-hardening
The command "./.ci/travis.run" exited with 1.
```
Perhaps --disable-hardening configure flag is not working?

configure.ac
```diff
@@ -207,7 +207,8 @@ AS_IF([test x"$enable_hardening" != x"no"], [
...
+      [-Werror]
```
thac0, line 211: I think the reason Travis et al are failing, @williamcroberts , is because of this line. I just ran `gcc -Wbool` for grins, and it's an invalid flag, so with this change to the macro, it's now an error at configure-time, and therefore a show-stopper.

configure.ac
thac0: See my comment on line 211

thac0:
Yeah, @tstellar , I think we're missing the forest for the trees in how cflags are discovered. The fact that you have to `--disable-hardening` feels wrong to me, given clang's feature set. I'm cooking up an idea, so expect a PR relatively soon.

thac0:
Looks like Cirrus failed because it didn't pull my modified & rebased change? Huh.

williamcroberts:
I would:
1. Make it best attempt to add a default set
2. If you don't want to attempt to add the default set, configure option --disable-defaultflags to match tpm2-tss
    
    - I would like to, when possible, merge behaviors between projects.

thac0:
I'm with you on all points, @williamcroberts . I was thinking hardest about the "default set" that would allow for easy building with gcc or clang. And if folks like it here, no reason why I couldn't apply that set/functionality to the other tpm projects.

In going through the motions, this PR has become less a proposed fix and more a discovery mission. So feel free to close if you like.

williamcroberts:
> I'm with you on all points, @williamcroberts . I was thinking hardest about the "default set" that would allow for easy building with gcc or clang. And if folks like it here, no reason why I couldn't apply that set/functionality to the other tpm projects.
> 
> In going through the motions, this PR has become less a proposed fix and more a discovery mission. So feel free to close if you like.

I'll let you close it or you can force push the branch with the newest changes when you're ready. Thanks for taking the time to look into this.

thac0:
Closing to pursue a more robust approach. Stay tuned for a new PR.

PR #2112, opened 2020-06-19
Enforce tool size and a few cleanups

williamcroberts:
tpm2_tool_register didn't make sure it was staying in bounds, so bad or missing CFLAGS could cause a wonky build.


williamcroberts:
cirrus is being a PITA....

PR #2114, opened 2020-06-20
tpm2_checkquote: Add option to input PCR-digest without TPML_PCR_SELECTION header

### LINKED ISSUE #2072: tpm2_pcrread is not compatible with tpm2_checkquote

idesai:
Fixes #2072


stephensmalley:
FWIW, this worked for me.

williamcroberts:
> This looks like good progress on this bug. Ill just take this as is.

Never Mind, flunked scan-build:
```
ar: `u' modifier ignored since `D' is the default (see `U')
  CCLD     tools/fapi/tss2_decrypt
../tools/misc/tpm2_checkquote.c:184:17: warning: Array subscript is undefined
                pcrs->pcr_values[0].digests[pcrs->pcr_values[0].count].size = read_size;
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

tools/misc/tpm2_checkquote.c
```diff
@@ -160,57 +161,109 @@ static TPM2B_ATTEST *message_from_file(const char *msg_file_path) {
...
-        goto out;
+    unsigned i = 0;
+    unsigned j = 0;
+    unsigned read_size = 0;
+    for (i = 0; i < pcr_select->count; i++) {
+        read_size = tpm2_alg_util_get_hash_size(pcr_select->pcrSelections[i].hash);
+
+        for (j = 0; j < pcr_select->pcrSelections[i].sizeofSelect * 8; j++) {
+            if ((pcr_select->pcrSelections[i].pcrSelect[j / 8] & 1 << (j % 8)) != 0) {
+                pcrs->pcr_values[0].digests[pcrs->pcr_values[0].count].size = read_size;
```
williamcroberts, line 184: I think scan-build is complaining because count can go beyond 8, which is the max size of field digests.

if pcr_select->count is 1, and sizeofSelect is 2,  count++ takes count to > 7. Additionally, if pcr_select->count is 2, on the second iteration of the loop count is also not reset causing the same issue.

tools/misc/tpm2_checkquote.c
```diff
@@ -160,57 +161,109 @@ static TPM2B_ATTEST *message_from_file(const char *msg_file_path) {
...
+                if (!(fread(pcrs->pcr_values[0].digests[pcrs->pcr_values[0].count].buffer,
```
williamcroberts, line 185: We don't hide functions in conditionals.

tools/misc/tpm2_checkquote.c
```diff
@@ -160,57 +161,109 @@ static TPM2B_ATTEST *message_from_file(const char *msg_file_path) {
...
+                    read_size, 1, pcr_input))) {
+                    LOG_ERR("Failed to read PCR digests from file");
+                    return false;
+                }
+                pcrs->pcr_values[0].count++;
+            }
+        }
```
williamcroberts, line 192: I hate PCR code....

idesai: me too :)

tools/misc/tpm2_checkquote.c
```diff
@@ -160,57 +161,124 @@ static TPM2B_ATTEST *message_from_file(const char *msg_file_path) {
...
+    unsigned j = 0;
+    unsigned read_size = 0;
+    size_t read_count = 0;
+    unsigned digest_list_count = 0;
+    for (i = 0; i < pcr_select->count; i++) {
+        read_size = tpm2_alg_util_get_hash_size(pcr_select->pcrSelections[i].hash);
+
+        for (j = 0; j < pcr_select->pcrSelections[i].sizeofSelect * 8; j++) {
+            if ((pcr_select->pcrSelections[i].pcrSelect[j / 8] & 1 << (j % 8)) != 0) {
+                pcrs->pcr_values[digest_list_count].digests[pcrs->pcr_values[digest_list_count].count].size = read_size;
```
williamcroberts, line 193: I **think** scan-build doesn't know that digest_list_count will never be grater than pcr_select->count, so you probably want to check that and make sure of that case.

PR #2123, opened 2020-06-29
Bundle the FAPI tss2 tools into a single executable

osresearch:
Reuses the same name registration and lookup functions as #2077 for the programs in `tools/fapi/tss2_*`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -44,7 +44,7 @@ noinst_LIBRARIES = $(LIB_COMMON)
 lib_libcommon_a_SOURCES = $(LIB_SRC)
 lib_libcommon_a_CFLAGS = -fPIC $(AM_CFLAGS)
 
-tools_fapi_tss2_CFLAGS = $(FAPI_CFLAGS)
+tools_fapi_tss2_CFLAGS = $(FAPI_CFLAGS) -DTSS2_TOOLS_MAX="$(words $(tss2_tools))"
 tools_fapi_tss2_LDFLAGS = $(TSS2_FAPI_LIBS)
 tools_fapi_tss2_SOURCES = \
 	tools/fapi/tss2_template.c \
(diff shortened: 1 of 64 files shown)


Makefile.am
```diff
@@ -36,195 +36,60 @@ TESTS =
...
 
 noinst_LIBRARIES = $(LIB_COMMON)
 lib_libcommon_a_SOURCES = $(LIB_SRC)
 lib_libcommon_a_CFLAGS = -fPIC $(AM_CFLAGS)
 
-tools_fapi_tss2_decrypt_CFLAGS = $(FAPI_CFLAGS)
-tools_fapi_tss2_decrypt_LDFLAGS = $(TSS2_FAPI_LIBS)
-tools_fapi_tss2_decrypt_SOURCES = tools/fapi/tss2_template.c tools/fapi/tss2_template.h \
-    tools/fapi/tss2_decrypt.c
+tools_fapi_tss2_CFLAGS = $(FAPI_CFLAGS)
```
williamcroberts, line 47: Shouldn't this have:
```C
tools_fapi_CFLAGS = $(AM_CFLAGS) -DTSS2_TOOLS_MAX="$(words $(tss2_tools))"
```

osresearch: Yes, although 1024 tss2 tools should be enough for anyone. Fixed in https://github.com/tpm2-software/tpm2-tools/pull/2123/commits/469e8518f99ddc8c85007d440bf6510d661beac8

williamcroberts:
@osresearch  Looking at the commit history, their are some things that could just be squashed down. But looking at these patches they don't build/test/run on their own, so they break git bisect (we try not to break that). So just squash this all down to one commit, it's not like we could really revert any of these commits individually anyways. Resubmit and ill take them as they are.

@idesai we need to make sure that we look at the bash completion scripts to make sure they are still working for all the tools. tss2 and tpm2 toolsets.

AndreasFuchsSIT:
THIS IS AWESOME !

williamcroberts:
Squashed and merged out of band: c5b15d93d62

Thanks,
Bill

PR #2145, opened 2020-07-14
build: add libgcrypt dependency for TSS-2.4

tstruk:


idesai:
@tstruk, does this [line](https://github.com/tpm2-software/tpm2-tools/blob/892faf27dbbe7171d26c46a81d342290603108dc/.cirrus.yml#L16) also needs to be conditional?

tstruk:
> @tstruk, does this [line](https://github.com/tpm2-software/tpm2-tools/blob/892faf27dbbe7171d26c46a81d342290603108dc/.cirrus.yml#L16) also needs to be conditional?

probably, yes. I will look into how to do it on cirrus.

tstruk:
> @tstruk, does this [line](https://github.com/tpm2-software/tpm2-tools/blob/892faf27dbbe7171d26c46a81d342290603108dc/.cirrus.yml#L16) also needs to be conditional?

Actually we can just remove it because on cirrus we build against TSS-master. I will push a PR soon

PR #2146, opened 2020-07-14
currus: update mssim version

tstruk:


tstruk:
Cirrus is still failing to build the tpm simulator for whatever reason:
```
clang: error: unable to execute command: Killed
clang: error: clang frontend command failed due to signal (use -v to see invocation)
```

Thavis should pass now.

PR #2153, opened 2020-07-29
tss2 *: Enhance Fapi Tools integration tests

cplappert:
This PR enhances the tss2 tools integration tests. Enhancements are:
1. The tests now additionally run with the ECC default profile.
2. The tests now test input boundaries with empty and big input files. Because of this, the TSS version in the CI needs to be updated to include latest fixes so that tests do not fail.

### DIFF
--- a/.ci/download-deps.sh
+++ b/.ci/download-deps.sh
@@ -3,7 +3,7 @@
 
 function get_deps() {
 
-	TSS_VERSION=${TPM2_TSS_VERSION:-2.4.0}
+	TSS_VERSION=${TPM2_TSS_VERSION:-2.4.2}
 	ABRMD_VERSION=2.3.1
 
 	echo "pwd starting: `pwd`"
--- a/bootstrap
+++ b/bootstrap
@@ -14,6 +14,18 @@ src_listvar () {
     echo ""
 }
 
+# duplicate fapi tests with ecc
+duplicate () {
+    basedir=$1
+    suffix=$2
+
+    find "${basedir}" \( -iname "${suffix}" ! -iname "*ecc.sh" \) | while read fname; do
+      cp $fname ${fname%.sh}_ecc.sh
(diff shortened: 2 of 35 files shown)


.ci/download-deps.sh
```diff
@@ -3,7 +3,7 @@
 
 function get_deps() {
 
-	TSS_VERSION=${TPM2_TSS_VERSION:-2.4.0}
+	TSS_VERSION=${TPM2_TSS_VERSION:-2.4.x}
```
tstruk, line 6: 2.4.x is a branch name. Don't you want a tag here? You may use 2.4.1 instead.

cplappert: The CI errors with `tss2 pcrextend` will be fixed as soon as https://github.com/tpm2-software/tpm2-tss/pull/1811 and https://github.com/tpm2-software/tpm2-tss/pull/1812 are merged. Thus, for the tests to run successfully, a version of the TSS is needed where these fixes are included. Probably 2.4.2?

tstruk: Yes, it will be in 2.4.2, which should be out soon.

cplappert: Ok. I will update the commit as soon as it is available.

cplappert: So, I updated to tag 2.4.2. Travis is fine now, but Cirrus reports a weird error. Maybe it was only temporary. Could retrigger the build?

tstruk: On Cirrus it always runs against TSS master so it doesn't really matter what TSS version is set in .ci/download-deps.sh
but it should work just fine, as it also runs fine on Travis against TSS master: https://travis-ci.org/github/tpm2-software/tpm2-tools/jobs/715117614
We have recently checked in a change to tcti-mssim to change it to async mode. Maybe this causes some trouble in Cirrus. I will look into it.

cplappert: Anything new?

williamcroberts:
Since cirrus is in general failing, I will merge this while @tstruk sorts out cirrus

PR #2154, opened 2020-07-29
set umask to 0117 (drop owner execute, group execute and all for other)

williamcroberts:
Don't create world readable files.... make them only group readable and fully controllable by owner.

### DIFF
--- a/tools/fapi/tss2_template.c
+++ b/tools/fapi/tss2_template.c
@@ -16,6 +16,9 @@
 #include <stdarg.h>
 #include <errno.h>
 
+#include <sys/types.h>
+#include <sys/stat.h>
+
 #include "tools/fapi/tss2_template.h"
 #include "lib/config.h"
 
@@ -397,6 +400,14 @@ static const tss2_tool *tss2_tool_lookup(int *argc, char ***argv)
  * specify which FAPI function to call.
  */
 int main(int argc, char *argv[]) {
+
+    /* get rid of:
+     *   owner execute (1)
+     *   group execute (1)
+     *   other write + read + execute (7)
+     */
+    umask(0117);
+
     const tss2_tool * const tool = tss2_tool_lookup(&argc, &argv);
     if (!tool) {
         LOG_ERR("%s: unknown tool. Available tss2 commands:\n", argv[0]);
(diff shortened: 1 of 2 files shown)


idesai:
@williamcroberts for file permission 0640, umask = 027, right?

williamcroberts:
> @williamcroberts for file permission 0640, umask = 027, right?

huh? It sets the default mask for open and creat. So this way the default for things like fopen will be a fie that is user readable/writeable, group readable, others no access.

idesai:
> > @williamcroberts for file permission 0640, umask = 027, right?
> 
> huh? It sets the default mask for open and creat. So this way the default for things like fopen will be a fie that is user readable/writeable, group readable, others no access.

https://www.webune.com/forums/umask-calculator.html

williamcroberts:
> > > @williamcroberts for file permission 0640, umask = 027, right?
> > 
> > 
> > huh? It sets the default mask for open and creat. So this way the default for things like fopen will be a fie that is user readable/writeable, group readable, others no access.
> 
> https://www.webune.com/forums/umask-calculator.html

Oh duh, "mask out the bits".... <facepalm>

williamcroberts:
Oh boy, fapi stores things globaly, this might not be the proper umask. @AndreasFuchsSIT what would be a good umask for tss2 tools?

diabonas:
Regarding FAPI, files under `/var/lib/tpm2-tss/system/keystore` should be group read-writable, see https://github.com/tpm2-software/tpm2-tss/issues/1810.

williamcroberts:
> Regarding FAPI, files under `/var/lib/tpm2-tss/system/keystore` should be group read-writable, see [tpm2-software/tpm2-tss#1810](https://github.com/tpm2-software/tpm2-tss/issues/1810).

Will update to 0117

williamcroberts:
Test program:
```C
#include <stdio.h>

#include <unistd.h>

#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>

int main() {

        umask(0117);

        int fd = open("foo.bar", O_CREAT, 0777);
        printf("fd: %d\n", fd);
        close(fd);

        return 0;
}
```
```bash
ls -la foo.bar 
-rw-rw---- 1 wcrobert wcrobert 0 Jul 30 15:03 foo.bar
```

williamcroberts:
Oh darn we only wanted this for tss2_ tools, but I think it applies to tpm2_ tools as well. You may wish to create things a group can get to and use.

Issue #2169, opened 2020-08-17
tpm2 verifysignature --format is no longer supported


osresearch:
I'm attempting to follow the [Intel guide for signing PCRs with an RSA key](https://software.intel.com/content/www/us/en/develop/articles/code-sample-protecting-secret-data-and-keys-using-intel-platform-trust-technology.html#inpage-nav-2-4).  However, `tpm2 verifysignature` fails with "`the TPM was unable to unmarshall a value because there were not enough octets in the input buffer`".

This is using https://github.com/tpm2-software/tpm2-tools/tree/c643ff688834d573772c9cc57fcbdf48a7e7735e and https://github.com/tpm2-software/tpm2-tss/tree/76be63d641e01e7a3fcdb987fedadf98e970ba8b

As a minimal test case:
```
#!/bin/bash -x
die() { echo >&2 "$*" ; exit 1 ; }

tpm2 flushcontext --transient-object
tpm2 flushcontext --loaded-session

# Generate RSA key
openssl genrsa -out key.priv 2048 || die genrsa
openssl rsa -in key.priv -out key.pem -pubout || die key.pem

# Create policy and sign it
tpm2 startauthsession --session session.ctx || die startsession

tpm2 policypcr \
	--session session.ctx \
	--pcr-list "sha256:0,11" \
	--policy pcr.policy \
|| die policypcr

tpm2 flushcontext session.ctx

openssl dgst \
	-sha256 \
	-sign key.priv \
	-out pcr.policy.sig \
	pcr.policy \
|| die openssl sign plicy

# Load the key into the TPM and validate on policy signature
tpm2 loadexternal \
	--key-algorithm rsa \
	--hierarchy o \
	--public key.pem \
	--key-context key.ctx \
	--name key.name \
|| die loadexternal

tpm2 verifysignature \
	--hash-algorithm sha256 \
	--format rsassa \
	--key-context key.ctx \
	--message pcr.policy \
	--signature pcr.policy.sig \
	--ticket verification.tkt \
|| die verifysignature
```

williamcroberts:
@osresearch I ran your test and am getting a slightly different output:
```
ERROR: Error deserializing signature structure: 0x9000b
ERROR: The input file needs to be a valid TPMT_SIGNATURE data structure
ERROR: Unable to run verifysignature
```
```bash
tpm2 rc_decode 0x9000b
mu:A parameter has a bad value
```
My tss version is:
```bash
[...]

PR #2170, opened 2020-08-17
tpm2_verifysignature: fix --format to --scheme fallback

williamcroberts:
--format was changed to --scheme in e275ba54e585c. However, the option
for --format was set to '0' (ASCII 0)  in the selector char. However,
the case statement was set to 0, thus --format was being silently
ignored. Fix this by setting the getopt selector charactar to 0.

Fixes: #2169

### DIFF
--- a/tools/tpm2_verifysignature.c
+++ b/tools/tpm2_verifysignature.c
@@ -203,6 +203,7 @@ static bool on_option(char key, char *value) {
     }
         break;
     case 0:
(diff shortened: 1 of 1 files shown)


osresearch:
Looks good to me.  Is there a longer term plan to deprecate `--format`? If so, would it make sense to print a warning?

williamcroberts:
No need for changelog entry, only on master and no tag contains it.
```bash
git branch --contains e275ba5
*  master

git tag --contains e275ba5
# no output
```

Issue #2171, opened 2020-08-18
tpm2_nvdefine without specifying the size of data area got "structure is the wrong size(0x2D5)"


Ivanhu5866:
I tried to test tpm2_nvdefine without specifying the size of data area with several latest machines, and lots of machines got "the structure is the wrong size(0x2D5)" error.
```
$sudo tpm2_nvdefine
WARNING:esys:src/tss2-esys/api/Esys_NV_DefineSpace.c:333:Esys_NV_DefineSpace_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_NV_DefineSpace.c:122:Esys_NV_DefineSpace() Esys Finish ErrorCode (0x000002d5)
ERROR: Failed to define NV area at index 0x1000000
ERROR: Esys_NV_DefineSpace(0x2D5) - tpm:parameter(2):structure is the wrong size
ERROR: Unable to run tpm2_nvdefine
```
I believe this is because the size defaults to MAX_NV_INDEX_SIZE which is typically 2048.
But under these machines, the value TPM2_PT_NV_INDEX_MAX is 1664(0x680). 
$ sudo tpm2_getcap properties-fixed
...
TPM2_PT_NV_INDEX_MAX:
  raw: 0x680
...
I also tried to specify the size with -s, when over 0x680, then got the same error returned.

Tpm2-tools version is 4.1.1
Ubuntu Focal(20.04)

williamcroberts:
@idesai this would be a good one to add to 5.0 for the behavioral changes. @Ivanhu5866 has a patch that perhaps he can share. I see a patch referenced at https://github.com/Ivanhu5866/tpm2-tools/commit/785cc544b30bb38cdd01917a86119409f474fcb9

idesai:
fixed via #2173

PR #2173, opened 2020-08-19
Structure is the wrong size(0x2d5) issue with tpm2_nvdefine without specifying the size

Ivanhu5866:
Please refer to
https://github.com/tpm2-software/tpm2-tools/issues/2171

### DIFF
--- a/tools/tpm2_nvdefine.c
+++ b/tools/tpm2_nvdefine.c
@@ -183,6 +183,48 @@ static void handle_default_attributes(void) {
     }
 }
 
+static void get_max_nv_index_size(ESYS_CONTEXT *ectx, UINT16 *size) {
+
+    *size = 0;
+
+    /* get the max NV index for the TPM */
+    TPMS_CAPABILITY_DATA *capabilities = NULL;
+    tool_rc tmp_rc = tpm2_getcap(ectx, TPM2_CAP_TPM_PROPERTIES, TPM2_PT_FIXED,
+            TPM2_MAX_TPM_PROPERTIES, NULL, &capabilities);
+    if (tmp_rc != tool_rc_success) {
+        *size = TPM2_MAX_NV_BUFFER_SIZE;
+        LOG_ERR("Could not get fixed TPM properties");
+        return;
+    }
+
+    TPMS_TAGGED_PROPERTY *properties = capabilities->data.tpmProperties.tpmProperty;
+    UINT32 count = capabilities->data.tpmProperties.count;
+
+    if (!count) {
+        *size = TPM2_MAX_NV_BUFFER_SIZE;
+        LOG_ERR("Could not get maximum NV index size");
+        goto out;
+    }
+
+    UINT32 i;
+    for (i=0; i < count; i++) {
+        if (properties[i].property == TPM2_PT_NV_INDEX_MAX) {
+            *size = properties[i].value;
+            break;
+        }
+    }
+
+    if (*size == 0) {
+        *size = TPM2_MAX_NV_BUFFER_SIZE;
+        LOG_ERR("Could not find max NV indices in capabilities");
+    }
+
+out:
+    free(capabilities);
+
+    return;
+}
+
 static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chosen) {
 
     tool_rc rc = tool_rc_general_error;
@@ -271,12 +313,12 @@ out:
     return rc;
 }
 
-static tool_rc validate_size(void) {
+static tool_rc validate_size(ESYS_CONTEXT *ectx) {
 
     switch ((ctx.nv_attribute & TPMA_NV_TPM2_NT_MASK) >> TPMA_NV_TPM2_NT_SHIFT) {
         case TPM2_NT_ORDINARY:
             if (!ctx.size_set) {
-                ctx.size = TPM2_MAX_NV_BUFFER_SIZE;
+                get_max_nv_index_size(ectx, &ctx.size);
             }
             break;
         case TPM2_NT_COUNTER:
@@ -339,7 +381,7 @@ static tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
         }
     }
 
-    rc = validate_size();
+    rc = validate_size(ectx);
     if (rc != tool_rc_success) {
         return rc;
     }


tools/tpm2_nvdefine.c
```diff
@@ -201,27 +200,50 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
...
+
+    TPM2_NV_INDEX max = 0;
+    UINT16 nv_max_sz = 0;
+
+    tool_rc tmp_rc = get_max_nv_index(ectx, &nv_max_sz);
+    if (tmp_rc != tool_rc_success) {
+        return tmp_rc;
+    }
+
+    max = TPM2_HR_NV_INDEX | nv_max_sz;
```
williamcroberts, line 239: bitwise or, why? Why are you adding the handle mask into the max size of an NV index? max will always be set and code on 240 will be dead. The original code checked that nv_max_size wasn't 0 before adding in this part of the handle, but thats for handles, not for size.

tools/tpm2_nvdefine.c
```diff
@@ -183,11 +183,10 @@ static void handle_default_attributes(void) {
     }
 }
 
-static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chosen) {
+static tool_rc get_max_nv_index(ESYS_CONTEXT *ectx, UINT16 *size) {
```
williamcroberts, line 186: get_max_nv_index_size

tools/tpm2_nvdefine.c
```diff
@@ -201,27 +200,50 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
     UINT32 count = capabilities->data.tpmProperties.count;
 
     if (!count) {
-        LOG_ERR("Could not get maximum NV index, try specifying an NV index");
+        LOG_ERR("Could not get maximum NV index");
```
williamcroberts, line 203: LOG_ERR("Could not get maximum NV index size");

tools/tpm2_nvdefine.c
```diff
@@ -271,12 +294,17 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
...
-static tool_rc validate_size(void) {
+static tool_rc validate_size(ESYS_CONTEXT *ectx) {
+
+    tool_rc rc = tool_rc_general_error;
 
     switch ((ctx.nv_attribute & TPMA_NV_TPM2_NT_MASK) >> TPMA_NV_TPM2_NT_SHIFT) {
         case TPM2_NT_ORDINARY:
             if (!ctx.size_set) {
-                ctx.size = TPM2_MAX_NV_BUFFER_SIZE;
+		if ((rc = get_max_nv_index(ectx, &ctx.size)) != tool_rc_success) {
```
williamcroberts, line 304: we don't hid functions call sin conditionals unless it makes the code simpler, this doesn't, so:
```C
rc = get_max_nv_index(ectx, &ctx.size);
if (rc != tool_rc_success) {
...
}
```

tools/tpm2_nvdefine.c
```diff
@@ -271,12 +294,17 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
...
+			ctx.size = TPM2_MAX_NV_BUFFER_SIZE;
+			return rc;
```
williamcroberts, line 306: So fail validate_size but set the ctx.size to max? I don't even think that will work as callers of validate_size will cause the tool to exit. I think you want to return tool_rc_success

tools/tpm2_nvdefine.c
```diff
@@ -183,11 +183,10 @@ static void handle_default_attributes(void) {
     }
 }
 
-static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chosen) {
+static tool_rc get_max_nv_index_size(ESYS_CONTEXT *ectx, UINT16 *size) {
```
williamcroberts, line 186: make this void, the return code isn't really used in any meaningful and consistent way. In one path, its only use is to set the default max size, in the other path it causes the tool to exit. Just make this on error set the *size to max via the define and print a warning.

tools/tpm2_nvdefine.c
```diff
@@ -201,27 +200,47 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
...
+
+    tool_rc rc = tool_rc_general_error;
+
+    TPM2_NV_INDEX max = 0;
+    UINT16 nv_max_sz = 0;
+
+    tool_rc tmp_rc = get_max_nv_index_size(ectx, &nv_max_sz);
+    if (tmp_rc != tool_rc_success) {
+        return tmp_rc;
+    }
```
williamcroberts, line 238: You shouldn't need to modify handle_no_index_specified(). It's just looking up the first free NV index, it has nothing to do with size.

tools/tpm2_nvdefine.c
```diff
@@ -201,27 +200,47 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
...
+
+    max = TPM2_HR_NV_INDEX + nv_max_sz;
```
williamcroberts, line 240: This makes no sense to me? Why would you take the maximum size of an nv index in bytes, turn it into an NV index, assign it to max, and then never use max again?

tools/tpm2_nvdefine.c
```diff
@@ -201,27 +200,47 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
...
+    rc = tool_rc_success;
+
+out:
     free(capabilities);
-    capabilities = NULL;
+
+    return rc;
+}
+
+static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chosen) {
```
williamcroberts, line 228: leave this function alone.

tools/tpm2_nvdefine.c
```diff
@@ -271,12 +291,17 @@ static tool_rc handle_no_index_specified(ESYS_CONTEXT *ectx, TPM2_NV_INDEX *chos
...
-static tool_rc validate_size(void) {
+static tool_rc validate_size(ESYS_CONTEXT *ectx) {
+
+    tool_rc rc = tool_rc_general_error;
 
     switch ((ctx.nv_attribute & TPMA_NV_TPM2_NT_MASK) >> TPMA_NV_TPM2_NT_SHIFT) {
         case TPM2_NT_ORDINARY:
             if (!ctx.size_set) {
-                ctx.size = TPM2_MAX_NV_BUFFER_SIZE;
+		rc = get_max_nv_index_size(ectx, &ctx.size);
```
williamcroberts, line 301: just make this void and set on failure to TPM2_MAX_NV_BUFFER_SIZE and print a warning on this case.

tools/tpm2_nvdefine.c
williamcroberts: In fact the error case where it causes the tool to exit shouldn't even exist. Their should be no changes to handle_no_index_specified AFAICT.

Ivanhu5866: Ok...will make get_max_nv_index_size simpler without modifying handle_no_index_specified.

tools/tpm2_nvdefine.c
Ivanhu5866: Ok, will skip this patch.

tools/tpm2_nvdefine.c
Ivanhu5866: I think max will be used for the range to fine the first nv index, maybe I am wrong, but it doesn't matter, this patch will be skipped.

tools/tpm2_nvdefine.c
Ivanhu5866: OK

tools/tpm2_nvdefine.c
Ivanhu5866: Ok

idesai:
> Please refer to
> #2171

Fixes #2171

tools/tpm2_nvdefine.c
```diff
@@ -183,6 +183,48 @@ static void handle_default_attributes(void) {
...
+	*size = TPM2_MAX_NV_BUFFER_SIZE;
+        LOG_ERR("Could not get fixed TPM properties");
+        return;
+    }
+
+    TPMS_TAGGED_PROPERTY *properties = capabilities->data.tpmProperties.tpmProperty;
+    UINT32 count = capabilities->data.tpmProperties.count;
+
+    if (!count) {
+	*size = TPM2_MAX_NV_BUFFER_SIZE;
```
williamcroberts, line 204: indent error, tab. We use 4 spaces.

tools/tpm2_nvdefine.c
```diff
@@ -183,6 +183,48 @@ static void handle_default_attributes(void) {
...
+    UINT32 i;
+    for (i=0; i < count; i++) {
+        if (properties[i].property == TPM2_PT_NV_INDEX_MAX) {
+		*size = properties[i].value;
+		break;
+        }
+    }
+
+    if (*size == 0) {
+	*size = TPM2_MAX_NV_BUFFER_SIZE;
```
williamcroberts, line 218: indent error, tab, use four spaces

tools/tpm2_nvdefine.c
```diff
@@ -183,6 +183,48 @@ static void handle_default_attributes(void) {
...
+static void get_max_nv_index_size(ESYS_CONTEXT *ectx, UINT16 *size) {
+
+    *size = 0;
+
+    /* get the max NV index for the TPM */
+    TPMS_CAPABILITY_DATA *capabilities = NULL;
+    tool_rc tmp_rc = tpm2_getcap(ectx, TPM2_CAP_TPM_PROPERTIES, TPM2_PT_FIXED,
+            TPM2_MAX_TPM_PROPERTIES, NULL, &capabilities);
+    if (tmp_rc != tool_rc_success) {
+	*size = TPM2_MAX_NV_BUFFER_SIZE;
```
williamcroberts, line 195: Here to, spaces instead of tab

PR #2174, opened 2020-08-19
Use correct index for policy branch selection

PeterHuewe:
The user is presented a menu starting from 1, while internally we are
zero based, thus the correct variable has to be decremented in order to
chose the right branch.


williamcroberts:
@PeterHuewe I split your patch, so if we have to revert the policy number change, we don't have to discard the spelling fix.

merged as:
- 88ff35f6bd4c Use correct index for policy branch selection
- 92dbca5b2c92 fapi: fix spelling mistake in tss2_template.c

PeterHuewe:
Thanks for splitting and merging.
It's better this way, yes.

PR #2182, opened 2020-08-27
Fix bashcompletion

### LINKED ISSUE #2129: verify bash completion
Ensure that bash completion works for tpm2 ans tss2 command sets since busybox style integration.

idesai:
Fixes #2129 

There was a growing problem of the tools-options and bash-completion scripts
diverging or becoming inconsistent. This PR is for addressing this issue. The
bash completion scripts are instead generated by directly parsing the man-pages.
Essentially it takes advantage of the checks performed by toggle_options.sh
integration tests which ensures the tool-options and man-pages are consistent.

While this approach has its merits, the following still need to be addressed
and are captured as separate tickets.

1. Since this utility generates bash-completion scripts for the tools, the
installation relies on the eval utility to construct the completion command.
Eg. eval `man_to_bashcompletion.sh`. Can the automake installation scripts
handle this?
#2196 

2. When specifying sub-options, it only picks one and continues. This is the
desired case in most tools but in some other tools this can be a problem.
Eg. tpm2_create --atrributes <tab> gives various attributes as options. However,
after choosing one for completion it immediately goes back to the tool options
as supposed to waiting for sub-options to complete. Eg. tpm2_create --attributes
noda|fixedtpm| etc. instead becomes tpm2_create --attributes noda -
#2195 

3. The busybox style utility uses the completion functions of the individual
tools. However, when specifying the options it pops back the tool name in the
completion. Eg. tpm2 getrandom --help getrandom --hex etc. as supposed to
tpm2 getrandom --help --hex etc.
#2194


idesai:
Need tss2 to fix issue #2183  where initialize routines are run prior to servicing options like --help and --version.

idesai:
> Fixes #2129
> 
> There was a growing problem of the tools-options and bash-completion scripts
> diverging or becoming inconsistent. This PR is for addressing this issue. The
> bash completion scripts are instead generated by directly parsing the man-pages.
> Essentially it takes advantage of the checks performed by toggle_options.sh
> integration tests which ensures the tool-options and man-pages are consistent.
> 
> While this approach has its merits, the following still need to be addressed.
> 
> * [ ]  Since this utility generates bash-completion scripts for the tools, the
>   installation relies on the eval utility to construct the completion command.
>   Eg. eval `man_to_bashcompletion.sh`. Can the automake installation scripts
>   handle this?
> * [ ]  When specifying sub-options, it only picks one and continues. This is the
>   desired case in most tools but in some other tools this can be a problem.
>   Eg. tpm2_create --atrributes  gives various attributes as options. However,
>   after choosing one for completion it immediately goes back to the tool options
>   as supposed to waiting for sub-options to complete. Eg. tpm2_create --attributes
>   noda|fixedtpm| etc. instead becomes tpm2_create --attributes noda -
> * [ ]   The busybox style utility uses the completion functions of the individual
>   tools. However, when specifying the options it pops back the tool name in the
>   completion. Eg. tpm2 getrandom --help getrandom --hex etc. as supposed to
>   tpm2 getrandom --help --hex etc.

@kaiserkyu, @dilyanpalauzov any suggestions?

dilyanpalauzov:
> Can the automake installation scripts handle this?

I do not remember the details on bash-completion or the tss internals, so I answer just the question above.

I propose that the completions are generated only on `make dist` (https://www.gnu.org/software/automake/manual/html_node/The-dist-Hook.html) and included in the tarball.gz.  Then `make install` can assume, that the completion scripts are present and correct.

dilyanpalauzov:
Apart from "dist hook" there are also “Automake-hooks” run by “make install”: https://www.gnu.org/software/automake/manual/html_node/Extending-Installation.html

PR #2185, opened 2020-08-31
fapi: add integration tests for policy or.

JuergenReppSIT:
* A test with different policies in or branches was added.
* A test with two nested or policies was added.

### DIFF
--- a/test/integration/fapi/fapi-nv-write-read-policy-or.sh
+++ b/test/integration/fapi/fapi-nv-write-read-policy-or.sh
@@ -1,0 +1,118 @@
+set -e
+source helpers.sh
+
+start_up
+
+CRYPTO_PROFILE="ECC"
+setup_fapi $CRYPTO_PROFILE
+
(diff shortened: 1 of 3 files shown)


JuergenReppSIT:
All travis test case should work after PR 1835 of tpm2-tss is merged.

AndreasFuchsSIT:
@idesai This fails because tests run against 2.4.2.
Would it make sense to run CI against 2.4.x branch so we always get latest ?

idesai:
> @idesai This fails because tests run against 2.4.2.
> 
> Would it make sense to run CI against 2.4.x branch so we always get latest ?

SGTM as long as a corresponding tpm2-tss release is cut from the .X branch close to the next tools release.

idesai:
@JuergenReppSIT @AndreasFuchsSIT I have a PR to update to tpm2-tss v3.0.1 and tpm2-abrmd v2.3.3, are these changes compatible?

Issue #2183, opened 2020-08-27
tss2 tools shouldn't try to run api initialize routines when options like --help or --version are invoked


idesai:

AndreasFuchsSIT:
@cplappert Could you take care of this please

cplappert:
@AndreasFuchsSIT : We currently use Fapi_GetInfo() to retrieve the api version string when using the --version option. In this case we need the API initialization. Should this be removed?

AndreasFuchsSIT:
I guess since the `tpm2_*` tools do not provide an esys-version, we don't need to provide a fapi-version here.

idesai:
Fixed via #2203

PR #2203, opened 2020-09-24
tss2 *: Avoid unnecessary API initialization

cplappert:
Do not initialize TSS API when options --help or --version
are invoked


cplappert:
This commit addresses #2183.

PR #2218, opened 2020-09-30
tpm2_startauthsession: add parameter --nonce-tpm

joholl:
In some cases, the user might want to get the `nonceTPM` generated at session start. For example, the message to be signed with PolicySigned could be constructed and signed remote (for which `nonceTPM` is needed).

Add a parameter `--nonce-tpm` to `tpm2_startauthsession` to enable saving `nonceTPM` to a file.


idesai:
@joholl please take a look at [this](https://github.com/tpm2-software/tpm2-tools/blob/4c0949bce5320eb43cad944635e50c5687b858f9/test/integration/tests/abrmd_policysigned.sh#L84-L103).

We went this route before and determined that if nonceTpm is needed it can be read internally within other tools.
However, I appreciate the policysigned use case wherein the signing doesn't happen on the same platform.

IMHO this should be a separate tool like **tpm2_readsession**, like that we won't restrict such usage only when the session is started since the nonces are rolled at every invocation.

+ @williamcroberts

joholl:
I agree. I hope I'll find some time to work on it some more, soon. I'll open a RFC PR then.

PR #2221, opened 2020-10-01
test: add support for swtpm simulator

diabonas:
tpm2-tss supports the [swtpm](https://github.com/stefanberger/swtpm) TPM simulator since version 3.0.0. Two small changes (adding Triple DES to the algorithm list, choosing a different unsupported algorithm for negative testing in [`testparms.sh`](https://github.com/tpm2-software/tpm2-tools/blob/26f60611a045320ceb2554b6929a54bb3e3de06b/test/integration/tests/testparms.sh#L66)) were necessary, other than that everything works out of the box after adapting [`test/integration/helpers.sh`](https://github.com/tpm2-software/tpm2-tools/compare/master...diabonas:add-swtpm?expand=1#diff-7e13f2fcb4dce9b9de0bb58c89995153).

Support for ibmswtpm (aka `tpm_server`) is kept for backwards compatibility, but swtpm is preferred if it is available. Since the Docker images come with swtpm preinstalled, this means that this PR [switches CI testing from ibmswtpm to swtpm](https://travis-ci.org/github/tpm2-software/tpm2-tools/jobs/732035679#L2286) as well.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -311,7 +311,7 @@ test_unit_test_tpm2_eventlog_yaml_LDADD = $(CMOCKA_LIBS) $(LDADD)
 
 AM_TESTS_ENVIRONMENT =	\
 	TPM2_ABRMD=tpm2-abrmd; export TPM2_ABRMD; \
-	TPM2_SIM=tpm_server; export TPM2_SIM; \
+	TPM2_SIM=$(TPM2_SIM); export TPM2_SIM; \
 	PATH=$(abs_builddir)/tools:$(abs_builddir)/tools/misc:$(abs_top_srcdir)/test/integration:$(PATH); \
 	TPM2_TOOLS_TEST_FIXTURES=$(abs_top_srcdir)/test/integration/fixtures; \
 	SRCDIR=$(abs_srcdir); export SRCDIR; BUILDDIR=$(abs_builddir); export BUILDDIR; \
--- a/configure.ac
+++ b/configure.ac
@@ -140,9 +140,12 @@ AS_IF([test "x$enable_unit" != xno], [
     AS_IF([test $tpm2_abrmd = no],
           [AC_MSG_ERROR([Required executable tpm2_abrmd not found, try setting PATH])])
 
+    AC_CHECK_PROG([swtpm], [swtpm], yes, no)
     AC_CHECK_PROG([tpm_server], [tpm_server], yes, no)
-    AS_IF([test $tpm_server = no],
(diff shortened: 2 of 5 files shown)


diabonas:
Interestingly, using swtpm even appears to be quite a bit quicker ([1:40 total CI time](https://travis-ci.org/github/tpm2-software/tpm2-tools/builds/732035678) vs. [2:20 total CI time with ibmswtpm](https://travis-ci.org/github/tpm2-software/tpm2-tools/builds/732032185)).

anatol:
Hi, I am trying to test `clevis` (that utilizes `tpm2-tools`) with `swtpm`. So I:
 * applied this patchseries to top of the master and then rebuilt the tools
 * created the tpm state `swtpm_setup --tpm-state /tmp/TestDecryptTpm2676516833/001 --tpm2 --ecc --display --create-ek-cert --create-platform-cert --lock-nvram`
 * ran swtpm as `swtpm socket --tpmstate dir=/tmp/TestDecryptTpm2676516833/001 --tpm2 --ctrl type=tcp,port=2322 --server type=tcp,port=2321 --flags not-need-init`
 * set the envvar `TPM2TOOLS_TCTI=swtpm:host=localhost,port=2321`.

But running `clevis` fails with

```
WARNING:esys:src/tss2-esys/api/Esys_StartAuthSession.c:390:Esys_StartAuthSession_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_StartAuthSession.c:136:Esys_StartAuthSession() Esys Finish ErrorCode (0x00000100) 
ERROR: Esys_StartAuthSession(0x100) - tpm:error(2.0): TPM not initialized by TPM2_Startup or already initialized
ERROR: Invalid parent key authorization
ERROR: Unable to run tpm2_createprimary
```

So the toolset was able to connect the TCP server but then it fails with a cryptic error `TPM not initialized by TPM2_Startup or already initialized`...

diabonas:
> So the toolset was able to connect the TCP server but then it fails with a cryptic error `TPM not initialized by TPM2_Startup or already initialized`...

You need to initialise the simulator by running `tpm2_startup -c` before it will accept further commands, `--flags not-need-init` only "physically" turns on the TPM, it does not run `TPM2_Startup`. As an alternative, using `--flags not-need-init,startup-clear` should work as well [according to the swtpm documentation](https://github.com/stefanberger/swtpm/blob/ab40d27ab138e7300a62d9a0367112c3d1824646/man/man8/swtpm.pod#user-content-flags-not-need-init-startup-clear-startup-state-startup-deactivated-startup-none).

Issue #2033, opened 2020-05-18
arm64 builds in travis


williamcroberts:
It might be worthwhile to do one build, perhaps not in a container, where we use the arch keyword in travis and specify arm.

See:
https://docs.travis-ci.com/user/multi-cpu-architectures

idesai:
Irrelevant now that we moved to github actions

PR #2224, opened 2020-10-02
Fix tpm2_kdfa for big endian systems

stefanberger:
This patch fixes an issue with tpm2_makeidentity when run on a big
endian system and the result is used by a little endian system, as
may be the case with keylime. The byte swapping must only be done
on little endian systems.

Signed-off-by: Stefan Berger <stefanb@linux.ibm.com>

### DIFF
--- a/lib/tpm2_kdfa.c
+++ b/lib/tpm2_kdfa.c
@@ -1,5 +1,7 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 
+#include <string.h>
+
 #include "log.h"
 #include "tpm2_kdfa.h"
 #include "tpm2_openssl.h"
@@ -8,10 +10,7 @@ TSS2_RC tpm2_kdfa(TPMI_ALG_HASH hash_alg, TPM2B *key, char *label,
         TPM2B *context_u, TPM2B *context_v, UINT16 bits,
         TPM2B_MAX_BUFFER *result_key) {
     TPM2B_DIGEST tpm2b_label, tpm2b_bits, tpm2b_i_2;
-    UINT8 *tpm2b_bits_ptr = &tpm2b_bits.buffer[0];
-    UINT8 *tpm2b_i_2_ptr = &tpm2b_i_2.buffer[0];
     TPM2B_DIGEST *buffer_list[8];
-    UINT32 bits_swizzled, i_swizzled;
     TSS2_RC rval = TPM2_RC_SUCCESS;
     int i, j;
     UINT16 bytes = bits / 8;
@@ -21,8 +20,8 @@ TSS2_RC tpm2_kdfa(TPMI_ALG_HASH hash_alg, TPM2B *key, char *label,
     tpm2b_i_2.size = 4;
 
     tpm2b_bits.size = 4;
-    bits_swizzled = tpm2_util_endian_swap_32(bits);
-    *(UINT32 *) tpm2b_bits_ptr = bits_swizzled;
+    UINT32 bits_be = tpm2_util_hton_32(bits);
+    memcpy(&tpm2b_bits.buffer[0], &bits_be, sizeof(bits_be));
 
     for(i = 0; label[i] != 0 ;i++ );
 
@@ -58,9 +57,8 @@ TSS2_RC tpm2_kdfa(TPMI_ALG_HASH hash_alg, TPM2B *key, char *label,
     while (result_key->size < bytes) {
         TPM2B_DIGEST tmpResult;
         // Inner loop
-
-        i_swizzled = tpm2_util_endian_swap_32(i);
-        *(UINT32 *) tpm2b_i_2_ptr = i_swizzled;
+        bits_be = tpm2_util_hton_32(i);
+        memcpy(&tpm2b_i_2.buffer[0], &bits_be, sizeof(bits_be));
 
         j = 0;
         buffer_list[j++] = (TPM2B_DIGEST *) &(tpm2b_i_2);


idesai:
> This patch fixes an issue with tpm2_makeidentity when run on a big
> endian system and the result is used by a little endian system, as
> may be the case with keylime. The byte swapping must only be done
> on little endian systems.
> 
> Signed-off-by: Stefan Berger [stefanb@linux.ibm.com](mailto:stefanb@linux.ibm.com)

@stefanberger since we still have to address #2033, did you happen to test the change on an arm64/32 build?

stefanberger:
> @stefanberger since we still have to address #2033, did you happen to test (run the unit/integrations tests) for the change on an arm64/32 build?

No, a s390 (z system). I don't have access to ARM.

I have  more patches in this direction: https://github.com/stefanberger/tpm2-tools/commits/4.X%2Bbea_fixes

One other issue is that whatever `tpm2_checkquote` seems to look at (from a file) is written in the native endianess of the system that wrote it, so typically little endian today. It's odd to see some of the TPM2 data structures written in little endian, but this is what it is and it's probably least disruptive for the projects to have big endian write the stuff (probably in tpm2_quote) in little endian as well before it's sent from the keylime agent side to the keylime_tenant side that at least looks at it the first time it seems...

lib/tpm2_kdfa.c
```diff
@@ -21,8 +18,7 @@ TSS2_RC tpm2_kdfa(TPMI_ALG_HASH hash_alg, TPM2B *key, char *label,
     tpm2b_i_2.size = 4;
 
     tpm2b_bits.size = 4;
-    bits_swizzled = tpm2_util_endian_swap_32(bits);
-    *(UINT32 *) tpm2b_bits_ptr = bits_swizzled;
+    *(UINT32 *)&tpm2b_bits.buffer[0] = tpm2_util_hton_32(bits);
```
idesai, line 21: Trap representation, makes it concise though.

stefanberger: What do you mean? 
Does it need a space before the `&`?

stefanberger: Or do you want me to revert the changes further above and use `tpm2b_bits_ptr` ?

idesai:
- [ ] Test build with qemu-arm32v7

lib/tpm2_kdfa.c
idesai: @stefanberger currently fails arm32v7 build
  CC       lib/lib_libcommon_a-tpm2_kdfa.o
../../lib/tpm2_kdfa.c: In function 'tpm2_kdfa':
../../lib/tpm2_kdfa.c:21:5: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     *(UINT32 *)&tpm2b_bits.buffer[0] = tpm2_util_hton_32(bits);
     ^
../../lib/tpm2_kdfa.c:57:9: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
         *(UINT32 *)&tpm2b_i_2.buffer[0] = tpm2_util_hton_32(i);
         ^

idesai, line 21:
```suggestion
    UINT8 *tpm2b_buffer_ptr = &tpm2b_bits.buffer[0];
    *(UINT32 *) tpm2b_buffer_ptr  = tpm2_util_hton_32(bits);
```

lib/tpm2_kdfa.c
```diff
@@ -58,9 +54,7 @@ TSS2_RC tpm2_kdfa(TPMI_ALG_HASH hash_alg, TPM2B *key, char *label,
     while (result_key->size < bytes) {
         TPM2B_DIGEST tmpResult;
         // Inner loop
-
-        i_swizzled = tpm2_util_endian_swap_32(i);
-        *(UINT32 *) tpm2b_i_2_ptr = i_swizzled;
+        *(UINT32 *)&tpm2b_i_2.buffer[0] = tpm2_util_hton_32(i);
```
idesai, line 57:
```suggestion
        tpm2b_buffer_ptr = &tpm2b_i_2.buffer[0];
        *(UINT32 *) tpm2b_buffer_ptr  = tpm2_util_hton_32(i);
```

idesai (changes requested):

lib/tpm2_kdfa.c
stefanberger: Your suggestion doesn't work for me (gcc 7.5.0 on s390x):
```
lib/tpm2_kdfa.c: In function ‘tpm2_kdfa’:
lib/tpm2_kdfa.c:22:5: error: invalid operands to binary * (have ‘BYTE * {aka unsigned char *}’ and ‘UINT32 * {aka unsigned int *}’)
     UINT8 *tpm2b_buffer_ptr = &tpm2b_bits.buffer[0]
                               ~~~~~~~~~~~~~~~~~~~~~
     *(UINT32 *) tpm2b_buffer_ptr = tpm2_util_hton_32(bits);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

stefanberger: Pushed another solution now. Could also fall back to the previous code...

williamcroberts:
For those brave enough, I would also accept a re-write of the KDFA code, that is a nightmare I inherited and never untangled....

lib/tpm2_kdfa.c
stefanberger: ... with the `tpm2_util_hton_32()` applied of course

idesai: > Pushed another solution now. Could also fall back to the previous/original code...

I will post feedback with the build status shortly.

idesai:
> For those brave enough, I would also accept a re-write of the KDFA code, that is a nightmare I inherited and never untangled....

First we will need to add unit/ integration tests including endianness scenarios like these. But I agree, this needs a rewrite.

PR #2230, opened 2020-10-06
Fix tpm2_nvreadpublic for big endian systems

stefanberger:
Fix the shifting of TPM2_HT_NV_INDEX to the most significant octet (MSO).
tpm2_util_hton_32(TPM2_HT_NV_INDEX) would only shift the TPM2_HT_NV_INDEX
on little endian systems and do nothing on big endian ones, so we use an
explicit shift toaccomplish this also on big endian systems.

Signed-off-by: Stefan Berger <stefanb@linux.ibm.com>


idesai:
- [ ] Build tested with qemu-arm32v7 build.

stefanberger:
Thanks!

Issue #2235, opened 2020-10-14
Having problem decoding yaml output file for eventlog


huang195:
Decoding of the yaml output file from eventlog is pretty smooth in most cases, but certain fields having certain values is causing issues for us. An example is

```
Event:
  VariableName: test
  VariableData: 010000000200
```

The field `VariableData` is supposed to be a hexstring, but due to the leading `0`, it's being treated as an octal. The underlying issue seems to be how should binary data be encoded in a yaml file. The canonical way seems to be using the Base64 encoding, but it's not so great for readability. I'm wondering whether or not as a compromise, we could encode the hexstring as a string by simply adding quotes around it, e.g., 

```
  VariableData: "010000000200"
```

There are other ways to cast a field to a string, but this seems to be the most readable.

PR #2236, opened 2020-10-14
adding quotes around hex string to avoid ambiguity during decoding

huang195:
To address https://github.com/tpm2-software/tpm2-tools/issues/2235.


idesai:
@huang195 can you please share the eventlog where you experienced the problem by:
1. Adding the eventlog blob file to the [test-fixture](https://github.com/tpm2-software/tpm2-tools/tree/master/test/integration/fixtures).
2. Add the file to the [integration-tests](https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/tests/eventlog.sh#L33)

huang195:
@idesai I uploaded the log that you can use to reproduce this problem. The particular part that's breaking for us is
```
      UnicodeName: BootOrder
      VariableData: 010000000200
```

`VariableData` was read in incorrectly as an octal.

williamcroberts (approved):

PR #2238, opened 2020-10-15
tpm2-checkqoute: fix wrong format specifier

PeterHuewe:
Fixes this compilation error:
```In file included from tools/misc/tpm2_checkquote.c:12:0:
tools/misc/tpm2_checkquote.c: In function ‘parse_selection_data_from_file’:
tools/misc/tpm2_checkquote.c:270:17: error: format ‘%zu’ expects
argument of type ‘size_t’, but argument 6 has type ‘__uint64_t {aka long
long unsigned int}’ [-Werror=format=]
         LOG_ERR("Malformed PCR file, pcr count cannot be greater than
%zu, got: %zu",
```
`le64toh `always returns an `uint64_t`, thus `%PRIu64`

### DIFF
--- a/tools/misc/tpm2_checkquote.c
+++ b/tools/misc/tpm2_checkquote.c
@@ -267,7 +267,7 @@ static bool parse_selection_data_from_file(FILE *pcr_input,
     }
 
     if (le64toh(pcrs->count) > ARRAY_LEN(pcrs->pcr_values)) {
-        LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %zu",
+        LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %" PRIu64 " ",
                 ARRAY_LEN(pcrs->pcr_values), le64toh(pcrs->count));
         return false;
     }


tools/misc/tpm2_checkquote.c
```diff
@@ -267,7 +267,7 @@ static bool parse_selection_data_from_file(FILE *pcr_input,
     }
 
     if (le64toh(pcrs->count) > ARRAY_LEN(pcrs->pcr_values)) {
-        LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %zu",
+        LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %llu",
```
idesai, line 270: The C standard only guarantees a minimal precision for integer types and so the specifier may not be portable for a fixed-width type.
```suggestion
        LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %" PRIu64 " ",
```

idesai (changes requested):

tools/misc/tpm2_checkquote.c
PeterHuewe: I wasn't sure whether we use these kinda ugly PRIu64 - but we do, so then yes this is more correct here.

idesai (approved):

tools/misc/tpm2_checkquote.c
```diff
@@ -267,7 +267,7 @@ static bool parse_selection_data_from_file(FILE *pcr_input,
     }
 
     if (le64toh(pcrs->count) > ARRAY_LEN(pcrs->pcr_values)) {
-        LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %zu",
+        LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %" PRIu64 " ",
                 ARRAY_LEN(pcrs->pcr_values), le64toh(pcrs->count));
```
williamcroberts, line 271: For fixed width 64 bit values, PRIu64 is the correct specifier. %zu should be reserved for size_t.

@idesai why is their a le64toh call? We have all that endian swapping code to avoid using endian.h. I avoided endian.h due to portability issues, but that might be moot using clang to compile on Windows. If we're going to use endian.h, I have no problems with it that I forsee, other than doing:
```C
#if defined(__FreeBSD__)
#include <sys/endian.h>
#else
#include <endian.h>
#endif
```
and then we can delete a lot of the endian swapping code and just use endian.h and perhaps consider something like:
  - https://gist.github.com/PkmX/63dd23f28ba885be53a5

for dealing with windows....

idesai: Sounds good. I like this direction. I will create a ticket.

idesai: Created #2241 to track these changes.

gotthardp:
Well, on little-endian systems (at least on my older raspberry) is `le64toh` defined as `#define le64toh(x) (x)`, hence it is **not** always `uint64_t`. Since on 32b systems (such as the raspberry) the `size_t` is `unsigned int`, the build fails with argument mismatch again
```
tools/misc/tpm2_checkquote.c:271:17: error: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 6 has type ‘size_t {aka unsigned int}’ [-Werror=format=]
         LOG_ERR("Malformed PCR file, pcr count cannot be greater than %zu, got: %" PRIu64 " ",
```

PR #2242, opened 2020-10-16
tpm2_eventlog.c: Fix for possible bad bit shift operation

idesai:
Because pcr_index is used for array indexing and bit-shift
operations it is 1 less than the max value.

Also moving the check to the top to fail fast/ early.

### DIFF
--- a/lib/tpm2_eventlog.c
+++ b/lib/tpm2_eventlog.c
@@ -35,8 +35,15 @@ bool foreach_digest2(tpm2_eventlog_context *ctx, unsigned pcr_index, TCG_DIGEST2
         return false;
     }
 
-    bool ret = true;
+    /* Because pcr_index is used for array indexing and bit-shift operations it
+       is 1 less than the max value */
+    if (pcr_index > (TPM2_MAX_PCRS - 1)) {
+        LOG_ERR("PCR Index %d is out of bounds for max available PCRS %d",
+        pcr_index, TPM2_MAX_PCRS);
+        return false;
+    }
 
+    bool ret = true;
     size_t i;
     for (i = 0; i < count; ++i) {
         if (size < sizeof(*digest)) {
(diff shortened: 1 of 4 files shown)


lib/tpm2_eventlog.c
```diff
@@ -35,8 +35,15 @@ bool foreach_digest2(tpm2_eventlog_context *ctx, unsigned pcr_index, TCG_DIGEST2
...
-    bool ret = true;
+    /* Because pcr_index is used for array indexing and bit-shift operations it
+       is 1 less than the max value */
+    if (pcr_index > (TPM2_MAX_PCRS - 1)) {
+        LOG_ERR("PCR Index %d is out of bounds for max available PCRS %d",
+        pcr_index, TPM2_MAX_PCRS);
+        return false;
+    }
 
+    bool ret = true;
```
idesai, line 46:
```suggestion
bool result = true;
```

lib/tpm2_eventlog.c
```diff
@@ -421,16 +421,18 @@ bool specid_event(TCG_EVENT const *event, size_t size,
 
 bool parse_eventlog(tpm2_eventlog_context *ctx, BYTE const *eventlog, size_t size) {
 
-    TCG_EVENT_HEADER2 *next;
-    TCG_EVENT *event = (TCG_EVENT*)eventlog;
-    bool ret;
+    if(!eventlog || (sizeof eventlog < sizeof(TCG_EVENT))) {
```
idesai, line 424: should be size < sizeof(TCG_EVENT)

lib/tpm2_eventlog.c
```diff
@@ -421,16 +421,18 @@ bool specid_event(TCG_EVENT const *event, size_t size,
...
+        return false;
+    }
 
+    TCG_EVENT *event = (TCG_EVENT*)eventlog;
     if (!event) {
         return false;
     }
 
     if (event->eventType == EV_NO_ACTION) {
```
idesai, line 433: why is this a valid access?— in that it is neither the effective type nor character type.

tools/misc/tpm2_certifyX509certutil.c
```diff
@@ -93,49 +95,46 @@ static struct name_fields names[] = {
...
-        goto err_free;
+        return tool_rc_general_error;
     }
 
-    r = read(fd, buf, size);
-    if (r != size) {
-        LOG_ERR("read failed");
-        goto err_free;
+    char* buf = calloc(1, size);
+    if (!buf) {
```
idesai, line 119: close fd here

lib/tpm2_eventlog.c
idesai: > why is this a valid access?— in that it is neither the effective type nor character type.

TCG_event is a packed structure.

PR #2255, opened 2020-10-28
lib/tpm2_session.c: Update tpm2_session_close

idesai:
1. Declare/ define variables closes to their scope of use
2. Set rc to tool_rc_general error explicitly on error paths
3. Do not use temporary return from function calls directly
   as function arguments.


williamcroberts:
Negative diffstat, nice

PR #2258, opened 2020-10-30
CI: switch to github container registry

williamcroberts:
Switch from Dockerhub to Github Container Registry. This is due to
Dockerhub limiting pulls for unauthenticated users. Also, change
DOCKER_TAG to DOCKER_IMAGE since it's no longer a "tag" and is the name
of the image itself.


williamcroberts:
@idesai I am merging this so your CI doesn't go down on Monday with the Dockerhub pull limits.

PR #2261, opened 2020-11-03
tpm2_eventlog: Fix post code

huang195:
One of our machines with SuperMicro BIOS is producing an eventlog with the following event:
```
  - EventNum: 2
    PCRIndex: 0
    EventType: EV_POST_CODE
    DigestCount: 2
    Digests:
      - AlgorithmId: sha1
        Digest: "2ae338407a3d9304e82360e9891ac12b89e838a6"
      - AlgorithmId: sha256
        Digest: "533a706b2c32ac3b42342959d4ff906135ebe9de52ecb4b0e0804c153aedcbe4"
    EventSize: 16
    Event: ''
```
The `Event` field is empty despite `EventSize` being 16. It seems the spec has multiple ways to interpret this field. One way is to treat it as an ASCII string, as documented in the comment here: https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_eventlog_yaml.c#L180-L191. Another way is to treat it as a 16-byte EV_EFI_PLATFORM_FIRMWARE_BLOB payload (added a comment in the PR). The latter seems to be how SuperMicro is encoding this event type. The PR handles both ways that the payload is encoded. A SuperMicro eventlog is included to reproduce this issue.


lib/tpm2_eventlog_yaml.c
```diff
@@ -188,16 +188,36 @@ static bool yaml_uefi_var_data(UEFI_VARIABLE_DATA *data) {
...
+ *
+ * Section 9.2.5 also says "...Below is the definition of the
+ * UEFI_PLATFORM_FIRMWARE_BLOB structure that the CRTM MUST put 
+ * into the Event Log entry TCG_PCR_EVENT2.event[1] field for
+ * event types EV_POST_CODE, EV_S_CRTM_CONTENTS, and
+ * EV_EFI_PLATFORM_FIRMWARE_BLOB."
  */
+
 static bool yaml_uefi_post_code(const TCG_EVENT2 * const event)
 {
```
idesai, line 200:
```suggestion
static bool yaml_uefi_post_code(const TCG_EVENT2* const event) {
```

lib/tpm2_eventlog_yaml.c
```diff
@@ -188,16 +188,36 @@ static bool yaml_uefi_var_data(UEFI_VARIABLE_DATA *data) {
...
-    const char * const data = (const char *) event->Event;
     const size_t len = event->EventSize;
 
-    tpm2_tool_output(
-        "    Event: '%.*s'\n",
-        (int) len,
-        data);
+    // if length is 16, we treat it as EV_EFI_PLATFORM_FIRMWARE_BLOB
```
idesai, line 203:
```suggestion
    /* if length is 16, we treat it as EV_EFI_PLATFORM_FIRMWARE_BLOB */
```

lib/tpm2_eventlog_yaml.c
```diff
@@ -188,16 +188,36 @@ static bool yaml_uefi_var_data(UEFI_VARIABLE_DATA *data) {
...
+                         blob->BlobLength);
+    }
+    // otherwise, we treat it as an ASCII string
+    else {
+        const char * const data = (const char *) event->Event;
+        tpm2_tool_output(
+            "    Event: '%.*s'\n",
+            (int) len,
+            data);
+    }
```
idesai, line 220:
```suggestion
    } else { // otherwise, we treat it as an ASCII string
        const char* const data = (const char *) event->Event;
        tpm2_tool_output("    Event: '%.*s'\n", (int) len, data);
    }
```

idesai (changes requested):

idesai:
@huang195 I would appreciate if you can also add the eventlog binary to our test fixtures. Thank you for this PR.

idesai (approved):

PR #2271, opened 2020-11-19
tpm2_getekcertificate: Fix tool failing to return error/non-zero for HTTP 404

idesai:


idesai:
Fixes #2273

PR #2284, opened 2020-12-07
tpm2_eventlog differentiate parsed events vs. unparsed and better handle parsed strings that could consistent multiple lines

huang195:
Now that more types of events are parsed, it would be nice to be able to differentiate (by a human or by a script) eventlog entries that are parsed vs. unparsed. One possible way is to look at the the data type of what is after `Event: `. If it's a string, then it's unparsed. Otherwise, it's parsed.

Additionally, added yaml annotation to handle parsed strings that could consist multiple lines.

### DIFF
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -210,9 +210,8 @@ static bool yaml_uefi_post_code(const TCG_EVENT2* const event) {
                          blob->BlobLength);
     } else { // otherwise, we treat it as an ASCII string
         const char* const data = (const char *) event->Event;
-        tpm2_tool_output("    Event:\n"
-                         "      Action: |-\n"
-                         "        %.*s\n", (int) len, data);
+        tpm2_tool_output("    Event: |-\n"
+                         "      %.*s\n", (int) len, data);
 
     }
     return true;
@@ -256,9 +255,8 @@ bool yaml_uefi_platfwblob(UEFI_PLATFORM_FIRMWARE_BLOB *data) {
 /* TCG PC Client PFP section 9.4.4 */
 bool yaml_uefi_action(UINT8 const *action, size_t size) {
 
-    tpm2_tool_output("    Event:\n"
-                     "      Action: |-\n"
-                     "        %.*s\n", (int) size, action);
+    tpm2_tool_output("    Event: |-\n"
+                     "      %.*s\n", (int) size, action);
 
     return true;
 }
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -210,7 +210,10 @@ static bool yaml_uefi_post_code(const TCG_EVENT2* const event) {
                          blob->BlobLength);
     } else { // otherwise, we treat it as an ASCII string
         const char* const data = (const char *) event->Event;
-        tpm2_tool_output("    Event: '%.*s'\n", (int) len, data);
+        tpm2_tool_output("    Event:\n"
+                         "      Action: |-\n"
+                         "        %.*s\n", (int) len, data);
+
     }
     return true;
 }
@@ -253,7 +256,9 @@ bool yaml_uefi_platfwblob(UEFI_PLATFORM_FIRMWARE_BLOB *data) {
 /* TCG PC Client PFP section 9.4.4 */
 bool yaml_uefi_action(UINT8 const *action, size_t size) {
 
-    tpm2_tool_output("    Event: '%.*s'\n", (int) size, action);
+    tpm2_tool_output("    Event:\n"
+                     "      Action: |-\n"
+                     "        %.*s\n", (int) size, action);
 
     return true;
 }
@@ -266,7 +271,7 @@ bool yaml_uefi_action(UINT8 const *action, size_t size) {
 bool yaml_ipl(UINT8 const *description, size_t size) {
 
     tpm2_tool_output("    Event:\n"
-                     "      String: |\n");
+                     "      String: |-\n");
 
     /* We need to handle when description contains multiple lines. */
     size_t i, j;


lib/tpm2_eventlog_yaml.c
```diff
@@ -210,7 +210,10 @@ static bool yaml_uefi_post_code(const TCG_EVENT2* const event) {
                          blob->BlobLength);
     } else { // otherwise, we treat it as an ASCII string
         const char* const data = (const char *) event->Event;
-        tpm2_tool_output("    Event: '%.*s'\n", (int) len, data);
+        tpm2_tool_output("    Event:\n"
+                         "      Action: |-\n"
+                         "        %.*s\n", (int) len, data);
```
williamcroberts, line 215: So instead of:
```yaml
Event: abcdefg...
```

its:
```yaml
Event:
    Action: |
      abcdefg...
```
Two things:
1. Doesn't that cause YAML consumers of this to have to know of the extra "Action" key, breaking backwards compat?
2. What does this actually change besides moving the data down a line?

huang195: > What does this actually change besides moving the data down a line?

You're right the content will be same, but if the string spans multiple lines, I don't believe we would be able to handle it. The added yaml annotation is to have the flexibility to handle both single and multiple line strings.

> Doesn't that cause YAML consumers of this to have to know of the extra "Action" key, breaking backwards compat?

Unfortunately, it would break backward compatibility. But you could argue that anything that parses additional event types also breaks backward compatibility, e.g.,

```
  Event: abcdef
```

would become

```
  Event:
    Field1: abc
    Field2: def
```

after we add the code to parse this new type of event. If a script depends on `Event` being a string of the raw bytes, this would break that script. I don't know if there's a good solution to avoid breaking backward compatibility.

williamcroberts: > > What does this actually change besides moving the data down a line?
> 
> You're right the content will be same, but if the string spans multiple lines, I don't believe we would be able to handle it. The added yaml annotation is to have the flexibility to handle both single and multiple line strings.

I see since the original handling released in 5.0 as commit 389f0609d was missing the handling for UEFI_PLATFORM_FIRMWARE_BLOB in fix post code as commit e4108a7c00cbb. However, that added a new sub type, and the else clause stayed the same. So old code would still get the old events under the else, but miss the new events. But changing the else clause structure will cause old code to have to figure out if they need the events or events->action in the YAML dictionary.

Is it possible for the data to contain a newline? If so then we would probably just want to add the multiline trailing carriage removed pipe dash character ("|-"). That wouldn't break readers of that data.

So adding the POST_CODE fix commit e4108a7c00cbb will cause folks to miss some events, but won't break their code.

> 
> > Doesn't that cause YAML consumers of this to have to know of the extra "Action" key, breaking backwards compat?
> 
> Unfortunately, it would break backward compatibility. But you could argue that anything that parses additional event types also breaks backward compatibility, e.g.,

That's the problem, the else isn't additional AFAICT. You can modify the if since it's not on a release, but the else is on the release. However, if their is a bug in the handling of the else, then we can say, it's broken and modify it anyways.

I'm trying to make sure we don't break folks, I want to keep the tools output as stable as possible for consumers.

> 
> ```
>   Event: abcdef
> ```
> 
> would become
> 
> ```
>   Event:
>     Field1: abc
>     Field2: def
> ```
> 
> after we add the code to parse this new type of event. If a script depends on `Event` being a string of the raw bytes, this would break that script. I don't know if there's a good solution to avoid breaking backward compatibility.

huang195: Thanks for that explanation. Breaking backward compatibility is bad, and should be avoided. I completely agree. Would the next release, i.e., `6.0`, be more open to breaking changes? If so, what is the process of merging breaking changes (if it really cannot be avoided)?

As the title suggests, there are 2 things this PR attempts to improve:
- Any string field of a parsed event should be able to handle multiple lines. As of today, `EV_EFI_ACTION` and `EV_EFI_PLATFORM_FIRMWARE_BLOB` would break if multi-line data is encoded. (Legit firmware usually does not produce multi-line data for these event types, but eventlog can always be manipulated to contain multi-line strings for various reasons)
- Be able to differentiate events that are parsed vs. raw bytes. As of today, a human operator should be able to differentiate them without problems in the parsed output, but for a script, it would be harder, if not impossible. My initial thought was if it's not parsed, the event field would be a string, e.g.,

```
Event: '123456780abcdef'
```

as opposed to a parsed field, e.g.,

```
Event:
  Action: '1234567890abcdef'
```

Perhaps, the 2nd point doesn't have a valid use case yet, so I can remove this part from the PR to keep backward compatibility while allowing the code to handle multi-line string field. @williamcroberts what do you think?

huang195:
@williamcroberts per our discussion above, I made the change that would keep backward compatibility while allowing the parser to handle string fields that could consist multiple lines.

lib/tpm2_eventlog_yaml.c
williamcroberts: > Thanks for that explanation. Breaking backward compatibility is bad, and should be avoided. I completely agree. Would the next release, i.e., `6.0`, be more open to breaking changes? 

Even though major releases are allowed to contain non-backwards compat changes, we want to avoid them as much as possible. The 5.0 release had backwards incompat changes, but they all had backwards compatible ways of dealing with it. 

> If so, what is the process of merging breaking changes (if it really cannot be avoided)?

Major version bump. But we need a *really* good justification for it. Generally we would have a option to opt into the new format, so old consumers get the output they expect.

> 
> As the title suggests, there are 2 things this PR attempts to improve:
> 
> * Any string field of a parsed event should be able to handle multiple lines. As of today, `EV_EFI_ACTION` and `EV_EFI_PLATFORM_FIRMWARE_BLOB` would break if multi-line data is encoded. (Legit firmware usually does not produce multi-line data for these event types, but eventlog can always be manipulated to contain multi-line strings for various reasons)
> * Be able to differentiate events that are parsed vs. raw bytes. As of today, a human operator should be able to differentiate them without problems in the parsed output, but for a script, it would be harder, if not impossible. My initial thought was if it's not parsed, the event field would be a string, e.g.,
> 
> ```
> Event: '123456780abcdef'
> ```
> 
> as opposed to a parsed field, e.g.,
> 
> ```
> Event:
>   Action: '1234567890abcdef'
> ```
> 
> Perhaps, the 2nd point doesn't have a valid use case yet, so I can remove this part from the PR to keep backward compatibility while allowing the code to handle multi-line string field. @williamcroberts what do you think?

Yeah you can add the multiline stuff. Just keep it as Event: '1234567890abdef'.

We can always add new outputs, just have to keep old outputs stable. In this case, folks might miss the new sub-type. But since we're fixing a bug in that we were missing UEFI_PLATFORM_FIRMWARE_BLOB as indicated by the spec, im less concerned about it.

williamcroberts (approved):

PR #2304, opened 2021-01-21
scripts: add mssim_cmd.sh

williamcroberts:
Add a script for manipulating the control socket on the ibmtpm
simulator.


williamcroberts:
@idesai this will hang if tpm2-abrmd is in place because the mssim TCTI takes control of the command socket.

williamcroberts:
Example use:
```bash
./mssim_cmd.sh on
```

idesai:
> Example use:
> 
> ```shell
> ./mssim_cmd.sh on
> ```
@williamcroberts should the script then detect abrmd running and warn?

PR #2300, opened 2021-01-13
tpm2_eventlog: Parse keys used for secure boot (i.e., PK, KEK, db, dbx)

huang195:
In the secureboot mode, the eventlog contains various keys (e.g., PK, KEK, db, and dbx) used to perform boot-time verification. For anyone doing attestation of the eventlog, it would be useful to verify that the keys are genuine. In the current state, the keys are part of a blob. This PR allows the keys (and other fields) to be parsed out separately, resulting in something like the following:

```
  - EventNum: 4
    PCRIndex: 7
    EventType: EV_EFI_VARIABLE_DRIVER_CONFIG
    DigestCount: 2
    Digests:
      - AlgorithmId: sha1
        Digest: "74695203091adbb40c8420f1b499a6ac1a723962"
      - AlgorithmId: sha256
        Digest: "b161e0347f5f040997f97ff52642d43b3a87b986dae8d776d6af27e6468675e6"
    EventSize: 1011
    Event:
      VariableName: 61dfe48b-ca93-d211-aa0d-00e098032b8c
      UnicodeNameLength: 2
      VariableDataLength: 975
      UnicodeName: PK
      VariableData:
      - SignatureType: a159c0a5-e494-a74a-87b5-ab155c2bf072
        SignatureListSize: 975
        SignatureHeaderSize: 0
        SignatureSize: 947
        Keys:
        - SignatureOwner: 5148dc26-5f19-e14a-9a19-fbf883bbb35e
[...]

PR #2291, opened 2021-01-07
tpm2_eventlog: parse EV_EFI_GPT_EVENT

huang195:
This PR parses EV_EFI_GPT_EVENT. In `test/integration/fixtures/event-bootorder.bin`, an example event is:

```
  - EventNum: 24                                                                         
    PCRIndex: 5                                                                          
    EventType: EV_EFI_GPT_EVENT                                                          
    DigestCount: 2                                                                       
    Digests:                                                                             
      - AlgorithmId: sha1                                                                
        Digest: "d35cb7d68eaa9de91261f18f3077b7ba9dd32974"                               
      - AlgorithmId: sha256                                                              
        Digest: "2b406513198abc7ffebdec8a744c1a8b828fe07489e000dbde27bf237aca3de2"       
    EventSize: 484                                                                       
    Event: "4546492050415254000001005c000000d8b2b15d000000000100000000000000ff5f6604000000002200000000000000de5f660400000000a21b0807c97c63468ecf4692ccb7e3ec02000000000000008000000080000000160e4f720300000000000000af3dc60f838472478e793d69d8477de47010a9784677964382b8ffc7ba90d8890078030000000000de5f66040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004861682149646f6e744e656564454649daf9d10afb18ac429f39fbba46240cbd0008000000000000ff27000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000028732ac11ff8d211ba4b00a0c93ec93b98e759c36a59b642b6f6eca290429d990028000000000000ff770300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000" 
```

This PR parses the blob at the bottom to human-readable format as follows:

```
  - EventNum: 24
    PCRIndex: 5
    EventType: EV_EFI_GPT_EVENT
    DigestCount: 2
    Digests:
      - AlgorithmId: sha1
        Digest: "d35cb7d68eaa9de91261f18f3077b7ba9dd32974"
      - AlgorithmId: sha256
        Digest: "2b406513198abc7ffebdec8a744c1a8b828fe07489e000dbde27bf237aca3de2"
    EventSize: 484
    Event:
      Header:
        Signature: "EFI PART"
        Revision: 0x10000
        HeaderSize: 92
[...]

PR #2308, opened 2021-01-26
tpm2_eventlog: parse secureboot variable

huang195:
`SecureBoot` is a variable that appears in every eventlog. This PR makes it more "obvious" what various numeric values mean. Instead of 

```
      UnicodeName: SecureBoot
      VariableData: "01"
```

it spells it out as

```
      UnicodeName: SecureBoot
      VariableData:
        Enabled:'Yes'
```

### DIFF
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -187,7 +187,7 @@ static bool yaml_uefi_var_data(UEFI_VARIABLE_DATA *data) {
  * event field string of "ACPI DATA" in all caps.
  *
  * Section 9.2.5 also says "...Below is the definition of the
- * UEFI_PLATFORM_FIRMWARE_BLOB structure that the CRTM MUST put 
+ * UEFI_PLATFORM_FIRMWARE_BLOB structure that the CRTM MUST put
  * into the Event Log entry TCG_PCR_EVENT2.event[1] field for
  * event types EV_POST_CODE, EV_S_CRTM_CONTENTS, and
  * EV_EFI_PLATFORM_FIRMWARE_BLOB."
@@ -218,7 +218,8 @@ static bool yaml_uefi_post_code(const TCG_EVENT2* const event) {
  * The tpm2_eventlog module validates the event structure but nothing within
  * the event data buffer so we must do that here.
  */
-static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
+static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type,
+                          int32_t yaml_version) {
 
     char uuidstr[37] = { 0 };
     size_t start = 0;
@@ -250,8 +251,8 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
     tpm2_tool_output("      UnicodeName: %s\n", ret);
 
     start += data->UnicodeNameLength*2;
-    /* Try to parse as much as we can without fail-stop. Bugs in firmware, shim, 
-     * grub could produce inconsistent metadata. As long as it is not preventing 
+    /* Try to parse as much as we can without fail-stop. Bugs in firmware, shim,
+     * grub could produce inconsistent metadata. As long as it is not preventing
      * us from parsing the data, we try to continue while giving a warning
      * message.
      */
@@ -259,92 +260,118 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
         LOG_ERR("EventSize is inconsistent with actual data\n");
     }
 
-    /* PK, KEK, db, and dbx are reserved variables names used to store platform
-     * keys, key exchange keys, database keys, and blacklisted database keys,
-     * respectively. 
-     */
-    if (type == EV_EFI_VARIABLE_DRIVER_CONFIG && 
-        ((strlen(ret) == 2 && strncmp(ret, "PK", 2) == 0) ||
-        (strlen(ret) == 3 && strncmp(ret, "KEK", 3) == 0) ||
-        (strlen(ret) == 2 && strncmp(ret, "db", 2) == 0) ||
-        (strlen(ret) == 3 && strncmp(ret, "dbx", 3) == 0))) {
+    if (yaml_version == 2) {
+        /* PK, KEK, db, and dbx are reserved variables names used to store platform
+         * keys, key exchange keys, database keys, and blacklisted database keys,
+         * respectively.
+         */
+        if (type == EV_EFI_VARIABLE_DRIVER_CONFIG) {
+            if ((strlen(ret) == 2 && strncmp(ret, "PK", 2) == 0) ||
+                (strlen(ret) == 3 && strncmp(ret, "KEK", 3) == 0) ||
+                (strlen(ret) == 2 && strncmp(ret, "db", 2) == 0) ||
+                (strlen(ret) == 3 && strncmp(ret, "dbx", 3) == 0)) {
 
-        free(ret);
-        tpm2_tool_output("      VariableData:\n");
-        uint8_t *variable_data = (uint8_t *)&data->UnicodeName[
-            data->UnicodeNameLength];
-        /* iterate through each EFI_SIGNATURE_LIST */
-        while (start < size) {
-            EFI_SIGNATURE_LIST *slist = (EFI_SIGNATURE_LIST *)variable_data;
-            if (start + sizeof(*slist) > size) {
-                LOG_ERR("EventSize is inconsistent with actual data\n");
-                break;
+                free(ret);
+                tpm2_tool_output("      VariableData:\n");
+                uint8_t *variable_data = (uint8_t *)&data->UnicodeName[
+                    data->UnicodeNameLength];
+                /* iterate through each EFI_SIGNATURE_LIST */
+                while (start < size) {
+                    EFI_SIGNATURE_LIST *slist = (EFI_SIGNATURE_LIST *)variable_data;
+                    if (start + sizeof(*slist) > size) {
+                        LOG_ERR("EventSize is inconsistent with actual data\n");
+                        break;
+                    }
+
+                    if (slist->SignatureSize < 16) {
+                        LOG_ERR("SignatureSize is too small\n");
+                        break;
+                    }
+
+                    uuid_unparse_lower(slist->SignatureType, uuidstr);
+                    tpm2_tool_output("      - SignatureType: %s\n"
+                                     "        SignatureListSize: %" PRIu32 "\n"
+                                     "        SignatureHeaderSize: %" PRIu32 "\n"
+                                     "        SignatureSize: %" PRIu32 "\n"
+                                     "        Keys:\n",
+                                     uuidstr, slist->SignatureListSize,
+                                     slist->SignatureHeaderSize,
+                                     slist->SignatureSize);
+
+                    start += (sizeof(*slist) + slist->SignatureHeaderSize);
+                    if (start + slist->SignatureSize > size) {
+                        LOG_ERR("EventSize is inconsistent with actual data\n");
+                        break;
+                    }
+
+                    int signature_size = slist->SignatureListSize -
+                        sizeof(*slist) - slist->SignatureHeaderSize;
+                    if (signature_size < 0 || signature_size % slist->SignatureSize != 0) {
+                        LOG_ERR("Malformed EFI_SIGNATURE_LIST\n");
+                        break;
+                    }
+
+                    uint8_t *signature = (uint8_t *)slist +
+                        sizeof(*slist) + slist->SignatureHeaderSize;
+                    int signatures = signature_size / slist->SignatureSize;
+                    /* iterate through each EFI_SIGNATURE on the list */
+                    int i;
+                    for (i = 0; i < signatures; i++) {
+                        EFI_SIGNATURE_DATA *s = (EFI_SIGNATURE_DATA *)signature;
+                        char *sdata = calloc (1,
+                            BYTES_TO_HEX_STRING_SIZE(slist->SignatureSize-16));
+                        if (sdata == NULL) {
+                            LOG_ERR("Failled to allocate data: %s\n", strerror(errno));
+                            return false;
+                        }
+                        bytes_to_str(s->SignatureData, slist->SignatureSize-16,
+                            sdata, BYTES_TO_HEX_STRING_SIZE(slist->SignatureSize-16));
+                        uuid_unparse_lower(s->SignatureOwner, uuidstr);
+                        tpm2_tool_output("        - SignatureOwner: %s\n"
+                                         "          SignatureData: %s\n",
+                                         uuidstr, sdata);
+                        free(sdata);
+
+                        signature += slist->SignatureSize;
+                        start += slist->SignatureSize;
+                        if (start > size) {
+                            LOG_ERR("Malformed EFI_SIGNATURE_DATA\n");
+                            break;
+                        }
+                    }
+                    variable_data += slist->SignatureListSize;
+                }
+                return true;
             }
-
-            if (slist->SignatureSize < 16) {
-                LOG_ERR("SignatureSize is too small\n");
-                break;
-            }
-
-            uuid_unparse_lower(slist->SignatureType, uuidstr);
-            tpm2_tool_output("      - SignatureType: %s\n"
-                             "        SignatureListSize: %" PRIu32 "\n"
-                             "        SignatureHeaderSize: %" PRIu32 "\n"
-                             "        SignatureSize: %" PRIu32 "\n"
-                             "        Keys:\n",
-                             uuidstr, slist->SignatureListSize,
-                             slist->SignatureHeaderSize,
-                             slist->SignatureSize);
-
-            start += (sizeof(*slist) + slist->SignatureHeaderSize);
-            if (start + slist->SignatureSize > size) {
-                LOG_ERR("EventSize is inconsistent with actual data\n");
-                break;
-            }
-
-            int signature_size = slist->SignatureListSize -
-                sizeof(*slist) - slist->SignatureHeaderSize;
-            if (signature_size < 0 || signature_size % slist->SignatureSize != 0) {
-                LOG_ERR("Malformed EFI_SIGNATURE_LIST\n");
-                break;
-            }
-
-            uint8_t *signature = (uint8_t *)slist +
-                sizeof(*slist) + slist->SignatureHeaderSize;
-            int signatures = signature_size / slist->SignatureSize;
-            /* iterate through each EFI_SIGNATURE on the list */
-            int i;
-            for (i = 0; i < signatures; i++) {
-                EFI_SIGNATURE_DATA *s = (EFI_SIGNATURE_DATA *)signature;
-                char *sdata = calloc (1,
-                    BYTES_TO_HEX_STRING_SIZE(slist->SignatureSize-16));
-                if (sdata == NULL) {
-                    LOG_ERR("Failled to allocate data: %s\n", strerror(errno));
+            else if ((strlen(ret) == 10 && strncmp(ret, "SecureBoot", 10) == 0)) {
+                free(ret);
+                tpm2_tool_output("      VariableData:\n"
+                                 "        Enabled:");
+                if (data->VariableDataLength == 0) {
+                    tpm2_tool_output("'No'\n");
+                }
+                else if (data->VariableDataLength > 1) {
+                    LOG_ERR("SecureBoot data length %" PRIu64 " > 1\n",
+                            data->VariableDataLength);
                     return false;
                 }
-                bytes_to_str(s->SignatureData, slist->SignatureSize-16, 
-                    sdata, BYTES_TO_HEX_STRING_SIZE(slist->SignatureSize-16));
-                uuid_unparse_lower(s->SignatureOwner, uuidstr);
-                tpm2_tool_output("        - SignatureOwner: %s\n"
-                                 "          SignatureData: %s\n",
-                                 uuidstr, sdata);
-                free(sdata);
-
-                signature += slist->SignatureSize;
-                start += slist->SignatureSize;
-                if (start > size) {
-                    LOG_ERR("Malformed EFI_SIGNATURE_DATA\n");
-                    break;
+                else {
+                    uint8_t *variable_data = (uint8_t *)&data->UnicodeName[
+                        data->UnicodeNameLength];
+                    if (*variable_data == 0) {
+                        tpm2_tool_output("'No'\n");
+                    }
+                    else {
+                        tpm2_tool_output("'Yes'\n");
+                    }
                 }
+                return true;
             }
-            variable_data += slist->SignatureListSize;
         }
-        return true;
     }
-    else {
-        free(ret);
-        return yaml_uefi_var_data(data);
-    }
+
+    free(ret);
+    return yaml_uefi_var_data(data);
 }
 /* TCG PC Client FPF section 9.2.5 */
 bool yaml_uefi_platfwblob(UEFI_PLATFORM_FIRMWARE_BLOB *data) {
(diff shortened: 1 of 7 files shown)


williamcroberts:
This would break existing consumers of the YAML.

huang195:
For any field that is currently in a raw format, i.e., shown as a hex string, if it is represented internally by a data structure, it will not be very consumable either by a script or a person. I'm not sure if backward compatibility is a big issue for this particular case. However, for any existing parsed fields, I do agree that schema changes should be avoided.

Here are a few examples to make this a bit more concrete:

Example 1:
```
Data: "0102030405060708"
```

We could interpret this as
```
Data:
  name: "x yz"
  address: "234 test rd"
```

This would be a breaking change, but I would argue that without reading the spec, it would be almost impossible for someone to interpret the raw data. Even if someone read the spec and wrote a script to parse the raw data, having this change in the `master` branch shouldn't have impacted him as he shouldn't peg against the `master` branch in the first place, but rather one of the release branches, which are more carefully groomed. When this breaking change eventually gets rolled into the next major release, rebasing to fix the breaking change should be fairly straightforward, and removing complicated logic in his custom script to interpret raw byte strings is most likely a good thing.

Example 2:
```
Data:
  name: "x yz"
  address: "234 test rd"
```

Now, if we want to change an existing schema to
```
Data:
  name:
    first: "x"
    last: "yz"
  address: "234 test rd"
```

This should be avoided, as this doesn't really "parse" data but just putting existing parsed data into a different format (i.e., schema).

@williamcroberts @idesai I'd like to parse some additional event types that are currently just raw bytes, so if you could clarify what a breaking change and a non-breaking change is, that would help me to better understand how I should parse data.

huang195:
These recently merged PRs are all causing breaking changes:
- https://github.com/tpm2-software/tpm2-tools/pull/2291
- https://github.com/tpm2-software/tpm2-tools/pull/2300

My understanding is that since they are parsing _raw bytes_, the probability of breaking backward compatibility is very low. This PR is of the same nature. Even though the `VariableData` field is only 1 byte long, it is still a raw byte string. In the most common cases, it is of 1 byte in size and of value of either 0 or 1. However, it could be 0 sized and have other values, in which case, the meaning of the field would be confusing. This PR tries to clarify all the different corner cases.

williamcroberts:
> For any field that is currently in a raw format, i.e., shown as a hex string, if it is represented internally by a data structure, it will not be very consumable either by a script or a person. I'm not sure if backward compatibility is a big issue for this particular case. However, for any existing parsed fields, I do agree that schema changes should be avoided.
> 
> Here are a few examples to make this a bit more concrete:
> 
> Example 1:
> 
> ```
> Data: "0102030405060708"
> ```
> 
> We could interpret this as
> 
> ```
> Data:
>   name: "x yz"
>   address: "234 test rd"
> ```
> 
> This would be a breaking change, but I would argue that without reading the spec, it would be almost impossible for someone to interpret the raw data. Even if someone read the spec and wrote a script to parse the raw data, having this change in the `master` branch shouldn't have impacted him as he shouldn't peg against the `master` branch in the first place, but rather one of the release branches, which are more carefully groomed. When this breaking change eventually gets rolled into the next major release, rebasing to fix the breaking change should be fairly straightforward, and removing complicated logic in his custom script to interpret raw byte strings is most likely a good thing.
> 
> Example 2:
> 
> ```
> Data:
>   name: "x yz"
>   address: "234 test rd"
> ```
> 
> Now, if we want to change an existing schema to
> 
> ```
> Data:
>   name:
>     first: "x"
>     last: "yz"
>   address: "234 test rd"
> ```
> 
> This should be avoided, as this doesn't really "parse" data but just putting existing parsed data into a different format (i.e., schema).
> 
> @williamcroberts @idesai I'd like to parse some additional event types that are currently just raw bytes, so if you could clarify what a breaking change and a non-breaking change is, that would help me to better understand how I should parse data.

If someone has an existing parser, they would need to update the parser based on tpm2-tools version. If it's older, they do it one way, if it's another, they do it another way, that's breaking. We don't want to break tool output, it needs to be stable for downstream consumers. We have been through this before, and downstream folks, while patient, were not appreciative.

I would propose adding some option, perhaps something like `--output-version=2` for specifying the YAML version that's output, that would:
1. embed the version into the YAML file.
2. Decode fields further.
3. Get as many changes you want to see merged before a release, becuase once it's released, then we would have to increment the YAML version number again.

Another approach would be to add, leaving the old data intact. As long as we only add, and do not modify or remove, it's always backwards compat.

williamcroberts:
> These recently merged PRs are all causing breaking changes:
> 
> * #2291
> * #2300
> 
> My understanding is that since they are parsing _raw bytes_, the probability of breaking backward compatibility is very low. This PR is of the same nature. Even though the `VariableData` field is only 1 byte long, it is still a raw byte string. In the most common cases, it is of 1 byte in size and of value of either 0 or 1. However, it could be 0 sized and have other values, in which case, the meaning of the field would be confusing. This PR tries to clarify all the different corner cases.

Okay, we will need to get those two changes onto a versioned YAML format.

huang195:
> If someone has an existing parser, they would need to update the parser based on tpm2-tools version. If it's older, they do it one way, if it's another, they do it another way, that's breaking. 

@williamcroberts I do maintain such a parser, but it's based on a tpm2-tools release branch, so all the changes on the `master` branch is not impacting us. Once tpm2-tool makes a new release with features that we need, we do a rebase of our parser on the new release branch. Since no breaking changes are done on release branches (as far as we can see), this has worked pretty well for us. We didn't need to rely on tpm-tools versions, but maybe other projects use a different dev process that could run into this issue. 

> Another approach would be to add, leaving the old data intact. As long as we only add, and do not modify or remove, it's always backwards compat.

I think this approach would work. However, after a while, we would have quite a bit of redundancy, i.e., both raw and parsed data. Is that desirable? As some events are not fully parsed, parsing deeper could require multiple iterations. This would leave raw/parsed redundancies in multiple levels. This might not matter to a script but when viewed by a person, it could be kind of messy.

williamcroberts:
> > If someone has an existing parser, they would need to update the parser based on tpm2-tools version. If it's older, they do it one way, if it's another, they do it another way, that's breaking.
> 
> @williamcroberts I do maintain such a parser, but it's based on a tpm2-tools release branch, so all the changes on the `master` branch is not impacting us. Once tpm2-tool makes a new release with features that we need, we do a rebase of our parser on the new release branch. Since no breaking changes are done on release branches (as far as we can see), this has worked pretty well for us. We didn't need to rely on tpm-tools versions, but maybe other projects use a different dev process that could run into this issue.
> 
> > Another approach would be to add, leaving the old data intact. As long as we only add, and do not modify or remove, it's always backwards compat.
> 
> I think this approach would work. However, after a while, we would have quite a bit of redundancy, i.e., both raw and parsed data. Is that desirable? As some events are not fully parsed, parsing deeper could require multiple iterations. This would leave raw/parsed redundancies in multiple levels. This might not matter to a script but when viewed by a person, it could be kind of messy.

It's not ideal, I think multiple versions might be better

huang195:
> It's not ideal, I think multiple versions might be better

@williamcroberts Can you give us some examples of how multiple versions would be implemented? For our downstream parser, it would help us anticipate what kind of effort we need to do to rebase our code when we pull in this change in the next release.

williamcroberts:
> > It's not ideal, I think multiple versions might be better
> 
> @williamcroberts Can you give us some examples of how multiple versions would be implemented? For our downstream parser, it would help us anticipate what kind of effort we need to do to rebase our code when we pull in this change in the next release.

```bash
tpm2_eventlog --yaml=version=2
```
Which would output the new yaml format with the version field and the new format below it.
```yaml
version: 2
<new format here>
```

A parser can handle this like so:
```python
# psuedo python code
y = yaml.safe_load(input)
version = y['version'] if 'version' in y else 1
# Either ensure you have the version you expect or do things conditionally on version
assert(version == expected_version)
if (version == 2):
   x = parse_version_2()
else
  x = parse_version_1()

do_things_with(x)
sys.exit(0)
```

huang195:
@williamcroberts yeah, that makes sense. By default, it's yaml version 1. Any breaking changes will go into yaml version 2. Before the next major release, breaking changes will all go into yaml version 2 (even if it's not backward compatible to yaml version 2). When the next major release happens, breaking changes to both yaml versions 1 and 2 will go into yaml version 3. 

The only thing I'm not sure about is what I wrote in parenthesis above:
> (even if it's not backward compatible to yaml version 2)

In this case, is it better to go straight to yaml version 3 when version 2's backward compatibility is broken? The problem with this is if eventlogs are incrementally parsed, there could be quite a bit of breaking changes, resulting in fast yaml version increment. 

It seems to me that the former makes more sense as `master` branch is considered development and things could change underneath until a cut is made for a release. Thoughts?

huang195:
@williamcroberts so I took the liberty to implement what you suggested above. By default,

```
tpm2_eventlog
```

will produce a yaml file with

```
version: 1
```

One can pass a `--yaml-version=2` parameter to change the yaml version. Currently only versions `1` and `2` are valid. Any other versions will throw an error.

The 2 previously merged PRs and this PR are now only activated when `--yaml-version=2` is explicitly passed as a parameter. Otherwise, they stay unparsed, so to not break backward compatibility. Let me know what you think.

huang195:
Any help on the failed test? I don't see how it's impacted by the PR

idesai:
> Any help on the failed test? I don't see how it's impacted by the PR

We have a couple of FAPI tests failing it is affecting all PRs at the moment.
@AndreasFuchsSIT, any thoughts?

cplappert:
> > Any help on the failed test? I don't see how it's impacted by the PR
> 
> We have a couple of FAPI tests failing it is affecting all PRs at the moment.
> @AndreasFuchsSIT, any thoughts?

This issue is fixed in PR #2320.

huang195:
@idesai @williamcroberts just want to check if you have any feedbacks on this PR?

lib/tpm2_eventlog_yaml.h
```diff
@@ -12,12 +12,13 @@ char const *eventtype_to_string (UINT32 event_type);
...
 bool yaml_digest2(TCG_DIGEST2 const *digest, size_t size);
 char *yaml_uefi_var_unicodename(UEFI_VARIABLE_DATA *data);
-bool yaml_event2data(TCG_EVENT2 const *event, UINT32 type);
+bool yaml_event2data(TCG_EVENT2 const *event, UINT32 type, INT32 yaml_version);
 bool yaml_digest2_callback(TCG_DIGEST2 const *digest, size_t size, void *data);
 bool yaml_event2hdr_callback(TCG_EVENT_HEADER2 const *event_hdr, size_t size,
                              void *data);
-bool yaml_event2data_callback(TCG_EVENT2 const *event, UINT32 type, void *data);
+bool yaml_event2data_callback(TCG_EVENT2 const *event, UINT32 type, void *data,
+                              INT32 yaml_version);
```
idesai, line 20: @huang195 can you use a uint32_t instead. The arithmetic is well defined and prevents overflows.

lib/tpm2_eventlog_yaml.c
```diff
@@ -250,101 +251,127 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
...
-                sizeof(*slist) - slist->SignatureHeaderSize;
-            if (signature_size < 0 || signature_size % slist->SignatureSize != 0) {
-                LOG_ERR("Malformed EFI_SIGNATURE_LIST\n");
-                break;
+    if (yaml_version == 2) {
+        /* PK, KEK, db, and dbx are reserved variables names used to store platform
+         * keys, key exchange keys, database keys, and blacklisted database keys,
+         * respectively.
+         */
+        if (type == EV_EFI_VARIABLE_DRIVER_CONFIG) {
```
idesai, line 268: To simplify the "if-else", is this a better equivalent if(yaml_version == 2 && type == EV_EFI_VARIABLE_DRIVER_CONFIG)

lib/tpm2_eventlog_yaml.c
```diff
@@ -250,101 +251,127 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
...
-                LOG_ERR("EventSize is inconsistent with actual data\n");
-                break;
-            }
-
-            int signature_size = slist->SignatureListSize -
-                sizeof(*slist) - slist->SignatureHeaderSize;
-            if (signature_size < 0 || signature_size % slist->SignatureSize != 0) {
-                LOG_ERR("Malformed EFI_SIGNATURE_LIST\n");
-                break;
+    if (yaml_version == 2) {
```
idesai, line 263: What happens when yaml_version == 1?

lib/tpm2_eventlog_yaml.c
```diff
@@ -250,101 +251,127 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
...
+         */
+        if (type == EV_EFI_VARIABLE_DRIVER_CONFIG) {
+            if ((strlen(ret) == 2 && strncmp(ret, "PK", 2) == 0) ||
+                (strlen(ret) == 3 && strncmp(ret, "KEK", 3) == 0) ||
+                (strlen(ret) == 2 && strncmp(ret, "db", 2) == 0) ||
+                (strlen(ret) == 3 && strncmp(ret, "dbx", 3) == 0)) {
+
+                free(ret);
+                tpm2_tool_output("      VariableData:\n");
+                uint8_t *variable_data = (uint8_t *)&data->UnicodeName[
```
idesai, line 276: Is the typecast required?

lib/tpm2_eventlog_yaml.c
```diff
@@ -250,101 +251,127 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
...
-            int signatures = signature_size / slist->SignatureSize;
-            /* iterate through each EFI_SIGNATURE on the list */
-            int i;
-            for (i = 0; i < signatures; i++) {
-                EFI_SIGNATURE_DATA *s = (EFI_SIGNATURE_DATA *)signature;
-                char *sdata = calloc (1,
-                    BYTES_TO_HEX_STRING_SIZE(slist->SignatureSize-16));
-                if (sdata == NULL) {
-                    LOG_ERR("Failled to allocate data: %s\n", strerror(errno));
+            else if ((strlen(ret) == 10 && strncmp(ret, "SecureBoot", 10) == 0)) {
```
idesai, line 346: this and everywhere, start the else on same line as the previous if closing.

lib/tpm2_eventlog_yaml.c
```diff
@@ -250,101 +251,127 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type) {
...
-                                 uuidstr, sdata);
-                free(sdata);
-
-                signature += slist->SignatureSize;
-                start += slist->SignatureSize;
-                if (start > size) {
-                    LOG_ERR("Malformed EFI_SIGNATURE_DATA\n");
-                    break;
+                else {
+                    uint8_t *variable_data = (uint8_t *)&data->UnicodeName[
```
idesai, line 359: Typecast, required?

lib/tpm2_eventlog_yaml.c
```diff
@@ -413,85 +440,93 @@ bool yaml_uefi_image_load(UEFI_IMAGE_LOAD_EVENT *data, size_t size) {
     free(buf);
     return true;
 }
+#define EVENT_BUF_MAX BYTES_TO_HEX_STRING_SIZE(1024)
 /* TCG PC Client PFP section 9.2.6 */
-bool yaml_gpt(UEFI_GPT_DATA *data, size_t size) {
+bool yaml_gpt(UEFI_GPT_DATA *data, size_t size, int32_t yaml_version) {
```
idesai, line 445: uint32_t?

lib/tpm2_eventlog_yaml.c
```diff
@@ -413,85 +440,93 @@ bool yaml_uefi_image_load(UEFI_IMAGE_LOAD_EVENT *data, size_t size) {
     free(buf);
     return true;
 }
+#define EVENT_BUF_MAX BYTES_TO_HEX_STRING_SIZE(1024)
```
idesai, line 443: Is this used?

lib/tpm2_eventlog_yaml.c
```diff
@@ -413,85 +440,93 @@ bool yaml_uefi_image_load(UEFI_IMAGE_LOAD_EVENT *data, size_t size) {
...
+        if (size != 0) {
+            LOG_ERR("EventSize is inconsistent with actual data\n");
+            return false;
+        }
     }
-
-    if (size != 0) {
-        LOG_ERR("EventSize is inconsistent with actual data\n");
-        return false;
+    else {
```
idesai, line 521: else on same line as closing of related if

lib/tpm2_eventlog_yaml.c
```diff
@@ -413,85 +440,93 @@ bool yaml_uefi_image_load(UEFI_IMAGE_LOAD_EVENT *data, size_t size) {
...
 
-    if (size < sizeof(*data)) {
-        LOG_ERR("EventSize(%zu) is too small\n", size);
-        return false;
-    }
+    if (yaml_version == 2) {
+        if (size < sizeof(*data)) {
+            LOG_ERR("EventSize(%zu) is too small\n", size);
+            return false;
+        }
```
idesai, line 451: This condition applies to both branches, shouldn't this be the first thing you check and return false?

tools/misc/tpm2_eventlog.c
```diff
@@ -27,10 +31,33 @@ static bool on_positional(int argc, char **argv) {
...
+    }
+    return true;
+}
+
 static bool tpm2_tool_onstart(tpm2_options **opts) {
 
-    *opts = tpm2_options_new(NULL, 0, NULL, NULL, on_positional,
-                             TPM2_OPTIONS_NO_SAPI);
+    static struct option topts[] = {
+         { "yaml-version",         required_argument, NULL, 'y' },
```
idesai, line 56: @huang195 @williamcroberts not trying to be picky, but the option yaml-version is a bit of detour from what we used in specifying the versioning for context and session, there we referenced them as session_version and context_version.

idesai: Also, maybe don't use a short option at all.

idesai: > @huang195 @williamcroberts not trying to be picky, but the option yaml-version is a bit of detour from what we used in specifying the versioning for context and session, there we referenced them as session_version and context_version.

Also because, yaml outputs exist for other tools as well and things can get confusing.

huang195:
@idesai Thanks for the feedbacks. Will take a look later today

lib/tpm2_eventlog_yaml.h
huang195: yep

lib/tpm2_eventlog_yaml.c
huang195: As we parse additional event types within the `if (yaml_version ==2)` clause, we will have to add more `if (type == xxx)` conditions. As I see more event types that could be parsed but break backward compatibility, it might be better we leave the current structure (even though it is a cleaner if-else statement).

lib/tpm2_eventlog_yaml.c
huang195: The `yaml_version==1` (default condition) is handled by `return yaml_uefi_var_data(data)` at the very bottom of the function, which basically just print the raw data in a hex string.

lib/tpm2_eventlog_yaml.c
huang195: The field in the struct is defined as 
```  
char16_t UnicodeName[];
```
I believe it gave a compiler warning without the explicit typecast.

lib/tpm2_eventlog_yaml.c
huang195: yep, will fix

lib/tpm2_eventlog_yaml.c
huang195: Same as above

lib/tpm2_eventlog_yaml.c
huang195: yep, will fix

lib/tpm2_eventlog_yaml.c
huang195: This macro was previously defined further down, and I just moved it up because the function `yaml_gpt()` also uses this macro.

lib/tpm2_eventlog_yaml.c
huang195: yep, will fix

lib/tpm2_eventlog_yaml.c
huang195: Yes, I was worried about backward compatibility again, as `version 1` didn't do this error check previously. I wonder if the extra error message could break anything. However, as this goes to stderr if the the condition is triggered, it shouldn't be a problem. Will move this outside of the `if (yaml_version == 2)` clause.

tools/misc/tpm2_eventlog.c
huang195: Sure, will rename `yaml-version` to `yaml_version`, and get rid of the short option.

huang195:
@idesai the issues you had raised are addressed either in the comments or in the last code push. Let me know if you have other comments.

tools/misc/tpm2_eventlog.c
idesai: > Sure, will rename `yaml-version` to `yaml_version`, and get rid of the short option.

@huang195 sorry for not being specific, I meant, something like `eventlog-version` as yaml versioning has a connotation of being applied to rest of the tools as well. @huang195 @williamcroberts what do you think?

tools/misc/tpm2_eventlog.c
```diff
@@ -27,10 +31,33 @@ static bool on_positional(int argc, char **argv) {
     return true;
 }
 
+static bool on_option(char key, char *value) {
+
+    uint32_t version;
+
+    switch (key) {
+    case 'y':
+        version = (uint32_t)atoi(value);
```
williamcroberts, line 40: use strtoul, because overflow detection, we have a helper for this, use: tpm2_util_string_to_uint32 from tpm2_util.h

tools/misc/tpm2_eventlog.c
huang195: @idesai ah gotcha. I agree - `eventlog-version` is more concise and would avoid the confusion. I've made the change

williamcroberts (changes requested):

tools/misc/tpm2_eventlog.c
```diff
@@ -27,10 +31,33 @@ static bool on_positional(int argc, char **argv) {
...
+        for (unsigned int i = 0; i < sizeof(allowed_eventlog_versions)/sizeof(int); i++) {
+            if (version == allowed_eventlog_versions[i]) {
+                eventlog_version = version;
+                return true;
+            }
+        }
```
williamcroberts, line 46: ETOOCOMPLICATED, just set two defines in tpm2_eventlog_yaml.h:
```C
#define MIN_EVLOG_YAML_VERSION 1
#define MAX_EVLOG_YAML_VERSION 2
```
Then here do:
```C
if (version < MIN_EVLOG_YAML_VERSION || version > MAX_EVLOG_YAML_VERSION) {
  FAIL
}
```

We also probably want that logic in the functions that take YAML version.

williamcroberts (changes requested):

tools/misc/tpm2_eventlog.c
huang195: ok, will do

tools/misc/tpm2_eventlog.c
huang195: ah good to know, will use the helper function instead of `atoi`

williamcroberts (approved):
LGTM, ill let @idesai make the merge

huang195:
@williamcroberts thanks for the help. Good catch on the backward compatibility issues. Now we have a mechanism in place to be able to parse new event types without impacting backward compatibility.

huang195:
@idesai thanks for the code review and your recommendations. Much appreciate it!

PR #2336, opened 2021-02-22
tpm2_eventlog: Standardize yaml spacing (cosmetic)

huang195:
In the current tpm2_eventlog YAML output, there are 2 ways items are indented, e.g.,

```
    SpecID:
      - Signature: Spec ID Event03
        platformClass: 0
        specVersionMinor: 0
        specVersionMajor: 2
        specErrata: 0
        uintnSize: 2
        numberOfAlgorithms: 2
        Algorithms:
          - Algorithm[0]:
            algorithmId: sha1
            digestSize: 20
          - Algorithm[1]:
            algorithmId: sha256
            digestSize: 32
        vendorInfoSize: 0
```

OTOH, it can also be indented like the following:

```
    Event:
      UnicodeNameLength: 2
      VariableDataLength: 975
      UnicodeName: PK
      VariableData:
      - SignatureType: a159c0a5-e494-a74a-87b5-ab155c2bf072
        SignatureListSize: 975
        SignatureHeaderSize: 0
        SignatureSize: 947
```

According to the YAML spec, they are both correct, but it would be good to stick with a single style. This PR standardizes indentations to the latter style. It's more compact and easier to read especially for items with more indentation layers. I used http://www.yamllint.com/ to that both styles of output generate semantically identical output.

### DIFF
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -371,7 +371,7 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type,
             EFI_SIGNATURE_DATA *s= (EFI_SIGNATURE_DATA *)&data->UnicodeName[
                 data->UnicodeNameLength];
             char *sdata = calloc (1,
-                BYTES_TO_HEX_STRING_SIZE(data->VariableDataLength-16));
+                BYTES_TO_HEX_STRING_SIZE(data->VariableDataLength - 16));
             if (sdata == NULL) {
                 LOG_ERR("Failled to allocate data: %s\n", strerror(errno));
                 return false;
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -93,9 +93,9 @@ void yaml_event2hdr(TCG_EVENT_HEADER2 const *eventhdr, size_t size) {
 
     (void)size;
 
-    tpm2_tool_output("    PCRIndex: %d\n"
-                     "    EventType: %s\n"
-                     "    DigestCount: %d\n",
+    tpm2_tool_output("  PCRIndex: %d\n"
+                     "  EventType: %s\n"
+                     "  DigestCount: %d\n",
                      eventhdr->PCRIndex,
                      eventtype_to_string(eventhdr->EventType),
                      eventhdr->DigestCount);
@@ -106,8 +106,8 @@ void yaml_sha1_log_eventhdr(TCG_EVENT const *eventhdr, size_t size) {
 
     (void)size;
 
-    tpm2_tool_output("    PCRIndex: %d\n"
-                     "    EventType: %s\n",
+    tpm2_tool_output("  PCRIndex: %d\n"
+                     "  EventType: %s\n",
                      eventhdr->pcrIndex,
                      eventtype_to_string(eventhdr->eventType));
 
@@ -121,8 +121,8 @@ bool yaml_digest2(TCG_DIGEST2 const *digest, size_t size) {
     char hexstr[DIGEST_HEX_STRING_MAX] = { 0, };
     bytes_to_str(digest->Digest, size, hexstr, sizeof(hexstr));
 
-    tpm2_tool_output("      - AlgorithmId: %s\n"
-                     "        Digest: \"%s\"\n",
+    tpm2_tool_output("  - AlgorithmId: %s\n"
+                     "    Digest: \"%s\"\n",
                      tpm2_alg_util_algtostr(digest->AlgorithmId, tpm2_alg_util_flags_hash),
                      hexstr);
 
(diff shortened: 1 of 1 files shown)


idesai:
> In the current tpm2_eventlog YAML output, there are 2 ways items are indented, e.g.,
> 
> ```
>     SpecID:
>       - Signature: Spec ID Event03
>         platformClass: 0
>         specVersionMinor: 0
>         specVersionMajor: 2
>         specErrata: 0
>         uintnSize: 2
>         numberOfAlgorithms: 2
>         Algorithms:
>           - Algorithm[0]:
>             algorithmId: sha1
>             digestSize: 20
>           - Algorithm[1]:
>             algorithmId: sha256
>             digestSize: 32
>         vendorInfoSize: 0
> ```
> 
> OTOH, it can also be indented like the following:
> 
> ```
>     Event:
>       VariableName: 61dfe48b-ca93-d211-aa0d-00e098032b8c
>       UnicodeNameLength: 2
>       VariableDataLength: 975
>       UnicodeName: PK
>       VariableData:
>       - SignatureType: a159c0a5-e494-a74a-87b5-ab155c2bf072
>         SignatureListSize: 975
>         SignatureHeaderSize: 0
>         SignatureSize: 947
> ```
> 
> According to the YAML spec, they are both correct, but it would be good to stick with a single style. This PR standardizes indentations to the latter style. It's more compact and easier to read especially for items with more indentation layers. I used http://www.yamllint.com/ to that both styles of output generate semantically identical output.

@huang195 is this consistent with the eventlog_versioning too?

huang195:
> @huang195 is this consistent with the eventlog_versioning too?

yes, both eventlog-versions 1 and 2 are converted to the same indentation style

huang195:
@idesai can you help retriggering the CI on this one too? I did a forced push, but that also failed. Since this PR only changes YAML indentation, it shouldn't fail those specific test scripts.

idesai:
> > @huang195 is this consistent with the eventlog_versioning too?
> 
> yes, both eventlog-versions 1 and 2 are converted to the same indentation style

I thought we were versioning these changes and so shouldn't these only apply to version 2?

huang195:
> I thought we were versioning these changes and so shouldn't these only apply to version 2?
@idesai if I understand it correctly, `--eventlog-version` is used to avoid putting breaking changes into the `master` branch. As this PR only changes the indentation of the output YAML file, it should not cause any backward compatibility problems. What do you think?

idesai:
> > I thought we were versioning these changes and so shouldn't these only apply to version 2?
> 
> @idesai if I understand it correctly, `--eventlog-version` is used to avoid putting breaking changes into the `master` branch. As this PR only changes the indentation of the output YAML file, it should not cause any backward compatibility problems. What do you think?

I was thinking we introduced versioning so as to track any changes to the previously released tpm2-tools versions. Any new changes to the YAML outputs on master will automatically be clubbed into the new version to be packaged for the next tpm2-tools version release. Right? @williamcroberts, your thoughts?

huang195:
@idesai sure, let's also get input from @williamcroberts. The point I was trying to make is that the output of tpm2_eventlog is in YAML format, which means it's expected to be subsequently taken as an input by one of the YAML libraries available in most languages. As we kept the YAML semantic the same, just different indentation styles, this PR should have no impact to this way of using tpm2_eventlog. OTOH, a person could be reading the YAML output file, e.g., to debug something. This would have an impact to them, because it would look slightly different, but I think for the better as it handles items with many layers of indentations more concisely.

williamcroberts:
> > > I thought we were versioning these changes and so shouldn't these only apply to version 2?
> > 
> > 
> > @idesai if I understand it correctly, `--eventlog-version` is used to avoid putting breaking changes into the `master` branch. As this PR only changes the indentation of the output YAML file, it should not cause any backward compatibility problems. What do you think?
> 
> I was thinking we introduced versioning so as to track any changes to the previously released tpm2-tools versions. Any new changes to the YAML outputs on master will automatically be clubbed into the new version to be packaged for the next tpm2-tools version release. Right? @williamcroberts, your thoughts?

The patch only changing indentation, without changing underlying types AFACT, so theirs no versioning issue here. 

Ideally, I'd like to get to a point where the tool is handed a YAML document from libyaml to populate, or NULL if --quiet, and then the tools just call libyaml to add scalar, etc. This way we don't have to worry about all this stuff.

williamcroberts (approved):

idesai:
Cirrus-CI results are so sporadic :angry:

williamcroberts (approved):

williamcroberts:
@idesai merge at your leisure

williamcroberts:
> Ideally, I'd like to get to a point where the tool is handed a YAML document from libyaml to populate, or NULL if --quiet, and then the tools just call libyaml to add scalar, etc. This way we don't have to worry about all this stuff.

FYI I started this here:
  - https://github.com/williamcroberts/tpm2.0-tools/tree/add-libyaml

If anyone wants to grab it and run, then we can get consistency without having to do this.

PR #2339, opened 2021-02-25
tpm2_eventlog: parse BootOrder variable

huang195:
The BootOrder variable is currently shown in raw bytes, e.g.,

```
- EventNum: 12
  PCRIndex: 1
  EventType: EV_EFI_VARIABLE_BOOT
  DigestCount: 2
  Digests:
  - AlgorithmId: sha1
    Digest: "ea3f530d2b261b5a945812c15858c09de04abe9c"
  - AlgorithmId: sha256
    Digest: "263d99957c7b574c63a265b32da7fff8b8ad831828946bfbff650d7074dd9198"
  EventSize: 56
  Event:
    UnicodeNameLength: 9
    VariableDataLength: 6
    UnicodeName: BootOrder
    VariableData: "010000000200"
```

This PR parses it in a more human-readable format, i.e.,

```
    VariableData:
    - Boot0001
    - Boot0000
    - Boot0002
```

showing the boot ordering of all the boot devices configured in the system.

This will only parse when `--eventlog-version` is set to `2`. The file `test/integration/fixtures/event-bootorder.bin` is used for the above example.

### DIFF
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -384,6 +384,24 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type,
                              uuidstr, sdata);
             free(sdata);
             return true;
+        } else if (type == EV_EFI_VARIABLE_BOOT) {
+            if ((strlen(ret) == 9 && strncmp(ret, "BootOrder", 9) == 0)) {
+                free(ret);
+                tpm2_tool_output("    VariableData:\n");
+                
+                if (data->VariableDataLength % 2 != 0) {
+                    LOG_ERR("BootOrder value length %" PRIu64 " is not divisible by 2\n",
+                            data->VariableDataLength);
+                    return false;
+                }
+
+                uint8_t *variable_data = (uint8_t *)&data->UnicodeName[
+                    data->UnicodeNameLength];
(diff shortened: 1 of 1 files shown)


lib/tpm2_eventlog_yaml.c
```diff
@@ -384,6 +384,24 @@ static bool yaml_uefi_var(UEFI_VARIABLE_DATA *data, size_t size, UINT32 type,
...
+                free(ret);
+                tpm2_tool_output("    VariableData:\n");
+                
+                if (data->VariableDataLength % 2 != 0) {
+                    LOG_ERR("BootOrder value length %" PRIu64 " is not divisible by 2\n",
+                            data->VariableDataLength);
+                    return false;
+                }
+
+                uint8_t *variable_data = (uint8_t *)&data->UnicodeName[
```
idesai, line 398:
```suggestion
                uint8_t *variable_data = &data->UnicodeName[
```
is the cast required?

huang195: Right, the UnicodeName field is defined as `char16_t UnicodeName[];` Compiler didn't like it without an explicit cast, and would give the following error:
```
lib/tpm2_eventlog_yaml.c:398:42: error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
                 uint8_t *variable_data = &data->UnicodeName[data->UnicodeNameLength];
```

PR #2386, opened 2021-03-28
tpm2_duplicate: support duplicating external non-TPM keys

osresearch:
Wrapping a key for a remote TPM2 right now requires loading the key into the local TPM. For some use cases it is not necessary to involve the local TPM, such as sending an openssl generated key to a remote machine.

`tpm2_duplicate` seems like the right place to put this functionality, although it shares a large amount of code with `tpm2_import`, and some with `tpm2_makecredential`.  If there is a better place to locate the code, please let me know.

Some todo's before it is ready for merging:

* [x] need to add documentation
* [x] write test cases
* [x] add output option for public section
* [x] move `DUPLICATE\0` seed generation into shared routine

### DIFF
--- a/lib/tpm2_identity_util.c
+++ b/lib/tpm2_identity_util.c
@@ -423,3 +423,43 @@ void tpm2_identity_util_calculate_outer_integrity(TPMI_ALG_HASH parent_name_alg,
             encrypted_duplicate_sensitive->size, pubname->name, pubname->size,
             protection_hmac_key->buffer, outer_hmac);
 }
+
+bool tpm2_identity_create_name(TPM2B_PUBLIC *public, TPM2B_NAME *pubname) {
+
+    /*
+     * A TPM2B_NAME is the name of the algorithm, followed by the hash.
+     * Calculate the name by:
+     * 1. Marshaling the name algorithm
+     * 2. Marshaling the TPMT_PUBLIC past the name algorithm from step 1.
+     * 3. Hash the TPMT_PUBLIC portion in marshaled data.
+     */
+
+    TPMI_ALG_HASH name_alg = public->publicArea.nameAlg;
+
+    // Step 1 - set beginning of name to hash alg
+    size_t hash_offset = 0;
+    Tss2_MU_UINT16_Marshal(name_alg, pubname->name, pubname->size,
+            &hash_offset);
+
+    // Step 2 - marshal TPMTP
+    TPMT_PUBLIC marshaled_tpmt;
+    size_t tpmt_marshalled_size = 0;
+    Tss2_MU_TPMT_PUBLIC_Marshal(&public->publicArea,
+            (uint8_t *) &marshaled_tpmt, sizeof(public->publicArea),
+            &tpmt_marshalled_size);
+
+    // Step 3 - Hash the data into name just past the alg type.
+    digester d = tpm2_openssl_halg_to_digester(name_alg);
+    if (!d) {
+        return false;
+    }
+
+    d((const unsigned char *) &marshaled_tpmt, tpmt_marshalled_size,
+            pubname->name + hash_offset);
+
+    //Set the name size, UINT16 followed by HASH
+    UINT16 hash_size = tpm2_alg_util_get_hash_size(name_alg);
+    pubname->size = hash_size + hash_offset;
+
+    return true;
+}
--- a/lib/tpm2_identity_util.c
+++ b/lib/tpm2_identity_util.c
@@ -99,7 +99,7 @@ static TPM2_KEY_BITS get_pub_asym_key_bits(TPM2B_PUBLIC *public) {
 }
 
 static bool share_secret_with_tpm2_rsa_public_key(TPM2B_DIGEST *protection_seed,
-        TPM2B_PUBLIC *parent_pub, unsigned char *label, int label_len,
+        TPM2B_PUBLIC *parent_pub, const unsigned char *label, int label_len,
         TPM2B_ENCRYPTED_SECRET *encrypted_protection_seed) {
     bool rval = false;
     RSA *rsa = NULL;
@@ -218,7 +218,7 @@ bool tpm2_identity_util_calc_outer_integrity_hmac_key_and_dupsensitive_enc_key(
 
 bool tpm2_identity_util_share_secret_with_public_key(
         TPM2B_DIGEST *protection_seed, TPM2B_PUBLIC *parent_pub,
-        unsigned char *label, int label_len,
+        const unsigned char *label, int label_len,
         TPM2B_ENCRYPTED_SECRET *encrypted_protection_seed) {
     bool result = false;
     TPMI_ALG_PUBLIC alg = parent_pub->publicArea.type;
--- a/lib/tpm2_identity_util.c
+++ b/lib/tpm2_identity_util.c
@@ -364,16 +364,29 @@ bool tpm2_identity_util_calculate_inner_integrity(TPMI_ALG_HASH name_alg,
         TPM2B_DATA *enc_sensitive_key, TPMT_SYM_DEF_OBJECT *sym_alg,
         TPM2B_MAX_BUFFER *encrypted_inner_integrity) {
 
+    TSS2_RC rval;
+
     //Marshal sensitive area
     uint8_t buffer_marshalled_sensitiveArea[TPM2_MAX_DIGEST_BUFFER] = { 0 };
     size_t marshalled_sensitive_size = 0;
-    Tss2_MU_TPMT_SENSITIVE_Marshal(&sensitive->sensitiveArea,
+    rval = Tss2_MU_TPMT_SENSITIVE_Marshal(&sensitive->sensitiveArea,
             buffer_marshalled_sensitiveArea + sizeof(uint16_t),
             TPM2_MAX_DIGEST_BUFFER, &marshalled_sensitive_size);
+    if (rval != TPM2_RC_SUCCESS)
+    {
+        LOG_ERR("Error serializing the sensitive data");
+        return false;
+    }
+
     size_t marshalled_sensitive_size_info = 0;
-    Tss2_MU_UINT16_Marshal(marshalled_sensitive_size,
+    rval = Tss2_MU_UINT16_Marshal(marshalled_sensitive_size,
             buffer_marshalled_sensitiveArea, sizeof(uint16_t),
             &marshalled_sensitive_size_info);
+    if (rval != TPM2_RC_SUCCESS)
+    {
+        LOG_ERR("Error serializing the sensitive size");
+        return false;
+    }
 
     //concatenate NAME
     memcpy(buffer_marshalled_sensitiveArea + marshalled_sensitive_size +
@@ -385,8 +398,13 @@ bool tpm2_identity_util_calculate_inner_integrity(TPMI_ALG_HASH name_alg,
                     + marshalled_sensitive_size_info + pubname->size;
     size_t digest_size_info = 0;
     UINT16 hash_size = tpm2_alg_util_get_hash_size(name_alg);
-    Tss2_MU_UINT16_Marshal(hash_size, marshalled_sensitive_and_name_digest,
+    rval = Tss2_MU_UINT16_Marshal(hash_size, marshalled_sensitive_and_name_digest,
             sizeof(uint16_t), &digest_size_info);
+    if (rval != TPM2_RC_SUCCESS)
+    {
+        LOG_ERR("Error serializing the name size");
+        return false;
+    }
 
     digester d = tpm2_openssl_halg_to_digester(name_alg);
     d(buffer_marshalled_sensitiveArea,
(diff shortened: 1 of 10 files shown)


williamcroberts:
> Wrapping a key for a remote TPM2 right now requires loading the key into the local TPM. For some use cases it is not necessary to involve the local TPM, such as sending an openssl generated key to a remote machine.
> 
> `tpm2_duplicate` seems like the right place to put this functionality, although it shares a large amount of code with `tpm2_import`, and some with `tpm2_makecredential`. If there is a better place to locate the code, please let me know.
> 
> Some todo's before it is ready for merging:
> 
> * [ ]  need to add documentation
> * [ ]  write test cases
> * [ ]  add output option for public section
> * [ ]  move `DUPLICATE\0` seed generation into shared routine

While I'll definetly take this as a feature. Just as an FYI, if you have ssh on the remote machine and can get tpm2_send to version 5.0 on the remote, you can use the command TCTI to use your local tools on a remote TPM2 over SSH.

Example:
```bash
tpm2 getrandom --tcti="cmd:ssh tpm2_send" --hex 4
```

lib/tpm2_identity_util.c
```diff
@@ -423,3 +423,43 @@ void tpm2_identity_util_calculate_outer_integrity(TPMI_ALG_HASH parent_name_alg,
...
+     * 1. Marshaling the name algorithm
+     * 2. Marshaling the TPMT_PUBLIC past the name algorithm from step 1.
+     * 3. Hash the TPMT_PUBLIC portion in marshaled data.
+     */
+
+    TPMI_ALG_HASH name_alg = public->publicArea.nameAlg;
+
+    // Step 1 - set beginning of name to hash alg
+    size_t hash_offset = 0;
+    Tss2_MU_UINT16_Marshal(name_alg, pubname->name, pubname->size,
```
williamcroberts, line 441: Since you're here, can you add error checks to these throughout.

lib/tpm2_identity_util.c
```diff
@@ -423,3 +423,43 @@ void tpm2_identity_util_calculate_outer_integrity(TPMI_ALG_HASH parent_name_alg,
...
+            &tpmt_marshalled_size);
+
+    // Step 3 - Hash the data into name just past the alg type.
+    digester d = tpm2_openssl_halg_to_digester(name_alg);
+    if (!d) {
+        return false;
+    }
+
+    d((const unsigned char *) &marshaled_tpmt, tpmt_marshalled_size,
+            pubname->name + 2);
```
williamcroberts, line 458: Isn't the plus two going to be the hash_offset variable, we could just add that. I think this applies to the +2 below as well.

osresearch:
Tunneling tpm commands over ssh with `tpm2_send` seems like a neat feature, although I'm not sure that it fits the same use case -- my goal is to ensure that the remote system never sees the key in cleartext, so sending it over via ssh protects it in transit, but not from a hostile process on the remote side who might be able to snoop on the communication.

I'm tracking down what I broke in the asymetric `tpm2_import` test right now; it works with the RSA keys I was testing, but fails with a public/sensitive portion skew:

```
./tpm2 import -Q -G aes -g sha256 -i sym.key -C primary.ctx -u import.pub -r import.priv
WARNING:esys:src/tss2-esys/api/Esys_Import.c:323:Esys_Import_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_Import.c:110:Esys_Import() Esys Finish ErrorCode (0x000000a5) 
ERROR: Esys_HMAC(0xA5) - tpm:handle(unk):public and sensitive portions of an object are not cryptographically bound
ERROR: Unable to run import
```

williamcroberts (commented):
> Tunneling tpm commands over ssh with `tpm2_send` seems like a neat feature, although I'm not sure that it fits the same use case -- my goal is to ensure that the remote system never sees the key in cleartext, so sending it over via ssh protects it in transit, but not from a hostile process on the remote side who might be able to snoop on the communication.

Fair point, if thats your threat model, then yes that won't work. I just trying to find you a now solution.

> 
> I'm tracking down what I broke in the asymetric `tpm2_import` test right now; it works with the RSA keys I was testing, but fails with a public/sensitive portion skew:
> 
> ```
> ./tpm2 import -Q -G aes -g sha256 -i sym.key -C primary.ctx -u import.pub -r import.priv
> WARNING:esys:src/tss2-esys/api/Esys_Import.c:323:Esys_Import_Finish() Received TPM Error 
> ERROR:esys:src/tss2-esys/api/Esys_Import.c:110:Esys_Import() Esys Finish ErrorCode (0x000000a5) 
> ERROR: Esys_HMAC(0xA5) - tpm:handle(unk):public and sensitive portions of an object are not cryptographically bound
> ERROR: Unable to run import
> ```

Godspeed. Other than your todo items, most of my current comments will be minor nits. Overall the structure and directions looks good.

williamcroberts (changes requested):
Some of the comments are on code you copied and pasted, but I figured since were there make some minor cleanups.

Overall structure looks good. Just keep working on them todo's and thanks as always.

lib/tpm2_identity_util.c
```diff
@@ -218,7 +218,7 @@ bool tpm2_identity_util_calc_outer_integrity_hmac_key_and_dupsensitive_enc_key(
 
 bool tpm2_identity_util_share_secret_with_public_key(
         TPM2B_DIGEST *protection_seed, TPM2B_PUBLIC *parent_pub,
-        unsigned char *label, int label_len,
+        const unsigned char *label, int label_len,
         TPM2B_ENCRYPTED_SECRET *encrypted_protection_seed) {
     bool result = false;
```
idesai, line 223: @osresearch, since you are here: Can you also set ```rval = false``` in the error returns.

lib/tpm2_identity_util.c
```diff
@@ -99,7 +99,7 @@ static TPM2_KEY_BITS get_pub_asym_key_bits(TPM2B_PUBLIC *public) {
 }
 
 static bool share_secret_with_tpm2_rsa_public_key(TPM2B_DIGEST *protection_seed,
-        TPM2B_PUBLIC *parent_pub, unsigned char *label, int label_len,
+        TPM2B_PUBLIC *parent_pub, const unsigned char *label, int label_len,
         TPM2B_ENCRYPTED_SECRET *encrypted_protection_seed) {
     bool rval = false;
```
idesai, line 104: @osresearch, since you are here: Can you also set rval = false in the error returns.

lib/tpm2_openssl.c
```diff
@@ -1271,3 +1274,118 @@ tpm2_openssl_load_rc tpm2_openssl_load_private(const char *path,
...
+    TPM2B_PUBLIC * public,
+    const char * input_key_file,
+    TPMI_ALG_PUBLIC key_type,
+    const char * auth_key_file,
+    const char * policy_file,
+    const char * key_auth_str,
+    char * attrs_str,
+    const char * name_alg_str
+)
+{
```
idesai, line 1290:
```suggestion
bool tpm2_openssl_import_keys(TPM2B_PUBLIC * parent_pub,
    TPM2B_SENSITIVE * private, TPM2B_PUBLIC * public,
    const char * input_key_file, TPMI_ALG_PUBLIC key_type,
    const char * auth_key_file, const char * policy_file,
    const char * key_auth_str, char * attrs_str, const char * name_alg_str) {
```

lib/tpm2_openssl.c
```diff
@@ -1271,3 +1274,118 @@ tpm2_openssl_load_rc tpm2_openssl_load_private(const char *path,
...
+    bool result;
```
idesai, line 1291:
```suggestion

```

lib/tpm2_openssl.c
```diff
@@ -1271,3 +1274,118 @@ tpm2_openssl_load_rc tpm2_openssl_load_private(const char *path,
...
+        tpm2_session_close(&tmp);
+    }
+
+    /*
+     * Set the object attributes if specified, overwriting the defaults, but hooking the errata
+     * fixups.
+     */
+    if (attrs_str) {
+        TPMA_OBJECT *obj_attrs = &public->publicArea.objectAttributes;
+        result = tpm2_attr_util_obj_from_optarg(attrs_str, obj_attrs);
```
idesai, line 1327:
```suggestion
        bool result = tpm2_attr_util_obj_from_optarg(attrs_str, obj_attrs);
```

tools/tpm2_import.c
```diff
@@ -388,96 +348,20 @@ static tool_rc openssl_import(ESYS_CONTEXT *ectx) {
...
+
+    result = tpm2_openssl_import_keys(parent_pub, &private, &public,
+        ctx.input_key_file,
+        ctx.key_type,
+        ctx.auth_key_file,
+        ctx.policy,
+        ctx.key_auth_str,
+        ctx.attrs,
+        ctx.name_alg
+    );
```
idesai, line 362:
```suggestion
        result = tpm2_openssl_import_keys(parent_pub, &private, &public,
        ctx.input_key_file, ctx.key_type, ctx.auth_key_file, ctx.policy,
        ctx.key_auth_str, ctx.attrs, ctx.name_alg);
```

lib/tpm2_openssl.h
```diff
@@ -251,6 +251,24 @@ tpm2_openssl_load_rc tpm2_openssl_load_private(const char *path,
...
+    TPM2B_SENSITIVE * private,
+    TPM2B_PUBLIC * public,
+    const char * input_key_file,
+    TPMI_ALG_PUBLIC key_type,
+    const char * auth_key_file,
+    const char * policy_file,
+    const char * key_auth_str,
+    char * attrs_str,
+    const char * name_alg_str
+);
```
idesai, line 270:
```suggestion
bool tpm2_openssl_import_keys(TPM2B_PUBLIC *parent_pub,
    TPM2B_SENSITIVE *private, TPM2B_PUBLIC *public,
    const char *input_key_file, TPMI_ALG_PUBLIC key_type,
    const char *auth_key_file, const char *policy_file,
    const char *key_auth_str, char *attrs_str, const char *name_alg_str);
```

tools/tpm2_duplicate.c
```diff
@@ -160,46 +179,53 @@ static bool check_options(void) {
...
-
-    if (ctx.flags.r == 0) {
-        LOG_ERR("Expected private key out filename to be specified via \"-r\","
-                " missing option.");
+    if (ctx.flags.U != ctx.flags.k)
+    {
+        LOG_ERR("Conflicting options: remote public key and local private key must both be specified");
         result = false;
+    } else
+    if (!ctx.flags.U) {
```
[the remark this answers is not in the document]
osresearch: This might need some further refinement - we need to be sure that the correct set of `-U/-u/-r/-s/-k` *or* `-C/-c` are set but not both.

lib/tpm2_identity_util.c
osresearch: I'm not sure I understand -- I think it is doing the right thing. `rval = true` only after everything succeeds, and then the cleanup code is shared with the error return case:

```
    return_code = RSA_public_encrypt(mod_size, encoded,
            encrypted_protection_seed->secret, rsa, RSA_NO_PADDING);
    if (return_code < 0) {
        LOG_ERR("Failed RSA_public_encrypt\n");
        goto error;
    }

    rval = true;

error:
    free(pub_modulus);
    RSA_free(rsa);
    return rval;
```

lib/tpm2_openssl.c
osresearch: When there are a dozen arguments it seems like that formatting becomes more challenging to read and identify the ordering.

lib/tpm2_openssl.c
osresearch: `result` is used in several places in this function. are you suggesting to have a `bool result = ...` inside each of the scopes where it is used instead?

lib/tpm2_identity_util.c
idesai: your approach is correct.

lib/tpm2_openssl.c
idesai: your approach is correct.

lib/tpm2_identity_util.c
osresearch: Checks to all the Tss2_MU_*_Marshal calls added in https://github.com/tpm2-software/tpm2-tools/pull/2386/commits/6a570acc653b8bdea5063bd398113660b4700dfc

osresearch:
Is this failure in the rsadecrypt test related to these changes or a separate issue? https://github.com/tpm2-software/tpm2-tools/pull/2386/checks?check_run_id=2245282166

williamcroberts:
> Is this failure in the rsadecrypt test related to these changes or a separate issue? https://github.com/tpm2-software/tpm2-tools/pull/2386/checks?check_run_id=2245282166

sometimes cirrus is odd, I don't see why this PR would cause any issue, so let me re-trigger that and see if it replicates. The fact that all the linux tests work makes me suspect that this is just cirrus flakeyness.

tools/tpm2_duplicate.c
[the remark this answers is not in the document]
williamcroberts: Ok but brackets on evertything.

williamcroberts:
@osresearch Cirrus is passing, so if it fails on another minor push I won't even bother re-triggering it. However, just that minor nit on brackets on everything (else without bracket) and LGTM. @idesai will decide when to merge. Thanks again.

PR #2673, opened 2021-03-30
Update example tpm2_policyauthorize.1.md

ernesto418:
The example was bad explained because in the trial session the authorization policy is being created, it is not authorizing anything (you don't need to include -i pcr.policy in the trial session)
So I changed that, restructured the example, and changed the example variables names to make this command more understandable


idesai:
@ernesto418 can you please sign the commit

ernesto418:
I will close this commit, and create a new signed commit

PR #2690, opened 2021-04-07
tpm2_eventlog: fix device path indentation issue

huang195:
DevicePath field can be included in multiple types of events, and at different indentation levels. Previous commit assumes it's always at the same indentation level, e.g.,

```
- EventNum: 15
  PCRIndex: 1
  EventType: EV_EFI_VARIABLE_BOOT
  DigestCount: 2
  Digests:
  - AlgorithmId: sha1
    Digest: "df5d6605cb8f4366d745a8464cfb26c1efdc305c"
  - AlgorithmId: sha256
    Digest: "4d387b02d63b2f4cd7f667feb0a387fe47a10a3e26bf3533ddd001c605f3dec5"
  EventSize: 136
  Event:
    UnicodeNameLength: 8
    VariableDataLength: 88
    UnicodeName: Boot0002
    VariableData:
      Enabled: 'Yes'
      FilePathListLength: 44
      Description: "EFI Internal Shell"
    DevicePath: "FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(7c04a583-9e3e-4f1c-ad65-e05268d0b4d1)"
```

The `DevicePath` field in the last line above should be one more level indented. This PR allows DevicePath field to be printed at the right indentation level.

### DIFF
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -226,12 +226,12 @@ static bool yaml_uefi_post_code(const TCG_EVENT2* const event) {
  * print the field in raw byte format
  */
 #ifdef HAVE_EFIVAR_EFIVAR_H
-bool yaml_devicepath(BYTE* dp, UINT64 dp_len) {
+char* yaml_devicepath(BYTE* dp, UINT64 dp_len) {
     int ret;
     ret = efidp_format_device_path(NULL, 0, (const_efidp)dp, dp_len);
     if (ret < 0) {
         LOG_ERR("failed to allocate memory: %s\n", strerror(errno));
-        return false;
+        return NULL;
     }
 
     int text_path_len;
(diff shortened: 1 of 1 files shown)


huang195:
Another issue related to `DevicePath` field is that it could contain backslashes, e.g., 
```
DevicePath: "PciRoot(0x0)/Pci(0x6,0x0)/HD(15,GPT,c359e798-596a-42b6-b6f6-eca290429d99,0x2800,0x35000)/File(\EFI\BOOT\BOOTX64.EFI)"
```

As double quotes in YAML would escape the next character, this is not what we want. Instead, we should use single quotes so the strings can be interpreted literally.

lib/tpm2_eventlog_yaml.c
```diff
@@ -226,12 +226,12 @@ static bool yaml_uefi_post_code(const TCG_EVENT2* const event) {
  * print the field in raw byte format
  */
 #ifdef HAVE_EFIVAR_EFIVAR_H
-bool yaml_devicepath(BYTE* dp, UINT64 dp_len) {
+char* yaml_devicepath(BYTE* dp, UINT64 dp_len) {
```
idesai, line 229:
```suggestion
char *yaml_devicepath(BYTE* dp, UINT64 dp_len) {
```

PR #2701, opened 2021-04-15
[WIP] Adds support for tpm2_contextload command

jopelima:
This is an attempt to add standalone support to the standard contexload command, so that one can convert a session stored in a context file into the correspondent session handle, and having the session resident in the TPM memory.

While I attempted to do the flip side of this command (tpm2_contextsave), I ended up hitting some issues I wasn't sure how to solve, so will put that to the side until confirmation it makes sense to include these tools.

### DIFF
--- a/.ci/coverity.run
+++ b/.ci/coverity.run
@@ -3,7 +3,8 @@
 
 set -eo pipefail
 
-echo "PROJECT=$PROJECT"
+export COV_PROJECT="01org/tpm2.0-tools"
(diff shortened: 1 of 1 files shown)


williamcroberts:
Why do we need a new tool for this? Why doesn't this work: https://github.com/tpm2-software/tpm2-tools/pull/2700

jopelima:
Sorry, was trying to fix the PR and it ended up getting closed, I'll answer your question on #2700.

PR #2717, opened 2021-05-05
Adds support for tpm2_contextload command

jopelima:
Add standalone support to the standard contexload command, so that one can convert a session stored in a context file into the correspondent session handle, and having the session resident in the TPM memory.

After context loading a session using this command, the session context file cannot be used anymore for further commands, as there is no guarantee the data stored in it is accurate anymore. Executing any tpm2_tools command using the session context file will also fail, since the tools will attempt to context load the session before executing, but it will be already loaded at that point. The only alternative to have the same session being used again would be to have a corresponding tpm2_contextsave command, which is not included in this PR.

I also noticed that, when using the in-kernel RM (haven't tested on abrmd, but assuming previous @williamcroberts feedback it shouldn't happen there) executing the contextload command results in the session being lost, as the RM flushes it out when the command finishes executing. That is explicitly called out on the tool man page, so that users can now they have to execute the tpm2_contextload command directly against the TPM using the TCTI option specifier.

### DIFF
--- a/man/tpm2_contextload.1.md
+++ b/man/tpm2_contextload.1.md
@@ -1,0 +1,53 @@
+% tpm2_contextload(1) tpm2-tools | General Commands Manual
+
+# NAME
+
+**tpm2_contextload**(1) - Loads the session into the TPM, so that it can be
+referenced using the TPM session handle.
+
+# SYNOPSIS
+
+**tpm2_contextload** [*OPTIONS*] [*ARGUMENT*]
+
+# DESCRIPTION
+
+**tpm2_contextload**(1) - Loads the session into the TPM, so that it can be
+referenced using the TPM session handle.
+
+The tool returns the TPM handle the session is currently loaded on, and that
+handle value can be used directly when communicating directly with the TPM.
+
+**NOTE**: The session context file will not be usable on subsequent calls. As
+all tools invoke contextload/contextsave before/after using the session, the
+contextload call for subsequent commands will fail as the session is loaded
+already.
+
+**NOTE 2**: When working with the in-kernel resource manager, the contextload
+command should be executed directly against the TPM (using the --tcti option)
+, as the in-kernel RM flushes the objects upon exiting.
+
+* **ARGUMENT** the session context file.
+
+## References
+
+[common options](common/options.md) collection of common options that provide
+information many users may expect.
+
+[common tcti options](common/tcti.md) collection of options used to configure
+the various known TCTI modules.
+
+# EXAMPLES
+
+## Create a policy session, contextload it into the TPM
+
+```bash
+
+tpm2 startauthsession -S session.ctx --policy-session
+
+tpm2 contextload -T "device:/dev/tpm0" session.ctx
+
+```
+
+[returns](common/returns.md)
+
+[footer](common/footer.md)
--- a/man/tpm2_contextload.1.md
+++ b/man/tpm2_contextload.1.md
@@ -42,9 +42,9 @@ the various known TCTI modules.
 
 ```bash
 
-tpm2 startauthsession -S session.ctx --policy-session
+tpm2_startauthsession -S session.ctx --policy-session
 
-tpm2 contextload -T "device:/dev/tpm0" session.ctx
+tpm2_contextload -T "device:/dev/tpm0" session.ctx
 
 ```
 
--- a/tools/tpm2_contextload.c
+++ b/tools/tpm2_contextload.c
@@ -53,7 +53,7 @@ static tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
 
     UNUSED(flags);
 
-   bool retval = is_input_option_args_valid();
+    bool retval = is_input_option_args_valid();
     if (!retval) {
         return tool_rc_option_error;
     }
--- a/tools/tpm2_contextload.c
+++ b/tools/tpm2_contextload.c
@@ -20,24 +20,31 @@ struct tpm2_contextload_ctx {
 
 static tpm2_contextload_ctx ctx;
 
-static bool on_option(char key, char *value) {
+static bool on_args(int argc, char **argv) {
 
-    switch (key) {
-    case 'S':
-        ctx.session_path = value;
-        break;
+    if (argc > 1) {
+        LOG_ERR("Argument takes one file name for session data");
+        return false;
     }
+
+    ctx.session_path = argv[0];
+
+    return true;
+}
+
+static bool is_input_option_args_valid(void) {
+
+    if (!ctx.session_path) {
+        LOG_ERR("Must specify session file as an argument.");
+        return false;
+    }
+
     return true;
 }
 
 static bool tpm2_tool_onstart(tpm2_options **opts) {
 
-    static struct option topts[] = {
-        { "session",  required_argument,  NULL, 'S' },
-    };
-
-    *opts = tpm2_options_new("S:", ARRAY_LEN(topts), topts, on_option,
-    NULL, 0);
+    *opts = tpm2_options_new(NULL, 0, NULL, NULL, on_args, 0);
 
     return *opts != NULL;
 }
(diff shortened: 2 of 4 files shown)


man/tpm2_contextload.1.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Create a policy session, contextload it into the TPM
+
+```bash
+
+tpm2 startauthsession -S session.ctx --policy-session
+
+tpm2 contextload -T "device:/dev/tpm0" session.ctx
+
+```
```
williamcroberts, line 49: Expand this example with your specific use case, so we can show an end-to-end scenario:
```bash
# use the session handle here for whatever things need it. In this case, we will set up the key for ....
your commands here
```

tools/tpm2_contextload.c
```diff
@@ -0,0 +1,86 @@
...
+    *opts = tpm2_options_new(NULL, 0, NULL, NULL, on_args, 0);
+
+    return *opts != NULL;
+}
+
+static tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
+
+    UNUSED(flags);
+
+   bool retval = is_input_option_args_valid();
```
williamcroberts, line 56: indent, 3 spaces, needs 4.

tools/tpm2_contextload.c
```diff
@@ -0,0 +1,86 @@
...
+        return tool_rc_general_error;
+    }
+
+    TPM2_HANDLE tpm_handle;
+    TSS2_RC rv = Esys_TR_GetTpmHandle(ectx, sessionhandle, &tpm_handle);
+    if (rv != TSS2_RC_SUCCESS) {
+        return tool_rc_general_error;
+    }
+
+    tpm2_tool_output("Session-Handle: 0x%.8"PRIx32"\n", tpm_handle);
```
williamcroberts, line 80: +1 for matching session config.

man/tpm2_contextload.1.md
```diff
@@ -0,0 +1,53 @@
...
+
+# EXAMPLES
+
+## Create a policy session, contextload it into the TPM
+
+```bash
+
+tpm2 startauthsession -S session.ctx --policy-session
+
+tpm2 contextload -T "device:/dev/tpm0" session.ctx
```
williamcroberts, line 47: make the commands have the _ between tpm2 and <commandlet>. Ie:

tpm2_startauthsession
tpm2_contextload

idesai:
These can be squashed into a single commit.

idesai:
So we don't have to maintain similar code in separate tools, I am wondering if we can augment tpm2_sessionconfig with an option like **--dont-unload** or **--keep-resident** or any other apt option name. We can then use this boolean to simply not close/save the session at the end of tpm2_sessionconfig.

williamcroberts:
> So we don't have to maintain similar code in separate tools, I am wondering if we can augment tpm2_sessionconfig with an option like **--dont-unload** or **--keep-resident** or any other apt option name. We can then use this boolean to simply not close/save the session at the end of tpm2_sessionconfig.
> 
> @williamcroberts @jopelima your thoughts?

I thought about that, I'm not opposed to it. @idesai ill let you make the final call. Initially I thought, if we add it to one tool, we have to add it to all tools. But I guess we don't, we could just keep it in session config.

jopelima:
> So we don't have to maintain similar code in separate tools, I am wondering if we can augment tpm2_sessionconfig with an option like **--dont-unload** or **--keep-resident** or any other apt option name. We can then use this boolean to simply not close/save the session at the end of tpm2_sessionconfig.
> 
> @williamcroberts @jopelima your thoughts?

I'm ok with that approach as well. I thought having it as an independent command would make sense, since contextsave exist as a native TPM/ESAPI command, but ultimately the functionality matters most.

I'm not sure if it makes it harder to do the contextsave part, which I understood William suggesting may make sense in the review of my code.

Let me know which way you'd like to proceed, so that I can include the requested changes if it ends up going with the independent tool version.

williamcroberts:
> > So we don't have to maintain similar code in separate tools, I am wondering if we can augment tpm2_sessionconfig with an option like **--dont-unload** or **--keep-resident** or any other apt option name. We can then use this boolean to simply not close/save the session at the end of tpm2_sessionconfig.
> > @williamcroberts @jopelima your thoughts?
> 
> I'm ok with that approach as well. I thought having it as an independent command would make sense, since contextsave exist as a native TPM/ESAPI command, but ultimately the functionality matters most.
> 
> I'm not sure if it makes it harder to do the contextsave part, which I understood William suggesting may make sense in the review of my code.
> 
> Let me know which way you'd like to proceed, so that I can include the requested changes if it ends up going with the independent tool version.

@jopelima I talked with @idesai yesterday, and I think we're going to take the PR with the new tool. We decided since this is such a specific use case, we didn't want to pollute the other tool(s) with options as it could make them more confusing. Thanks again for resubmitting.

man/tpm2_contextload.1.md
jopelima: Fixed, thanks!

tools/tpm2_contextload.c
jopelima: Fixed, thanks!

man/tpm2_contextload.1.md
jopelima: Ok, there is an issue here though.

The end-to-end use-case doesn't quite work, because there are a few things that are slightly broken on the Kernel keyring trusted keys implementation for TPM2 (which was our original use case for this). It's unlikely that it will be fixed anytime soon, and when it gets fixed the subsystem may end up with a different interface to do these things.

So, at this point, I can't present a valid use-case for this tool. There may be others other people may have, but I'm not sure you'd want to keep the tool just in case?

man/tpm2_contextload.1.md
```diff
@@ -0,0 +1,53 @@
...
+**tpm2_contextload**(1) - Loads the session into the TPM, so that it can be
+referenced using the TPM session handle.
+
+The tool returns the TPM handle the session is currently loaded on, and that
+handle value can be used directly when communicating directly with the TPM.
+
+**NOTE**: The session context file will not be usable on subsequent calls. As
+all tools invoke contextload/contextsave before/after using the session, the
+contextload call for subsequent commands will fail as the session is loaded
+already.
```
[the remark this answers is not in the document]
jopelima: I did look at this sometime back, but was having a lot more difficulties getting the contextsave working compared to the contextload.

My understanding was that it required changes at the tools library level, since the tpm2_session.c functions that had part of the behaviour to do contextsave seemed to have additional, undesired side-effects.

I'll take another look at it and will try to come up with that as well (provided that it still makes sense to keep this tool pair around, based on my comment below).

jopelima:
Thanks for your reply @williamcroberts and apologies for taking some time to get back.

As mentioned as part of the review comments, the use case we initially considered for this tool isn't really valid anymore, since the Kernel trusted keys implementation works quite differently to the way we expected it to (to the point of making it very difficult to use policy sessions in practice at the moment - session HMAC is not being calculated on the kernel, and calculating it in userspace to be passed on to the kernel, while possible, seems quite finicky). 

I don't have another use case for this tool (or tool pair if we consider contextsave), so I'm not sure you'll want to keep this around, just in case someone else happens to have a different use-case? I'm happy to go whichever way you prefer.

man/tpm2_contextload.1.md
williamcroberts: > Ok, there is an issue here though.
> 
> The end-to-end use-case doesn't quite work, because there are a few things that are slightly broken on the Kernel keyring trusted keys implementation for TPM2 (which was our original use case for this). It's unlikely that it will be fixed anytime soon, and when it gets fixed the subsystem may end up with a different interface to do these things.
> 
> So, at this point, I can't present a valid use-case for this tool. There may be others other people may have, but I'm not sure you'd want to keep the tool just in case?

uhh, that sucks. The beautiful thing about PRs is that we can keep em around. Just don't kill your branch off. We can close this, but if the kernel stuff gets fixed and we have a full end-to-end need, I have no issues with merging this.

jopelima: Ok, I'll close this PR for now, not to have it dangling around here, but will leave the code if we later on end up being able to get the kernel side working.

Thanks for your help!

PR #2719, opened 2021-05-05
[RFC] session: dont start HMAC session on empty auth

williamcroberts:
Their are 1509 HMAC sessions started during the test code on empty
passwords. We dont need an HMAC session for an empty password.

Second thought: maybe we do? To hide that the object has an empty auth?
Granted thats security through obscurity and easily figured out.

@idesai thoughts?

This broke cphash calculations, not sure why.

### DIFF
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -177,10 +177,9 @@ tool_rc tpm2_nv_read(ESYS_CONTEXT *esys_context,
             goto tpm2_nvread_free_name1_name2;
         }
 
-        cp_hash->size = tpm2_alg_util_get_hash_size(
-            tpm2_session_get_authhash(auth_hierarchy_obj->session));
+        cp_hash->size = tpm2_alg_util_get_hash_size(TPM2_ALG_SHA256);
         rc = tpm2_sapi_getcphash(sys_context, name1, name2, NULL,
-            tpm2_session_get_authhash(auth_hierarchy_obj->session), cp_hash);
+            TPM2_ALG_SHA256, cp_hash);
         /*
          * Exit here without making the ESYS call since we just need the cpHash
          */
@@ -553,10 +552,9 @@ tool_rc tpm2_policy_authorize_nv(ESYS_CONTEXT *esys_context,
             goto tpm2_policyauthorizenv_free_name1_name2_name3;
         }
 
-        cp_hash->size = tpm2_alg_util_get_hash_size(
-            tpm2_session_get_authhash(auth_hierarchy_obj->session));
+        cp_hash->size = tpm2_alg_util_get_hash_size(TPM2_ALG_SHA256);
         rc = tpm2_sapi_getcphash(sys_context, name1, name2, name3,
-            tpm2_session_get_authhash(auth_hierarchy_obj->session), cp_hash);
+            TPM2_ALG_SHA256, cp_hash);
 
         /*
          * Exit here without making the ESYS call since we just need the cpHash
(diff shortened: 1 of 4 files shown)


idesai:
> Their are 1509 HMAC sessions started during the test code on empty
> passwords. We dont need an HMAC session for an empty password.
> 
> Second thought: maybe we do? To hide that the object has an empty auth?
> Granted thats security through obscurity and easily figured out.
> 
> Signed-off-by: William Roberts [william.c.roberts@intel.com](mailto:william.c.roberts@intel.com)
> 
> @idesai thoughts?
> 
> This broke cphash calculations, not sure why.

Not sure. Let me check.

idesai:
@williamcroberts, This PR has exposed an interesting bug in the tools not yet moved to the new layout/format that uses common routines to calculate cphash size. And so the cp_hash size is wrong because the session is a password session which means its session hash alg size is always returned zero.

There is a new routine/algo in 5.1/ master to calculate session digest sizes when this scenario happens. I will have to move the nvsetbits and policysigned tool to use it and we can then rebase commits in this PR on those changes. 

In general, though, the direction looks good unless security from obscurity is intended or preferred in this case. Some options,
1. Default always use hmac on blank passwords
2. Configure time option to disable hmac on blank passwords
Or
Vice versa.

williamcroberts:
> @williamcroberts, This PR has exposed an interesting bug in the tools not yet moved to the new layout/format that uses common routines to calculate cphash size. And so the cp_hash size is wrong because the session is a password session which means its session hash alg size is always returned zero.
> 
> There is a new routine/algo in 5.1/ master to calculate session digest sizes when this scenario happens. I will have to move the nvsetbits and policysigned tool to use it and we can then rebase commits in this PR on those changes.
> 
> In general, though, the direction looks good unless security from obscurity is intended or preferred in this case. Some options,
> 
> 1. Default always use hmac on blank passwords
> 2. Configure time option to disable hmac on blank passwords
>    Or
>    Vice versa.

If you were snooping the traffic, you have all the bits to add the secret into and see what it would HMAC to. So if I was trying to brute force this, I would start with a common passwords db and perform the HMAC and see if it matches. I think the first thing I would try is an empty password. Some might also consider probing various objects loaded in the TPM with am empty auth and see if you get access, but you could trip the DA lockout.

So I don't see obscurity really adding anything here (like usual).

idesai:
> > @williamcroberts, This PR has exposed an interesting bug in the tools not yet moved to the new layout/format that uses common routines to calculate cphash size. And so the cp_hash size is wrong because the session is a password session which means its session hash alg size is always returned zero.
> > There is a new routine/algo in 5.1/ master to calculate session digest sizes when this scenario happens. I will have to move the nvsetbits and policysigned tool to use it and we can then rebase commits in this PR on those changes.
> > In general, though, the direction looks good unless security from obscurity is intended or preferred in this case. Some options,
> > 
> > 1. Default always use hmac on blank passwords
> > 2. Configure time option to disable hmac on blank passwords
> >    Or
> >    Vice versa.
> 
> If you were snooping the traffic, you have all the bits to add the secret into and see what it would HMAC to. So if I was trying to brute force this, I would start with a common passwords db and perform the HMAC and see if it matches. I think the first thing I would try is an empty password. Some might also consider probing various objects loaded in the TPM with am empty auth and see if you get access, but you could trip the DA lockout.
> 
> So I don't see obscurity really adding anything here (like usual).

Okay sounds good. Additionally, we can always bind sessions to objects if need be. I will fix the two tools that are failing by using the new session hash algorithm derivation routines and then rebase this PR.

idesai:
@williamcroberts I added 0669667e as a workaround to be able to absorb this PR. The actual fix will be to move all tools to the new template where phash alg is determined based on session and attribute information.

Please rebase your commit on to  0669667e  to avoid git-bisect pointing to this PR commits as the failing point.

idesai:
@williamcroberts also have a scan bug
```../lib/tpm2_session.c:223:9: warning: Use of memory after it is freed
2843
        s->input->auth_data.size = strlen(password);
```

williamcroberts:
> @williamcroberts also have a scan bug
> 
> ```shell
> ../lib/tpm2_session.c:223:9: warning: Use of memory after it is freed
> 2843
>         s->input->auth_data.size = strlen(password);
> ```

Can you fix it and pull this PR over the finish line when you get to it? If not I can carve out time if needed. Theirs no rush on this patch, its an enhancement.

idesai:
> > @williamcroberts also have a scan bug
> > ```shell
> > ../lib/tpm2_session.c:223:9: warning: Use of memory after it is freed
> > 2843
> >         s->input->auth_data.size = strlen(password);
> > ```
> 
> Can you fix it and pull this PR over the finish line when you get to it? If not I can carve out time if needed. Theirs no rush on this patch, its an enhancement.

Awesome. I will delay this until I have the other tools to the new framework so my workaround commit will not be needed. I will take over this one and fix the scan bug too. Thanks Bill.

PR #2728, opened 2021-05-19
ci: Use the latest fedora stable release and rawhide

nullr0ute:
Rather than specifying a specific release, Fedora 30 is long obsolete,
Fedora 32 is EOL on May 25th, use the latest Fedora stable release which
is the equivilent of fedora:latest, and fedora:rawhide which is the
latest development release which saves updating this all the time.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     if: "!contains(github.ref, 'coverity_scan')"
     strategy:
       matrix:
-        docker_image: [ubuntu-18.04, ubuntu-20.04, fedora-30, opensuse-leap, fedora-32]
+        docker_image: [ubuntu-18.04, ubuntu-20.04, fedora, fedora:rawhide, opensuse-leap]
         compiler: [gcc, clang]
     steps:
       - name: Check out repository


.github/workflows/main.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     if: "!contains(github.ref, 'coverity_scan')"
     strategy:
       matrix:
-        docker_image: [ubuntu-18.04, ubuntu-20.04, fedora-30, opensuse-leap, fedora-32]
+        docker_image: [ubuntu-18.04, ubuntu-20.04, fedora, fedora:rawhide, opensuse-leap]
```
williamcroberts, line 10: I don't think this is going to work as you intend. The docker_Images are pulled from https://github.com/tpm2-software/tpm2-software-container. So we would want to:
1. create a fedora-rwhide image on the tpm2-software-container-project
2. remove all uses of fedora-30 from various CI's if in use.
3. drop the fedora-30 docker file.
4. delete the docker-30 image from Github Packages.

williamcroberts (changes requested):

.github/workflows/main.yml
williamcroberts: We also like to test on current stable releases, so we would want to keep fedora-32

nullr0ute: Fedora 32 is EOL in 6 days, it won't get any more updates from the tpm2 PoV. The current stable release is Fedora 34, Fedora 33 is still stable.

Issue #2738, opened 2021-05-25
Fix CVE-2021-3565


williamcroberts:
Fix the fixed key in tpm2_import. This allows an attacker to MITM key imports into the TPM.

idesai:
Create releasese 5.1.1 and 4.3.2

idesai:
5.1.1 and 4.3.2 release in progress.

ajakk:
> 5.1.1 and 4.3.2 release in progress.

I guess this got held up?

williamcroberts:
4.3.2: https://github.com/tpm2-software/tpm2-tools/releases/tag/4.3.2
5.1.1: https://github.com/tpm2-software/tpm2-tools/releases/tag/5.1.1

ajakk:
Thank you!

PR #2739, opened 2021-05-25
Fix import CVE-2021-3565

williamcroberts:
tpm2_import: fix fixed AES key CVE-2021-3565
    
tpm2_import used a fixed AES key for the inner wrapper, which means that
a MITM attack would be able to unwrap the imported key. Even the
use of an encrypted session will not prevent this. The TPM only
encrypts the first parameter which is the fixed symmetric key.
    
To fix this, ensure the key size is 16 bytes or bigger and use
OpenSSL to generate a secure random AES key.
    
Fixes: #2738

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -174,7 +174,7 @@ AS_IF([test "x$enable_unit" != xno], [
     AS_IF([test $BASH_SHELL = no],
           [AC_MSG_ERROR([Required executable bash not found, system tests require a bash shell!])])
 
-    AM_PATH_PYTHON([2.7],
+    AM_PATH_PYTHON([],
         [],
         [AC_MSG_ERROR([Required executable python not found, some system tests will fail!])]
     )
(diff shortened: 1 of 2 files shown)


williamcroberts:
@idesai merge and mailing list announce? We should probably back port this to stable branches and cut some minor releases so packagers can update easily. Ie if their on 4.3.2, they can easily go to 4.2.3, etc.

configure.ac
```diff
@@ -174,7 +174,7 @@ AS_IF([test "x$enable_unit" != xno], [
     AS_IF([test $BASH_SHELL = no],
           [AC_MSG_ERROR([Required executable bash not found, system tests require a bash shell!])])
 
-    AM_PATH_PYTHON([2.7],
+    AM_PATH_PYTHON([],
```
idesai, line 177: @williamcroberts is this related to this fix?

williamcroberts: It's a separate commit that happens to be in the PR, we don't need to backport it.

PR #2745, opened 2021-05-28
tpm2_eventlog: Display GUIDs correctly

### LINKED ISSUE #2744: tpm2_eventlog: GUIDs are displayed incorrectly
When #1950 introduced parsing for various UEFI events, it also included parsing/displaying GUIDs using `libuuid`.

Unfortunately, libuuid's [`uuid_t` type](https://github.com/karelzak/util-linux/blob/ba3b93525920f2782e12016e7444fca89fb5bb9c/libuuid/src/uuid.h#L44) and UEFI's [`EFI_GUID` type](https://github.com/tianocore/edk2/blob/0ecdcb6142037dd1cdd08660a2349960bcf0270a/MdePkg/Include/Base.h#L218-L227) do not have the same memory layout. `uuid_t` is byte buffer (big endian), while `EFI_GUID` is a struct w/ 4 fields (mixed endian). 

`EFI_GUID` is [used in structs like `UEFI_VARIABLE_DATA`](https://github.com/tianocore/edk2/blob/0ecdcb6142037dd1cdd08660a2349960bcf0270a/MdePkg/Include/IndustryStandard/UefiTcgPlatform.h#L246-L252) which are then [directly `memcpy`ed to the eventlog](https://github.com/tianocore/edk2/blob/e1999b264f1f9d7230edf2448f757c73da567832/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c#L2027). This means any parsing software also has to use the weird `EFI_GUID` structure.

As `tpm2 eventlog` doesn't currently do this, the global variable GUID (`8be4df61-93ca-11d2-aa0d-00e098032b8c`) displays as `61dfe48b-ca93-d211-aa0d-00e098032b8c` and the Microsoft GUID (`77fa9abd-0359-4d32-bd60-28f4e78f784b`) displays as `bd9afa77-5903-324d-bd60-28f4e78f784b`. See [this GUID list](https://github.com/rhboot/efivar/blob/master/src/guids.txt) for more info.

josephlr:
Fixes #2744 

This PR adds the definition of `EFI_GUID` and uses it instead of `uuid_t`. It also implements `guid_unparse_lower` using sprintf. This then removes any need to depend on libuuid.

Running the data against a sample event log gives correct GUIDs which are consistent with `efibootmgr -v`, `ls /sys/firmware/efi/efivars`, and `efi-readvar`.

Signed-off-by: Joe Richey <joerichey@google.com>


lib/efi_event.h
```diff
@@ -48,6 +48,18 @@
 #define PACKED __attribute__((__packed__))
 #endif
 
+/* GUIDs used by UEFI/GPT are stored in a mixed-endian layout, where the first
+ * three fields are serialized as little-endian and the final two fields are
+ * serialized as big-endian. This differs from uuid_t which is entirely a
+ * big-endian layout (just an array of 16 bytes).
+ */
```
idesai, line 54: That is an interesting arrangement of structure members in UEFI/GPT. If we are not using the uuid library we may have to revision and track this structure. @williamcroberts your thoughts?

josephlr: Do we have a way to revision and track all the other UEFI structures? Given that this definition is from the spec, if we do revision tracking, we should track all of the UEFI-spec structures together.

idesai: True, we don't do it as yet.

lib/tpm2_eventlog_yaml.c
```diff
@@ -23,6 +23,13 @@
 #include <efivar/efivar.h>
 #endif
 
+static void guid_unparse_lower(EFI_GUID guid, char guid_buf[37]) {
+    snprintf(guid_buf, 37, "%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x",
```
idesai, line 27:
```suggestion

    snprintf(guid_buf, 37, "%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x",
```

josephlr:
@idesai do you want to do the struct revision/tracking in this PR? If so, how would I go about doing that?

idesai:
> @idesai do you want to do the struct revision/tracking in this PR? If so, how would I go about doing that?

This PR is good, please can you add the integration test in #2747?

idesai:
This can be tested/seen with test-fixture event-bootorder.bin

PR #2762, opened 2021-06-10
policy: Support TPMLess commands

williamcroberts:
For now this works with policysecret, owner hiearchy and NULL auth.
It's hardcoded to ignore the TPM.

tpm2 policysecret -S session.ctx -c o
0d84f55daf6e43ac97966e62c9bb989d3397777d25c5f749868055d65394f952

TODO:
For each policy command:
  - Support --tcti=none
  - When --tcti is none, require:
    --name/-n
    -L/--policy for old value
    -- Require hash algorithm.
    - Don't output a TICKET, since we can't?

Todo, consider creating a session.ctx structure that can be passed
from startauthsession with --tcti=none? This was we can encapsulate
the state instead of requiring -L and hash algorithm?

### DIFF
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -685,6 +685,106 @@ tool_rc tpm2_policy_countertimer(ESYS_CONTEXT *esys_context,
     return tool_rc_success;
 }
 
+static tool_rc policy_update(TPMI_ALG_HASH halg, TPM2_CC cc, TPM2B_NAME *name, TPM2B_DIGEST *old, TPM2B_NONCE *ref, TPM2B_DIGEST *new) {
+
+    tool_rc result = tool_rc_general_error;
+    const EVP_MD *md = tpm2_openssl_halg_from_tpmhalg(halg);
+    if (!md) {
+        LOG_ERR("Could not find digester for algorithm 0x%x", halg);
+        return tool_rc_general_error;
+    }
+
+    EVP_MD_CTX *mdctx = EVP_MD_CTX_create();
+    if (!mdctx) {
+        LOG_ERR("%s", tpm2_openssl_get_err());
+        return tool_rc_general_error;
+    }
+
+    int rc = EVP_DigestInit_ex(mdctx, md, NULL);
+    if (!rc) {
+        LOG_ERR("%s", tpm2_openssl_get_err());
+        goto out;
+    }
+
+    rc = EVP_DigestUpdate(mdctx, old->buffer, old->size);
+    if (!rc) {
+        LOG_ERR("%s", tpm2_openssl_get_err());
+        goto out;
+    }
+
+    TPM2_CC becc = tpm2_util_hton_32(cc);
+    rc = EVP_DigestUpdate(mdctx, &becc, sizeof(becc));
+    if (!rc) {
+        LOG_ERR("%s", tpm2_openssl_get_err());
+        goto out;
+    }
+
+    if (name) {
+        rc = EVP_DigestUpdate(mdctx, name->name, name->size);
+        if (!rc) {
+            LOG_ERR("%s", tpm2_openssl_get_err());
+            goto out;
+        }
+    }
+
+    unsigned size = EVP_MD_size(md);
+    rc = EVP_DigestFinal_ex(mdctx, new->buffer, &size);
+    if (!rc) {
+        LOG_ERR("%s", tpm2_openssl_get_err());
+        goto out;
+    }
+    /* hash sizes are not bigger than 16 bits, safe truncate */
+    new->size = (UINT16)size;
+
+    if(ref) {
+        mdctx = EVP_MD_CTX_create();
+        if (!mdctx) {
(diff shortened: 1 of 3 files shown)


idesai:
@williamcroberts calculating cphash when rphash isn't required could also use a similar approach. In that case though, cphash alg has to specified using the provision halg:/path/to/cphash.dat

williamcroberts:
For this, I think we could to something like:
```bash
tpm2 startauthsession --tcti=none --trial-session-no-tpm -S session.ctx
# this will propagate a new session.ctx format that can store halg and current policy hash
# do we want to enforce that if --trial-session-no-tpm is specified that --tcti=none is also specified?

# Policy tools then can load the new session format, and get the halg/policy hash WHEN --tcti=none
tpm2 policytoolX --tcti=none -S session.ctx

# flush will just ignore these contexts or error?
```

Thoughts here?

idesai:
> For this, I think we could to something like:
> 
> ```shell
> tpm2 startauthsession --tcti=none --trial-session-no-tpm -S session.ctx
> # this will propagate a new session.ctx format that can store halg and current policy hash
> # do we want to enforce that if --trial-session-no-tpm is specified that --tcti=none is also specified?
> 
> # Policy tools then can load the new session format, and get the halg/policy hash WHEN --tcti=none
> tpm2 policytoolX --tcti=none -S session.ctx
> 
> # flush will just ignore these contexts or error?
> ```
> 
> Thoughts here?

How about this approach:
1. Specify the policy halg with the file-path for policy using something like -L halg:/path/to/policyfile
2. Mandate this way of specifying the policy file path if tcti=none is specified
3. If the policyfile has non-zero content we extend the data in the policyfile with the one created by the policy tool consuming the inputs above.
4. Like cpHash, we don't actually execute the policy command and exit out writing out the policyfile.

nicowilliams:
> For this, I think we could to something like:
> 
> ```shell
> tpm2 startauthsession --tcti=none --trial-session-no-tpm -S session.ctx
> # this will propagate a new session.ctx format that can store halg and current policy hash
> # do we want to enforce that if --trial-session-no-tpm is specified that --tcti=none is also specified?
> 
> # Policy tools then can load the new session format, and get the halg/policy hash WHEN --tcti=none
> tpm2 policytoolX --tcti=none -S session.ctx
> 
> # flush will just ignore these contexts or error?
> ```
> 
> Thoughts here?

Yes:

> # do we want to enforce that if --trial-session-no-tpm is specified that --tcti=none is also specified?

No, I think that's implied.  Even if `TPM2TOOLS_TCTI` is set in the environment to something other than `none`.  Alternatively, `--tcti=none` implies a trial, software session, and it's already used in `tpm2 duplicate` and `tpm2 makecredential` to signify "do a software-only operation that TPMs can do".  So I would say "lose the `-trial-session-no-tpm` argument.

> # Policy tools then can load the new session format, and get the halg/policy hash WHEN --tcti=none

That works, but presumably the policy tools could tell that the session is a `--tcti=none` session, so maybe the `--tcti=none` option can be elided in this case.  But it's OK `--tcti=none` is required anyways -- consistency.

williamcroberts:
> > For this, I think we could to something like:
> > ```shell
> > tpm2 startauthsession --tcti=none --trial-session-no-tpm -S session.ctx
> > # this will propagate a new session.ctx format that can store halg and current policy hash
> > # do we want to enforce that if --trial-session-no-tpm is specified that --tcti=none is also specified?
> > 
> > # Policy tools then can load the new session format, and get the halg/policy hash WHEN --tcti=none
> > tpm2 policytoolX --tcti=none -S session.ctx
> > 
> > # flush will just ignore these contexts or error?
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thoughts here?
> 
> Yes:
> 
> > # do we want to enforce that if --trial-session-no-tpm is specified that --tcti=none is also specified?
> 
> No, I think that's implied. Even if `TPM2TOOLS_TCTI` is set in the environment to something other than `none`. Alternatively, `--tcti=none` implies a trial, software session, and it's already used in `tpm2 duplicate` and `tpm2 makecredential` to signify "do a software-only operation that TPMs can do". So I would say "lose the `-trial-session-no-tpm` argument.
> 
> > # Policy tools then can load the new session format, and get the halg/policy hash WHEN --tcti=none
> 
> That works, but presumably the policy tools could tell that the session is a `--tcti=none` session, so maybe the `--tcti=none` option can be elided in this case. But it's OK `--tcti=none` is required anyways -- consistency.

I think we would have to go with `--tcti=none`. As that's understood by the common framework to not try and invoke a TPM connection. If we make it a different tool argument, the framework won't know about that and attempt to invoke a TPM connection which is what we want to avoid. So I think `--tcti=none`, starting in `tpm2_startauthsession`, will create a session.ctx file that maintains the state the TPM normally would (so we can track the hash, and alg). Then we can just run the hash commands an update the context file until someone does a `-L` to dump the policy.

PR #2765, opened 2021-06-11
lgtm: fix for tss version

williamcroberts:


williamcroberts:
@idesai also need to update LGTM config, so after this all CI's should be passing AFAIK.

PR #2769, opened 2021-06-14
Support pcrs output file formats "values" and "serialized"

hansonchar:
for tpm2_pcrread and tpm2_quote, with backward compatible defaults.

Signed-off-by: Hanson Char <hchar@amazon.com>

### DIFF
--- a/lib/pcr.c
+++ b/lib/pcr.c
@@ -237,30 +237,119 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
     return result;
 }
 
-bool pcr_print_pcr_struct(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
+bool pcr_fwrite_serialized(const TPML_PCR_SELECTION *pcr_select,
+    const tpm2_pcrs *ppcrs, FILE *output_file) {
 
-    UINT32 vi = 0, di = 0, i;
-    bool result = true;
+    TPML_PCR_SELECTION pcr_selection = *pcr_select;
+    for (UINT32 i = 0; i < pcr_selection.count; i++) {
+        TPMS_PCR_SELECTION *p = &pcr_selection.pcrSelections[i];
+        p->hash = htole16(p->hash);
+    }
+    pcr_selection.count = htole32(pcr_selection.count);
 
-    tpm2_tool_output("pcrs:\n");
+    // Export TPML_PCR_SELECTION structure to pcr outfile
+    size_t fwrite_len = fwrite(&pcr_selection, sizeof(TPML_PCR_SELECTION),
+        1, output_file);
+    if (fwrite_len != 1) {
+        LOG_ERR("write to output file failed: %s", strerror(errno));
+        return false;
+    }
 
+    // Export PCR digests to pcr outfile
+    UINT32 count = htole32(ppcrs->count);
+    fwrite_len = fwrite(&count, sizeof(UINT32), 1, output_file);
+    if (fwrite_len != 1) {
+        LOG_ERR("write to output file failed: %s", strerror(errno));
+        return false;
+    }
+
+    tpm2_pcrs pcrs = *ppcrs;
+    for (size_t j = 0; j < pcrs.count; j++) {
+        TPML_DIGEST *pcr_value = &pcrs.pcr_values[j];
+
+        for (size_t k = 0; k < pcr_value->count; k++) {
+            TPM2B_DIGEST *p = &pcr_value->digests[k];
+            p->size = htole16(p->size);
+        }
+        pcr_value->count = htole32(pcr_value->count);
+        fwrite_len = fwrite(pcr_value, sizeof(TPML_DIGEST), 1, output_file);
+        if (fwrite_len != 1) {
+            LOG_ERR("write to output file failed: %s", strerror(errno));
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool pcr_fwrite_values(const TPML_PCR_SELECTION *pcr_select,
+    const tpm2_pcrs *pcrs, FILE *output_file) {
+
+    size_t vi = 0;  /* value index */
+    UINT32 di = 0;  /* digest index */
     // Loop through all PCR/hash banks
-    for (i = 0; i < pcr_select->count; i++) {
-        const char *alg_name = tpm2_alg_util_algtostr(
-                pcr_select->pcrSelections[i].hash, tpm2_alg_util_flags_hash);
+    for (UINT32 i = 0; i < pcr_select->count; i++) {
+        const TPMS_PCR_SELECTION *pcr_selection = &pcr_select->pcrSelections[i];
+        // Loop through all PCRs in this bank
+        for (unsigned int pcr_id = 0; pcr_id < pcr_selection->sizeofSelect * 8u;
+            pcr_id++) {
+
+            bool is_pcr_select_bit_set = tpm2_util_is_pcr_select_bit_set(
+                pcr_selection, pcr_id);
+            if (! is_pcr_select_bit_set) {
+                // skip non-selected banks
+                continue;
+            }
+            const TPML_DIGEST *pcr_value = &pcrs->pcr_values[vi];
+            if (vi >= pcrs->count || di >= pcr_value->count) {
+                LOG_ERR("Something wrong, trying to print but nothing more");
+                return false;
+            }
+
+            // Print out current PCR digest value
+            const TPM2B_DIGEST *digest = &pcr_value->digests[di];
+            size_t fwrite_len = fwrite(digest->buffer, digest->size, 1,
+                output_file);
+            if (fwrite_len != 1) {
+                LOG_ERR("write to output file failed: %s", strerror(errno));
+                return false;
+            }
+
+            if (++di >= pcr_value->count) {
+                di = 0;
+                ++vi;
+            }
+        } /* end looping through all PCRs in a bank */
+    }  /* end looping through all PCR banks */
+
+    return true;
+}
+
+bool pcr_print_values(const TPML_PCR_SELECTION *pcr_select,
+    const tpm2_pcrs *pcrs) {
+
+    size_t vi = 0;  /* value index */
+    UINT32 di = 0;  /* digest index */
+    // Loop through all PCR/hash banks
+    for (UINT32 i = 0; i < pcr_select->count; i++) {
+        const TPMS_PCR_SELECTION *pcr_selection = &pcr_select->pcrSelections[i];
+        const char *alg_name = tpm2_alg_util_algtostr(pcr_selection->hash,
+            tpm2_alg_util_flags_hash);
 
         tpm2_tool_output("  %s:\n", alg_name);
 
         // Loop through all PCRs in this bank
-        unsigned int pcr_id;
-        for (pcr_id = 0; pcr_id < pcr_select->pcrSelections[i].sizeofSelect * 8u;
-                pcr_id++) {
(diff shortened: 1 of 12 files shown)


lib/pcr.c
```diff
@@ -237,26 +237,114 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
     return result;
 }
 
-bool pcr_print_pcr_struct(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
+bool pcr_fwrite_serialized(const TPML_PCR_SELECTION *ppcr_select,
+                           const tpm2_pcrs *ppcrs, FILE *output_file) {
```
idesai, line 241:
```suggestion
    const tpm2_pcrs *ppcrs, FILE *output_file) {
```

lib/pcr.c
```diff
@@ -237,26 +237,114 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
+    // Make sure the file content is written in little endian format
+    pcr_selections.count = htole32(pcr_selections.count);
+    UINT32 i;
+    for (i = 0; i < le32toh(pcr_selections.count); i++) {
+        pcr_selections.pcrSelections[i].hash = htole16(
+                pcr_selections.pcrSelections[i].hash);
+    }
+
+    // Export TPML_PCR_SELECTION structure to pcr outfile
+    if (fwrite(&pcr_selections, sizeof(TPML_PCR_SELECTION), 1,
```
idesai, line 253:
```suggestion
    size_t fwrite_len = fwrite(&pcr_selections, sizeof(TPML_PCR_SELECTION), 1, output_file);
    if (fwrite_len != 1) {
```

lib/pcr.c
```diff
@@ -237,26 +237,114 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
+               output_file) != 1) {
+        LOG_ERR("write to output file failed: %s", strerror(errno));
+        return false;
+    }
+
+    const UINT32 count = htole32(ppcrs->count);
+    // Export PCR digests to pcr outfile
+    if (fwrite(&count, sizeof(UINT32), 1, output_file) != 1) {
```
idesai, line 261: Don't hide function returns in a condition.

lib/pcr.c
```diff
@@ -237,26 +237,114 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
+    tpm2_pcrs pcrs = *ppcrs;
+    for (UINT32 j = 0; j < pcrs.count; j++) {
+        pcrs.pcr_values[j].count = htole32(pcrs.pcr_values[j].count);
+
+        for (UINT32 k = 0; k < le32toh(pcrs.pcr_values[j].count); k++) {
+            pcrs.pcr_values[j].digests[k].size = htole16(
+                    pcrs.pcr_values[j].digests[k].size);
+        }
+
+        if (fwrite(&pcrs.pcr_values[j], sizeof(TPML_DIGEST), 1,
```
idesai, line 275: Don't hide function returns in a condition.

lib/pcr.c
```diff
@@ -237,26 +237,114 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
+                           const tpm2_pcrs *ppcrs, FILE *output_file) {
+
+    TPML_PCR_SELECTION pcr_selections = *ppcr_select;
+    // Make sure the file content is written in little endian format
+    pcr_selections.count = htole32(pcr_selections.count);
+    UINT32 i;
+    for (i = 0; i < le32toh(pcr_selections.count); i++) {
+        pcr_selections.pcrSelections[i].hash = htole16(
+                pcr_selections.pcrSelections[i].hash);
+    }
```
williamcroberts, line 250: Rather than doing yet another binary format, can't we just make this YAML?

Plus, since in general things are host endianess or big endianness in this code base, BE over LE.

lib/pcr.c
```diff
@@ -237,26 +237,114 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
+bool pcr_fwrite_serialized(const TPML_PCR_SELECTION *ppcr_select,
+                           const tpm2_pcrs *ppcrs, FILE *output_file) {
+
+    TPML_PCR_SELECTION pcr_selections = *ppcr_select;
+    // Make sure the file content is written in little endian format
+    pcr_selections.count = htole32(pcr_selections.count);
+    UINT32 i;
+    for (i = 0; i < le32toh(pcr_selections.count); i++) {
+        pcr_selections.pcrSelections[i].hash = htole16(
+                pcr_selections.pcrSelections[i].hash);
```
idesai, line 249:
```suggestion
        pcr_selections.pcrSelections[i].hash);
```

lib/pcr.c
hansonchar: This particular binary format ("serialized") is currently the format used by `tpm2_quote` to output the selected list of PCR's.  In contrast, the binary format used by `tpm2_pcrread` is currently equivalent to the proposed "values" format.  Since the existing output format behavior of the two commands are not are different and not selectable, it is not possible to compare the output files (of the same PCR values, for example) directly to see if they are the same or not.

For backward compatibility, this patch doesn't change the default behavior of the two commands, while allowing the users to choose the specific output formats to use.  This helps alleviate the comparison problems described above.

In other words, this patch hasn't created "yet another binary format", but rather makes the existing binary formats available for selection and reuse. To that end, I have not changed the endianess of the existing binary formats.

We can certainly support additional formats, such as YAML, but I think that belongs to a future enhancement.

williamcroberts: Ahh that knocks the dust off my brain. I remember @idesai and I having a discussion on this way back when we noticed that the tools had differing PCR formats. So this supports an interchange between those tools. +2 from me.

lib/tpm2_convert.h
```diff
@@ -25,6 +25,25 @@ enum tpm2_convert_sig_fmt {
     signature_format_err
 };
 
+typedef enum tpm2_convert_pcrs_output_fmt tpm2_convert_pcrs_output_fmt;
+enum tpm2_convert_pcrs_output_fmt {
+    pcrs_output_format_values,
+    pcrs_output_format_serialized,
```
williamcroberts, line 31: I wish we could find a better naming scheme than values and serialized, but since I don't have a better suggestion and the binary formats are obtuse anyways, it lives on.

idesai:
@hansonchar, the commits can be squashed. Please can you also add an integration test.

lib/pcr.c
```diff
@@ -237,26 +237,116 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
-    tpm2_tool_output("pcrs:\n");
+    // Loop through all PCR/hash banks
+    for (i = 0; i < pcr_select->count; i++) {
+        // Loop through all PCRs in this bank
+        unsigned int pcr_id;
+        for (pcr_id = 0;
+             pcr_id < pcr_select->pcrSelections[i].sizeofSelect * 8u;
+             pcr_id++) {
+            if (!tpm2_util_is_pcr_select_bit_set(&pcr_select->pcrSelections[i],
+                    pcr_id)) {
```
idesai, line 301:
```suggestion
            bool is_pcr_select_bit_set = tpm2_util_is_pcr_select_bit_set(&pcr_select->pcrSelections[i], pcr_id)
            if (! is_pcr_select_bit_set) {
```

lib/pcr.c
```diff
@@ -237,26 +237,116 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
+            }
+            if (vi >= pcrs->count || di >= pcrs->pcr_values[vi].count) {
+                LOG_ERR("Something wrong, trying to print but nothing more");
+                return false;
+            }
+
+            // Print out current PCR digest value
+            TPM2B_DIGEST *digest = &pcrs->pcr_values[vi].digests[di];
+
+            if (fwrite(digest->buffer, digest->size, 1, output_file) != 1) {
```
idesai, line 313: Don't hide fn returns in condition.

lib/pcr.c
```diff
@@ -237,26 +237,116 @@ bool pcr_print_pcr_struct_le(TPML_PCR_SELECTION *pcr_select, tpm2_pcrs *pcrs) {
...
         tpm2_tool_output("  %s:\n", alg_name);
 
         // Loop through all PCRs in this bank
         unsigned int pcr_id;
-        for (pcr_id = 0; pcr_id < pcr_select->pcrSelections[i].sizeofSelect * 8u;
-                pcr_id++) {
+        for (pcr_id = 0;
+             pcr_id < pcr_select->pcrSelections[i].sizeofSelect * 8u;
+             pcr_id++) {
             if (!tpm2_util_is_pcr_select_bit_set(&pcr_select->pcrSelections[i],
```
idesai, line 348: Don't hide fn returns in condition.

tools/tpm2_pcrread.c
```diff
@@ -1,4 +1,4 @@
-/* SPDX-License-Identifier: BSD-3-Clause */
+ /* SPDX-License-Identifier: BSD-3-Clause */
```
idesai, line 1:
```suggestion
/* SPDX-License-Identifier: BSD-3-Clause */
```

tools/tpm2_quote.c
```diff
@@ -277,12 +245,13 @@ static bool tpm2_tool_onstart(tpm2_options **opts) {
         { "signature",      required_argument, NULL, 's' },
         { "message",        required_argument, NULL, 'm' },
         { "pcr",            required_argument, NULL, 'o' },
+        { "pcrs_format",     required_argument, NULL, 'F' },
```
idesai, line 248:
```suggestion
        { "pcrs_format",    required_argument, NULL, 'F' },
```

idesai:
> @hansonchar, the commits can be squashed. Please can you also add an integration test.

We want to make sure the default output format hasn't changed to be backwards compatible.

idesai:
Thank you @hansonchar. The changes look good. Can you please squash these into a single commit?
And also add an integration test, may be one that shows how this commit bridges quote and pcrread would be a good one. Like so: ```diff pcrread.pcrdata quote.pcrdata```

hansonchar:
Hi @idesai and @williamcroberts, please let me know if there are additional changes you'd like to make to this pull request.

idesai:
> Hi @idesai and @williamcroberts, please let me know if there are additional changes you'd like to make to this pull request.

@hansonchar This looks good. Just need to add the integration test.

```bash
tpm2_quote -Q -c key.ctx -l 0x0004:16,17,18+0x000b:16,17,18 -o quoted.pcr
tpm2_pcrread -Q  0x0004:16,17,18+0x000b:16,17,18 -o pcrread.pcr
diff pcrread.pcr quoted.pcr .........< must fail>
tpm2_pcrread -Q  0x0004:16,17,18+0x000b:16,17,18 -o serialized.pcrread.pcr -F serialized
diff serialized.pcrread.pcr quoted.pcr ............<must pass>

hansonchar:
Oops, didn’t see this.  Already squashed.  Will work on the integration test.  

Thanks,
Hanson

hansonchar:
Integration test `pcrs_format.sh` added.

hansonchar:
@idesai is there anything else needed before having this pull request merged?

Issue #2764, opened 2021-06-11
TPM2_CC_NV_Read/Write pHash data for larger buffers is in error


idesai:
Max read size can be specified up to UINT16_MAX, however there some caveats:
1. If calculating pHash without executing the actual NV read, a read buffer
   size bigger than: max-command-response-buffer OR max-NV-space OR max-NV-Index
   size is allowed. But such a pHash will not be useful when specifying its use in policies or audits.
2. When the actual read occurs, it is broken down in multiple blocks of:
   max-command-response-buffer up to max-NV-space  OR up to max-NV-Index size.
   In this case if pHash is additionally required, then:
   the actual audit/ policy session buffer, if one is specified, must be updated
   with multiple TPM2_CC_NV_READ and so the phash output must be a digest of
   all the phash to match audit/ policy session buffer action at this instance.

idesai:
Addressed in the linked commits shown in this ticket.

PR #2782, opened 2021-06-29
lib/tpm2_nv_util.h: parse info for NV op specific max nv buffer sz

idesai:
Progresses #2764 

The NV index size is constrained to different data sets read from
the GetCapability structures because there is a dependency on the
NV operation type: Define vs Read vs Write vs Extend.

This commit addresses those parameters and chooses a sane default
in the case where GetCapability fails or fails to report the
specific property/ data set. This is especially true because some
of the properties are TPM implementation dependent.

### DIFF
--- a/lib/tpm2_nv_util.h
+++ b/lib/tpm2_nv_util.h
@@ -40,42 +40,90 @@ static inline tool_rc tpm2_util_nv_read_public(ESYS_CONTEXT *context,
     if (tmp_rc != tool_rc_success) {
         rc = tmp_rc;
     }
-
     return rc;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


lib/tpm2_nv_util.h
```diff
@@ -40,42 +40,91 @@ static inline tool_rc tpm2_util_nv_read_public(ESYS_CONTEXT *context,
     if (tmp_rc != tool_rc_success) {
         rc = tmp_rc;
     }
-
+return rc;
     return rc;
```
williamcroberts, line 44: multiple returns?

Issue #185, opened 2017-01-17
Enhance tpm2_takeownership to reconfigure lockout related properties


gwei3:
Since tpm2_takeownership -c was changed to use lockout auth to clear tpm ownership, it is very easy to trigger the disabled status of lockout auth if using wrong lockout auth to clear tpm. The default value for TPM_PT_LOCKOUT_RECOVERY might be very large or just 0(require reset), so it would be easier to give use choices to reconfigure it as needed.

Meanwhile, other properties like TPM_PT_MAX_AUTH_FAIL, TPM_PT_LOCKOUT_INTERVAL should be able to be easily reconfigured per needs by users.

idesai:
@gwei3 this has been factored into the tpm2_dictionarylockout tool. Can close this one? Thanks!

gwei3:
yes, let's close it since tpm2_dictionarylockout tool already supported the required functions.

PR #2783, opened 2021-07-01
tpm2_nvread: add support to calculate cphash in extended mode

idesai:
Progresses #2764

### DIFF
--- a/lib/tpm2_nv_util.h
+++ b/lib/tpm2_nv_util.h
@@ -6,6 +6,8 @@
 #include <tss2/tss2_esys.h>
 #include <string.h>
 
+#include <openssl/sha.h>
+
 #include "log.h"
 #include "tpm2.h"
 #include "tpm2_capability.h"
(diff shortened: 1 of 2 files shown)


idesai:
> Progresses #2764
> 
> ```
> /*
>  * It is possible that the NV Read operation has a buffer size smaller than
>  * that of an NV Index. In such scenarios, the cpHash digest is extended to
>  * emulate the cpHash calculation on the TPM.
>  *
>  * The extension comes from the fact that when there is a larger buffer,
>  * this tool dispatches multiple NV Reads until all the chunks are read.
>  */
> ```
> 
> Signed-off-by: Imran Desai [imran.desai@intel.com](mailto:imran.desai@intel.com)

Well, cphash are per command basis. So instead of extending hashes, the commands should actually generate separate cphash blobs.

lib/tpm2.c
```diff
@@ -127,16 +127,39 @@ tool_rc tpm2_getcap(ESYS_CONTEXT *esys_context, TPM2_CAP capability,
...
+    bool is_cmd_dispatched = cp_hash->size ? false : true;
+
+    bool is_nvindex_name_input_required =
+        (!is_cmd_dispatched && cp_hash->size);
+    if (is_nvindex_name_input_required && !precalc_nvname) {
+        LOG_ERR("Must specify the nvindex name.");
+        return tool_rc_option_error;
+    }
+
+    bool is_auth_hierarchy_nv_index =
```
idesai, line 147: Specify this to its closest use, line #185

lib/tpm2_nv_util.h
```diff
@@ -6,6 +6,8 @@
 #include <tss2/tss2_esys.h>
 #include <string.h>
 
+#include <openssl/sha.h>
```
idesai, line 9: Don't need this.

idesai:
This is currently failing CI as it breaks backwards compatibility for tpm2_nvread.
This is because with this PR enables calculating cpHash without having to go to the TPM, however that mandates specifying the NV-index name through the options. This mandate conflicts for 4.X tpm2_nvread tool.
Solution:
Because tcti is still checked for in esys and sys initializations, we need to pass a fake tcti.

lib/tpm2_options.c
```diff
@@ -233,6 +233,29 @@ void tpm2_print_usage(const char *command, struct tpm2_options *tool_opts) {
...
 }
 
+static TSS2_RC tcti_fake_receive (TSS2_TCTI_CONTEXT *tctiContext, size_t *response_size,
+        unsigned char *response_buffer, int32_t timeout) {
+	(void) tctiContext;
+	(void) response_size;
+	(void) response_buffer;
+	(void) timeout;
+
+	return TSS2_TCTI_RC_NOT_IMPLEMENTED;
```
williamcroberts, line 243: I would toss some error logs here, because if this ever gets invoked we want to know exactly why it failed simply by looking at logs.

lib/tpm2_options.c
```diff
@@ -233,6 +233,29 @@ void tpm2_print_usage(const char *command, struct tpm2_options *tool_opts) {
...
+static void tcti_fake_finalize (TSS2_TCTI_CONTEXT *tctiContext) {
+	(void) tctiContext;
+}
+
+static TSS2_RC tcti_fake_transmit (TSS2_TCTI_CONTEXT *tcti_ctx, size_t size,
+        const uint8_t *cmd_buf) {
+	(void) tcti_ctx;
+	(void) size;
+	(void) cmd_buf;
+
```
williamcroberts, line 255: Same log error here

lib/tpm2_options.h
```diff
@@ -66,6 +66,7 @@ typedef bool (*tpm2_arg_handler)(int argc, char **argv);
  */
 #define TPM2_OPTIONS_NO_SAPI 0x1
 #define TPM2_OPTIONS_OPTIONAL_SAPI 0x2
+#define TPM2_OPTIONS_OPTIONAL_SAPI_AND_FAKE_TCTI 0x4
```
williamcroberts, line 69: Offhand, I would just make NO_SAPI the fake sapi, so if someone does something like
`tpm2_nvread --tcti=none --cphash xxx` it just works. But existing tools that check if ectx is null, like makecredential would need to have a flag set or something to not try to use the ectx.

How exactly do you see tpm2_nvread being executed? Do you envision that if --cphash is requested that you just send the fake tcti down and avoid users needing to do --tcti=none?

idesai: We have to rely on tcti=none being called so as to not break compatibility with 4.X

williamcroberts: If tcti=none, I would just send the FAKE tcti and use it's magic to determine if it's fake tcti for things like makecred.

Issue #1597, opened 2019-07-16
tpm2_import: hmac support


williamcroberts:
Add support for importing HMAC keys

osresearch:
Was this ever implemented? In #1556 it is implied that HMAC keys can be imported, although the man page doesn't mention it and `tpm2 import -G hmac` results in `ERROR: Cannot handle algorithm, got: hmac`

salrashid123:
Just checking if this capability is in the works for an upcoming release?

salrashid123:
thanks.   suggest adding in an examples here for the import too:
*   https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_import.1.md#examples
*   https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_hmac.1.md

```bash
export secret="key"
export plain="The quick brown fox jumps over the lazy dog"

echo -n $secret > hmac.key
hexkey=$(xxd -p -c 256 < hmac.key)
echo $hexkey

echo -n $plain > data.in

openssl dgst -sha256 -mac hmac -macopt hexkey:$hexkey data.in
   HMAC-SHA256(data.in)= f7bc83f430538424b13298e6aa6fb143ef4d59a14946175997479dbc2d1a3cd8
   

tpm2 createprimary -Q -G rsa -g sha256 -C e -c primary.ctx

tpm2 import -C primary.ctx -G hmac -i hmac.key -u hmac.pub -r hmac.priv
tpm2 load -C primary.ctx -u hmac.pub -r hmac.priv -c hmac.ctx

echo -n $plain | tpm2_hmac -g sha256 -c hmac.ctx | xxd -p -c 256
    f7bc83f430538424b13298e6aa6fb143ef4d59a14946175997479dbc2d1a3cd8

```

---

as a side note, importing hmac could be used to generate AWS S4 signed requests and for S3 and Google storage signedurls.
eg, for aws request, i used go-tpm [here](https://github.com/salrashid123/aws_hmac) to create an AWS v4 signer

PR #2791, opened 2021-07-15
tpm2_import: add support for importing HMAC keys and sealed data blobs

whooo:
This enables wrapping an external HMAC key or a data blob under a key hierarchy.

Fixes https://github.com/tpm2-software/tpm2-tools/issues/1597

### DIFF
--- a/lib/tpm2_openssl.c
+++ b/lib/tpm2_openssl.c
@@ -1018,6 +1018,39 @@ static bool load_public_AES_from_file(FILE *f, const char *path,
     return tpm2_util_calc_unique(name_alg, key, seed, unique);
 }
 
+#define KEYEDHASH_MAX_SIZE 128
+
+static bool load_public_KEYEDHASH_from_file(FILE *f, const char *path,
+        TPM2B_PUBLIC *pub, TPM2B_SENSITIVE *priv) {
+
+    /*
+     * Get the file size and validate that it is within the allowed size for keyed hash objects
+     */
+    unsigned long file_size = 0;
+    bool result = files_get_file_size(f, &file_size, path);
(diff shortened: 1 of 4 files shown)


lib/tpm2_openssl.c
```diff
@@ -1018,6 +1018,37 @@ static bool load_public_AES_from_file(FILE *f, const char *path,
...
+    /*
+     * Get the file size and validate that it is within the allowed size for keyed hash objects
+     */
+    unsigned long file_size = 0;
+    bool result = files_get_file_size(f, &file_size, path);
+    if (!result) {
+        return false;
+    }
+
+    if (file_size > 128 || file_size == 0) {
```
idesai, line 1033: is the magic number 128 a TSS defined max value? use that instead.

whooo: I couldn't find any define when looking in tss2_tpm2_types.h

idesai: I thought it was MAX_SYM_DATA. However, In the 11.1.3 section of the part II of the specifications it allows for a max value of 128 but in tss2_tpm2_types.h it is 256. @AndreasFuchsSIT, @williamcroberts  do you know why the difference?

idesai: @whooo can you macro define a label for the number 128?

whooo: Sure, so something like ```#define KEYEDHASH_MAX_SIZE 128``` in lib/tpm2_openssl.c ?

whooo: @idesai it's now added

idesai (approved):

PR #2801, opened 2021-07-29
tpm2_eventlog: fix buffer offset when reading the event log

### LINKED ISSUE #2778: tpm2_eventlog segfault
I'm playing around with a Dell Latitude 7520 and i see:

tpm2_eventlog /sys/kernel/security/tpm0/binary_bios_measurements
---
version: 1
events:
Segmentation fault

https://github.com/ValdikSS/binary_bios_measurements_parser gives me:
```
./binary_bios_measurements_parser.py -l -1 -f binary_bios_measurements
Num PCR EV_type                         PCR_value                                Size Valid Data
0   0   no_action                       0000000000000000000000000000000000000000 33   0     b'Spec ID Event03\x00\x00\x00\x00\x00\x00\x02\x00\x02\x01\x00\x00\x00\x0b\x00 \x00'
1   0   no_action                       010000000b000000000000000000000000000000 0    0     b''
2   0   preboot_cert                    00000000000011000000537461727475704c6f63 1953066081 0     ....
```

So I'm guessing that somethings not as good as its expected. I expect tpm2_eventlog not to segfault on a value being wrong, but  rather just state its invalid size.

Any ideas on now to continue to diagnose this issue would also be appreciated.

aplanas:
The event log is read in chunks of CHUNK_SIZE blocks (16KB), always
checking when the EOF is reached, so is compatible with virtual files
that lives in securefs and we do not know the full size.  The current
code is not taking care of adjusting the offset when the next chunk is
read.

This patch add "size" to the base buffer where the event log is stored
in memory.

Fix #2778

### DIFF
--- a/tools/misc/tpm2_eventlog.c
+++ b/tools/misc/tpm2_eventlog.c
@@ -92,7 +92,7 @@ static tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
     unsigned long size = 0;
     bool is_file_read = false;
     do {
-        is_file_read = files_read_bytes_chunk(fileptr, eventlog, CHUNK_SIZE, &size);
+        is_file_read = files_read_bytes_chunk(fileptr, eventlog + size, CHUNK_SIZE, &size);
         UINT8 *eventlog_tmp = realloc(eventlog, size + CHUNK_SIZE);
         if (!eventlog_tmp){
             LOG_ERR("failed to allocate %lu bytes: %s", size + CHUNK_SIZE, strerror(errno));


Vogtinator (approved):
Works, thanks!

Nitpicks: The commit message is missing an "is" and s/securefs/securityfs/

aplanas:
> Nitpicks: The commit message is missing an "is" and s/securefs/securityfs/

I found a missing "it", and fixed the securityfs reference. Thanks!

idesai:
@aplanas this looks good. I am waiting for some CI issues to resolve prior to merge.

PR #2810, opened 2021-08-09
Convert deprecated digesters to EVP_Digest

gotthardp:
Before, two approaches were used for digest calculation:
1) `tpm2_openssl_halg_from_tpmhalg` and the EVP_Digest
2) `tpm2_openssl_halg_to_digester` and (deprecated) digesters such as SHA256 or SHA384 

After this PR, only a single approach is used: `tpm2_openssl_md_from_tpmhalg` and the EVP_Digest. The function name was changed to highlight it returns `EVP_MD *` (the term "halg" is not much used in openssl).

### DIFF
--- a/lib/tpm2_identity_util.c
+++ b/lib/tpm2_identity_util.c
@@ -134,7 +134,7 @@ static bool share_secret_with_tpm2_rsa_public_key(TPM2B_DIGEST *protection_seed,
     unsigned char encoded[TPM2_MAX_DIGEST_BUFFER];
     return_code = RSA_padding_add_PKCS1_OAEP_mgf1(encoded, mod_size,
             protection_seed->buffer, protection_seed->size, label, label_len,
-            tpm2_openssl_halg_from_tpmhalg(parent_name_alg), NULL);
+            tpm2_openssl_md_from_tpmhalg(parent_name_alg), NULL);
     if (return_code != 1) {
         LOG_ERR("Failed RSA_padding_add_PKCS1_OAEP_mgf1\n");
         goto error;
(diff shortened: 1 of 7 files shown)


lib/tpm2_identity_util.c
```diff
@@ -481,16 +489,20 @@ bool tpm2_identity_create_name(TPM2B_PUBLIC *public, TPM2B_NAME *pubname) {
...
+    const EVP_MD *md = tpm2_openssl_md_from_tpmhalg(name_alg);
+    if (!md) {
+        LOG_ERR("Algorithm not supported: %x", name_alg);
         return false;
     }
 
-    d((const unsigned char *) &marshaled_tpmt, tpmt_marshalled_size,
-            pubname->name + hash_offset);
+    unsigned int hash_size;
+    if (!EVP_Digest(&marshaled_tpmt, tpmt_marshalled_size,
```
idesai, line 499: We don't hide function returns in conditionals. This and elsewhere...

idesai (changes requested):

lib/tpm2_identity_util.c
gotthardp: OK, I have fixed that.

No big deal here, but openssl heavily uses "streams of commands" also elsewhere, relying on lazy evaluation. For example, see the code at the end of this page: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_fromdata.html.
It is of course possible to rewrite it with explicit `rc = `, but the beauty of a concise code will be lost.

idesai (approved):

PR #2814, opened 2021-08-15
openssl: Use EVP_MAC_update instead of HMAC_Update on OpenSSL >= 3.0.0

gotthardp:
The HMAC_Update is deprecated in OpenSSL 3.0, but the replacement EVP_MAC_update was added in OpenSSL 3.0, so version specific code is needed.

### DIFF
--- a/lib/tpm2_kdfa.c
+++ b/lib/tpm2_kdfa.c
@@ -2,6 +2,13 @@
 
 #include <string.h>
 
+#include <openssl/evp.h>
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
+#include <openssl/hmac.h>
+#else
+#include <openssl/core_names.h>
+#endif
+
 #include "log.h"
 #include "tpm2_kdfa.h"
 #include "tpm2_openssl.h"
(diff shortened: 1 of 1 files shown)


lib/tpm2_kdfa.c
```diff
@@ -40,13 +47,27 @@ TSS2_RC tpm2_kdfa(TPMI_ALG_HASH hash_alg, TPM2B *key, char *label,
...
 
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
     int rc = HMAC_Init_ex(ctx, key->buffer, key->size, md, NULL);
+#else
+    OSSL_PARAM params[2];
+
+    params[0] = OSSL_PARAM_construct_utf8_string(OSSL_ALG_PARAM_DIGEST,
+                                                 (char *)EVP_MD_get0_name(md), 0);
+    params[1] = OSSL_PARAM_construct_end();
+    int rc = EVP_MAC_init(ctx, key->buffer, key->size, params);
```
idesai, line 69: Should this be unsigned? PS: Return looks and behaves like a boolean.

PR #2815, opened 2021-08-15
openssl: Implement EVP_PKEY based key import

gotthardp:
The `RSA_KEY` and `EC_KEY` are not publicly available in OpenSSL 3.0 and the generic `EVP_PKEY` must be used instead.
Since import of raw keys still requires access to the internal structures the OpenSSL 3.0 introduced a completely new approach to access key internals.

This PR mainly:
* removes the unnecessary BN_bn2binpad compatibility function
* changes `PEM_read_RSA_PUBKEY`/`PEM_read_RSAPublicKey` and `PEM_read_EC_PUBKEY` to a generic `PEM_read_PUBKEY`
* changes `PEM_read_RSAPrivateKey` and `PEM_read_ECPrivateKey` to `PEM_read_PrivateKey`
* adds functions for key import from OpenSSL 3.0 key internals via the `EVP_PKEY_get_xxx_param` functions
* removes NULL checks before OpenSSL `xxx_free` functions as these checks are done internally; the OpenSSL itself heavily relies on that

### DIFF
--- a/lib/tpm2_openssl.c
+++ b/lib/tpm2_openssl.c
@@ -6,7 +6,11 @@
 #include <string.h>
 
 #include <openssl/pem.h>
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
 #include <openssl/rand.h>
+#else
+#include <openssl/core_names.h>
+#endif
 
 #include "files.h"
 #include "log.h"
@@ -520,8 +524,9 @@ static bool handle_ossl_pass(const char *passin, char **pass) {
     return pfn(passin, pass);
 }
 
-static bool load_public_RSA_from_key(RSA *k, TPM2B_PUBLIC *pub) {
+static bool load_public_RSA_from_key(EVP_PKEY *key, TPM2B_PUBLIC *pub) {
 
+    bool result = false;
     TPMT_PUBLIC *pt = &pub->publicArea;
     pt->type = TPM2_ALG_RSA;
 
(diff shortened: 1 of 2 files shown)


lib/tpm2_openssl.c
```diff
@@ -710,46 +696,80 @@ int tpm2_ossl_curve_to_nid(TPMI_ECC_CURVE curve) {
...
 
     int ret = EC_POINT_get_affine_coordinates_tss(group, point, x, y, NULL);
     if (!ret) {
         LOG_ERR("Could not get X and Y affine coordinates");
         goto out;
     }
+#else
+    char curveName[80];
+
+    int rc = EVP_PKEY_get_utf8_string_param(key, OSSL_PKEY_PARAM_GROUP_NAME,
```
idesai, line 744: Should rc be unsigned?

lib/tpm2_openssl.c
```diff
@@ -710,46 +696,80 @@ int tpm2_ossl_curve_to_nid(TPMI_ECC_CURVE curve) {
...
+    }
+    nid = OBJ_txt2nid(curveName);
+    keysize = (EVP_PKEY_bits(key) + 7) / 8;
+
+    rc = EVP_PKEY_get_bn_param(key, OSSL_PKEY_PARAM_EC_PUB_X, &x);
+    if (!rc) {
+        LOG_ERR("Could not read public X coordinate");
+        goto out;
+    }
+    rc = EVP_PKEY_get_bn_param(key, OSSL_PKEY_PARAM_EC_PUB_Y, &y);
```
idesai, line 758:
```suggestion

    rc = EVP_PKEY_get_bn_param(key, OSSL_PKEY_PARAM_EC_PUB_Y, &y);
```

lib/tpm2_openssl.c
```diff
@@ -799,43 +817,30 @@ static bool load_public_ECC_from_key(EC_KEY *k, TPM2B_PUBLIC *pub) {
...
         return false;
     }
 
-    bool result = load_public_ECC_from_key(k, pub);
+    bool result;
+    if (EVP_PKEY_base_id(k) == EVP_PKEY_EC) {
+        result = load_public_ECC_from_key(k, pub);
+    } else {
+        result = false;
+    }
```
idesai, line 841:
```suggestion
    bool result = false;
    if (EVP_PKEY_base_id(k) == EVP_PKEY_EC) {
        result = load_public_ECC_from_key(k, pub);
    }
```

lib/tpm2_openssl.c
```diff
@@ -639,15 +619,21 @@ static bool load_public_RSA_from_pem(FILE *f, const char *path,
...
         return false;
     }
 
-    bool result = load_public_RSA_from_key(k, pub);
+    bool result;
+    if (EVP_PKEY_base_id(k) == EVP_PKEY_RSA) {
+        result = load_public_RSA_from_key(k, pub);
+    } else {
+        result = false;
+    }
```
idesai, line 634:
```suggestion
    bool result = false;
    if (EVP_PKEY_base_id(k) == EVP_PKEY_RSA) {
        result = load_public_RSA_from_key(k, pub);
    }
```

idesai (changes requested):

lib/tpm2_openssl.c
gotthardp: The documentation says the result is `int`. https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_get_utf8_string_param.html

Do you see a reason why `rc` should be unsigned?

idesai: I see that makes sense. Let's stick with int.
I thought the return was never going to be negative.

idesai (approved):

PR #2821, opened 2021-08-23
[RFC] support keyedhash and hmac seperatley

williamcroberts:
Currently -Ghmac in tpm2_create will produce a keyedhash object with a
scheme of hmac-sha256. However, import -G hmac will produce a keyedhash
with a scheme of null. Have import use hmac to create a fully specified
keyedhash object like create, and have keyedhash perform a null alg.

For HMAC the hashing algorithm will be the same as the key size, so only
hmac keys of a certain size will be accepted. This might not be
desirable. We might want to extend the parsing routines to handle
hmac:sha256, etc, and then default to the normal sha256 if left off.


williamcroberts:
@cplappert thoughts?

PR #2825, opened 2021-08-30
Import hmac keyedhash

williamcroberts:
This PR rectified the -G behavior between `tpm2_create` and `tpm2_createprimary` with `tpm2_import. -G keyedhash` will give you something akin to a "seal object" in `tpm2_create` and `tpm2_createprimary`, and hmac will give you an hmac object (keyedhash with an hmac hash alg). This also makes things like `-Ghmac:sha512` possible.


williamcroberts:
@idesai negative diffstat with added features, love it :-p

idesai (approved):

Issue #2813, opened 2021-08-14
generate_partial_X509 fails as i2d_X509_bio cannot produce invalid data in OSSL >=3.0


gotthardp:
The `tpm2_certifyX509certutil.c`:`generate_partial_X509()` creates a "partial" (in fact, an incomplete) certificate and stores that with `i2d_X509_bio`. The OpenSSL team decided that starting from 3.0 the *i2d should not create invalid encodings*. In other words, the encoding of a partial certificate will fail.
```
00429D20477F0000:error:068000DE:asn1 encoding routines:asn1_template_ex_i2d:illegal zero content::0:
00429D20477F0000:error:068C0100:asn1 encoding routines:ASN1_item_i2d_bio:malloc failure::0:
ERROR: i2d_X509_bio
ERROR: Unable to run certifyX509certutil
```
The change is originally motivated by https://github.com/openssl/openssl/issues/16026, implemented by https://github.com/openssl/openssl/pull/16036 and further discussed in https://github.com/openssl/openssl/issues/16257. The later even includes a suggestion to create a `TPM_PARTIAL_CERT` structure.

Please have a look at the proposal https://github.com/openssl/openssl/issues/16257#issuecomment-895448370. Is that a correct way to proceed?

williamcroberts:
Yeah I have been following the mailing list for openssl more closely because poor Ken has been one of the first testing OSSL 3.0 against common TPM things... so I would imagine whatever Ken gets suggested to do by the community is the right thing to do.

PR #2826, opened 2021-08-31
Generate partial X.509 certificate using own struct

gotthardp:
The `tpm2_certifyX509certutil.c`:`generate_partial_X509()` creates a partial (in fact, an incomplete) certificate, see Part 3, 18.8.1. The OpenSSL team decided that starting from 3.0 the i2d should not create invalid encodings, so we need to create an own ASN.1 structure for this.

This PR:
 * Switches from `X509 *` to `TPM2_PARTIAL_CERT *`, which is adapted from https://github.com/openssl/openssl/issues/16257#issuecomment-895448370.
 * Removes unnecessary `X509_get_XX` - `X509_set_XX` pairs. (The pointer was retrieved and then put back, which is not necessary as the pointer was already set when retrieved.)
 * Removes the `fixup_cert` which I believe is not necessary now.

Fixes: #2813

### DIFF
--- a/tools/misc/tpm2_certifyX509certutil.c
+++ b/tools/misc/tpm2_certifyX509certutil.c
@@ -8,12 +8,54 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include <openssl/asn1.h>
+#include <openssl/asn1t.h>
 #include <openssl/x509.h>
 #include <openssl/x509v3.h>
 #include <openssl/bio.h>
 
 #include "log.h"
 #include "tpm2_tool.h"
+
+typedef struct {
+    ASN1_TIME *notBefore;
+    ASN1_TIME *notAfter;
+} TPM2_PARTIAL_CERT_VALIDITY;
+
+typedef struct {
+    X509_ALGOR *algorithm;
+    X509_NAME *issuer;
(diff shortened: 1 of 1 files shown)


gotthardp:
I do *not* have a way to verify the output certificate is correct.

williamcroberts:
> I do _not_ have a way to verify the output certificate is correct.

I guess we could create the TPM2_CertifyX509 command line tool and pass this into it?

williamcroberts:
> > I do _not_ have a way to verify the output certificate is correct.
> 
> I guess we could create the TPM2_CertifyX509 command line tool and pass this into it?

So the tpm2-tss project has a test in test/integration/esys-certifyX509.int.c that takes the output of this tool and adds it as a included byte buffer, you could paste the output and test it.

gotthardp:
> So the tpm2-tss project has a test in test/integration/esys-certifyX509.int.c that takes the output of this tool and adds it as a included byte buffer, you could paste the output and test it.

That `test/integration/esys-certifyX509.int` (`Esys_CertifyX509`) test passed with the "new" data. Of course, the `openssl asn1parse -in partial_cert.der -inform DER` happily prints the cert content, so it seems to be valid.

williamcroberts:
That sounds good enough until we actually get a tool in place, I can add this tool. @idesai sound good to you?

Issue #1599, opened 2019-07-17
TPM2 tools export keys in a parsable form


nmav:
The TPM2 tools like tpm2_createprimary and tpm2_create output keys in a form that cannot be identified by other applications. There is a defacto standard from the tpm2 openssl engines which is a PEM (base64) encoding with header "BEGIN TSS2 PRIVATE KEY" of the DER encoding of the ASN.1 definition below:

```
 TPMKey ::= SEQUENCE {
    type            OBJECT IDENTIFIER,
    emptyAuth       [0] EXPLICIT BOOLEAN OPTIONAL,
    parent          INTEGER,
    pubkey          OCTET STRING,
    privkey         OCTET STRING
}
```

(see https://github.com/tpm2-software/tpm2-tss-engine/blob/master/src/tpm2-tss-engine-common.c#L44 )

The advantage of this form in comparison with the existing form is that:
     1. a single file is used to store public and private key
     2. the file can be detected from the PEM header

It would be useful for other applications not necessarily depending on openssl engine to be able to use TPM2 keys, and the move by tpm2-tools to outputing format will make this simpler. In particular it will allow opening such keys as part of a generic key opening routine rather than a specialized tpm2 key loading which will require multiple files and options.

That change can be backwards compatible and would improve interaction with tpm2-tools as well, for the same reasons (a single file needed, future revisions can rely on headers to load older formats).

[amended 2019-08-7: I realized that the a similar issue occurs for the public portion of the key; the format exported by `tpm2_create` is unknown (to me) and a standard SubjectPublicKeyInfo would have been much more useful.]

dwmw2:
It would be extremely useful to be able to import/export such keys with the PKCS#11 module too.

williamcroberts:
> It would be extremely useful to be able to import/export such keys with the PKCS#11 module too.

FYI That is a goal.

williamcroberts:
@AndreasFuchsSIT we mentioned a tpm2 util library. The first 2 really useful candidates are:
1. load from disk in multiple forms
2. save to disk in multiple forms

AndreasFuchsSIT:
Well, yes, that format is definately useful. Don't know if it's the perfect match for the tools, since they are oriented on part 3 of tpm-spec. But as **additional** option, this could be useful for people...

williamcroberts:
useful for tools, we do PEM conversions

williamcroberts:
Plus that's also code common to FAPI

[...]

PR #2838, opened 2021-09-17
tpm2_encodeobject: New tool to encode TPM2 object

ueno:
This adds a new tool tpm2_encodeobject in tools/misc.  It takes public and private portions of an object and encode them in a combined PEM form used by tpm2-tss-engine and other applications.

Related: #1599

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -103,6 +103,7 @@ tools_tpm2_SOURCES = \
 tpm2_tools = \
     tools/misc/tpm2_certifyX509certutil.c \
     tools/misc/tpm2_checkquote.c \
+    tools/misc/tpm2_encodeobject.c \
     tools/misc/tpm2_eventlog.c \
     tools/misc/tpm2_print.c \
     tools/misc/tpm2_rc_decode.c \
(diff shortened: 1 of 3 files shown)


idesai:
@ueno what would be a good test for it?

ueno:
> That's not entirely true, as it currently contains the features to print the output in PEM for objects that can be converted like TPMT_PUBLIC and TPM2B_PUBLIC structures.

Indeed; sorry, I missed that because `-f` option is only mentioned in the EXAMPLES section. Still, I am wondering if this is the right place to this functionality: as the new behavior would take multiple objects (public key, private blob, and primary context), while the current `tpm2_print` processes only one object file at a time. Would it be ok to extend the command to take, say `-C` (for primary context), `-u' (for public key), and `-r` (for private blob), if `-f pem` is specified (or any new option that indicates this "combining" operation)?

williamcroberts:
We can leave this as it's own tool, since it would clutter tpm2_print AND print is a "NO SAPI" tool. However, things under misc have historically been NO_SAPI tools, I don't see why we cannot add this here. Fine with me.

PR #2839, opened 2021-09-20
Update tpm2_pcrevent.1.md

PhilippHomann:
Fixed order of arguments in example


idesai:
@PhilippHomann it can be specified in any order. Is it not working in your case?

PhilippHomann:
My bad.. Haven't read that the order doesn't matter.
Haven't tried it yet, but I believe it will work :+1:

PR #2878, opened 2021-11-10
tools/tpm2_nvread: fix reads > TPM2_PT_NV_BUFFER_MAX with policy session

### LINKED ISSUE #2875: tpm2_nvread with PCR authorization fails
Using tpm2-tools v5.2:

```
tpm2_createpolicy --policy-pcr -l sha1:0 -L policy.digest
tpm2_nvdefine -L policy.digest -a 'policyread|authwrite' 0x1000000
echo foo > foo
tpm2_nvwrite -i foo 0x1000000
```

The above steps all work as I would expect.  However, I can't turn around and nvread:

```
$ tpm2_nvread -P pcr:sha1:0 0x1000000
WARN: Reading full size of the NV index
WARNING:esys:src/tss2-esys/api/Esys_NV_Read.c:315:Esys_NV_Read_Finish() Received TPM Error 
ERROR:esys:src/tss2-esys/api/Esys_NV_Read.c:105:Esys_NV_Read() Esys Finish ErrorCode (0x0000099d) 
ERROR: Esys_NV_Read(0x99D) - tpm:session(1):a policy check failed
ERROR: Failed to read NVRAM area at index 0x1000000
ERROR: Unable to run tpm2_nvread
```

The value of PCR 0 has not changed throughout the entire operation according to tpm2_pcrread (nor would I expect it to)

idesai:
Fixes #2875

Policy authorization session must be re-instantiated on multiple
iterations of nv read if the read size exceeds the allowed operating
size. To avoid this, only a single iteration is allowed when using
policy authorization session.

Re-instantiation is avoided because compounded policies are not
possible to be replayed to arrive at the same policy digest.

### DIFF
--- a/lib/tpm2_nv_util.h
+++ b/lib/tpm2_nv_util.h
@@ -228,7 +228,18 @@ static inline tool_rc tpm2_util_nv_read(ESYS_CONTEXT *ectx,
         "Make seperate tpm2_nvread calls with proper offsets specified to "
         "specify all NV index data to be read");
 
-        rc =-tool_rc_option_error;
+        rc = tool_rc_option_error;
+        goto out;
+    }
+
+    bool is_auth_a_policy_session = is_nvread_dispatched &&
+        tpm2_session_get_type(auth_hierarchy_obj->session) == TPM2_SE_POLICY;
+    if (is_auth_a_policy_session && is_nv_index_data_larger_than_max_read) {
+        LOG_ERR("Cannot continue as the policy auth session must be "
+                "reinstantiated for multiple iterations of NV read. "
+                "Specify a max read size of %lu or specify an offset and max "
+                "read size of %lu", max_data_size, max_data_size);
+        rc = tool_rc_option_error;
         goto out;
     }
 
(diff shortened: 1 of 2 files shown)


lib/tpm2_nv_util.h
```diff
@@ -228,7 +228,18 @@ static inline tool_rc tpm2_util_nv_read(ESYS_CONTEXT *ectx,
...
+    }
+
+    bool is_auth_a_policy_session = is_nvread_dispatched &&
+        tpm2_session_get_type(auth_hierarchy_obj->session) == TPM2_SE_POLICY;
+    if (is_auth_a_policy_session && is_nv_index_data_larger_than_max_read) {
+        LOG_ERR("Cannot continue as the policy auth session must be "
+                "reinstantiated for multiple iterations of NV read. "
+                "Specify a max read size of TPM2_PT_NV_BUFFER_MAX or specify "
+                "an offset and max read size of TPM2_PT_NV_BUFFER_MAX");
+        rc = tool_rc_option_error;
```
williamcroberts, line 242: Why not print out the actual number the user needs, which IIUC, is: max_data_size
So:
```C
    if (is_auth_a_policy_session && is_nv_index_data_larger_than_max_read) {
        LOG_ERR("Cannot continue as the policy auth session must be "
                "reinstantiated for multiple iterations of NV read. "
                "Specify a max read size of %lu or specify "
                "an offset and max read size of %lu", max_data_size, max_data_size);
```

williamcroberts (changes requested):

PR #2890, opened 2021-11-25
Fix argument parsing in tpm2_policylocality

trchen1033:
This patch fixes a bug that caused tpm2_policylocality to almost
always generate PolicyLocality(0).

There was a logical inversion that caused almost any argument
(including invalid ones) to be interpreted as zero, except "zero"
would be interpreted as one.

### DIFF
--- a/tools/tpm2_policylocality.c
+++ b/tools/tpm2_policylocality.c
@@ -54,15 +54,15 @@ static bool on_arg(int argc, char **argv) {
         return false;
     }
 
-    if (strcmp(argv[0], "zero")) {
+    if (strcmp(argv[0], "zero") == 0) {
         ctx.locality = TPMA_LOCALITY_TPM2_LOC_ZERO;
-    } else if (strcmp(argv[0], "one")) {
+    } else if (strcmp(argv[0], "one") == 0) {
         ctx.locality = TPMA_LOCALITY_TPM2_LOC_ONE;
-    } else if (strcmp(argv[0], "two")) {
+    } else if (strcmp(argv[0], "two") == 0) {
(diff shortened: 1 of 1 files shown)


osresearch:
It might be worth looking for other uses of `strcmp()` that have the same pattern, such as https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_options.c#L527 (which I think is correct, but might be easier to read if it had an explicit `!= 0`):

```
if (tcti_conf_option && strcmp(tcti_conf_option, "none"))
```

idesai:
Additionally per the coding guidelines we shouldn't be hiding a return from a function. Thanks for these finds.

trchen1033:
Yep I wrote in this way to be consistent with other part of the codebase, and I'm used to the `!strcmp()` idiom. However I second that `strcmp() == 0` probably easier to read, as this bug is an evidence that treating `strcmp()` as bool could be error prone.

I changed this PR to `strcmp() == 0` if that's the direction this project moving towards. I won't touch other part of the codebase for now. I could make a separate patch if needed.

> It might be worth looking for other uses of `strcmp()` that have the same pattern, such as https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_options.c#L527 (which I think is correct, but might be easier to read if it had an explicit `!= 0`):
> 
> ```
> if (tcti_conf_option && strcmp(tcti_conf_option, "none"))
> ```
> 
> also https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_options.c#L150 might clearer with `!= 0`:
> 
> ```
> if (strncmp(manpage, "tpm2_", strlen("tpm2_")))
> ```

PR #2893, opened 2021-12-02
ci: enable multi-arch builds

williamcroberts:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,32 @@ jobs:
       - name: failure
         if: ${{ failure() }}
         run: cat build/test-suite.log || true
+  multi-arch:
+    runs-on: ubuntu-latest
+    if: "!contains(github.ref, 'coverity_scan')"
+    strategy:
+      matrix:
+        DOCKER_IMAGE: [
+          "ubuntu-20.04.arm32v7",
+          "ubuntu-20.04.arm64v8",
+          "fedora-32.ppc64le"
+        ]
+    steps:
+      - name: Setup QEMU
+        run: |
+          sudo apt-get update
+          sudo apt-get install qemu binfmt-support qemu-user-static
+          docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Launch Container
+        env:
+          DOCKER_IMAGE: ${{ matrix.docker_image }}
+          CC: ${{ matrix.compiler }}
+        run: ./.ci/ci.run
+      - name: failure
+        if: ${{ failure() }}
+        run: cat build/test-suite.log || true
   coverage-test:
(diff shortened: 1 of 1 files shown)


williamcroberts:
timeout, we might need to do a more compact build/test run on emulated architectures.

idesai:
> timeout, we might need to do a more compact build/test run on emulated architectures.

@williamcroberts, should only a successful build suffice?

williamcroberts:
> > timeout, we might need to do a more compact build/test run on emulated architectures.
> 
> @williamcroberts, should only a successful build suffice?

yes

malikabhi05:
@williamcroberts @idesai does "make distcheck" [here](https://github.com/tpm2-software/tpm2-tools/blob/0f11c70f70fe846cef8fe7fed93a66f205fd0e9b/.ci/docker.run#L58) force the documentation to be generated? I was expecting it to not generate documentation if pandoc isn't present, however, it errors out saying `You do not have pandoc, a requirement for the distribution of manpages`.

Can I replace make distcheck and make distclean with make -jN for multiarch builds or does that cause material problems for the build? :)

williamcroberts:
> @williamcroberts @idesai does "make distcheck" [here](https://github.com/tpm2-software/tpm2-tools/blob/0f11c70f70fe846cef8fe7fed93a66f205fd0e9b/.ci/docker.run#L58) force the documentation to be generated? I was expecting the build to not generate documentation if pandoc isn't present, however, the build errors out saying `You do not have pandoc, a requirement for the distribution of manpages`.
> 
> Can I replace make distcheck and make distclean with make -jN for multiarch builds or does that cause material problems for the build? :)

Yes distcheck requires pandoc as it's a dist sub-command and to make the distribution tarball you need pandoc.

malikabhi05:
> Yes distcheck requires pandoc as it's a dist sub-command and to make the distribution tarball you need pandoc.

I'm removing pandoc from the build in order to speed it up as it is currently timing out. Please let me know if I can use make -jN instead of make distcheck for multiarch builds?

williamcroberts:
> > Yes distcheck requires pandoc as it's a dist sub-command and to make the distribution tarball you need pandoc.
> 
> I'm removing pandoc from the build in order to speed it up as it is currently timing out. Please let me know if I can use make -jN instead of make distcheck for multiarch builds?

Yeah that would work.

malikabhi05:
The build is now failing due to integration tests [here](https://github.com/malikabhi05/tpm2-tools/blob/1b940000536268059e46dd82e5d3e80f840a1481/.ci/docker.run#L95). I keep on seeing the error `Could not start simulator at port`. Do the integration tests have to be run for multiarch builds?

This is the [log](https://github.com/malikabhi05/tpm2-tools/runs/6782655709?check_suite_focus=true)

idesai:
Tracking this in #3110

PR #2898, opened 2021-12-14
tpm2_createek/tpm2_createak: Support alternative ECC curves

strongX509:
Newer TPM devices like the ST33HTPHAHD4 model from STMicroelectronics support **ECC NIST_P384** besides the default **ECC NIST_P256** curve.

This patch allows to create Endorsement Keys for all ECC curves for which default EK templates are defined by the **TCG EK Credential Profile Specification Version 2.1 Rev. 13** (NIST_P256, NIST_P384, NIST_P521, and SM2_P256). 

The attached file shows that the derived **ECC NIST_P384** EK key corresponds to the public key contained in the STMicroelectronics EK certificate:
[ECC_NIST_P384_EK.txt](https://github.com/tpm2-software/tpm2-tools/files/7712062/ECC_NIST_P384_EK.txt)

### DIFF
--- a/lib/tpm2_alg_util.h
+++ b/lib/tpm2_alg_util.h
@@ -46,6 +46,16 @@ TPM2_ALG_ID tpm2_alg_util_strtoalg(const char *name, tpm2_alg_util_flags flags);
 const char *tpm2_alg_util_algtostr(TPM2_ALG_ID id, tpm2_alg_util_flags flags);
 
 /**
+ * If the input is a numerical string then convert it to a nice-name.
+ * Otherwise return the unmodified nice-name.
+ * @param str
+ *  The numerical string to convert.
+ * @return
+ *  The nice-name.
+ */
+const char *tpm2_alg_util_numtoalgstr(const char* str, tpm2_alg_util_flags flags);
+
+/**
  * XXX DOC AND TESTME
  * @param id
  * @return
--- a/lib/tpm2_alg_util.h
+++ b/lib/tpm2_alg_util.h
@@ -185,8 +185,9 @@ tool_rc tpm2_alg_util_handle_rsa_ext_alg(const char *alg_spec,
  * @param public
  * @return
  */
-tool_rc tpm2_alg_util_public_init(char *alg_details, char *name_halg, char *attrs,
-        char *auth_policy,  TPMA_OBJECT def_attrs, TPM2B_PUBLIC *public);
+tool_rc tpm2_alg_util_public_init(const char *alg_details, const char *name_halg,
+        char *attrs, char *auth_policy,  TPMA_OBJECT def_attrs,
+        TPM2B_PUBLIC *public);
 
 /**
  * Returns an ECC curve as a friendly name.
--- a/man/tpm2_createek.1.md
+++ b/man/tpm2_createek.1.md
@@ -42,7 +42,8 @@ Refer to:
   * **-G**, **\--key-algorithm**=_ALGORITHM_:
 
     The endorsement key algorithm. Supports:
-    * **ecc** - An P256 key.
+    * **ecc** - A NIST_P256 key by default. Alternative curves can be selected
+      using algorithm specifiers (e.g. **ecc384** or **ecc_nist_p384**) .
     * **rsa** - An RSA2048 key.
     * **keyedhash** - hmac key.
 
@@ -81,9 +82,14 @@ the various known TCTI modules.
 
 # EXAMPLES
 
-### Create an Endorsement Key and make it persistent
+### Create an RSA Endorsement Key and make it persistent
 ```bash
 tpm2_createek -P abc123 -w abc123 -c 0x81010001 -G rsa -u ek.pub
+```
+
+### Create an ECC NIST_P384 Endorsement Key and make it persistent
+```bash
+tpm2_createek -G ecc384 -c 0x81010002
 ```
 
 ### Create a transient Endorsement Key, flush it, and reload it.
--- a/tools/tpm2_createak.c
+++ b/tools/tpm2_createak.c
@@ -407,46 +407,32 @@ out_session:
 
 static bool on_option(char key, char *value) {
 
-    TPM2_ALG_ID alg_id;
-
     switch (key) {
     case 'C':
         ctx.ek.ctx_arg = value;
         break;
     case 'G':
-        if (tpm2_util_string_to_uint16(value, &alg_id)) {
-            ctx.ak.in.alg.type = tpm2_alg_util_algtostr(alg_id,
-                            tpm2_alg_util_flags_base);
-            if (!ctx.ak.in.alg.type) {
-                LOG_ERR("Could not convert key algorithm, got \"%s\"", value);
-                return false;
-            }
-        } else {
-            ctx.ak.in.alg.type = value;
+        ctx.ak.in.alg.type = tpm2_alg_util_numtoalgstr(value,
+                                 tpm2_alg_util_flags_base);
+        if (!ctx.ak.in.alg.type) {
+            LOG_ERR("Could not convert key algorithm, got \"%s\"", value);
+            return false;
         }
         break;
     case 'g':
-        if (tpm2_util_string_to_uint16(value, &alg_id)) {
-            ctx.ak.in.alg.digest = tpm2_alg_util_algtostr(alg_id,
-                            tpm2_alg_util_flags_hash);
-            if (!ctx.ak.in.alg.digest) {
-                LOG_ERR("Could not convert digest algorithm, got \"%s\"", value);
-                return false;
-            }
-        } else {
-            ctx.ak.in.alg.digest = value;
+        ctx.ak.in.alg.digest = tpm2_alg_util_numtoalgstr(value,
+                                tpm2_alg_util_flags_hash);
+        if (!ctx.ak.in.alg.digest) {
+            LOG_ERR("Could not convert digest algorithm, got \"%s\"", value);
+            return false;
         }
         break;
     case 's':
-        if (tpm2_util_string_to_uint16(value, &alg_id)) {
-            ctx.ak.in.alg.sign = tpm2_alg_util_algtostr(alg_id,
-                            tpm2_alg_util_flags_sig);
-            if (!ctx.ak.in.alg.sign) {
-                LOG_ERR("Could not convert signing algorithm, got \"%s\"", value);
-                return false;
-            }
-        } else {
-            ctx.ak.in.alg.sign = value;
+         ctx.ak.in.alg.sign = tpm2_alg_util_numtoalgstr(value,
+                                tpm2_alg_util_flags_sig);
+        if (!ctx.ak.in.alg.sign) {
+            LOG_ERR("Could not convert signing algorithm, got \"%s\"", value);
+            return false;
         }
         break;
     case 'P':
--- a/tools/tpm2_createak.c
+++ b/tools/tpm2_createak.c
@@ -2,14 +2,85 @@
 
 #include <stdbool.h>
 #include <stdlib.h>
+#include <string.h>
 
 #include "files.h"
 #include "log.h"
 #include "object.h"
+#include "tpm2.h"
 #include "tpm2_alg_util.h"
 #include "tpm2_auth_util.h"
 #include "tpm2_convert.h"
 #include "tpm2_tool.h"
+
+#define ATTRS  \
+    TPMA_OBJECT_RESTRICTED|TPMA_OBJECT_USERWITHAUTH| \
+    TPMA_OBJECT_SIGN_ENCRYPT|TPMA_OBJECT_FIXEDTPM| \
+    TPMA_OBJECT_FIXEDPARENT|TPMA_OBJECT_SENSITIVEDATAORIGIN
+
+/* Templates from TCG EK Credential Profile TPM 2.0, Version 2.4 Rev. 3, 2021 */
+
+static const TPM2B_DIGEST policy_a_sha384 = {
+    .size = 48,
+    .buffer = {
+        0x8b, 0xbf, 0x22, 0x66, 0x53, 0x7c, 0x17, 0x1c, 0xb5, 0x6e,
+        0x40, 0x3c, 0x4d, 0xc1, 0xd4, 0xb6, 0x4f, 0x43, 0x26, 0x11,
+        0xdc, 0x38, 0x6e, 0x6f, 0x53, 0x20, 0x50, 0xc3, 0x27, 0x8c,
+        0x93, 0x0e, 0x14, 0x3e, 0x8b, 0xb1, 0x13, 0x38, 0x24, 0xcc,
+        0xb4, 0x31, 0x05, 0x38, 0x71, 0xc6, 0xdb, 0x53 }
+};
+
+static const TPM2B_DIGEST policy_a_sha512 = {
+    .size = 64,
+    .buffer = {
+        0x1e, 0x3b, 0x76, 0x50, 0x2c, 0x8a, 0x14, 0x25, 0xaa, 0x0b,
+        0x7b, 0x3f, 0xc6, 0x46, 0xa1, 0xb0, 0xfa, 0xe0, 0x63, 0xb0,
+        0x3b, 0x53, 0x68, 0xf9, 0xc4, 0xcd, 0xde, 0xca, 0xff, 0x08,
+        0x91, 0xdd, 0x68, 0x2b, 0xac, 0x1a, 0x85, 0xd4, 0xd8, 0x32,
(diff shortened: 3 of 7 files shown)


man/tpm2_createek.1.md
```diff
@@ -42,10 +42,15 @@ Refer to:
   * **-G**, **\--key-algorithm**=_ALGORITHM_:
 
     The endorsement key algorithm. Supports:
-    * **ecc** - An P256 key.
+    * **ecc** - A NIST_P256 key by default. Alternative ECC curves can be selected
+      with the **-e** option.
     * **rsa** - An RSA2048 key.
     * **keyedhash** - hmac key.
 
+  * **-e**, **\--ecc-curve**=_ECC_CURVE_ID_:
```
idesai, line 50: there is an [algorithm-parser ](https://github.com/tpm2-software/tpm2-tools/blob/master/lib/tpm2_alg_util.c#L313:#L346)for handling ECC curves so we don't have to add another option.

strongX509: Thanks for the hint. But how do can the non-NIST ECC curve **SM2_P256** be parsed for which an EK template exists? Can the algorithm-parser be extended?

idesai: @strongX509 yes you can extend the algorithm-parser as long as the curve is defined in the tss.  Looks like 25519 isn't yet added to the tss. @AndreasFuchsSIT do you know?

strongX509:
I refactored the code to use the algorithm-parser and extended it to support the **ecc_sm2_p256** non-NIST curve. In the future any new curves like **curve_25519** could be added easily.

Some unit tests for the ecc algorithm-parser could be added in `/test/unit/test_tpm2_alg_util.c` to increase code coverage.

tools/tpm2_createek.c
```diff
@@ -55,6 +55,7 @@ static createek_context ctx = {
     .objdata = TPM2_HIERARCHY_DATA_INIT,
     .flags = { 0 },
     .find_persistent_handle = false
+
```
williamcroberts, line 58: superfluous whitspace change, drop.

tools/tpm2_createek.c
```diff
@@ -178,6 +209,89 @@ static tool_rc set_ek_template(ESYS_CONTEXT *ectx, TPM2B_PUBLIC *input_public) {
...
+            0xC9, 0x27, 0xAC, 0x56, 0xC0, 0x32, 0xF3, 0xB3, 0xA6, 0x46,
+            0x2F, 0x8D, 0x03, 0x79, 0x98, 0xE7, 0xA1, 0x0F, 0x77, 0xFA,
+            0x45, 0x4A
+        }
+    };
+
+    TPM2B_DIGEST *authp = &ctx.objdata.in.public.publicArea.authPolicy;
+
+    switch (ctx.objdata.in.public.publicArea.nameAlg) {
+    case TPM2_ALG_SHA256:
```
williamcroberts, line 275: You need a falls-thru comment here to keep some compilers happy.

tools/tpm2_createek.c
```diff
@@ -244,15 +367,27 @@ static bool on_option(char key, char *value) {
...
         ctx.auth_ek.auth_str = value;
         break;
-    case 'G': {
-        TPMI_ALG_PUBLIC type = tpm2_alg_util_from_optarg(value,
-                tpm2_alg_util_flags_base);
-        if (type == TPM2_ALG_ERROR) {
-            LOG_ERR("Invalid key algorithm, got \"%s\"", value);
-            return false;
+    case 'G':
+       if (!strncmp("ecc", value, 3)) {
```
williamcroberts, line 371: Wouldn't it just be simpler to delay handling of this and set a field to string "rsa2048:null:aes128cfb." and override it here?

Then in the main flow you can just set up the object with a call to this and the _ext variant supports raw integer values IIRC.

tools/tpm2_createek.c
```diff
@@ -399,15 +507,6 @@ static tool_rc tpm2_tool_onrun(ESYS_CONTEXT *ectx, tpm2_option_flags flags) {
...
 
-    /* override the default attrs */
-    set_default_obj_attrs();
-
-    /* set the auth policy */
-    set_default_auth_policy();
-
-    /* set the default hierarchy */
-    set_default_hierarchy();
-
```
williamcroberts, line 410: @idesai went through some work items to make all the tools look more similar, I think this works against his end goal.

williamcroberts (changes requested):
I wouldn't reorder stuff in this change and I also think you can defer processing of `-G` and use `tpm2_alg_util_public_init` to handle the vast majority of the setup. Really note the `tpm2_alg_util_public_init`, that should be able to take in all the defaults and user options and give you what you expect, I think with the exception of the policy hash, so you need to determine default or supplied and set that yourself.

tools/tpm2_createek.c
idesai: This tool can take the patch as the changes haven't been made yet.

tools/tpm2_createek.c
strongX509: Dropped

tools/tpm2_createek.c
strongX509: I have never met such compilers but I dropped the line anyway.

tools/tpm2_createek.c
strongX509: I don't understand this approach about delayed handling of the parameters. probably because I'm not too familiar with your parsers. `tpm2_alg_util_handle_ext_alg` is not capable of handling raw integer values.

tools/tpm2_createek.c
strongX509: `set_default_hierarchy()` could keep its old location because it doesn't depend on the `nameAlg` settings.

tools/tpm2_createek.c
williamcroberts: https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7

tools/tpm2_createek.c
williamcroberts: Apparently I went to edit my comment and never hit submit and lost that tab.... anyways sorry for the confusion here.

This should really be around using tpm2_alg_util_public_init(). Where you can specify all the parts and get back a public as expected. Right now, the code manually populates all the fields, when this function will do the heavy lifting for you and tons of the code in the tool just go away.

tools/tpm2_createek.c
strongX509: Actually there is no fall-thru in the switch statement above. `TPM2B_DIGEST *authp = ...` is a pointer definition outside the switch statement.

tools/tpm2_createek.c
strongX509: The problem is that ` tpm2_alg_util_public_init(char *alg_details, char *name_halg, ...)` requires `name_halg` as an input argument because otherwise `nameAlg` would currently be set to the default `TPM2_ALG_SHA256`   in the populated `public` structure . Also I would need to know the symmetric encryption algorithm in advance in order to append it to the  `alg_details` input string. Thus I would either have to preparse `alg_details` first to determine all the input parameters needed by  `tpm2_alg_util_public_init()` or add a switch statement very similar to the existing `set_key_algorithm()` after the call so that not much is gained by calling  `tpm2_alg_util_public_init()` first.

The only way of saving code lines would be for `tpm2_alg_util_public_init()` to set reasonable defaults for nameAlg and symmetric encryption depending on the value of the key algorithm type.

williamcroberts: I think we can map the expected input string, like `ecc384` to what matches the templates:
1. the full specifier that works for `tpm2_alg_util_public_init`
2. The nameAlg per the template

So the vast majority of the code goes away and now you have something like:
```C
typedef struct alg_map alg_map;
struct alg_map {
   const char *input;
   const char *alg;
   const char *namealg;
   uint8_t *policy;
   const char *attributes;
};

static const uint8_t policy384[] = {
0xB2, 0x6E, 0x7D, 0x28, 0xD1, 0x1A,
0x50, 0xBC, 0x53, 0xD8, 0x82, 0xBC,
0xF5, 0xFD, 0x3A, 0x1A, 0x07, 0x41,
0x48, 0xBB, 0x35, 0xD3, 0xB4, 0xE4,
0xCB, 0x1C, 0x0A, 0xD9, 0xBD, 0xE4,
0x19, 0xCA, 0xCB, 0x47, 0xBA, 0x09,
0x69, 0x96, 0x46, 0x15, 0x0F, 0x9F,
0xC0, 0x00, 0xF3, 0xF8, 0x0E, 0x12
};

/*
 * See: https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_EKCredentialProfile_v2p4_r3.pdf
 */
alg_map algs[] = {
// from B.4.6 Template H-3: ECC NIST P384 (Storage) (I don;t know how 256 works since storage parents are required to 
// be aes128cfb for duplication IIUC).
{ "ecc384", "ecc384:aes256cfb", "sha384", policy384, "fixedtpm|fixedparent|sensitivedataorigin|userwithauth|adminwithpolicy|restricted|decrypt" },
...
};

alg_map m = lookup(userinput);
if (!m) {
 do_invalid_error();
}

TPM2B_PUBLIC public;
tpm2_alg_util_public_init(m->alg, m->namealg, m->attrs, NULL, def_attrs, &public);

/* set policy hash here */
memset(...);

....

create_primary()
```

strongX509:
ERROR: Unable to run createek
tpm2 createek -Q -c $Handle_ek_quote -G 0x01 on line 88 failed: 5

Hmm, tests are broken because numeric alg IDs are not supported any more.

strongX509:
Numeric alg ID support added.

strongX509:
**tpm2_createak** now also supports alternative ECC curves. Because of the new EK templates in the high range, an optional `PolicyOR()` function call had to be added to the session code.

strongX509:
@williamcroberts: Hi, what is the status of the pull request to extend `tpm2_createek` and `tpm2_createak` with the support of additional ECC curves?

tools/tpm2_createek.c
```diff
@@ -52,59 +108,81 @@ struct createek_context {
...
+        }
     }
+    return NULL;
+}
 
-    return true;
+static tool_rc init_ek_public(const char *key_alg, TPM2B_PUBLIC *public) {
+
+    const alg_map *m = lookup_alg_map(key_alg);
+
```
williamcroberts, line 163: no blank line here

tools/tpm2_createek.c
```diff
@@ -52,59 +108,81 @@ struct createek_context {
...
+    }
+
+    public->publicArea.authPolicy = *m->policy;
+
+    if (public->publicArea.type == TPM2_ALG_ECC &&
+        public->publicArea.parameters.eccDetail.curveID == TPM2_ECC_NIST_P256) {
+        public->publicArea.unique.ecc.x.size = 32;
+        public->publicArea.unique.ecc.y.size = 32;
+    } else  if (public->publicArea.type == TPM2_ALG_RSA &&
+                public->publicArea.parameters.rsaDetail.keyBits == 2048) {
```
williamcroberts, line 182: why only do we care on 2048 RSA and ECCP256, can't their be other keys?

williamcroberts (changes requested):
Looking much better. nothing large in the code review now.

tools/tpm2_createek.c
strongX509: This is a requirement in the `TCG EK Credential Profile` for `EK Templates in the Low Range` (B.3.3 Template L-1: RSA 2048) and B.3.4 Template L-2: ECC NIST P256). For the newer `EK Templates in the High Range` the `unique size` must be initialized to `0`.

lib/tpm2_alg_util.h
```diff
@@ -185,8 +185,9 @@ tool_rc tpm2_alg_util_handle_rsa_ext_alg(const char *alg_spec,
  * @param public
  * @return
  */
-tool_rc tpm2_alg_util_public_init(char *alg_details, char *name_halg, char *attrs,
-        char *auth_policy,  TPMA_OBJECT def_attrs, TPM2B_PUBLIC *public);
+tool_rc tpm2_alg_util_public_init(const char *alg_details, const char *name_halg,
+        char *attrs, char *auth_policy,  TPMA_OBJECT def_attrs,
+        TPM2B_PUBLIC *public);
```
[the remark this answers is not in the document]
strongX509: The new `struct alg_map` used for algorithm lookup consists of `const` strings for the `alg` and `namealg` fields that are used as input parameters the `tpm2_alg_util_public_init()` call. Changing `*alg_details` and `*name_alg` to `const char*` eliminates compiler warnings.

tools/tpm2_createek.c
strongX509: Ok

tools/tpm2_createak.c
```diff
@@ -192,23 +166,55 @@ static tool_rc create_ak(ESYS_CONTEXT *ectx) {
...
+    TPM2B_PUBLIC in_public;
+    TPMI_ALG_HASH ek_name_alg;
+    TPML_DIGEST pHashList = { .count = 2 };
 
-    bool result = set_key_algorithm(&in_public);
-    if (!result) {
-        return tool_rc_general_error;
+    /* get the nameAlg of the EK */
+    tool_rc tmp_rc = tpm2_readpublic(ectx, ctx.ek.ek_ctx.tr_handle, &ek_public,
+            NULL, NULL);
```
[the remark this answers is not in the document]
strongX509: `-g` is already assigned to specify the digest algorithm used for signing. There is no special option for name halg.

 `Part 3: Commands` of the `TCG Trusted Platform Module Library` defines in section `12.1 TPM2_Create`:
> If the object being created is a Storage Key and fixedParent is SET in the attributes of inPublic, then the
symmetric algorithms and parameters of inPublic are required to match those of the parent. The
algorithms that must match are inPublic.nameAlg, ...

Actually we are creating a Signature Key so that the requirement above does not need to fulfilled. But it would make sense that the nameAlg of the Attestation Key secured be the Endorsement Key would have the same strength as the parent. Otherwise a new option for setting the name halg would be needed for `tpm2_createak`

tools/tpm2_createak.c
```diff
@@ -386,32 +407,46 @@ static tool_rc create_ak(ESYS_CONTEXT *ectx) {
...
-        if (ctx.ak.in.alg.sign == TPM2_ALG_ERROR) {
-            LOG_ERR("Could not convert signing algorithm.");
-            return false;
+        if (tpm2_util_string_to_uint16(value, &alg_id)) {
+            ctx.ak.in.alg.sign = tpm2_alg_util_algtostr(alg_id,
+                            tpm2_alg_util_flags_sig);
+            if (!ctx.ak.in.alg.sign) {
+                LOG_ERR("Could not convert signing algorithm, got \"%s\"", value);
+                return false;
+            }
```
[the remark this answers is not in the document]
strongX509: Ok, can be done

tools/tpm2_createak.c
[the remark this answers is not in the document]
strongX509: The nameAlg of the EK is required for the selection of the matching EK Template policy hashes, so we need the readpublic call anyway.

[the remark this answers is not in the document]
williamcroberts: > `-g` is already assigned to specify the digest algorithm used for signing. There is no special option for name halg.
> 
Ahh this is createak not ek.

tools/tpm2_createek.c
williamcroberts: Ahh yeah, I remember that fun, I'm seeing pop back up over in the Python code.

williamcroberts:
@strongX509 I think we're down to two items and we will merge, the comments outstanding are:
1. "probably make a local helper function for this so we're not repeating these blocks everywhere."
2. "no blank line here"

Thanks for playing along and putting in a ton of work.

strongX509:
@williamcroberts Thanks for the review. I pushed the two requested changes.

tools/tpm2_createek.c
williamcroberts: @strongX509 did you get this? I see the helper function.

strongX509: @williamcroberts I removed the blank line in the tools/tpm2_createek: Eliminate whitespace patch

williamcroberts (approved):

williamcroberts:
> @strongX509 I am going to make a change as a separate change on top of this. It occurred to me, that we can use Esys_TR_GetName() to get the name of the EK and then use that to get the name alg. The name is a UINT16 of the alg number followed by the hash. Since this is an optimization, I won't require it in this patch series.
> 
> @idesai let me know if I can merge this, and if so, ill follow up with the patch.

FYI I have this change on a branch at:
https://github.com/williamcroberts/tpm2-tools/tree/strongX509/master

PR #2903, opened 2021-12-21
Fix pr2890 suggests

idesai:


idesai:
@osresearch, any thoughts on why the symlink test fails with this change?

tools/tpm2_tool.c
```diff
@@ -149,22 +149,71 @@ int main(int argc, char **argv) {
...
+        exit(tool_rc_success);
+    }
 
+    bool is_opt_str_help_no_man = (is_one_opt_specified &&
+        (strcmp(argv[1],"--help=no-man") == 0));
+
+    if (is_opt_str_help_no_man) {
+        tpm2_tool_output("Specify [ -v | --version ], [-h | --help] or one of "
+            "the following tool name with corresponding tool options:\n"
+            "[ certifyX509certutil ]\n[ checkquote ]\n[ eventlog ]\n[ print ]"
```
osresearch, line 174: This list seems like it will be difficult to maintain; there is a run-time generated list in `tpm2_tools.c` that could be traversed to generate it instead.

idesai: @osresearch, thanks I will take a look at that. Why is the symlink test failing though?

osresearch:
Looks like the failure was:
```
Expected 'tmp2_bad-command' to fail.
Expected 'bad-command' to fail.
```

I think the reason is that `tpm2_bad-command` will take this path In the new code:

```
    bool is_no_opts = (argc == 1);
....
    if (is_no_opts || is_opt_str_help) {
        char *options[2] = {"tpm2","--help=man"};
        tpm2_handle_options(2, options, 0, 0, 0);
        exit(tool_rc_success);
```

which results in an success exit code, so the negative test did not fail as expected.

idesai:
> Looks like the failure was:
> 
> ```
> Expected 'tmp2_bad-command' to fail.
> Expected 'bad-command' to fail.
> ```
> 
> I think the reason is that `tpm2_bad-command` will take this path In the new code:
> 
> ```
>     bool is_no_opts = (argc == 1);
> ....
>     if (is_no_opts || is_opt_str_help) {
>         char *options[2] = {"tpm2","--help=man"};
>         tpm2_handle_options(2, options, 0, 0, 0);
>         exit(tool_rc_success);
> ```
> 
> which results in an success exit code, so the negative test did not fail as expected.

Thanks!

tools/tpm2_tool.c
```diff
@@ -149,22 +149,45 @@ int main(int argc, char **argv) {
...
+        ((strcmp(argv[1],"--help") == 0) || (strcmp(argv[1],"-h") == 0) ||
+         (strcmp(argv[1],"--help=man") == 0)));
+
+    bool is_no_opts = (is_str_tpm2 && argc == 1);
+
+    if (is_no_opts || is_opt_str_help) {
+        char *options[2] = {"tpm2","--help=man"};
+        tpm2_handle_options(2, options, 0, 0, 0);
+        tool_rc rc = is_no_opts ? tool_rc_option_error : tool_rc_success;
+        exit(rc);
```
idesai, line 166: @osresearch this look good?

PR #2909, opened 2022-01-06
tools/tpm2_pcrevent: Improve handling of procfs (0 byte) files

### LINKED ISSUE #2895: tpm2_pcrevent - Improve handling of procfs (0 byte) files
Hey there,

I've added a small script to my initramfs to read `/proc/cmdline` and hash it to tpm.
After some playing around I've noticed that `tpm2_pcrevent` is not reading the contents of `/proc/cmdline` if provided as argument (`tpm2_pcrevent /proc/cmdline`).
Looking at the code shows that the file size is determined (which is 0 in case of procfs) and due to that no data is read at all.
To workaround that you have to pipe data proc `/proc/cmdline` to `tpm2_pcrevent` (`cat /proc/cmdline | tpm2_pcrevent`) or read data to a temporary file and let `tpm2_pcrevent` use this file for reading.

I don't think that this is very intuitional as this also produces a hash and not just a bunch of zeros.
So one (like me) wouldn't even recognize that something is not working as expected.

idesai:
Fixes #2895

### DIFF
--- a/tools/tpm2_pcrevent.c
+++ b/tools/tpm2_pcrevent.c
@@ -39,8 +39,7 @@ static tool_rc tpm_pcrevent_file(ESYS_CONTEXT *ectx,
     bool res = files_get_file_size(input, &file_size, NULL);
 
     /* If we can get the file size and its less than 1024, just do it in one hash invocation */
-    if (res && file_size <= BUFFER_SIZE(TPM2B_EVENT, buffer)) {
-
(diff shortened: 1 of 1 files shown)


williamcroberts:
So what exactly does this do now, cause us to emit an event of size 0?

idesai:
> So what exactly does this do now, cause us to emit an event of size 0?

It reads till EOF is reached.

williamcroberts:
> > So what exactly does this do now, cause us to emit an event of size 0?
> 
> It reads till EOF is reached.

What happens if size == 0?

PR #2914, opened 2022-01-12
tpm2_createak: use Esys_TR_GetName over ReadPublic

williamcroberts:
Rather than going to the TPM to get the name alg for the EK via a
TPM2_ReadPublic call, just use the name as cached by ESYS. The name of
an object is the name algorithm followed by the hash itself.

Not only is this faster, but prevents a MITM from lying to use about the
EK properties as the TPM2_Readpublic was not protected with a session.


williamcroberts:
@idesai you good with me merging this?

PR #2918, opened 2022-01-18
configure.ac: Fix automagic depency on libefivar

idesai:
Signed-off-by: Christopher Byrne <salah.coronya@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -62,8 +62,12 @@ PKG_CHECK_MODULES([CRYPTO], [libcrypto >= 1.1.0])
 PKG_CHECK_MODULES([CURL], [libcurl])
 
 # pretty print of devicepath if efivar library is present
-PKG_CHECK_MODULES([EFIVAR], [efivar],,[true])
-AC_CHECK_HEADERS([efivar/efivar.h])
+AC_ARG_WITH([efivar], AS_HELP_STRING([--without-efivar], [Build without efivar library (default: test)]))
+
+AS_IF([test "x$with_efivar" != "xno"], [
+	PKG_CHECK_MODULES([EFIVAR], [efivar])
+	AC_CHECK_HEADERS([efivar/efivar.h])
+])
 
 # backwards compat with older pkg-config
 # - pull in AC_DEFUN from pkg.m4


salahcoronya:
The way I wrote the patch, the default sense is to enable it unless it is explicitly disabled. Because efivars isn't in the container, and --disable-efivars is not passed, the build will fail. I could reverse the sense of the patch, in that case efivars would need to explicitly enable, but the checks would then succeed. 

Here's the patch in the reversed sense:

```
diff --git a/configure.ac b/configure.ac
index 0f8332af..415eb176 100644
--- a/configure.ac
+++ b/configure.ac
@@ -68,9 +68,9 @@ AC_CHECK_LIB(crypto, [EVP_sm4_cfb128], [
 PKG_CHECK_MODULES([CURL], [libcurl])
 
 # pretty print of devicepath if efivar library is present
-AC_ARG_WITH([efivar], AS_HELP_STRING([--without-efivar], [Build without efivar library (default: test)]))
+AC_ARG_WITH([efivar], AS_HELP_STRING([--with-efivar], [Build with efivar library (default: disabled)]))
 
-AS_IF([test "x$with_efivar" != "xno"], [
+AS_IF([test "x$with_efivar" = "xyes"], [
        PKG_CHECK_MODULES([EFIVAR], [efivar])
        AC_CHECK_HEADERS([efivar/efivar.h])
 ])

```

configure.ac
```diff
@@ -62,8 +62,12 @@ PKG_CHECK_MODULES([CRYPTO], [libcrypto >= 1.1.0])
...
 
 # pretty print of devicepath if efivar library is present
-PKG_CHECK_MODULES([EFIVAR], [efivar],,[true])
-AC_CHECK_HEADERS([efivar/efivar.h])
+AC_ARG_WITH([efivar], AS_HELP_STRING([--without-efivar], [Build without efivar library (default: test)]))
+
+AS_IF([test "x$with_efivar" != "xno"], [
+	PKG_CHECK_MODULES([EFIVAR], [efivar])
+	AC_CHECK_HEADERS([efivar/efivar.h])
+])
```
williamcroberts, line 70: shouldn't this be:
```m4
AS_IF([test "x$with_efivar" != "xtest"], [
	PKG_CHECK_MODULES([EFIVAR], [efivar],,[true])
	AC_CHECK_HEADERS([efivar/efivar.h])
],
[test "x$with_efivar" != "xyes"], [
	PKG_CHECK_MODULES([EFIVAR], [efivar],,[true])
	AC_CHECK_HEADERS([efivar/efivar.h])
])
```

Probably don't need the AC_CHECK_HEADERS, that can be dropped.

salahcoronya: The AC_CHECK_HEADERS can probably be dropped, but the point of the patch is to NOT autoprobe for efivar and include it if found, which the above does. See https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies.The way he original patch is setup the default is efivar is expected unless --without-efivar is stated. If efivar is not present, an error is generated, and the builder must either include it or specify --without-efivar to exclude it.

williamcroberts: Given my approach, which is autoprobe by default, but can be user overridden, Why can't Gentoo's build system add the configure flag `--without-efivar` when building. Then the build system is non-automagic for their build.

salahcoronya: The current build actually does include efivar, and with your changes above, if --with-efivar is passed, but the efivar library is not present, it just ignores and builds out, rather than error out. If efivar support is requested (either by default or explicitly) then ./configure process should error out if efivar isn't present .

williamcroberts: > The current build actually does include efivar, and with your changes above, if --with-efivar is passed, but the efivar library is not present, it just ignores and builds out, rather than error out. If efivar support is requested (either by default or explicitly) then ./configure process should error out if efivar isn't present .

That was the behavior I wanted, but that's a trivial fix. Ultimately the behavior should be:

| config options | efi present | result              |
|------------------- |----------------|--------------------|
| auto (default) | yes            | efi in build       |
| auto (default) | no             | efi not in build | 
| --with-efi        | yes            | efi in build       |
| --with-efi       | no              | fail configure   | 
| --without-efi  | yes            | efi not in build |
| --without-efi  | yes            | efi not in build |

salahcoronya: Yes, that works

williamcroberts:
See: https://github.com/tpm2-software/tpm2-tools/pull/2933

PR #2924, opened 2022-01-24
fapi: add new command to enable the use of fapi objects for tpm2 tools

JuergenReppSIT:
* The new command tss2_gettpm2object was added. With this command context
  files which can be used for tpm2 tool commands can be created.
* For persistent objects only the textual representation of the handle as
  hex number will be written and for keys a tpm2 tool context file will
  be written.
* Creation of a context file: tss2_gettpm2object -p /HS/SRK/mykey -c mykey.ctx
* Output of an NV index: tss2_gettpm2object -p /nv/Owner/mynv -c-
* An Integration test and a man page with examples is also added.

### DIFF
--- a/dist/bash-completion/tpm2-tools/tss2_gettpm2object
+++ b/dist/bash-completion/tpm2-tools/tss2_gettpm2object
@@ -1,0 +1,26 @@
+ bash completion for tss2_gettpm2object              -*- shell-script -*-
+
+_tss2_gettpm2object()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+    case $prev in
+        -!(-*)h | --help)
+            COMPREPLY=( $(compgen -W "man no-man" -- "$cur") )
+            return;;
+        -!(-*)[c] | --context)
+            _filedir
+            if [ x"$cur" = x ]; then COMPREPLY+=( '-' ); fi
+            return;;
+        -!(-*)[p] | --path)
+            return;;
+    esac
+
+    $split && return
+
+    COMPREPLY=( $(compgen -W "-h --help -v --version --force -f -p --path= --context= -c" -- "$cur") )
+    [[ $COMPREPLY == *= ]] && compopt -o nospace
+} &&
+complete -F _tss2_gettpm2object tss2_gettpm2object
+
+# ex: filetype=sh
--- a/tools/fapi/tss2_gettpm2object.c
+++ b/tools/fapi/tss2_gettpm2object.c
@@ -1,0 +1,184 @@
+/* SPDX-License-Identifier: BSD-3-Clause */
+
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <tss2/tss2_tctildr.h>
+#include <tss2/tss2_esys.h>
+#include <tss2/tss2_mu.h>
+
+#include "tools/fapi/tss2_template.h"
+#include "lib/tpm2.h"
+#include "lib/files.h"
+
+/* Context struct used to store passed commandline parameters */
+static struct cxt {
+    char const *data;
+    char const *path;
+    char const *tcti;
+    bool        overwrite;
+} ctx;
+
+/* Parse commandline parameters */
+static bool on_option(char key, char *value) {
+    switch (key) {
+    case 'c':
+        ctx.data = value;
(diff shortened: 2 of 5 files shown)


idesai:
@JuergenReppSIT would the tool name tss2_gettpm2object be more apt?

JuergenReppSIT:
@idesai yes that sounds better. I have changed the name.

tools/fapi/tss2_gettpm2object.c
```diff
@@ -0,0 +1,202 @@
...
+    if (strcmp(ctx.data, "-")) {
+        if (!ctx.overwrite) {
+            FILE *fp = fopen(ctx.data, "rb");
+            if (fp) {
+                fclose(fp);
+                LOG_ERR("Path: %s already exists. Please rename or delete the file!\n",
+                ctx.data);
+                goto error;
+            }
+        }
```
williamcroberts, line 98: isn't this a time of check time of use bug... You would really want to use the exclusive flag. I know all the libc's I've used support the "x" flag for that.

I don't even really know why we need this force option. If someone gives you an output path, just truncate and write it.

tools/fapi/tss2_template.c
```diff
@@ -563,7 +567,13 @@ int main(int argc, char *argv[]) {
      * Cleanup contexts & memory allocated for the modified argument vector
      * passed to execute_tool.
      */
-    Fapi_Finalize (&fctx);
+    if (strcmp(tool_name, "gettpm2object")) {
+        /*
+         * Fapi_Finalize was already executed by gettpm2object.
+         */
+        Fapi_Finalize (&fctx);
```
williamcroberts, line 574: Why do we have to change the flow for this tool? I would imagine it could still call finalize in the framework flow like normal.

dist/bash-completion/tpm2-tools/tss2_gettpm2object
```diff
@@ -0,0 +1,26 @@
...
+
+_tss2_gettpm2object()
+{
+    local cur prev words cword split
+    _init_completion -s || return
+    case $prev in
+        -!(-*)h | --help)
+            COMPREPLY=( $(compgen -W "man no-man" -- "$cur") )
+            return;;
+        -!(-*)[o] | --tpm2object)
```
AndreasFuchsSIT, line 11: How about using `-c | --key-context` or something similar to match the tpm2_* tools ?

tools/fapi/tss2_template.c
JuergenReppSIT: > Why do we have to change the flow for this tool? I would imagine it could still call finalize in the framework flow like normal.

The tool needed an esys context. Unfortunately it's not possible to get the esys context from the fapi context via fapi API.
With the socket interface I had problems to create the esys context. Therefore the fapi context was closed before calling Esys_Initialize.

tools/fapi/tss2_gettpm2object.c
JuergenReppSIT: > isn't this a time of check time of use bug... You would really want to use the exclusive flag. I know all the libc's I've used support the "x" flag for that.
> 
> I don't even really know why we need this force option. If someone gives you an output path, just truncate and write it.

I also do not know why we need this option and could remove it ;-)

dist/bash-completion/tpm2-tools/tss2_gettpm2object
JuergenReppSIT: yes I agree, that would be better.

tools/fapi/tss2_template.c
JuergenReppSIT: Sorry I forgot that we have Fapi_GetTcti, which can be used to create an esys context. I will change it.

JuergenReppSIT:
I did remove the ugly special handling in tss2_template.c and it works. Do all agree to use -c instead of -o and to remove --force?

williamcroberts:
> I did remove the ugly special handling in tss2_template.c and it works. Do all agree to use -c instead of -o and to remove --force?

+2 from me

idesai:
> @idesai yes that sounds better. I have changed the name, but I do not know the reason for the following fail?
> fatal: Couldn't find remote ref refs/tags/4.3.0

I have seen that show up sporadically, most times I simply rerun the CI.

PR #2963, opened 2022-03-09
tools/tpm2_tool.c: Fix an issue where LOG_WARN is always displayed

### LINKED ISSUE #2961: SAPI warning from makecredential
When using `tpm2 makecredential --tcti=none` there is always a warning and it does not appear possible to silence it:
```
WARN: Tool optionally uses SAPI. Continuing with tcti=none
```

Since this operation does not require a TPM, it seems like this warning is extraneous?

Added in https://github.com/tpm2-software/tpm2-tools/commit/f7cbbd0d4d81a771887966ffb59809fccf8ecca0

idesai:
Fixes #2961 

Despite setting the 'quiet' flag with -Q the warning messages were
always displayed.

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -469,12 +469,16 @@ tpm2_option_code tpm2_handle_options(int argc, char **argv,
 
         /* tool doesn't request a sapi, don't initialize one */
         if (flags->tcti_none && is_optional_sapi) {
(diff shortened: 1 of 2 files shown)


osresearch:
With `--quiet` this makes `makecredential` no longer warn about the TCTI and `-G`.  Thanks!

PR #2967, opened 2022-03-15
tools/tpm2_checkquote: Fix compilation failures  with openSSL 3

idesai:


gotthardp:
Please note that `OPENSSL_VERSION_MAJOR < 3` is an unusual construct because the `OPENSSL_VERSION_MAJOR` was first introduced in OpenSSL 3.0 (https://github.com/openssl/openssl/commit/3a63dbef15b62b121c5df8762f8cb915fb06b27a). It is not defined for OpenSSL versions prior 3.

PR #2975, opened 2022-04-08
Adding new event types to tpm2_eventlog

galmasi:
While looking at a boot event log from a recent AMD Milan system, I discovered that tpm2_eventlog does not implement the UEFI_PLATFORM_FIRMWARE_BLOB2 structure introduced in ~ 2020 by a TCG specification update.

This patch adds 3 new definitions -- UEFI_PLATFORM_FIRMWARE_BLOB2, EV_EFI_HANDOFF_TABLES2, EV_EFI_VARIABLE_BOOT2.

Correct handling of "blob2" is included in this patch. The handling of 'boot2' is the same as the old 'boot'; added an entry in the switch statement, but have not tested.

Reference: https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v22_02dec2020.pdf


galmasi:
I'll close this PR and reopen it with signed commits. This always gets me for some reason.

PR #2976, opened 2022-04-08
Adding new event types to tpm2_eventlog

galmasi:
Patch set for handling of new event types in tpm2_eventlog: EV_EFI_PLATFORM_FIRMWARE_BLOB2, EV_EFI_HANDOFF_TABLES2, EV_EFI_VARIABLE_BOOT2

Found the need for handling newer entry types while looking at the UEFI boot event logs on recent HW acquisitions in IBM cloud.

Reference: https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v22_02dec2020.pdf

### DIFF
--- a/lib/tpm2_eventlog_yaml.c
+++ b/lib/tpm2_eventlog_yaml.c
@@ -536,7 +536,7 @@ bool yaml_uefi_platfwblob(UEFI_PLATFORM_FIRMWARE_BLOB *data) {
 /* TCG PC Client PFP (02 dec 2020) section 10.2.5 */
 bool yaml_uefi_platfwblob2(UEFI_PLATFORM_FIRMWARE_BLOB2 *data) {
   UINT8 blobdescsize = data->BlobDescriptionSize;
-  UEFI_PLATFORM_FIRMWARE_BLOB * data2 = (UEFI_PLATFORM_FIRMWARE_BLOB *)((UINT8 *)data + sizeof(UINT8) + blobdescsize);
+  UEFI_PLATFORM_FIRMWARE_BLOB * data2 = (UEFI_PLATFORM_FIRMWARE_BLOB *)((UINT8 *)data + sizeof(data->BlobDescriptionSize) + blobdescsize);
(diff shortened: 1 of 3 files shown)


galmasi:
Reference update: it was pointed out to me that the reference in the PR comment is a draft.
The published version is

https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf

This change is not impacting any of the code in the patch.

lib/tpm2_eventlog_yaml.c
```diff
@@ -526,6 +532,37 @@ bool yaml_uefi_platfwblob(UEFI_PLATFORM_FIRMWARE_BLOB *data) {
                      data->BlobLength);
     return true;
 }
+
+/* TCG PC Client PFP (02 dec 2020) section 10.2.5 */
+bool yaml_uefi_platfwblob2(UEFI_PLATFORM_FIRMWARE_BLOB2 *data) {
+  UINT8 blobdescsize = data->BlobDescriptionSize;
+  UEFI_PLATFORM_FIRMWARE_BLOB * data2 = (UEFI_PLATFORM_FIRMWARE_BLOB *)((UINT8 *)data + sizeof(UINT8) + blobdescsize);
```
williamcroberts, line 539: UEFI_PLATFORM_FIRMWARE_BLOB * data2 = (UEFI_PLATFORM_FIRMWARE_BLOB *)((UINT8 *)data + sizeof(data->BlobDescriptionSize + blobdescsize);

williamcroberts (changes requested):
minor nit

lib/tpm2_eventlog_yaml.c
galmasi: Hi @williamcroberts . I had to look carefully to figure out what you mean, but I get it -- replacing the sizeof(UINT8) with the sizeof of the data member. I will make that change right away.

PR #3008, opened 2022-06-09
tpm2_eventlog: fix read loop for files bigger than one chunk

mgerstner:
The read loop is not correctly accumulating the total file size which
will result in erroneous behaviour should a file larger than one chunk
be encountered.

While fixing this also remove redundancy in the allocation logic and
avoid an extraneous chunk allocation even if the EOF has already been
reached.

Signed-off-by: Matthias Gerstner <matthias.gerstner@suse.de>


mgerstner:
Nevermind, the code is actually correct, only the interface of the `files_read_bytes_chunk()` function is a bit uncommon in that it accumulates the bytes in the output variable.

PR #3009, opened 2022-06-09
fix about type

sephiroth6:
error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘UINT64’ {aka ‘long long unsigned int’}


sephiroth6 (commented):
error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 2 has type ‘UINT64’ {aka ‘long long unsigned int’}

Signed-off-by: Angelo Giovinazzo [sephiroth.a17@gmail.com](https://github.com/tpm2-software/tpm2-tools/pull/sephiroth.a17@gmail.com)

PR #3024, opened 2022-06-24
tools/tpm2_ecephemeral: add support for calculating cpHash

ccihula:


idesai:
@ccihula  this commit must be signed and the conflict must be resolved.

PR #3034, opened 2022-07-07
tools/tpm2_flushcontext: add support for calculating cpHash

idesai:
Signed-off-by: cihula <connor.cihula@intel.com>

### DIFF
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -352,22 +352,54 @@ tool_rc tpm2_flush_context(ESYS_CONTEXT *esys_context, ESYS_TR flush_handle,
             return rc;
         }
 
+        TPM2_HANDLE sapi_flush_handle = 0;
+        rval = Esys_TR_GetTpmHandle(esys_context, flush_handle,
+            &sapi_flush_handle);
+        if (rval != TPM2_RC_SUCCESS) {
+            LOG_ERR("Failed to acquire SAPI handle");
+            return tool_rc_general_error;
+        }
+
         TSS2_RC rval = Tss2_Sys_FlushContext_Prepare(
-        sys_context, flush_handle);
+        sys_context, sapi_flush_handle);
         if (rval != TPM2_RC_SUCCESS) {
             LOG_PERR(Tss2_Sys_FlushContext_Prepare, rval);
             return tool_rc_general_error;
         }
 
-        TPM2B_NAME *name1 = 0;
-        rc = tpm2_tr_get_name(esys_context, flush_handle, &name1);
-        if (rc != tool_rc_success) {
-            LOG_ERR("Failed to acquire SAPI handle");
+        /*
+         * There is a bug in SAPI where in the flush handle is placed in the
+         * handle area instead of the parameter area.
+         * Ref: https://github.com/tpm2-software/tpm2-tss/issues/2382
+         *
+         * We determine this scenario by reading the parameter size in the
+         * cpBuffer which is returned as zero due to the bug. 
+         *
+         * When calculating the cpHash, the workaround for this scenario is to
+         * provide the flush handle as a name.
+         */
+        const uint8_t *command_parameters;
+        size_t command_parameters_size;
+        rval = Tss2_Sys_GetCpBuffer(sys_context, &command_parameters_size,
+            &command_parameters);
+        if (rval != TPM2_RC_SUCCESS) {
+            LOG_PERR(Tss2_Sys_GetCpBuffer, rval);
             return tool_rc_general_error;
         }
 
-        rc = tpm2_sapi_getcphash(sys_context, name1, NULL, NULL,
-            parameter_hash_algorithm, cp_hash);
+        TPM2B_NAME name1 = { 0 };
+        if (!command_parameters_size) {
+            name1.size = sizeof(TPM2_HANDLE);
+            rval = Tss2_MU_TPM2_HANDLE_Marshal(sapi_flush_handle, name1.name,
+                name1.size, 0);
+            if (rval != TPM2_RC_SUCCESS) {
+                LOG_ERR("Failed to populate SAPI handle");
+                return tool_rc_general_error;
+            }
+        }
+
+        rc = tpm2_sapi_getcphash(sys_context, name1.size ? &name1 : NULL, NULL,
+            NULL, parameter_hash_algorithm, cp_hash);
         /*
          * Exit here without making the ESYS call since we just need the cpHash
          */
--- a/lib/tpm2.c
+++ b/lib/tpm2.c
@@ -336,15 +336,52 @@ tool_rc tpm2_context_load(ESYS_CONTEXT *esys_context,
     return tool_rc_success;
 }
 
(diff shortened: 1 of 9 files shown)


lib/tpm2.c
```diff
@@ -336,15 +336,75 @@ tool_rc tpm2_context_load(ESYS_CONTEXT *esys_context,
...
+        /*
+         * There is a bug in SAPI where in the flush handle is placed in the
+         * handle area instead of the parameter area.
+         * Ref: https://github.com/tpm2-software/tpm2-tss/issues/2382
+         *
+         * We determine this scenario by reading the parameter size in the
+         * cpBuffer which is returned as zero due to the bug. 
+         *
+         * When calculating the cpHash, the workaround for this scenario is to
+         * provide the flush handle name.
```
williamcroberts, line 371: flush handle **as a** name

lib/tpm2.c
```diff
@@ -336,15 +336,75 @@ tool_rc tpm2_context_load(ESYS_CONTEXT *esys_context,
...
+        rval = Tss2_Sys_GetCpBuffer(sys_context, &command_parameters_size,
+            &command_parameters);
+        if (rval != TPM2_RC_SUCCESS) {
+            LOG_PERR(Tss2_Sys_GetCpBuffer, rval);
+            return tool_rc_general_error;
+        }
+
+        TPM2B_NAME *name1 = 0;
+        bool is_param_size_zero = (command_parameters_size == 0);
+        if (is_param_size_zero) {
```
williamcroberts, line 384: Their is no need to assign to a bool, a simple `if (command_parameters_size == 0)` or `(!command_parameters_size)` is sufficient.

lib/tpm2.c
```diff
@@ -336,15 +336,75 @@ tool_rc tpm2_context_load(ESYS_CONTEXT *esys_context,
...
+        const uint8_t *command_parameters;
+        size_t command_parameters_size;
+        rval = Tss2_Sys_GetCpBuffer(sys_context, &command_parameters_size,
+            &command_parameters);
+        if (rval != TPM2_RC_SUCCESS) {
+            LOG_PERR(Tss2_Sys_GetCpBuffer, rval);
+            return tool_rc_general_error;
+        }
+
+        TPM2B_NAME *name1 = 0;
```
williamcroberts, line 382: NULL, it's C standard.

lib/tpm2.c
```diff
@@ -336,15 +336,75 @@ tool_rc tpm2_context_load(ESYS_CONTEXT *esys_context,
...
+        bool is_param_size_zero = (command_parameters_size == 0);
+        if (is_param_size_zero) {
+            rc = tpm2_tr_get_name(esys_context, flush_handle, &name1);
```
williamcroberts, line 385: This may yield the wrong result. When no RM is present a transient handle is OK, which will then cause the name of the object versus the handle value to be encoded. Name's for session's and other objects just happen to be the same as the handle value and it just works out.

So you need to:
1. Call `Esys_TR_GetTpmHandle` to get the TPM handle going to the TPM.....
2. Use function `Tss2_MU_TPM2_HANDLE_Marshal` from `tss2_mu.h` to marshal the handle into the TPM2B_NAME structure.

tools/tpm2_flushcontext.c
```diff
@@ -79,6 +92,19 @@ static tool_rc process_outputs(ESYS_CONTEXT *ectx) {
     /*
      * 1. Outputs that do not require TPM2_CC_<command> dispatch
      */
+    bool is_file_op_success = true;
+    if (ctx.cp_hash_path) {
+        is_file_op_success = files_save_digest(&ctx.cp_hash, ctx.cp_hash_path);
+
```
williamcroberts, line 98: no newline here...

tools/tpm2_flushcontext.c
```diff
@@ -79,6 +92,19 @@ static tool_rc process_outputs(ESYS_CONTEXT *ectx) {
     /*
      * 1. Outputs that do not require TPM2_CC_<command> dispatch
      */
+    bool is_file_op_success = true;
+    if (ctx.cp_hash_path) {
+        is_file_op_success = files_save_digest(&ctx.cp_hash, ctx.cp_hash_path);
```
williamcroberts, line 97: This is only used in the scope of the if statement, so just do:
```C
if (ctx.cp_hash_path) {
  bool is_file_op_success = files_save_digest();
  if (!is_file_op_success) {
    return tool_rc_general_error;
  }
}
```
And delete line 95 and make it an empty line.

tools/tpm2_flushcontext.c
```diff
@@ -180,10 +206,22 @@ static tool_rc process_inputs(ESYS_CONTEXT *ectx) {
     /*
      * 4.a Determine pHash length and alg
      */
+    tpm2_session *all_sessions[MAX_SESSIONS] = {
+        0,
+        0,
+        0
+    };
```
williamcroberts, line 213: Isn't this the same as:
```C
tpm2_session *all_sessions[MAX_SESSIONS] = { 0 };
```

tools/tpm2_flushcontext.c
```diff
@@ -250,17 +288,21 @@ static bool on_option(char key, char *value) {
...
         ctx.property[ctx.encountered_option] = TPM2_TRANSIENT_FIRST;
+        ctx.encountered_option++;
         break;
     case 'l':
         ctx.property[ctx.encountered_option] = TPM2_LOADED_SESSION_FIRST;
+        ctx.encountered_option++;
         break;
     case 's':
         ctx.property[ctx.encountered_option] = TPM2_ACTIVE_SESSION_FIRST;
+        ctx.encountered_option++;
```
williamcroberts, line 299: This isn't a bug with this patch, but this needs bounds checking... ill submit a patch since I made the bug:
```bash
tpm2_flushcontext -tttttttt
Segmentation fault (core dumped)

**Review by williamcroberts — CHANGES_REQUESTED** [review]

File: tools/tpm2_flushcontext.c (hunk 8 above)
Comment (idesai) [reply tools/tpm2_flushcontext.c:299]: okay I will merge and rebase on that PR.

File: tools/tpm2_flushcontext.c [hunk 9]
```diff
@@ -253,13 +291,15 @@ static bool on_option(char key, char *value) {
         ctx.encountered_option_flags |= 1 << 0;
         break;
     case 'l':
-
         ctx.encountered_option_flags |= 1 << 1;
         ctx.encountered_option_flags++;
```
idesai, line 295: @williamcroberts is this a typo?

tools/tpm2_flushcontext.c
```diff
@@ -253,13 +291,15 @@ static bool on_option(char key, char *value) {
...
         break;
     case 's':
         ctx.encountered_option_flags |= 1 << 3;
```
idesai, line 298:
```suggestion
        ctx.encountered_option_flags |= 1 << 2;
```

tools/tpm2_flushcontext.c
williamcroberts: Yep, no more increment just set the flags

PR #3041, opened 2022-07-08
tests/getekcertificate.sh: Skip the test if curl is not present

aplanas:


idesai:
@aplanas, the tpm2_getekcertificate tool uses libcurl which may still be available.

aplanas:
> @aplanas, the tpm2_getekcertificate tool uses libcurl which may still be available.

Ah. The issue here is that we are executing the tests in a build system without internet access, so the current `curl` test seem natural to produce an `exit` to skip it.

PR #3047, opened 2022-07-13
test: add missing shut_down call on cleanup

### LINKED ISSUE #3042: Test cannot be executed in paralell
Seems that executing the tests in parallel (`make -j8 check`, for example) left some `tpm_server` instances running behind, making the tests to never end.

aplanas:
The tests "gettime", "readclock" and "symlink" are not calling
"shut_down" during the "cleanup" stage, making the TPM simulator to keep
their process during the "make check".  Somehow this produces problems
when the tests are executed in parallel under certain conditions, with
the effect of "make" not ending and waiting to those process to die.

This commit and the mising call in the "cleanup" stage.

Fix #3042


idesai:
@aplanas we had an issue with the CI. Can you please rebase to the current master?

PR #3056, opened 2022-07-14
tools/tpm2_pcrreset: redo tool's layout

### LINKED ISSUE #2430: tools/tpm2_pcrreset: Redo tool layout to standard format

idesai:
Fixes #2430

### DIFF
--- a/tools/tpm2_pcrreset.c
+++ b/tools/tpm2_pcrreset.c
@@ -10,30 +10,29 @@
 
 typedef struct tpm_pcr_reset_ctx tpm_pcr_reset_ctx;
 struct tpm_pcr_reset_ctx {
+    /*
+     * Inputs
+     */
     bool pcr_list[TPM2_MAX_PCRS];
+
+    /*
+     * Outputs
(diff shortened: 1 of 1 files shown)


tools/tpm2_pcrreset.c
```diff
@@ -10,71 +10,176 @@
...
+    /*
+     * 1. Outputs that do not require TPM2_CC_<command> dispatch
+     */
+
+    /*
+     * 2. Outputs generated after TPM2_CC_<command> dispatch
+     */
+
+    return tool_rc_success;
+}
```
williamcroberts, line 56: So here and in a few other places we have functions that get defined but just return success without doing anything, is it worth keeping these for future expansion in connor's cpHash work or is it better to not clutter the code and delete them?

idesai: Then there is rpHash, session support, auditing, tcti=none for cpHash. With the template it makes it a little easier to replicate the patches. In short term, it may seem a clutter for sure.

Issue #3063, opened 2022-07-18
Segmentation fault when running tpm2_policyor with empty argument


ptsolakos:

idesai:
Closed with #3064

PR #3064, opened 2022-07-18
tpm2_policyor: fix unallocated policy list

williamcroberts:
The TPML_DIGEST policy list was calloc'd for some reason, however it
could just be statically allocated in the context. The side effect is
that when no options or arguments were given a NPD occured when checking
the count of the policy list. TO fix this, just statically allocate it.

### DIFF
--- a/tools/tpm2_policyor.c
+++ b/tools/tpm2_policyor.c
@@ -14,7 +14,7 @@ struct tpm2_policyor_ctx {
     //File path for the session context data
     const char *session_path;
     //List of policy digests that will be compounded
-    TPML_DIGEST *policy_list;
+    TPML_DIGEST policy_list;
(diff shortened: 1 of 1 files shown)


williamcroberts:
Fixes: #3063

williamcroberts:
@idesai what should the behavior be if someone specifies option `-l` with the argument? Looks like parser always increments count. Is the parser code robust enough to handle this?

idesai:
> @idesai what should the behavior be if someone specifies option `-l` with the argument? Looks like parser always increments count. Is the parser code robust enough to handle this?

The -l option was retained for backwards compatibility. It appears that we allow for the policy list to be specified both ways and we do not check if both ways are specified simultaneously. We should definitely check for this scenario.

PR #3093, opened 2022-08-02
tpm2_eventlog; Handle EV_IPL event in PLCR 12 for sd-boot >= 251

berrange:
When using sd-boot,  tpm2-eventlog throws warnings:

WARN: Event 23 is unexpectedly not extending either PCR 8, 9, or 14
WARN: Event 24 is unexpectedly not extending either PCR 8, 9, or 14

Not only is the data put in PCR 12, but sd-boot doesn't add an unhashed prefix to the data, it hashes the entire event contents.

This pairs of commits addresses that problem


idesai:
@berrange  can you please add the eventlog binary [here](https://github.com/tpm2-software/tpm2-tools/tree/master/test/integration/fixtures) and or add a test.

berrange:
> @berrange can you please add the eventlog binary [here](https://github.com/tpm2-software/tpm2-tools/tree/master/test/integration/fixtures) and or add a test.

New version

 * adds an example eventlog binary, and al
 * extends the eventlog.sh test to compare against desired expected YAML output and stderr warnings
 * fixes the eventlog YAML generator to emit more spec compliant string data
 *

idesai:
@berrange, is there a test that uses the yaml files included in the PR?

berrange:
> @berrange, is there a test that uses the yaml files included in the PR?

Yes, look at the `test/integration/tests/eventlog.sh` changes in the commit [test: track expected YAML output for eventlog](https://github.com/tpm2-software/tpm2-tools/pull/3093/commits/6d6fdbae6b6f6068f68c5495297142ba041a866e) For each sample eventlog bin file, it generates YAML output and compares to the expected YAML stored in git.

PR #3096, opened 2022-08-02
tpm2_eventlog: clean up some magic numbers

snits:
Make the code a bit clearer by making it clear
we are subtracting the size of the EFI_GUID member
from the EFI_SIGNATURE_DATA size.


snits:
Bah, I apparently transposed DATA and SIGNATURE. Cleaned that up.

idesai:
> @idesai am I understanding the ci failure correctly that there was some infrastructure issue with ci, and it couldn't find something it needed so it failed?
Right. Perhaps with GitHub actions, I will investigate if it doesn't work in a couple of hours.

PR #3110, opened 2022-08-09
CI Multiarch

malikabhi05:
Fixes #3118
Fixes #3112

PR #3113, opened 2022-08-15
lib/object.c: Fix persistent parent load errors

### LINKED ISSUE #3112: tpm2_load: fails due to 64 bit openssl call
TPM2_LOAD fails while retrieving parent value from a TSS Privkey PEM file. Code calls to ASN1_INTEGER_GET_UNIT64 from the openssl API, but the function seems to not work correctly on a 32 bit system.

`
persistent-handle: 0x81100000
action: persisted
ERROR:esys:src/tss2-esys/esys_mu.c:813:iesys_MU_IESYS_RESOURCE_Unmarshal() Error unmarshaling subfield name ErrorCode (0x00090006) 
ERROR:esys:src/tss2-esys/esys_tr.c:96:Esys_TR_Deserialize() Unmarshal resource object ErrorCode (0x00090006) 
ERROR: Esys_TR_Deserialize(0x90006) - mu:A buffer isn't large enough
ERROR: Incorrect handle value, got: "mykey.pem", expected expected [o|p|e|n|l] or a handle number
ERROR: Unable to convert parent to integer/value too large
ERROR: Cannot make sense of object context "mykey.pem"
ERROR: Unable to run load
tpm2 load -r $pem_file.pem -c $pem_file.ctx on line 90 failed: 1
`

malikabhi05:
Fixes #3112

Signed-off-by: malikabhi05 <abhishek.malik@intel.com>

### DIFF
--- a/lib/object.c
+++ b/lib/object.c
@@ -196,14 +196,26 @@ static tool_rc tpm2_util_object_fetch_parent_from_tpk(const char *objectstr,
     tool_rc rc = tool_rc_general_error;
     BIO *input_bio = 0;
     TSSPRIVKEY_OBJ *tpk = 0;
+    BIGNUM *bn_parent = 0;
     tpm2_util_object_fetch_tssprivkey_from_file(objectstr, &input_bio, &tpk);
     if (!input_bio || !tpk) {
         goto ret;
     }
 
-    uint64_t parent;
-    int ret = ASN1_INTEGER_get_uint64(&parent, tpk->parent);
-    if (ret == 0 || parent > UINT32_MAX) {
+    bn_parent = ASN1_INTEGER_to_BN(tpk->parent, NULL);
+    if (!bn_parent) {
+        goto ret;
+    }
+
+    TPM2_HANDLE parent = 0;
+    bool is_bn_parent_negative = BN_is_negative(bn_parent);
+    if (is_bn_parent_negative) {
+        parent = ASN1_INTEGER_get(tpk->parent);
+    } else {
+        parent = BN_get_word(bn_parent);
+    }
+
+    if (parent == 0) {
         LOG_ERR("Unable to convert parent to integer/value too large");
         goto ret;
     }
@@ -211,6 +223,7 @@ static tool_rc tpm2_util_object_fetch_parent_from_tpk(const char *objectstr,
     rc = tool_rc_success;
 
 ret:
+    BN_free(bn_parent);
     if (input_bio) {
         BIO_free(input_bio);
     }
--- a/tools/misc/tpm2_encodeobject.c
+++ b/tools/misc/tpm2_encodeobject.c
@@ -127,6 +127,7 @@ static int encode(void) {
     tool_rc rc = tool_rc_general_error;
     BIO *bio = NULL;
     TSSPRIVKEY_OBJ *tpk = NULL;
+    BIGNUM *bn_parent = NULL;
     TSS2_RC rval = Tss2_MU_TPM2B_PRIVATE_Marshal(&ctx.object.private,
         private_buf, sizeof(private_buf), &private_len);
     if (rval != TPM2_RC_SUCCESS) {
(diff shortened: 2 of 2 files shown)


idesai:
@malikabhi05 there is an empty file added to this PR, please remove.

lib/object.c
```diff
@@ -201,13 +201,12 @@ static tool_rc tpm2_util_object_fetch_parent_from_tpk(const char *objectstr,
         goto ret;
     }
 
-    uint64_t parent;
-    int ret = ASN1_INTEGER_get_uint64(&parent, tpk->parent);
-    if (ret == 0 || parent > UINT32_MAX) {
+    TPM2_HANDLE ret = ASN1_INTEGER_get(tpk->parent);
```
williamcroberts, line 204: This returns a long, which is signed, which then get converted to the unsigned TPM2_HANDLE, which would be a weird handle. also the man page states that, "The deprecated function [ASN1_INTEGER_get](https://man.openbsd.org/ASN1_INTEGER_get.3#ASN1_INTEGER_get)() converts a to the long type."

I remember their being some issues with negative handles in tpm2-tss-engine code, see:
  - https://github.com/tpm2-software/tpm2-tss-engine/pull/222

It't important to note that they saved a pem file for testing that has the old negative handle in it to ensure backwards compatibility. We should do the same thing here, they use the BN routines and then BN to ASN1 and also check for negative if it's negative use ASN1_INTEGER_get and if positive use BN_get_word. We still probably want to use ASN1_INTEGER_get_int64 or something because of the deprecation notes in the man pages (may be OSSL version dependent).

williamcroberts (changes requested):

lib/object.c
idesai:
```suggestion
    BIGNUM *bn_parent =  ASN1_INTEGER_to_BN(tpk->parent, NULL);
    if (!bn_parent) {
         goto error;
    }
    
    TPM2_HANDLE ret = 0;
    bool is_bn_parent_negative = BN_is_negative(bn_parent);
    if (is_bn_parent_negative) {
         ret = ASN1_INTEGER_get(tpk->parent);
     } else {
         ret = BN_get_word(bn_parent);
     }
```

lib/object.c
```diff
@@ -196,21 +196,36 @@ static tool_rc tpm2_util_object_fetch_parent_from_tpk(const char *objectstr,
     tool_rc rc = tool_rc_general_error;
     BIO *input_bio = 0;
     TSSPRIVKEY_OBJ *tpk = 0;
+    BIGNUM *bn_parent = 0;
```
williamcroberts, line 199: NULL, @idesai knock off this 0 pointer non-sense :-p. stdlib.h defines the proper macro, "NULL" for "zeroing" of a pointer.

lib/object.c
```diff
@@ -196,21 +196,36 @@ static tool_rc tpm2_util_object_fetch_parent_from_tpk(const char *objectstr,
...
+
+    if (parent == 0) {
         LOG_ERR("Unable to convert parent to integer/value too large");
         goto ret;
     }
     *val = parent;
     rc = tool_rc_success;
 
 ret:
+    if (bn_parent) {
```
williamcroberts, line 226: you don't need this check
"BN_free() frees the components of the BIGNUM, and if it was created by BN_new(), also the structure itself. BN_clear_free() additionally overwrites the data before the memory is returned to the system. If a is NULL, nothing is done."

lib/object.c
```diff
@@ -196,21 +196,36 @@ static tool_rc tpm2_util_object_fetch_parent_from_tpk(const char *objectstr,
...
-    if (ret == 0 || parent > UINT32_MAX) {
+    bn_parent = ASN1_INTEGER_to_BN(tpk->parent, NULL);
+    if (!bn_parent) {
+        goto ret;
+    }
+
+    TPM2_HANDLE parent = 0;
+    bool is_bn_parent_negative = BN_is_negative(bn_parent);
+    if (is_bn_parent_negative) {
+        parent = ASN1_INTEGER_get(tpk->parent);
```
williamcroberts, line 213: I guess this is how the other code does it, wonder how it's going to fair with the deprecation mention, but the header files have yet to tag it with a deprecated attribute to cause a warning AFAICT.

williamcroberts (changes requested):
minor nits

lib/object.c
idesai: https://www.stroustrup.com/bs_faq2.html#null

idesai: > https://www.stroustrup.com/bs_faq2.html#null

@malikabhi05 you can use NULL, no problem :)

williamcroberts: > https://www.stroustrup.com/bs_faq2.html#null

Barne Stroustrup is the creator of C++ and his blog is for C++, we are C programmers. Therefore we follow the K&R Book on C: https://www.amazon.com/Programming-Language-2nd-Brian-Kernighan/dp/0131103628, which states on page 102:

"Pointers and Integers are not interchangeable. Zero is the sole exception: the constant zero may be assigned to a pointer, and a pointer may be compared with the constant zero. The symbolic constant NULL is often used in the place of zero , as a mnemonic to indicate more clearly that this is a special value for a pointer."

Now in Bourne's blog he mentions, "In pre-standard code, NULL was/is sometimes defined to something unsuitable and therefore had/has to be avoided. That's less common these **days**". I have been bit by that before where 0 was a valid and in use address (which made things like `if(!fooptr) { /* do this /* }` to be broken, so its very not wise to do that as porting software is a nightmare.

williamcroberts (approved):

williamcroberts (changes requested):
Oh wait, duh we need a test. So somewhere we generate a bad pem file as well.... Theirs two parts to this equation, generation and consumption. We need to ensure their is nothing we need to do at the generation spot, and we also should take a pem file that triggers this and add it to the test. I guess it would only get tripped on 32 bit builds, but still.

williamcroberts:
@malikabhi05 test coming?

idesai:
> @malikabhi05 test coming?

I think we need to add NVChip and a negative value pem file to our test fixtures for testing backwards compatibility.

idesai:
> > @malikabhi05 test coming?
> 
> 
> 
> I think we need to add NVChip and a negative value pem file to our test fixtures for testing backwards compatibility.

Or we could also write wrap functions.

idesai:
continuing this pr in #3110

Issue #2299, opened 2021-01-12
Avoid casts when dereferencing session structures in sessionconfig tool


idesai:

williamcroberts:
So it looks like you get all the information from an IESYS_METADATA via populate_iesys_session_info(). What we need to do is find a way to get the information you care about, which you currently dump it all out, in an upstream compatible way. You need to work with Andreas and if we need a new interface, the TCG TSS working group to build this interface. I only see you use the IESYS_METADATA to print the data, can we just remove this functionality and still have a working tool or do you need some of those outputs for something? The simplest way to deal with it, is to avoid it ;-p

PR #3119, opened 2022-09-02
tpm2_sessionconfig: print digest

williamcroberts:
Currently their is no way to get the session digest without running a
policy command.


idesai:
@williamcroberts as discussed in #2299 we were planning to keep sessionconfig for setting/reading attributes only, no?

williamcroberts:
> @williamcroberts as discussed in #2299 we were planning to keep sessionconfig for setting/reading attributes only, no?

I thought that discussion was just on IESYS_METADATA usage, get the session digest is fine IMO versus making a whole new tool for it

PR #3157, opened 2022-11-04
tpm2_startauthsession: no warn: if tpmkey transient

aospan:
Previous behavior produces misleading warning message in the following scenario:
$ tpm2_createprimary -c primary.ctx
$ tpm2_evictcontrol -C o -c primary.ctx 0x81010000 $ tpm2_flushcontext -t
$ tpm2_startauthsession --policy-session -c 0x81010000 -S mysession.ctx 
WARN: check public portion of the tpmkey manually

According to the TCG TPM 2.0 Spec only bind object checked: 11.1 TPM2_StartAuthSession:
 If bind references a transient object, then the TPM shall return
 TPM_RC_HANDLE if the sensitive portion of the object is not loaded.

User will receive TPM_RC_HANDLE error code from TPM chip immediately, no additional warning required.

Additional info:
tpmKey & bind fields checks on the TPM chip side (reference tpm_server from Ken Goldman): https://bit.ly/3fAnjRS
Only bind object checked for TPM_HT_TRANSIENT.

Signed-off-by: Abylay Ospan <aospan@amazon.com>

### DIFF
--- a/tools/tpm2_startauthsession.c
+++ b/tools/tpm2_startauthsession.c
@@ -266,16 +266,6 @@ static tool_rc process_input_data(ESYS_CONTEXT *ectx) {
         if (rc != tool_rc_success) {
             return rc;
         }
-
-        /* if loaded object is non-permanant, it should ideally be persistent */
-        if (ctx.session.tpmkey.key_context_object.handle) {
-
-            bool is_transient = (ctx.session.tpmkey.key_context_object.handle
-                    >> TPM2_HR_SHIFT) == TPM2_HT_TRANSIENT;
-            if (!is_transient) {
-                LOG_WARN("check public portion of the tpmkey manually");
-            }
-        }
     }
 
     /*


williamcroberts:
> $ tpm2_startauthsession --policy-session -c 0x81010000 -S mysession.ctx

NACK. That's not what that warning means.

So you're starting a session using `-c` which enables encrypted sessions. However, you started the session off of a persistent handle and the tools never verify that the persistent handle you specified is associated with the proper object. So, this means an attacker could be in control of the private key you established the session with.

To avoid this, use a serialized esys_tr which is both the handle and the name for verifying the public key. You can do something along the lines of:
```bash
tpm2_createprimary -c primary.ctx
tpm2_evictcontrol -c primary.ctx 0x81010000 -o primary.tr
tpm2_startauthsession --policy-session -c primary.tr -S mysession.ctx
```
Then no warning is issued.

I would take a patch to perhaps clarify the warning or to support a `-n` option where you can specify the name.

williamcroberts (changes requested):
This removes a functional and important warning, see my comment for details.

aospan:
William, thanks for the explanation! Now it's clear what this warning is about. 

But still not clear how adversaries can exploit this:
>So, this means an attacker could be in control of the private key you established the session with.
According to the TPM 2.0 spec:
24.1 TPM2_CreatePrimary
"The command will create and load a Primary Object. The sensitive area is not returned.
Since the sensitive data is not returned, the key cannot be reloaded. It can either be made
persistent or it can be recreated."

So, we're safe here to use persistent keys - no private portion of the key is exposed. Is this correct?

Thanks!

williamcroberts:
> William, thanks for the explanation! Now it's clear what this warning is about.
> 
> But still not clear how adversaries can exploit this:
> 
> > So, this means an attacker could be in control of the private key you established the session with.

Think of this like a normal https connection. If you don't verify the server's public key you only verify a encrypted session is established, not the entity it was established with. If you think of this like ssh, you wouldn't verify the remote's public key and add it to your trusted list.

> 
> According to the TPM 2.0 spec: 24.1 TPM2_CreatePrimary "The command will create and load a Primary Object. The sensitive area is not returned. Since the sensitive data is not returned, the key cannot be reloaded. It can either be made persistent or it can be recreated."
> 
> So, we're safe here to use persistent keys - no private portion of the key is exposed. Is this correct?

All keys are protected by the storage protections as discussed in the Architecture (part 1) of the TPM spec. No private portion of a key is exposed unless you duplicate the key in the clear and the attributes to allow duplication would need to be set. By default the tools create non-duplicable keys (fixedTPM and fixedParent).

However, you want to verify that the key at handle X is the key you expect to establish the connection with. If you don't verify the public portion of the key, an attacker could be interposing between the application and the TPM. The attacker could use whatever key they have of the same algorithm and the application would happily create that connection. Therefore to use an encrypted channel properly clients should always verify the remote key using the public key in some fashion. For https this is verifying the remote certificate up to a trusted root CA. For SSH this is adding the remote's public key fingerprint to your ssh/known_hosts file and for TPM this is verifying the public key using the object name as stored in an ESYS_TR serialized blob (-o output on evictcontrol). When you set this up you should verify the name is right.

> 
> Thanks!

williamcroberts:
> > William, thanks for the explanation! Now it's clear what this warning is about.
> > But still not clear how adversaries can exploit this:
> > > So, this means an attacker could be in control of the private key you established the session with.
> 
> Think of this like a normal https connection. If you don't verify the server's public key you only verify a encrypted session is established, not the entity it was established with. If you think of this like ssh, you wouldn't verify the remote's public key and add it to your trusted list.
> 
> > According to the TPM 2.0 spec: 24.1 TPM2_CreatePrimary "The command will create and load a Primary Object. The sensitive area is not returned. Since the sensitive data is not returned, the key cannot be reloaded. It can either be made persistent or it can be recreated."
> > So, we're safe here to use persistent keys - no private portion of the key is exposed. Is this correct?
> 
> All keys are protected by the storage protections as discussed in the Architecture (part 1) of the TPM spec. No private portion of a key is exposed unless you duplicate the key in the clear and the attributes to allow duplication would need to be set. By default the tools create non-duplicable keys (fixedTPM and fixedParent).
> 
> However, you want to verify that the key at handle X is the key you expect to establish the connection with. If you don't verify the public portion of the key, an attacker could be interposing between the application and the TPM. The attacker could use whatever key they have of the same algorithm and the application would happily create that connection. Therefore to use an encrypted channel properly clients should always verify the remote key using the public key in some fashion. For https this is verifying the remote certificate up to a trusted root CA. For SSH this is adding the remote's public key fingerprint to your ssh/known_hosts file and for TPM this is verifying the public key using the object name as stored in an ESYS_TR serialized blob (-o output on evictcontrol). When you set this up you should verify the name is right.

I added support to tpm2_print to output the details of an ESYS_TR handle if your curious to what's in there:
  - https://github.com/tpm2-software/tpm2-tools/pull/3160

I guess now we need a `name` option to startauthsession and if name is set, verify it and not print the warning.

> 
> > Thanks!

williamcroberts:
> I added support to tpm2_print to output the details of an ESYS_TR handle if your curious to what's in there:
> 
> * [tpm2_print: support printing serialized ESYS_TR's #3160](https://github.com/tpm2-software/tpm2-tools/pull/3160)
> 
> I guess now we need a `name` option to startauthsession and if name is set, verify it and not print the warning.
> 
I have added this support, so now tpm2_startauthsession can be provided a name file and check the name is as expected:
  - https://github.com/tpm2-software/tpm2-tools/pull/3161

williamcroberts:
So I think we can close this, we can continue any discussions but the needed support to ease this check is now on other PRs.

aospan:
I love your analogy with HTTPS. Thanks for the clarification and corresponding PRs!

Issue #2812, opened 2021-08-12
startauthsession needs a way to verify persistent objects


williamcroberts:
If you start an encrypted session with a persistent key their is no verification that it is the correct key. We should support a public key option that can take:
1. The name.
2. the public key as a pem or der and then verify that the public modulus or points match.

PR #3161, opened 2022-11-07
tpm2_startauthsession: VERIFY persistent handles

williamcroberts:
When an AuthSession is used with a tpmKey, that key MUST be verified in order to trust the connection. Think of this as analagous to known hosts with ssh. With Serialized ESYS_TR's or Context Files, the name is associated with that handle and ESAPI is instrumented to perform all the required name checks. When loaded from a persistent handle, ESAPI does not have the associated metadata and reads the name from the handle with a ReadPublic call. Becuase of this, an attacker COULD MiTM the connection with their own key. Thus, with persistent handles we must verify the name from some other source. The TPM name could be thought of as an ssh key fingerprint.

To fix this, provide a -n argument for a name file as output by readpublic or could be generated as a binary array of the objects name. When a tpmKey is provided, check the name and fail.

Fixes: #2812

### DIFF
--- a/test/integration/tests/abrmd_extended-sessions.sh
+++ b/test/integration/tests/abrmd_extended-sessions.sh
@@ -132,4 +132,34 @@ tpm2 sessionconfig session.ctx --enable-encrypt --enable-decrypt
 tpm2 getrandom 8 -S session.ctx
 tpm2 flushcontext session.ctx
 
+# test that name verification works
+tpm2 evictcontrol -c prim.ctx 0x81000000
+tpm2 readpublic -c prim.ctx -n name.bin
+tpm2 startauthsession -S session.ctx --policy-session \
+  --tpmkey-context 0x81000000 -n name.bin
+# verify -c works
+tpm2 startauthsession -S session.ctx --policy-session \
+  -c 0x81000000 -n name.bin
+
+# test if name is bad, generate a random name of of the same size
+# Note FreeBSD stat has no printf and wc prints spaces hence the sed.
+size=$(wc -c < name.bin | sed 's/ //g')
+dd if=/dev/urandom bs=${size} count=1 of=name2.bin
+
+if tpm2 startauthsession -S session.ctx --policy-session \
+  --tpmkey-context 0x81000000 -n name2.bin 2>/dev/null; then
+    echo "Expected tpm2 startauthsession to fail with bad name"
+    exit 1
+else
+    true
+fi
+
+if tpm2 startauthsession -S session.ctx --policy-session \
+  -c 0x81000000 -n name2.bin 2>/dev/null; then
+    echo "Expected tpm2 startauthsession to fail with bad name"
+    exit 1
+else
+    true
+fi
+
 exit 0
--- a/tools/tpm2_startauthsession.c
+++ b/tools/tpm2_startauthsession.c
@@ -1,7 +1,9 @@
 /* SPDX-License-Identifier: BSD-3-Clause */
 
 #include <stdbool.h>
+#include <string.h>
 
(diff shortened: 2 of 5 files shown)


tools/tpm2_startauthsession.c
```diff
@@ -272,9 +292,34 @@ static tool_rc process_input_data(ESYS_CONTEXT *ectx) {
...
+                rc = tpm2_tr_get_name(ectx, ctx.session.tpmkey.key_context_object.tr_handle,
+                        &got_name);
+                if (rc != tool_rc_success) {
+                    Esys_Free(got_name);
+                    return rc;
+                }
+
+                if (ctx.name.size != got_name->size ||
+                        memcmp(got_name->name, ctx.name.name, ctx.name.size)) {
+                    LOG_ERR("Expected name does not match");
```
idesai, line 316:
```suggestion
                bool is_expected_object = (memcmp(got_name->name, ctx.name.name, ctx.name.size) == 0);
                if (ctx.name.size != got_name->size || !is_expected_object) {
                    LOG_ERR("Expected name does not match");
```

williamcroberts: > 
You MUST check size before you do memcmp. Size should also be bounded by the buffer size, I wonder how many of those bugs we have floating around. I'll write a compare TPM2B function.

test/integration/tests/abrmd_extended-sessions.sh
```diff
@@ -132,4 +132,33 @@ tpm2 sessionconfig session.ctx --enable-encrypt --enable-decrypt
...
+tpm2 evictcontrol -c prim.ctx 0x81000000
+tpm2 readpublic -c prim.ctx -n name.bin
+tpm2 startauthsession -S session.ctx --policy-session \
+  --tpmkey-context 0x81000000 -n name.bin
+# verify -c works
+tpm2 startauthsession -S session.ctx --policy-session \
+  -c 0x81000000 -n name.bin
+
+# test if name is bad, generate a random name of of the same size
+size=$(stat --printf="%s" name.bin)
```
idesai, line 145: @williamcroberts, I am not sure why cirrus-ci doesn't like what appears to be a valid command line option for the stat tool.
```
stat: illegal option -- -
usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file|handle ...]
size=$(stat --printf="%s" name.bin) on line 145 failed: 1
```

williamcroberts: > @williamcroberts, I am not sure why cirrus-ci doesn't like what appears to be a valid command line option for the stat tool.
> 
> ```
> stat: illegal option -- -
> usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file|handle ...]
> size=$(stat --printf="%s" name.bin) on line 145 failed: 1
> ```

Ha yeah, bsd doesn't have all the gnu extensions on tools.

williamcroberts:
Dang:
```
0000: 00000000                          ....
dd: no value specified for bs
dd if=/dev/urandom bs=${size} count=1 of=name2.bin on line 146 failed: 1
Shutting down
```

williamcroberts:
> Dang:
> 
> ```
> 0000: 00000000                          ....
> dd: no value specified for bs
> dd if=/dev/urandom bs=${size} count=1 of=name2.bin on line 146 failed: 1
> Shutting down
> ```

Looks like bsd wc -c has spaces in the output I need to strip. dd bs=32 is not the same as dd= 32 because the shell passes it as 1 arg of key-value-pair vs 2 args as the space is a delimiter for argv parsing by the shell.

PR #3162, opened 2022-11-07
tpm2_getekcertificate: drop redundant function call

williamcroberts:
print_intel_ek_certificate_warning is already called in process_input, no need to call it on this path as well.

@idesai I think we can drop this but I am not 100% sure, can you look?

### DIFF
--- a/tools/tpm2_getekcertificate.c
+++ b/tools/tpm2_getekcertificate.c
@@ -526,12 +526,6 @@ static tool_rc get_ek_certificates(ESYS_CONTEXT *ectx) {
     /*
      * Following is everything applicable to ctx.is_cert_on_nv = false.
      */
-
-    rc = print_intel_ek_certificate_warning();
-    if (rc != tool_rc_success) {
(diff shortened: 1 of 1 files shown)


idesai:
> print_intel_ek_certificate_warning is already called in process_input, no need to call it on this path as well.
> 
> @idesai I think we can drop this but I am not 100% sure, can you look?

nack - still need  the warning text to point users to revocation software when EK is revoked.

williamcroberts:
> > print_intel_ek_certificate_warning is already called in process_input, no need to call it on this path as well.
> > @idesai I think we can drop this but I am not 100% sure, can you look?
> 
> nack - still need the warning text to point users to revocation software when EK is revoked.

Im removing a duplicate call so it should still print the error. Function `print_intel_ek_certificate_warning` is called in process_inputs, why do we need to print it twice? All the state that decides if the print occurs occurs in the beginning of process_inputs and print_intel_ek_certificate_warning is a tail call in that function.

PR #3167, opened 2022-11-16
tpm2_getcap: support vendor properties

williamcroberts:
Requires tpm2-tss 4.0

Support vendor properties as a vendor agnostic TPM2B buffer. This also introduces the -i or --ignore-more option to ignore the moreData byte returned from the TPM.

Introduce -i to ignore moreData when it's set to 1 but moreData does not follow.

Example output from my NUC:
./tools/tpm2 getcap -i --tcti=pcap:device:/dev/tpmrm0 vendor:1 0000000100000003

### DIFF
--- a/tools/tpm2_getcap.c
+++ b/tools/tpm2_getcap.c
@@ -123,6 +123,14 @@ capability_map_entry_t capability_map[] = {
         .property          = TPM2_ACTIVE_SESSION_FIRST,
         .count             = TPM2_MAX_CAP_HANDLES,
     },
+#if defined(ESYS_4_0)
+    {
+        .capability_string = "vendor",
+        .capability        = TPM2_CAP_VENDOR_PROPERTY,
+        .property          = 1,
+        .count             = TPM2_MAX_CAP_BUFFER,
+    },
+#endif
 };
 /*
  * Structure to hold options for this tool.
@@ -133,6 +141,7 @@ typedef struct capability_opts {
     UINT32 property;
     UINT32 count;
     bool list;
+    bool ignore_moredata;
 } capability_opts_t;
 
 static capability_opts_t options;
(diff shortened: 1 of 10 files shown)


williamcroberts:
We should build a TCTI that takes data from stdin and writes to stdout so we can echo in a fake response buffer. I wonder if we can do this in bash and use the subprocess TCTI.

tools/tpm2_getcap.c
```diff
@@ -801,6 +834,9 @@ static bool on_option(char key, char *value) {
     switch (key) {
     case 'l':
         options.list = true;
+        break;
+    case 'i':
+        options.ignore_more = true;
```
idesai, line 839: Elsewhere in all the tools `-i` is regarded as input. So maybe drop the short option.

tools/tpm2_getcap.c
```diff
@@ -820,11 +856,12 @@ static bool on_arg(int argc, char *argv[]) {
 
 static bool tpm2_tool_onstart(tpm2_options **opts) {
 
-    const struct option topts[] = { { "list", no_argument, NULL, 'l' },
-
+    const struct option topts[] = {
+        { "list",        no_argument, NULL, 'l' },
+        { "ignore-more", no_argument, NULL, 'i' },
```
idesai, line 861:
```suggestion
        { "ignore-moredata", no_argument, NULL, 'i' },
```

tools/tpm2_getcap.c
williamcroberts: I felt dirty about using -i, but it didn't even occur to me to just keep the long option.

williamcroberts:
Nooooo, I forgot to submit the test tcti and now I am afraid I lost it :-(

williamcroberts:
> Nooooo, I forgot to submit the test tcti and now I am afraid I lost it :-(

Thank you eclipse local history feature

williamcroberts:
Pulling becuase tpm2-pytss CI is down.

PR #3174, opened 2022-12-02
bugfix: sm2 public key in openssl3

mayuanchenma:
1. sm2 public key in openssl3
2. add sm2 ek template support


williamcroberts (approved):
I'll take this for rc1 on the tools release.

PR #3184, opened 2023-01-05
Skip test_tpm2_header tests in BE architectures

aplanas:
The tests are casting a fixed stream of bytes into a struct.  This array is already expressed in little endian order, so big endian processors will fail the test (PPC, PPC64, S390X, for example)

This commit skip the test in those architectures

### DIFF
--- a/test/unit/test_tpm2_header.c
+++ b/test/unit/test_tpm2_header.c
@@ -8,6 +8,7 @@
 #include <cmocka.h>
 
 #include "tpm2_header.h"
+#include "tpm2_util.h"
 
 static void test_tpm_command_header(void **state) {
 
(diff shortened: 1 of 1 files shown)


JuergenReppSIT (changes requested):
see last comment.

test/unit/test_tpm2_header.c
```diff
@@ -18,6 +19,10 @@ static void test_tpm_command_header(void **state) {
       0x00, 0x06, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00, 0x7f
     };
 
+    /* the test can only be executed in little endian architectures */
+    if (tpm2_util_is_big_endian())
+      skip();
+
     tpm2_command_header *c = tpm2_command_header_from_bytes(command_bytes);
 
     assert_true(c->tag == 0x0180);
```
williamcroberts, line 28: It's really that the bytestream is in big endian as thats what the TPM produces and the checks on the scalars in the struct being casted on it are gonna be weird based on byte ordering. I guess this test was to ensure that `tpm2_command_header_from_bytes` didn't modify anything, which I guess means that:
1. tpm2_command_header_from_bytes should take a const pointer `const UINT8 *`.
2. These tests could be removed as the compiler won't allow modification.

I can spin these up real quick.

williamcroberts:
@aplanas could you test my fix is on #3323, it should fix it and we can drop this PR.

williamcroberts:
Closing in favor of #3323

PR #3186, opened 2023-01-06
revert: sm2 sign and verifysignature.

mayuanchenma:
Although tpm2 tools sm2 sign and verifysignature conform to sm2 standard, sm2 only sign and verifysignature the digest in the tpm2. so keep pace with tpm2 and remove the code.


mayuanchenma:
@williamcroberts

williamcroberts:
> @williamcroberts

Why are we removing this? What does it mean to "keep pace with the tpm2", did the TPM 2.0 spec remove SM2 support?

williamcroberts:
> the TPM 2.0 spec don't remove SM2 support. TPM2 sign the digest of message, but the code in present sign (id || the digest of message), 

Does that mean it signs the `id` concatenated on `digest`.

> that cannot keep pace with the tpm2. @williamcroberts

I still have no idea what keep pace with the TPM2 means? Can't we just fix the code rather than removing it?

mayuanchenma:
removing it = fixing it，because the sign of sm2 is similar to ecdsa，don't the special processing.

keep pace with the TPM2 means
for example，using quote command in tpm2，quote sign the digest of the tpm information, the sm2 id is not concatenated on digest in tpm2, but the sm2 id is concatenated in verifysignature command of the tpm2 tool.

joholl:
As far as I understand @mayuanchenma correctly, the tools do not only hash user-provided data, but also an id `SM2_DEFAULT_ID`). They don't see why there should be an id included in the hash to be signed.

https://github.com/tpm2-software/tpm2-tools/blob/1529b0bf99f4b41608770b011effa5bda35a3976/lib/tpm2_alg_util.c#L1210

mayuanchenma:
hash(zid || message), This conform to sm2 standard.  But tpm only sign the final digest,  so it is not neccessary to hash message included zid. @joholl

williamcroberts:
> removing it = fixing it，because the sign of sm2 is similar to ecdsa，don't the special processing.
> 
> keep pace with the TPM2 means for example，using quote command in tpm2，quote sign the digest of the tpm information, the sm2 id is not concatenated on digest in tpm2, but the sm2 id is concatenated in verifysignature command of the tpm2 tool. @williamcroberts

So removing this code just removes the special processing for SM2 and it's treated like all other hashing algorithms and doesn't remove SM2 support, is that correct?

mayuanchenma:
yes @williamcroberts

williamcroberts:
> yes @williamcroberts

Fine by me then, I'll leave it to @JuergenReppSIT to decide when to merge.

williamcroberts (approved):

Issue #3191, opened 2023-01-30
`tpm2_startauthsession --bind-auth file:-` does not prompt for password


nicowilliams:
Most tpm2-tools use `tpm2_auth_util_from_optarg()`, but not `tpm2_startauthsession`, so it won't prompt.

PR #3199, opened 2023-02-01
policy authorization: allow the file attribute for policy authorization.

JuergenReppSIT:
Currently for several commands it's necessary to add password to shell commands. Now the file attribute can be used for session parameters and for the bind-auth parameter of tpm2_startauthsession. So also in these cases the hidden input from stdin can be used.

Examples:

  tpm2_startauthsession --policy-session -S session2.ctx \\
                        --bind-context=prim.ctx --bind-auth=file:-

  tpm2_sign -c key.ctx -p session:session2.ctx+file:- -g sha256 \\
            -o sig -d /tmp/digest

Fixes: #3191


nicowilliams:
Don't forget to update the man page.

lib/tpm2_auth_util.c
```diff
@@ -48,8 +50,88 @@ static bool handle_hex_password(const char *password, TPM2B_AUTH *auth) {
     return true;
 }
 
+static bool handle_password(const char *password, TPM2B_AUTH *auth);
+
+static tool_rc get_auth_for_file_param(const char* password, TPM2B_AUTH *auth) {
```
nicowilliams, line 55: Would `tpm2_auth_util_from_optarg()` be better here to avoid code duplication?

JuergenReppSIT: `tpm2_auth_util_from_optarg` also allows the prefix `session` or `pcr` which would not be appropriate.
There is not so much code duplication `get_auth_for_file_param` uses and fixes code from `handle_file` which now calls `get_auth_for_file_param`.

lib/tpm2_auth_util.c
```diff
@@ -48,8 +50,91 @@ static bool handle_hex_password(const char *password, TPM2B_AUTH *auth) {
...
+        LOG_ERR("oom");
+        return tool_rc_general_error;
+    }
+
+    if (path) {
+        path += FILE_PREFIX_LEN;
+        path = strcmp("-", path) ? path : NULL;
+    }
+    if (!is_a_tty || path) {
+        UINT16 fsize = sizeof(buffer) - 1;
```
nicowilliams, line 71: `buffer` is not an array but a heap pointer.  `fsize` will be `7` on 64-bit systems and `3` on 32-bit systems.

lib/tpm2_auth_util.c
```diff
@@ -48,8 +50,91 @@ static bool handle_hex_password(const char *password, TPM2B_AUTH *auth) {
...
+        bool ret = files_load_bytes_from_buffer_or_file_or_stdin(NULL, path,
+                                                                 &fsize, buffer);
+        if (!ret) {
+            free(buffer);
+            return tool_rc_general_error;
+        }
+        /* bash here strings and many commands add a trailing newline, if its stdin, kill the newline */
+        if (!path && buffer[fsize - 1] == '\n') {
```
nicowilliams, line 79: You should check that `fsize` is not `0`.

nicowilliams: Also, you might just make that a while loop.  And you might chomp carriage returns too (think of Windows users).

lib/tpm2_auth_util.c
```diff
@@ -48,8 +50,91 @@ static bool handle_hex_password(const char *password, TPM2B_AUTH *auth) {
...
+            buffer[fsize - 1] = '\0';
+        }
+    } else {
+        /*
+         * It is a TTY and we're reading from stdin.
+         * Prompt the user for the password with echoing
+         * disabled.
+         */
+        tool_rc rc = console_display_echo_control(false);
```
nicowilliams, line 88: Do we need a signal handler to restore echo-on mode on interrupt?

lib/tpm2_auth_util.c
williamcroberts: Another option would be to make an _ex variant that takes flags for what it could support, but I'll leave that judgment call on you. Both roads lead to Rome.

lib/tpm2_auth_util.c
JuergenReppSIT: @nicowilliams Thank you for the review, and apologies for the first sloppy version of the PR ;-)
There is already a signal handler for the FAPI commands. Next week I will create a separate PR for the tool commands if there is no handler.

lib/tpm2_auth_util.c
JuergenReppSIT: @williamcroberts next week i'll think about what's the best way to rome ;-)

JuergenReppSIT: @williamcroberts I would prefer the current solution, because there is a clear separation
in the way of the password handling between `tpm2_auth_util_from_optarg` and  `handle_str_password`.
So there is no need for addtional flags...

lib/tpm2_auth_util.c
JuergenReppSIT: @nicowilliams I have changed the buffer allocation for getline from stack allocated to heap allocated to avoid reallocation errors from getline.
But I think we do not need an additional signal handler, because the program
is always terminated by SIGTERM or SIGINT and so all resources will be reclaimed.

nicowilliams: You need a signal handler to restore echo-on, otherwise if you ^C while being prompted then you'll get your shell prompt back with echo-off.

lib/tpm2_auth_util.c
nicowilliams: I would prefer a `tpm2_auth_util_from_optarg_ex()` that takes a flag like @williamcroberts proposes.  The reasons for this are: a) the delta to `tpm2_auth_util_from_optarg()` will be smaller than the delta here, b) you might need the same functionality again later, and doing that now saves you or someone else the refactor later.

lib/tpm2_auth_util.c
JuergenReppSIT: Yes that's true. I had tested ^C in the bash but in sh there is the problem with echo-off.
I have added the solution from the fapi tool commands.

lib/tpm2_auth_util.c
JuergenReppSIT: I would say there is no big delta. Some code was moved to the new function `get_auth_for_file_param` with
the fixes you asked for and the fix for reallocation in getline. And I would say it'ts better do use the function
`handle_str_password` to get the auth structure than using a function which returns a session which is used to get the auth structure.

lib/tpm2_auth_util.h
```diff
@@ -73,5 +73,6 @@ tool_rc tpm2_auth_util_get_shandle(ESYS_CONTEXT *ectx, ESYS_TR for_auth,
  *   Boolean indicating the success of the operation.
  */
 bool handle_str_password(const char *password, TPM2B_AUTH *auth);
```
cplappert, line 75: Is this function used anymore? If not you should remove it.

cplappert (changes requested):

PR #3207, opened 2023-02-11
Add AMD EK certificate fetching support

subtervisor:
Adds support for fetching AMD EK certificates via tpm2_getekcertificate. Tested on my AMD Ryzen 9 3900XT fTPM. I haven't had the opportunity to check on Intel/other platforms.

### DIFF
--- a/tools/tpm2_getekcertificate.c
+++ b/tools/tpm2_getekcertificate.c
@@ -20,12 +20,52 @@
 #include "tpm2_nv_util.h"
 #include "tpm2_tool.h"
 
+
+typedef enum tpm_manufacturer tpm_manufacturer;
+enum tpm_manufacturer {
+    VENDOR_AMD       = 0x414D4400,
+    VENDOR_ATMEL     = 0x41544D4C,
+    VENDOR_BROADCOM  = 0x4252434D,
+    VENDOR_CISCO     = 0x4353434F,
+    VENDOR_FLYSLICE  = 0x464C5953,
+    VENDOR_ROCKCHIP  = 0x524F4343,
+    VENDOR_GOOGLE    = 0x474F4F47,
+    VENDOR_HPE       = 0x48504500,
+    VENDOR_HUAWEI    = 0x48495349,
+    VENDOR_IBM       = 0x49424D00,
+    VENDOR_IBMSIM    = 0x49424D20, // Used only by mssim/ibmswtpm2
+    VENDOR_INFINEON  = 0x49465800,
+    VENDOR_INTEL     = 0x494E5443,
+    VENDOR_LENOVO    = 0x4C454E00,
+    VENDOR_MICROSOFT = 0x4D534654,
+    VENDOR_NSM       = 0x4E534D20,
+    VENDOR_NATIONZ   = 0x4E545A00,
+    VENDOR_NUVOTON   = 0x4E544300,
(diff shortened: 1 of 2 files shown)


tools/tpm2_getekcertificate.c
```diff
@@ -435,15 +657,20 @@ static tool_rc nv_read(ESYS_CONTEXT *ectx, TPMI_RH_NV_INDEX nv_index) {
 
     TPM2B_DIGEST cp_hash = { 0 };
     TPM2B_DIGEST rp_hash = { 0 };
-    rc = nv_index == RSA_EK_CERT_NV_INDEX ?
-
+    uint16_t nv_buf_size;
```
JuergenReppSIT, line 660: should be initialized with 0 because of scan-build error

tools/tpm2_getekcertificate.c
```diff
@@ -64,55 +154,130 @@ static unsigned char *hash_ek_public(void) {
...
+            uint32_t exp = ctx.out_public->publicArea.parameters.rsaDetail.exponent;
+            if (exp == 0) {
+                exp = 0x00010001; // 0 indicates default
+            } else {
+                LOG_WARN("non-default exponent used");
+            }
+            buf[3] = (BYTE)exp;
+            buf[2] = (BYTE)(exp>>=8);
+            buf[1] = (BYTE)(exp>>=8);
+            buf[0] = (BYTE)(exp>>=8);
```
JuergenReppSIT, line 189: should be set to `buf[0] = (BYTE)(exp);` because of scan-build error.

tools/tpm2_getekcertificate.c
```diff
@@ -605,30 +833,29 @@ static tool_rc process_output(void) {
...
+    if (!is_intel_cert && ctx.ecc_cert_buffer) {
+        is_intel_cert = !(strncmp((const char *)ctx.ecc_cert_buffer,
             "{\"pubhash", strlen("{\"pubhash")));
     }
 
     /*
      * Intel EK certificates on the NV-index are already in standard DER format.
      */
-    if (ctx.is_intc_cert && ctx.is_cert_on_nv) {
+    if (is_intel_cert && !ctx.is_cert_on_nv) {
```
JuergenReppSIT, line 850: should not be changed, to prevent failures in integration tests. I think the complete if statement could be removed. But that could be done in later PR which implements correct usage of the raw option.

JuergenReppSIT (changes requested):
The integration test test/integration/tests/getekcertificate.sh did work because the default url was set to the intel address. After your change one function in this test has to be changed:
```
# Setup retries for when EK certificate cannot be retrieved on first try
getekcert_with_retries() {
    trap - ERR
    INTL_URL=https://ekop.intel.com/ekcertservice/
    cert_done=
    for i in 1 2 3;
        do
        tpm2 getekcertificate -u $1 -x -X -o $2 $INTL_URL
        tpm2 getekcertificate -u $1 -x -X $INTL_URL > $3
        # Test that stdout output is the same as output to file
        cmp $2 $3
        if [ $? == 0 ]; then
            cert_done=1
            break
        fi
    done
    trap onerror ERR
    test $cert_done
}
```

JuergenReppSIT:
@subtervisor I have created pr #3236 with some small fixes of your pr to enable successful execution of the tests in the CI.
Therefore I will close the PR. Thank you again.

PR #3214, opened 2023-02-27
readpublic: fix reading and writing serialized trs

williamcroberts:
Fix reading and writing a serialized persistent ESYS_TR handles. This occurs becuase the TPM2_HANDLE is never set after loading and decisions are made on it.

Fixes:
tpm2_readpublic -t handle2.tr -c handle.tr
ERROR: Can only output a serialized handle for persistent object handles ERROR: Unable to run tpm2_readpublic


man/tpm2_tr_encode.1.md
```diff
@@ -0,0 +1,58 @@
...
+
+[common tcti options](common/tcti.md) collection of options used to configure
+the various known TCTI modules.
+
+# EXAMPLES
+
+## Serialize a public and handle as an ESYS_TR
+
+```bash
+tpm2_createprimary -c primary.ctx
```
JuergenReppSIT, line 50: The _ should be replaced with a blank in all tool commands

JuergenReppSIT (changes requested):

man/tpm2_tr_encode.1.md
williamcroberts: All the man pages are written with underscores since all the distros install the symlinks.,

williamcroberts:
I need to sanity check the options before I use them

PR #3229, opened 2023-03-31
tests: defaut to Python 3 in helpers code

aplanas:
Also add "dbus-run-session" in the example code that explain how to run the test.


JuergenReppSIT:
@aplanas Could you please rebase and push force your PR. The problem with the CI now should be fixed.

JuergenReppSIT (approved):

PR #3231, opened 2023-04-04
abrmd_policynv.sh Skip test which uses sign operator for comparison

JuergenReppSIT:
The tests whch use sign comparison operators will be skipped. After a CI
update these tests did produce strange error(0x126) which did not occur
outside the new docker images. Downgrade to the version of swtpm where
this error did not occur did not change this behaviour.


AndreasFuchsTPM (approved):
So ugly, but needed for now...
Let's also open an issue to track that we need to revert this at some point in the future.

Issue #3146, opened 2022-10-10
Tpm2_eventlog not initializing PCR[0] correctly


vrpate2:
For PCR[0], the default initial condition may be all bits CLEAR, all bits SET, the locality at which TPM2_Startup() was received, or an indicator that the first measurement came from an H-CRTM. Refer to Hardware Core Root of Trust Measurement (H-CRTM) Event Sequence in TPM Library specification part 1.

Max-ad-Wu:
Hi,
I saw it. May I know is it an issue and can it be fixed in this application?
Thanks

PR #3242, opened 2023-04-25
tpm2_eventlog: add support for PCR0 replay with different StartupLocality

THS-on:
According to the "TCG PC Client Platform Firmware Profile Specification Level 00 Version 1.05 Revision 23" section 10.4.5.3 the startup locality is the starting value of PCR0. This can be currently either 0 or 3.

This implements support for replaying logs where locality is not 0. 

Fixes: https://github.com/tpm2-software/tpm2-tools/issues/3146

### DIFF
--- a/lib/efi_event.h
+++ b/lib/efi_event.h
@@ -45,6 +45,7 @@
 #define EV_EFI_HANDOFF_TABLES2           EV_EFI_EVENT_BASE + 0xb
 #define EV_EFI_VARIABLE_BOOT2            EV_EFI_EVENT_BASE + 0xc
 
+#define EV_EFI_HCRTM_EVENT               EV_EFI_EVENT_BASE + 0x10
 #define EV_EFI_VARIABLE_AUTHORITY        EV_EFI_EVENT_BASE + 0xe0
 
 #ifndef PACKED
--- a/lib/tpm2_eventlog.c
+++ b/lib/tpm2_eventlog.c
@@ -31,7 +31,7 @@ bool digest2_accumulator_callback(TCG_DIGEST2 const *digest, size_t size,
  * 'size' parameter.
  */
 bool foreach_digest2(tpm2_eventlog_context *ctx, UINT32 eventType, unsigned pcr_index,
-                     TCG_DIGEST2 const *digest, size_t count, size_t size) {
+                     TCG_DIGEST2 const *digest, size_t count, size_t size, uint8_t locality) {
 
     if (digest == NULL) {
         LOG_ERR("digest cannot be NULL");
@@ -79,6 +79,10 @@ bool foreach_digest2(tpm2_eventlog_context *ctx, UINT32 eventType, unsigned pcr_
             ctx->sm3_256_used |= (1 << pcr_index);
         } else {
             LOG_WARN("PCR%d algorithm %d unsupported", pcr_index, alg);
+        }
+
+        if (eventType == EV_NO_ACTION && pcr && pcr_index == 0 && locality > 0 ){
+            pcr[alg_size -1] = locality;
         }
 
         if (eventType != EV_NO_ACTION && pcr &&
@@ -184,7 +188,7 @@ bool parse_event2(TCG_EVENT_HEADER2 const *eventhdr, size_t buf_size,
     ret = foreach_digest2(&ctx, eventhdr->EventType,
                           eventhdr->PCRIndex, 
                           eventhdr->Digests, eventhdr->DigestCount,
-                          buf_size - sizeof(*eventhdr));
+                          buf_size - sizeof(*eventhdr), 0);
     if (ret != true) {
         return false;
     }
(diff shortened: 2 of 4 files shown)


JuergenReppSIT:
@THS-on pcr0 is set to the locality for every event `EV_NO_ACTION` in `foreach_digest2`. Shouldn't there be a check of the signature (`STARTUP_LOCALITY_SIGNATURE`)?
The locality should be added to the calls of `foreach_digest2` in the unit test test/unit/test_tpm2_eventlog.c

THS-on:
> pcr0 is set to the locality for every event `EV_NO_ACTION` in `foreach_digest2`.

Only if the the locality flag it set to > 0, which is the case we need to handle. I know this is somewhat hacky. Ideally I would like to have a unset value, but did not find a good way to do that in C.

> Shouldn't there be a check of the signature (STARTUP_LOCALITY_SIGNATURE)?

Yes the check is now added.

>  The locality should be added to the calls of foreach_digest2 in the unit test test/unit/test_tpm2_eventlog.c

Now fixed.

JuergenReppSIT:
@THS-on Some more questions:
* The spec states "In the case where an HCRTM event did not
occur, this event may not be used" for  the StartupLocality event. Doesn't this mean that the PCR0 will only be set to the locality if there is a HCRTM event before?
* You check the first event whether it's NO_ACTION with  StartupLocality and produce an error if the signature is not StartupLocality. Couldn't there be another NO_ACTION event the first event?

THS-on:
> The spec states "In the case where an HCRTM event did not
occur, this event may not be used" for the StartupLocality event. Doesn't this mean that the PCR0 will only be set to the locality if there is a HCRTM event before?

If I understand the spec correctly this seems to be the case, is is currently not implemented in the code.

> You check the first event whether it's NO_ACTION with StartupLocality and produce an error if the signature is not StartupLocality. Couldn't there be another NO_ACTION event the first event?

I don't think so, because this is the only other usage for NO_ACTION. Technically there is also TCG_Sp800_155_PlatformId_Event, but that seems to be deprecated. 
But it might be not the first event in the log overall if we can have a HCRTM before. 
Do we have any logs with HCRTM events?

I can move that check into foreach_event2(), which might be a better place and we then just can set a flag if we found a HCRTM event before.

JuergenReppSIT:
> Do we have any logs with HCRTM events?
> 
No unfortunately not.
In the tpm2 tool test files we have only EV_S_CRTM_VERSION and EV_NO_ACTION with "Signature: Spec ID Event03" events. 
It would be nice if we could add such a file to the test files.

> I can move that check into foreach_event2(), which might be a better place and we then just can set a flag if we found a HCRTM event before.

yes that would be better. Thanks.

THS-on:
>  It would be nice if we could add such a file to the test files.

Do you know if any devices generate actually logs with that event? I have not seen that.

I've now moved the special handling into `foreach_event2(..)`.

JuergenReppSIT:
Very good. Now there are no scan-build errors. 
Wouldn't it be better not to create an error. The spec allows vendor specific  EV_NO_ACTION events:  "10.4.5.4 Vendor Specific Events"
So the check could be
```
   if (memcmp(locality_event->Signature, STARTUP_LOCALITY_SIGNATURE, sizeof(STARTUP_LOCALITY_SIGNATURE)) == 0) {
      locality = locality_event->Cases.StartupLocality;
   }
```

"

THS-on:
The vendor specific events are in PCR6, but I changed it now not to error.

mayuanchenma:
The spec states "In the case where an HCRTM event did not occur, this event may not be used" for the StartupLocality event. 

if (eventhdr->EventType == EV_EFI_HCRTM_EVENT && eventhdr->PCRIndex == 0) {
            found_hcrtm = true;
        }

        /* Handle StartupLocality in replay for PCR0 */
        if (!found_hcrtm && eventhdr->EventType == EV_NO_ACTION && eventhdr->PCRIndex == 0) {
            if (event_size < sizeof(EV_NO_ACTION_STRUCT)) {
                LOG_ERR("EventSize is too small\n");
                return false;
            }

            EV_NO_ACTION_STRUCT *locality_event = (EV_NO_ACTION_STRUCT*)event->Event;

            if (memcmp(locality_event->Signature, STARTUP_LOCALITY_SIGNATURE,
                       sizeof(STARTUP_LOCALITY_SIGNATURE)) == 0) {
                locality = locality_event->Cases.StartupLocality;
            }
        }
The code means that an HCRTM event did occur, this event may not be used? spec is incorrect？@THS-on @JuergenReppSIT

JuergenReppSIT:
The comment  "In the case where an HCRTM event did not occur, this event may not be used" 
in the discussion was wrong. The spec states:
"In the case where an HCRTM event occurred, this event may not be used."
So the implementation is correct.

mayuanchenma:
Thanks @JuergenReppSIT

PR #3243, opened 2023-04-28
configure: check for efivar/efivar.h manually

THS-on:
I noticed during my testing `DevicePath`s are not decoded even when compiled with `--with-efivar`.
The reason was that the code checks for `HAVE_EFIVAR_EFIVAR_H`, but only `PKG_CHECK_MODULES` is used to detect if `libefivar.so` is present, not if the header is also there. This commit adds that check.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,9 +92,9 @@ AC_ARG_WITH([efivar],
 
 # use the true program to avoid failing hard
 AS_IF([test "x$with_efivar" == "xauto"],
-  [PKG_CHECK_MODULES([EFIVAR], [efivar],,[true])],
+  [PKG_CHECK_MODULES([EFIVAR], [efivar], [AC_CHECK_HEADERS([efivar/efivar.h], , [true])], [true])],
   [test "x$with_efivar" == "xyes"],
-  [PKG_CHECK_MODULES([EFIVAR], [efivar])],
+  [PKG_CHECK_MODULES([EFIVAR], [efivar], [AC_CHECK_HEADERS([efivar/efivar.h])])],
 )
 
 # backwards compat with older pkg-config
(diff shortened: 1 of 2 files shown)


JuergenReppSIT (changes requested):
You should replace your change with:
```
AS_IF([test "x$with_efivar" == "xauto"],
  [PKG_CHECK_MODULES([EFIVAR], [efivar], [AC_CHECK_HEADERS([efivar/efivar.h], , [true])], [true])],
  [test "x$with_efivar" == "xyes"],
  [PKG_CHECK_MODULES([EFIVAR], [efivar], [AC_CHECK_HEADERS([efivar/efivar.h])])],
)
```
and rebase your branch to avoid CI errors. After this change, if the efivar dev package is installed,  you will get an compilation error in:
https://github.com/tpm2-software/tpm2-tools/blob/115bbc0581d8d2f4f99d028b184cddb6fb5b151a/lib/tpm2_eventlog_yaml.c#L280
To fix it you could replace `(unsigned char *)` with `(char *)`

THS-on:
The signature of `efidp_format_device_path(..)` changed from `char*` to `unsigned char*` in this commit https://github.com/rhboot/efivar/commit/ba165d1482558005a16f4b44abaa1df5169727d1, so with the latest version this is `unsigned char*`.
What is the best way to handle this?

JuergenReppSIT:
> The signature of `efidp_format_device_path(..)` changed from `char*` to `unsigned char*` in this commit [rhboot/efivar@ba165d1](https://github.com/rhboot/efivar/commit/ba165d1482558005a16f4b44abaa1df5169727d1), so with the latest version this is `unsigned char*`. What is the best way to handle this?

The signature changed from version 37 to 38, but it was not possible to use the version compare `[efivar >= 38 ]` in `PKG_CHECK_MODULES`. So perhaps we should use and comment the ugly hack `(void *)` instead of `(unsigned char *)`.

THS-on:
@JuergenReppSIT I now included the hack to support versions < 38.

JuergenReppSIT:
@THS-on Thanks. So we don't have to rework when the first distributions use 38.

PR #3249, opened 2023-05-15
man/tpm2_verifysignature.1.md: correct example command tpm2_sign

penghao-dingdao:
There is actually no -m option in tpm2_sign, the example command has made a mistake, so correct it, link what's in man/tpm2_sign.1.md.


penghao-dingdao:
OK. Thanks for reply. I will do it tomorrow.

PR #3250, opened 2023-05-17
tools/tpm2_unseal.c: add qualifier static to enhance privacy of unseal function

penghao-dingdao:
Other tpm2_xxx.c have qualifier "static" before their "main function",
for example: static tool_rc load(ESYS_CONTEXT *ectx) in tpm2_load.c,
only function unseal in tpm2_unseal.c lacks it, so add it before unseal
function to enhance privacy.


penghao-dingdao:
Is there something wrong?

JuergenReppSIT:
> Is there something wrong?

@penghao-dingdao  No it's ok.

PR #3257, opened 2023-06-12
Forward seal and more than 8 PCRs

jandryuk:
Forward sealing is sealing TPM data to precomputed values instead of the
current PCR values.  This can be useful for sealing to new values during
an upgrade process.

This PR adds support for forward sealing and sealing against more than 8 PCRs.

New struct tpm2_fowards and struct tpm2_forward are added.  struct tpm2_forward is a PCR selection and PCR values for a single bank, while struct tpm2_forwards is a set of multiple banks.  tpm2_forward is space-inefficient in that it has the full PCR range, but it allows for more straightforward code to parse, track and use the PCR values.  The 8 digest restriction of TPML_DIGEST isn't really suitable for use when tracking potentially many PCR values.

Forward sealing is implemented for tpm2_policypcr.  Maybe it should also be added to tpm2_createpolicy?  I am not familiar with that command, and tpm2_policypcr was all I needed for my use case.  Hence, I did not add it there.

https://github.com/tpm2-software/tpm2-tools/commit/000201c1d32d4c257456ea16e83a77d3fc30aa1a isn't really necessary, but I included it since it is already written.  It is step toward expanding test_tpm2_policy.c.

In addition to the unit tests, I tested on a real TPM2, but it only has a single SHA256 bank.


jandryuk:
Hi.  Any feedback on this PR?

JuergenReppSIT:
> Hi. Any feedback on this PR?

@jandryuk I started with the review and hope that I will find time next week to continue it.
One first remark: ` pcr_parse_list` allows sizes smaller than the digest size of the hashalg. It would be better to check whether size of the digest has the size of the corresponding hashalg.

jandryuk:
Thanks, @JuergenReppSIT .  I labeled a unit test as looking for a small digest, but it was actually "passing" because the length was odd.

jandryuk:
> In addition to the unit tests, I tested on a real TPM2, but it only has a single SHA256 bank.

I modified test/integration/tests/unseal.sh to try adding some forward sealing integration tests.  Single bank, sha1 or sha256, works (unseals successfully).  Curiously, `sha1:0=6fd13bfa9ec8bc42e39d262810bbb912373ca5f9,1,2,3+sha256:0,1,2,3` successfully unseals while switching the bank order `sha256:0,1,2,3+sha1:0=6fd13bfa9ec8bc42e39d262810bbb912373ca5f9,1,2,3` fails unseal with `Esys_Unseal(0x99D) - tpm:session(1):a policy check failed`

JuergenReppSIT:
@jandryuk Could you please push force your version with the changed unseal test?

jandryuk:
Pushed.  @JuergenReppSIT I fixed up the digest size checking like you spotted.  The forward-seal integration test is added.  Some unit tests are expanded.  I also fixed a bug with 'idx' which was the cause of the multi-bank sealing error.  I also switched to `pcrs->count == ARRAY_LEN(pcrs->pcr_values)` (with equality) since that is the correct bounds on the array size.

JuergenReppSIT:
@jandryuk Thank you for the big PR. LGTM for me now.
You did use `tpm2_pcrs` to store the values in tpm2_policypcr.
During the review, the only thing that struck me was the strange dimension of  `pcr_values` in ` tpm2_pcrs` which should be `(TPM2_MAX_PCRS + 7) / 8 * TPM2_NUM_PCR_BANKS` instead of `TPM2_MAX_PCRS`? But that was already used before your PR. So it could be changed later.
Perhaps you could also test more than 8 PCRs in the integration test?
```
pcr_specification=sha256:1,2,3,4,16+sha1:1,2,3,4,16
pcr_fwd_specification=sha256:1,2,3,4,16=bba91ca85dc914b2ec3efb9e16e7267bf9193b14350d20fba8a8b406730ae30a+sha1:1,2,3,4,16=6fd13bfa9ec8bc42e39d262810bbb912373ca5f9
pcr_sha1_specification=sha1:1,2,3,16
```

JuergenReppSIT (approved):

jandryuk:
Hi, @JuergenReppSIT .  Thanks for the approval.  Do you want me to force push with the integration test change to use more PCRs?

> tpm2_pcrs which should be (TPM2_MAX_PCRS + 7) / 8 * TPM2_NUM_PCR_BANKS instead of TPM2_MAX_PCRS

I think your calculation is correct, but is there another factor?  TPM2_PCR_Read here https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf has
>The TPM will continue processing bits until all have been processed or until pcrValues would be too large to fit into the output buffer if additional values were added.

Might some TPMs not return 8 entries in each call?  In that case, the larger number of entries could be needed if only a single PCR is returned in each call.  But I'm just guessing.

JuergenReppSIT:
Hi, @jandryuk, Thanks for the hint to the spec. I do not understand  the meaning of "until pcrValues would be too large to fit into the output buffer if additional values were added". Maybe your guess is correct. So I will merge the current version.

PR #3258, opened 2023-06-12
Support duplicate sealing

williamcroberts:


idesai (approved):
@williamcroberts If you have a test at hand, add it.

PR #3261, opened 2023-06-15
tpm2_send: Fix EOF detection on input file

AndreasFuchsTPM:
If an input file is shorter that the size expressed in the command header, then we will not do a correct abort. The fix is to also call feof() and error out.

### DIFF
--- a/tools/tpm2_send.c
+++ b/tools/tpm2_send.c
@@ -70,8 +70,8 @@ static int read_command_from_file(FILE *f, tpm2_command_header **c,
     LOG_INFO("command code: 0x%08x", tpm2_command_header_get_code(command));
 
     ret = fread(command->data, data_size, 1, f);
-    if (ret != 1 && ferror(f)) {
-        LOG_ERR("Failed to read command body: %s", strerror (errno));
+    if (ret != 1) {
+        LOG_ERR("Failed to read command body: %s", feof(f) ? "EOF" : strerror (errno));
         free(command);
         return -1;
     }


tools/tpm2_send.c
```diff
@@ -70,6 +70,10 @@ static int read_command_from_file(FILE *f, tpm2_command_header **c,
     LOG_INFO("command code: 0x%08x", tpm2_command_header_get_code(command));
 
     ret = fread(command->data, data_size, 1, f);
+    if (feof(f)) {
+        LOG_ERR("Faild to read command body: EOF reached");
+        return -1;
+    }
     if (ret != 1 && ferror(f)) {
```
jandryuk, line 77: Hi, @AndreasFuchsTPM.  Looks like you should also `free(command)`.  Maybe you can just simplify to `if (ret != 1) {` and then drop the strerror in the logging?  You would lose the specifics though, so maybe that isn't desired.

JuergenReppSIT: @AndreasFuchsTPM I would agree to @jandryuk  and could merge the PR if the free is added.

williamcroberts: Done.

tools/tpm2_send.c
```diff
@@ -70,10 +70,10 @@ static int read_command_from_file(FILE *f, tpm2_command_header **c,
...
 
     ret = fread(command->data, data_size, 1, f);
-    if (ret != 1 && ferror(f)) {
-        LOG_ERR("Failed to read command body: %s", strerror (errno));
-        free(command);
-        return -1;
+    if (ret != 1) {
+        LOG_ERR("Failed to read command body: %s", feof(f) ? "EOF" : strerror (errno));
+	free(command);
+	return -1;
```
jandryuk, line 76: Looks like hard tabs slipped in.  Otherwise, LGTM.

williamcroberts: > Looks like hard tabs slipped in. Otherwise, LGTM.

new machine... didn't set up vim thnaks

jandryuk (approved):
LGTM.  Thanks!

PR #3262, opened 2023-06-21
tpm2_duplicate: add --key-algorithm option.

### LINKED ISSUE #3259: tpm2_duplicate -G not used consistently
Variable "ctx.key_type" is used to set both the symmetric algorithm:

https://github.com/tpm2-software/tpm2-tools/blob/6e484f59bc65b01ec5b1f81b5bf0897c53a1ff4b/tools/tpm2_duplicate.c#L377

and the key type 

https://github.com/tpm2-software/tpm2-tools/blob/6e484f59bc65b01ec5b1f81b5bf0897c53a1ff4b/tools/tpm2_duplicate.c#L217-L218

The encryption key and the key type specifier should have different options. This is likely a backwards compat breaking change :-(

JuergenReppSIT:
The -G --wrapper-algorithm option was used for the symmetric algorithm used for the inner wrapper and the algorithm of the public key used to wrap an openSSL key for the destination TPM.
To match this parameter with the options used for tpm2_create the long option --key-algorithm is now also mapped to -G. Fixes #3259

### DIFF
--- a/man/tpm2_duplicate.1.md
+++ b/man/tpm2_duplicate.1.md
@@ -21,9 +21,15 @@ These options control the key importation process:
 
   * **-G**, **\--wrapper-algorithm**=_ALGORITHM_:
 
-    The symmetric algorithm to be used for the inner wrapper. Supports:
+    The symmetric algorithm to be used for the inner wrapper if -U is not used.
+    Supports:
     * aes - AES 128 in CFB mode.
     * null - none
+    The key algorithm associated with the public parent if -U is used.
+
+  * **-G**, **\--key-algorithm**=_ALGORITHM_:
+
+    The key algorithm associated with the public parent if -U is used.
 
   * **-i**, **\--encryptionkey-in**=_FILE_:
 
@@ -45,6 +51,8 @@ These options control the key importation process:
 
     Specifies the file path to the public key of the parent object on the
     destination TPM.  This should be a `TPM2B_PUBLIC` formatted file.
+    This public key is used for the wrapping of a PEM or DER key
+    which will be exported for a remote TPM.
 
   * **-k**, **\--private-key**=_FILE_:
 
--- a/tools/tpm2_duplicate.c
+++ b/tools/tpm2_duplicate.c
@@ -559,6 +559,7 @@ static bool tpm2_tool_onstart(tpm2_options **opts) {
       { "auth",              required_argument, 0, 'p'},
       { "policy",            required_argument, 0, 'L'},
       { "wrapper-algorithm", required_argument, 0, 'G'},
+      { "key-algorithm",     required_argument, 0, 'G'},
       { "private",           required_argument, 0, 'r'},
       { "public",            required_argument, 0, 'u'},
       { "private-key",       required_argument, 0, 'k'},


whooo (approved):

man/tpm2_duplicate.1.md
```diff
@@ -21,9 +21,15 @@ These options control the key importation process:
...
-    The symmetric algorithm to be used for the inner wrapper. Supports:
+    The symmetric algorithm to be used for the inner wrapper if -U is not used.
+    Supports:
     * aes - AES 128 in CFB mode.
     * null - none
+    The key algorithm associated with the public parent if -U is used.
+
+  * **-G**, **\--key-algorithm**=_ALGORITHM_:
+
+    The key algorithm associated with the public parent if -U is used.
```
williamcroberts, line 32: I wonder if we can phase this option out. Wouldn't we just gate this based on `-i` input? if `-i` is used, inner wrapper mode is aes128cfb (as their is no other choice), and if not set the inner wrapper is null?

For backwards compat we can just NOP this option, perhaps issue a warn deprecated, mark deprecated in man page?

man/tpm2_duplicate.1.md
```diff
@@ -45,6 +51,8 @@ These options control the key importation process:
 
     Specifies the file path to the public key of the parent object on the
     destination TPM.  This should be a `TPM2B_PUBLIC` formatted file.
+    This public key is used for the wrapping of an openssl key
+    which will be exported for a remote TPM.
```
williamcroberts, line 55: An openssl key is too specific, it'll support any formatted PEM or DER key.
```markdown
This public key will be used for wrapping a plaintext key, eg PEM or DER, that is being duplicated to a new TPM storage parent.
See [Section 22 of TPM Spec 1.59](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part1_Architecture_pub.pdf) for further details.
```

Couldn't we also NOP -G for this option and figure it out? If it's an RSA key we use OAEP and if it's an ECC parent we perform that Diffe-Helman like process for the Z point.

williamcroberts (changes requested):
I'm thinking we can NOP the `-G` option and in the context of `-U` or `-i` options set the correct algorithm. This would also allow an inner wrapping to occur with a storage parent specified through `-U`. Ie `-U` and `-i` options are set.

It's more work, but it simplifies the tooling greatly. Existing code should just work, as they would only be able to specify one or the other (`-i` or `-U`) and new code could use both options.

Unless I groked something way off here

man/tpm2_duplicate.1.md
williamcroberts: We can also use `-o` to determine if inner wrapping is performed, `-i` or `-o` implies inner wrapping.

williamcroberts: Oh but we still need `-G` for determining what to do with symmetric keys since we can't parse a PEM/DER file and figure it out. They may want an AES or KEYEDHASH key, etc....

I guess disregard this train of thought, but this whole tool could be done nicer on the interface but then we have to bump major vers number

man/tpm2_duplicate.1.md
JuergenReppSIT: @williamcroberts @whooo I changed openssl key to PEM or DER key. Did I understand correctly that the PR can now be merged?

williamcroberts: > @williamcroberts @whooo I changed openssl key to PEM or DER key. Did I understand correctly that the PR can now be merged?

yes thank you

PR #3268, opened 2023-07-17
tpm2_eventlog: add H-CRTM event support.

mayuanchenma:


JuergenReppSIT:
@mayuanchenma Do you have an example event log file with an H-CRTM event?

JuergenReppSIT:
@mayuanchenma  Thank you for the eventlog. Could you please also attach the output of:
`tpm2_pcrread`

mayuanchenma:
[pcrread.log](https://github.com/tpm2-software/tpm2-tools/files/12101746/pcrread.log) @JuergenReppSIT

JuergenReppSIT:
@mayuanchenma Thank you for the PR and the uploaded files. I will add this change also to tss FAPI. Can I use the uploaded eventlog for  the FAPI integration tests?

JuergenReppSIT (approved):

JuergenReppSIT:
@mayuanchenma Thank you for making the files available.

JuergenReppSIT:
@mayuanchenma  I tried to use your eventlog.log in the the TSS2-FAPI unit test for event logging. There were two events where the digest list did not correspond to the event data: EV_EFI_HCRTM_EVENT and EV_EFI_VARIABLE_BOOT.
Do you have an explanation for this difference?

mayuanchenma:
variable_boot
@JuergenReppSIT

mayuanchenma:
hcrtm_event_data
HCRTM event data is a fixed string, so the digest is fixed.  If so, the event will be meaningless.  I put the digest of HCRTM code in the digest list. @JuergenReppSIT

JuergenReppSIT:
@mayuanchenma many thanks for the explanations. During the tests I did check what digests do match the event data. For the event log of my laptop the event data of EV_EFI_VARIABLE_BOOT always did correspond to the computed digest values. But as you pointed out this is not expected. 
For the HCRTM event I did expect the sha1 digest would be:
`echo -n "HCRTM" | xxd -r -p| sha1sum`
`da39a3ee5e6b4b0d3255bfef95601890afd80709`
but was `1f29fc784497b2858bd7cf698e0a949c836a0720` in your eventlog?

mayuanchenma:
hcrtm_event_data1
I think that it is correct for recording the digest of the HCRTM code in the hcrtm event. So, don't need do check for hcrtm event.
@JuergenReppSIT

PR #3270, opened 2023-07-21
eventlog: fix YAML string formatting

berrange:
This fixes my previous commit:
    
      commit c78d258b2588aee535fd17594ad2f5e808056373
      Author: Daniel P. Berrangé <berrange@redhat.com>
      Date:   Thu Sep 29 10:48:36 2022 -0400
    
        tpm2_eventlog_yaml: fix malformed YAML for EV_IPL data

That commit addressed the malformed YAML, but used an incorrect
choice for the YAML string encoding. The "block scalar" string
encoding in YAML is a fairly raw representation which lets us
split across multiple lines, but does not in fact support escape
sequences, nor quoting. So the escapes/quotes emitted were taken
literally in the YAML parsing of anything consuming the eventlog.

To enable those parsing the YAML eventlog to get the original
data accurately, we must use the YAML "flow scalar" with double
quoting. The complication with "flow scalar" encoding is that
leading whitespace is deleted on lines. Fortunately it is
possible to avoid this whitespace deletion using '\t' for tab,
and '\x20' for space characters.

The unit test is extended to validate the result of parsing an
event data multi-line string with leading whitespace.


JuergenReppSIT (approved):
@berrange sorry it took so long with the review. During the review I tested your changes with an eventlog file created by fapi. fapi stores only the hex data of the events in a json file.
I created a yaml file from this file and compared all events with the results in
./test/integration/fixtures/event-moklisttrusted.bin.yaml. I think that's better than picking one event for the comparison. All events were equal. I will create a separate PR with this test. Perhaps you could review the PR #3277?

berrange:
> I created a yaml file from this file and compared all events with the results in
./test/integration/fixtures/event-moklisttrusted.bin.yaml. I think that's better than picking one event for the comparison. All events were equal. I will create a separate PR with this test. Perhaps you could review the PR https://github.com/tpm2-software/tpm2-tools/pull/3277?

I was on holiday, so that PR was already merged, but none the less I reviewed it and like what you did there for validation.

JuergenReppSIT:
@berrange Thanks for the review. I did merge it because @whooo already approved the PR.

PR #3279, opened 2023-09-03
tpm2_createek: Fix usage of nonce.

### LINKED ISSUE #3278: tpm2_createek: Does not follow TPM specification
In the current TCG EK Credential Profile [0] in section **2.2.1.6 TPMT_PUBLIC Calculation** there are four cases specified on how to build the endorsement key template (or when to fail). By now, tpm2_createek does not support the second case which is:

> If the EK Template is Populated and the EK Nonce is Absent, the EK Template is used unmodified as the TPMT_PUBLIC. 

This appears since if the `-t` option is given to `tpm2_createek`, the code reads unconditionally both the EK template and the EK nonce (except for the condition whether the algorithm generally supports a nonce, which is the case for RSA).

https://github.com/tpm2-software/tpm2-tools/blob/23114c9f5d3463f68464ecbd10593df2c863437f/tools/tpm2_createek.c#L221-L249

I suppose the tool should first check whether the EK Nonce is even defined, and if not, just use the EK Template as it was retrieved from the EK Template NV Index.

The exact command I use: `tpm2_createek -f pem -t -u key.pub -c key.ctx -V`

[0] https://trustedcomputinggroup.org/wp-content/uploads/TCG-EK-Credential-Profile-V-2.5-R2_published.pdf

JuergenReppSIT:
The current TCG EK Credential Profile [0] in section "2.2.1.6 TPMT_PUBLIC Calculation" states:
"If the EK Template is Populated and the EK Nonce is Absent, the EK Template is used unmodified as the TPMT_PUBLIC." Currently the command did produce an error if there is a template but no nonce.
Now the template is used but it's checked whether the nonce exists. Fixes #3278.

### DIFF
--- a/tools/tpm2_createek.c
+++ b/tools/tpm2_createek.c
@@ -237,20 +237,42 @@ static tool_rc set_ek_template(ESYS_CONTEXT *ectx, TPM2B_PUBLIC *input_public) {
         goto out;
     }
 
-    // Read EK nonce
-    UINT16 nonce_size = 0;
+    // Check whether nonce exists
     if (nonce_nv_index) {
-        rc = tpm2_util_nv_read(ectx, nonce_nv_index, 0, 0,
-                &ctx.auth_owner_hierarchy.object, &nonce, &nonce_size, &cp_hash,
-                &rp_hash, TPM2_ALG_SHA256, 0, ESYS_TR_NONE, ESYS_TR_NONE, NULL);
+        TPMS_CAPABILITY_DATA *capabilities = NULL;
+
+        rc = tpm2_getcap(ectx, TPM2_CAP_HANDLES, nonce_nv_index,
+                         1, NULL, &capabilities);
         if (rc != tool_rc_success) {
             goto out;
         }
+
+        if (capabilities->data.tpmProperties.count == 0 ||
+            capabilities->data.handles.handle[0] != nonce_nv_index) {
+            free(capabilities);
+            // The EK Template is used unmodified
+            goto out;
+        }
+        free(capabilities);
+    } else {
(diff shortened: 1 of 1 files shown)


tools/tpm2_createek.c
```diff
@@ -237,19 +237,41 @@ static tool_rc set_ek_template(ESYS_CONTEXT *ectx, TPM2B_PUBLIC *input_public) {
...
+    if (rc != tool_rc_success) {
+        goto out;
     }
 
     if (input_public->publicArea.type == TPM2_ALG_RSA) {
-        memcpy(&input_public->publicArea.unique.rsa.buffer, &nonce, nonce_size);
+        if (nonce_size) {
+            memcpy(&input_public->publicArea.unique.rsa.buffer, &nonce, nonce_size);
+        }
         input_public->publicArea.unique.rsa.size = 256;
```
akorb, line 275: I think the size of the unique field shouldn't be overwritten as well.

akorb (commented):
I think the size of the unique field shouldn't be overwritten as well.

tools/tpm2_createek.c
JuergenReppSIT: If the nonce does not exist there is a `goto out`. So the nonce will not be overwritten.

JuergenReppSIT: @akorb  I just copied the initialization to the part where the nonce size is gt 0 as you suggested. So the RSA case is treated equally to the ECC case.

akorb: Thanks! I don't know how to declare this review as `Done` though. Somehow the button is missing.

PR #3280, opened 2023-09-05
tpm2_createek: Fix integrating nonce

akorb:
I externally create the EK with a given template and nonce. And I wondered why `tpm2_createek` yields a different EK than what I retrieve. I eventually figured out that the `tpm2_createek` code falsely copies the pointer into the final template, instead of the actual nonce.

By fixing this with these changes, the EK from `tpm2_createek` and my code is consistent.


akorb:
Should I sign off the commit, or does someone else have to do it?

JuergenReppSIT:
> > Should I sign off the commit, or does someone else have to do it?
> 
> You should sign off the commit, git commit --amend -s should do it.

And you can change the commit  with `push --force ...` Thank you.

akorb:
Done

JuergenReppSIT (approved):

PR #3281, opened 2023-09-21
tpm2_createek: Correct man page example

haxtibal:
The example
```
tpm2_createek -G rsa -u ek.pub
```
from `man tpm2_createek` misses to provide `-c`.

It thus fails with
```
ERROR: Expected option -c
```

According to [tools/tpm2_createek.c:441](https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_createek.c#L441), the option must be present in any case. Adjust the man page example accordingly.


JuergenReppSIT (approved):
Thank you for the PR. The complete example also does not work if a resource manager is used. I will create a second PR to fix also this problem.

PR #3282, opened 2023-09-25
tpm_createek: Fix man page example.

JuergenReppSIT:
The last example does not work if a resource manager is used. Thus the option -Tdevice:/dev/tpm0 is added to the commands.


man/tpm2_createek.1.md
```diff
@@ -1,7 +1,7 @@
 % tpm2_createek(1) tpm2-tools | General Commands Manual
 
 # NAME
-
+https://www.aka55plus.de/materialien/Programm_35.pdf
```
idesai, line 4: @JuergenReppSIT something spurious left here

JuergenReppSIT: @idesai Thanks for the hint and sorry for adding this link by mistake.

AndreasFuchsTPM (approved):

PR #3293, opened 2023-10-06
Coverity scan

ajaykish:


williamcroberts:
merged in 4998ecfea and 0cb093f1d

Issue #1511, opened 2019-05-20
Remove transient object from TPM after context-saving


AndreasFuchsSIT:
IMHO, it would be good practice for the tools to remove transient objects from the TPM after they were saved and stored to disk. Since the kernel as well as the user-space resource managers will do so anyways, the functional flow will not change. On the opposite, this will allow tools to be used without a resource manager on a TPM or simulator directly.

IMHO even better would be, to conditionally save and flush transient object, only if a context output file was provided and leave them inside the TPM if none was provided.

From my analysis, this should be the easiest of fixes, by adding something like
```
if ((handle & TPM2_HR_RANGE_MASK) == TPM2_HR_TRANSIENT) {
    TSS2_RC rval = Esys_FlushContext(ectx, handle);
    if (rval != TPM2_RC_SUCCESS) {
        LOG_PERR(Eys_ContextFlush, rval);
        return false;
}
```
after
https://github.com/tpm2-software/tpm2-tools/blob/6163942372cd9da7ed6687730d0d814ded56df1b/lib/files.c#L218-L222

@williamcroberts Would you agree with either of the approaches ? Shall I send the PR ?

idesai:
@AndreasFuchsSIT 
1. Agree with flushing the transient object from TPM when saved to the disk. Question is does the resource manager try to keep track of such a handle that has been explicitly flushed by the application. Would it then:
a. Attempt to flush the handle again
b. Flush the handle# without realizing it was re-allocated to another application

2. IIUC this would mean making the output context optional in tools like load, createload, loadexternal. This approach would work only when a resource manager is not in play. Leaving any in the TPM would be categorically flushed by the resource manager.

AndreasFuchsSIT:
1. If something was flushed by the application, the RM sees that and seizes tracking that object; aka will not flush.
2. Yes. The intention was, if we want to somehow keep the current capability of the tools to create transient object inside the TPM if no RM is around. The reason being that I don't know, if people are relying on this feature. In that case, we don't want to take it away. But I cannot really say, whether it is being used.

Alternatively to 2., we could introduce a "--don't-flush" option.

idesai:
Agreed.  Another approach could be since tools know the TCTI it is talking to and can process missing command line option for the context file, i guess we could do something like:
if tcti!=RM and !tool.cmdline.contextfile
 <don't flush>
All other cases following a contextsave of the transient object we always flush.

AndreasFuchsSIT:
[...]

PR #3298, opened 2023-10-22
Save Context: Add the possibility for autoflush.

JuergenReppSIT:
If the environment variable TPM2TOOLS_AUTOFLUSH exists transient objects will be removed after they were saved and stored to disk. Also a transient parent will be removed if a context file for this parent was used by this command.
For the commands which will check the autoflush also an option -R is added to enable the autoflush independent from the environment variable.
Addresses: #1511

### DIFF
--- a/lib/files.h
+++ b/lib/files.h
@@ -84,12 +84,14 @@ bool files_save_bytes_to_file(const char *path, UINT8 *buf, UINT16 size);
  *  The object handle for the object to save.
  * @param path
  *  The output path of the file.
+ * @param autoflush
+ *  Flush the tpm object after context save.
  *
  * @return
  *  tool_rc indicating status.
  */
 tool_rc files_save_tpm_context_to_path(ESYS_CONTEXT *context, ESYS_TR handle,
-        const char *path);
+        const char *pathm, bool autoflush);
 
 /**
  * Like files_save_tpm_context_to_path() but saves a tpm session to a FILE stream.
@@ -99,11 +101,13 @@ tool_rc files_save_tpm_context_to_path(ESYS_CONTEXT *context, ESYS_TR handle,
  *  The object handle for the object to save.
  * @param stream
  *  The FILE stream to save too.
+ * @param autoflush
+ *  Flush the tpm object after context save.
  * @return
  *  tool_rc indicating status.
  */
 tool_rc files_save_tpm_context_to_file(ESYS_CONTEXT *context, ESYS_TR handle,
-        FILE *stream);
+        FILE *stream, bool autoflush);
 
 /**
  * Loads a ESAPI TPM object context from disk or an ESAPI serialized ESYS_TR object.
--- a/lib/tpm2_util.c
+++ b/lib/tpm2_util.c
@@ -632,6 +632,14 @@ char *tpm2_util_getenv(const char *name) {
     return getenv(name);
 }
 
+bool tpm2_util_env_yes(const char *name) {
+
+    char *value = getenv(name);
+    return (value && (strcasecmp(name, "yes") == 0 ||
+                      strcasecmp(name, "1") == 0 ||
+                      strcasecmp(name, "true") == 0));
+}
+
 /**
  * Parses a hierarchy value from an option argument.
  * @param value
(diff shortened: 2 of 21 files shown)


cplappert (approved):

man/common/options.md
```diff
@@ -27,3 +27,8 @@ information that many users may expect.
     Enable the application of errata fixups. Useful if an errata fixup needs to be
     applied to commands sent to the TPM. Defining the environment
     TPM2TOOLS\_ENABLE\_ERRATA is equivalent.
+  * **-R**, **\--autoflush**:
+    Enable autoflush for transient objects created by the command. If a parent
+    object is loaded from a context file also the transient parent object will
+    be flushed. Autoflush can also be activated if the environment variable
+    TPM2TOOLS\_AUTOFLUSH is is set.
```
AndreasFuchsTPM, line 34: I would argue that we want "yes" or "true" as values that must be set.
What do you think ?

lib/tpm2_util.c
```diff
@@ -632,6 +632,13 @@ char *tpm2_util_getenv(const char *name) {
     return getenv(name);
 }
 
+bool tpm2_util_env_yes(const char *name) {
+
+    char *value = getenv(name);
+    return (value && (strcasecmp(name, "yes") == 0 ||
+                      strcasecmp(name, "true") == 0));
```
williamcroberts, line 639: also support "1"

lib/files.h
```diff
@@ -628,4 +632,16 @@ bool files_load_attest_file(FILE *f, const char *path, TPMS_ATTEST *attest);
 tool_rc files_load_unique_data(const char *file_path,
 TPM2B_PUBLIC *public_data);
 
+/**
+ * @brief
+ * Check whether file can be read
```
williamcroberts, line 637: this is introduces a Time of Check Time of Use Bug. The best way to detect if a file cannot be read is when fopen() fails when you go to read it. Why do we need this?

williamcroberts (changes requested):

lib/files.h
JuergenReppSIT: The idea for the check was that the flush should only be executed when a context file for the parent exists. If only the TPM handle for the parent was passed the flash should not be executed.
I wanted to avoid calling the close function after the file test. Therefore i added this function.

williamcroberts: OK so the problem is, a user passes in a transient handle not in a context file and under those conditions we don't flush it. I guess this is possible when an RM is not being used, did I grok this right?

I think the right fix is to look at the `tpm2_loaded_object` structure s path field, it should be NULL if I understand the flow in `tpm2_util_object_load2`, as it calls `tpm2_util_object_load2` only if the `fopen` succeeds and that function sets the objects path. So if that path is set, then it's a context file. You could also add an attribute to the object `bool is_ctx_file` and set that and then check that over checking the path. Might be more robust as code changes.

JuergenReppSIT: * yes you did grok it right. 
* I have added "1" to the possible values of the env variable.
* It's a good idea to check whether the parent path is a context file. Because there is no common structure for the context of the affected commands I would prefer to add the check to the file_exists function. To make the use of the function clearer, I have renamed the function to `ctx_file_exists`.

tools/misc/tpm2_encodeobject.c
```diff
@@ -186,6 +190,16 @@ static int encode(void) {
     }
 
     PEM_write_bio_TSSPRIVKEY_OBJ(bio, tpk);
+
+    if ((ctx.autoflush || tpm2_util_env_yes(TPM2TOOLS_ENV_AUTOFLUSH)) &&
+        ctx_file_exists(ctx.parent.ctx_path) &&
```
williamcroberts, line 195: @JuergenReppSIT no need to check `ctx_file_exists(ctx.parent.path)` here, because you can check `ctx.parent.object.path`. If that field is set than you know for a fact it is a context file and it existed when the tool was opened and avoids the race condition of the check and the filesystem. Suppose the tool has the context file unlinked from out and underneath it, we still want to flush the transient handle.

williamcroberts (changes requested):
Same comment as before but with more detail on the proper fix. No checking for paths existing.

tools/misc/tpm2_encodeobject.c
JuergenReppSIT: @williamcroberts Sorry for being a bit confused here ;-) You are right, the test of the path in  `tpm2_loaded_object` is sufficient.

AndreasFuchsTPM (approved):

PR #3304, opened 2023-11-20
test eventlog: fix check eventlog.sh if efivar.h exists

### LINKED ISSUE #3302: Failing test: test/integration/tests/eventlog.sh
Output is as follows:

```
FAIL: test/integration/tests/eventlog
=====================================

ERROR: Missing required positional parameter, try -h / --help
Usage: eventlog [<options>] <arguments>
Where <options> are:
    [ --eventlog-version=<value>]
ERROR: Unable to run eventlog
ERROR: Expected one file name as a positional parameter. Got: 2
ERROR: Unable to run eventlog
--- ./test/integration/fixtures/event-uefiservices.bin.yaml     2023-11-08 07:19:36.000000000 -0000
+++ event-uefiservices.bin.out  2023-11-17 21:07:51.699999766 -0000
@@ -37,7 +37,7 @@
     ImageLengthInMemory: 171464
     ImageLinkTimeAddress: 0x0
     LengthOfDevicePath: 52
-    DevicePath: '02010c00d041030a0000000001010600000201010600000004081800000000000026010000000000ffc30300000000007fff0400'
+    DevicePath: 'PciRoot(0x0)/Pci(0x2,0x0)/Pci(0x0,0x0)/Offset(0x12600,0x3c3ff)'
 pcrs:
   sha1:
     2  : 0x5b5f4d5c31664f01670a98a5796a36473671befc
YAML output matching ./test/integration/fixtures/event-uefiservices.bin.yaml changed, set TEST_REGENERATE_OUTPUT=1 to re-create
FAIL test/integration/tests/eventlog.sh (exit status: 1)
```

JuergenReppSIT:
If efivar.h exist a pretty print function for the DevicePath is executed. Therefore two yaml test files are needed for the bin test file uefiservices.
Fixes #3302.


AndreasFuchsTPM (changes requested):
We should instead make sure that the outputs are always the same, no matter if evivar.h is here or not.

AndreasFuchsTPM:
So if I understand this correctly:
- If efivar.h is present on the system, the tool outputs a pretty print.
- If efivar.h is not present on the system, the tool outputs a raw format print.
This is not acceptable, we need to decide on either pretty print or raw print or make this a CML-line option (e.g. --pretty is only available if efivar.h is present).

@JuergenReppSIT @williamcroberts Thoughts ?

JuergenReppSIT:
>       This is not acceptable, we need to decide on either pretty print or raw print or make this a CML-line option (e.g. --pretty is only available if efivar.h is present).
> 

> @JuergenReppSIT @williamcroberts Thoughts ?

@AndreasFuchsTPM  @williamcroberts none of the alternatives is really good. I would prefer the one with the CML option --pretty

williamcroberts:
@JuergenReppSIT or @AndreasFuchsTPM looks like FreeBSD in Cirrus is unhappy with:
```
0000: 00000000                          ....
This test failed due to a TPM bug regarding signed comparison as described
in TCG's Errata for TCG Trusted Platform Module Library Revision 1.59 Version 1.4,
Section 2.5 TPM_EO – two’s complement
```
Have you seen this before? If so what's the fix? update the tpm simulator binary?

JuergenReppSIT:
> Have you seen this before? If so what's the fix? update the tpm simulator binary?

@williamcroberts This problem was fixed with #3322. I didn't rebase the PR because I wanted to hear your opinion to the pretty print problem  first.

williamcroberts:
> > Have you seen this before? If so what's the fix? update the tpm simulator binary?
> 
> @williamcroberts This problem was fixed with #3322. I didn't rebase the PR because I wanted to hear your opinion to the pretty print problem first.

We just leave the behavior as is. While not desirable, we will fix for 6.0.0. I removed breaking change, because this isn't a breaking change.

Issue #3314, opened 2023-12-09
tcti handling is broken


williamcroberts:
A command that should work with no options would fail:
    
    ./tools/tpm2 nvdefine
    ERROR:esys:src/tss2-esys/api/Esys_GetCapability.c:164:Esys_GetCapability_Async() esyscontext is NULL.
    ERROR:esys:src/tss2-esys/api/Esys_GetCapability.c:83:Esys_GetCapability() Error in async function ErrorCode (0x00070005)
    ERROR: Esys_NV_ReadPublic(0x70005) - esapi:A pointer is NULL that isn't allowed to be NULL.
    ERROR: Unable to run nvdefine

PR #3315, opened 2023-12-09
options: fix TCTI handling

williamcroberts:
A command that should work with no options would fail:

./tools/tpm2 nvdefine
ERROR:esys:src/tss2-esys/api/Esys_GetCapability.c:164:Esys_GetCapability_Async() esyscontext is NULL. ERROR:esys:src/tss2-esys/api/Esys_GetCapability.c:83:Esys_GetCapability() Error in async function ErrorCode (0x00070005) ERROR: Esys_NV_ReadPublic(0x70005) - esapi:A pointer is NULL that isn't allowed to be NULL. ERROR: Unable to run nvdefine

The handling for the optional and fake TCTI states would get conflated and not initialize an ESAPI context. Clean up the handling and make supporting a fake TCTI it's own option bit to make it less confusing to code.

After this the tool works as expected:
./tools/tpm2 nvdefine
nv-index: 0x1000000

Fixes: #3314

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -434,44 +434,43 @@ tpm2_option_code tpm2_handle_options(int argc, char **argv,
          * SAPI
          */
         bool is_sapi =
-            (!tool_opts || !tool_opts->flags);
+            !tool_opts || !(tool_opts->flags & TPM2_OPTIONS_NO_SAPI);
 
-        /*
-         * NO_SAPI
-         */
-        bool is_no_sapi =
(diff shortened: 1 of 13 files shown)


JuergenReppSIT (changes requested):
lgtm.  Now nobody has to think about it anymore why is_sapi and is_no_sapi both can be false ;-)
In the commit message you did describe an error of the old version.
For me this command worked with the current master. So perhaps this part should be removed?

AndreasFuchsTPM (approved):

Issue #3295, opened 2023-10-19
tpm2_sessionconfig --disable-continuesession can't be used


JuergenReppSIT:
The following example produces an error because the session is flushed before save context is executed by `tpm2_create`:
```
 tpm2_createprimary -R -c primary.ctx
 tpm2_startauthsession  --hmac-session -S session.ctx
 tpm2_sessionconfig session.ctx --enable-decrypt --disable-continuesession
 tpm2_create  -S session.ctx -C primary.ctx -u key.pub -r key.priv -c key.ctx
```

AndreasFuchsTPM:
AI here: If continueSession == false, delete the session.ctx file at the end of command handling

PR #3320, opened 2023-12-17
tpm2_sessionconfig fix usage of --disable-continuesession

JuergenReppSIT:
If continue session was disabled a error did occur in the function for restoring the session context.
Now after usage of an session with continue session disabled the context will not be saved and the session context file will be deleted.
In one integration test continue session is now disabled and the flush for this session is removed.

Fixes: #3295

### DIFF
--- a/lib/tpm2_session.c
+++ b/lib/tpm2_session.c
@@ -35,6 +35,7 @@ struct tpm2_session {
         char *path;
         ESYS_CONTEXT *ectx;
         bool is_final;
+        bool delete;
     } internal;
 };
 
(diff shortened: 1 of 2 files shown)


williamcroberts (approved):
I didn't like the delete idea at first, but it grew on me

williamcroberts:
> I didn't like the delete idea at first, but it grew on me

Dang it, I shouldn't have let that happen. Actually @JuergenReppSIT and @AndreasFuchsTPM we can't delete it right now, as that would be a change of behavior. We just avoid the savecontext and in the next major we can bump it. See #3324, we can just git revert that for 6.0

Issue #3060, opened 2022-07-15
Tests error in PPC64


aplanas:
I am running the tests (units and integration) under a PPC64
architecture and I have errors in 4 tests.

I create the issue to report them and maybe work on them in separate
issues later.

```
[  322s] FAIL: test/unit/test_tpm2_header
[  322s] ================================
[  322s] 
[  322s] [==========] Running 2 test(s).
[  322s] [ RUN      ] test_tpm_command_header
[  322s] [  ERROR   ] --- c->tag == 0x0180
[  322s] [   LINE   ] --- test/unit/test_tpm2_header.c:23: error: Failure!
[  322s] [  FAILED  ] test_tpm_command_header
[  322s] [ RUN      ] test_tpm_response_header
[  322s] [  ERROR   ] --- r->tag == 0x0180
[  322s] [   LINE   ] --- test/unit/test_tpm2_header.c:99: error: Failure!
[  322s] [  FAILED  ] test_tpm_response_header
[  322s] [==========] 2 test(s) run.
[  322s] [  PASSED  ] 0 test(s).
[  322s] [  FAILED  ] 2 test(s), listed below:
[  322s] [  FAILED  ] test_tpm_command_header
[  322s] [  FAILED  ] test_tpm_response_header
[  322s] 
[  322s]  2 FAILED TEST(S)
[  322s] FAIL test/unit/test_tpm2_header (exit status: 2)
```

```
[  322s] FAIL: test/integration/tests/abrmd_policyticket
[  322s] ===============================================
[  322s] 
[  322s] creating simulator working dir: /tmp/tpm2_test_KiMy7R
[  322s] /tmp/tpm2_test_KiMy7R ~/rpmbuild/BUILD/tpm2-tools-5.2
[  322s] Switched to CWD: /tmp/tpm2_test_KiMy7R
[  322s] Starting the simulator
[  322s] Attempting to start simulator on port: 49151
[  322s] LISTEN 0      0            0.0.0.0:49151      0.0.0.0:*    users:(("tpm_server",pid=7624,fd=6))
[  322s] LISTEN 0      0            0.0.0.0:49152      0.0.0.0:*    users:(("tpm_server",pid=7624,fd=8))
[  322s] Started simulator on port 49151 in dir "/tmp/tpm2_test_KiMy7R"
[  322s] tpm2tools_tcti="mssim:host=localhost,port=49151"
[  322s] Started the simulator
[  322s] Starting tpm2-abrmd
[...]

PR #3323, opened 2023-12-20
test/unit: fix tpm2_header test on BE machines

williamcroberts:
Fixes the test on BE machines by making the pointers const and thus not needing to check if underlying function is modifying things by checking the expected LE output of casting a BE scalar. Make it all const interface and let the compiler enforce that idiom.


williamcroberts:
@aplanas can you check this commit? Partial fix for #3060

aplanas (approved):
Interesting fix

PR #3331, opened 2024-01-08
files_load_unique_data: fix unique data copy to ecc point

gomesj:
Split of unique data into x and y components of ecc point currently fumbles the indexes leading to incomplete copy.
Using "IAK" (size=3) as input, this leads to:
> x.size = (unique_size / 2) = 1
> x.buffer = unique_size[0,1[ = ['I']
> y.size = (unique_size / 2) + (unique_size % 2) = 2
> y.buffer = unique_size[1,2[ + 0 = ['A', 0]

Change copy bounds to copy "bigger" first half in x and remainder in y to reach a more reasonable result:
> x.size = (unique_size / 2) + (unique_size % 2) = 2
> x.buffer = unique_size[0,2[ = ['I', 'A']
> y.size = (unique_size / 2) + (unique_size % 2) = 2
> y.buffer = unique_size[2,3[ + 0 = ['K', 0]

### DIFF
--- a/lib/files.c
+++ b/lib/files.c
@@ -843,20 +843,19 @@ tool_rc files_load_unique_data(const char *file_path, TPM2B_PUBLIC *public_data)
 
     /* Unstructured stdin unique data paired with an ECC object */
     if (public_data->publicArea.type == TPM2_ALG_ECC) {
+        int pivot = (unique_size / 2) + (unique_size % 2);
         /* The size limitation is TPM implementation dependent. */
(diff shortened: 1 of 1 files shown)


AndreasFuchsTPM:
I think this makes a lot of sense.

BUT, this looks like a change in behavior, i.e. different key derivation and such.
So shall we postpone this until 6.0 ?

@williamcroberts Your thoughts ?

PR #3337, opened 2024-01-11
Add pcrauth status

### LINKED ISSUE #3330: Getting PCR properties with tpm2_getcap
It would be useful to display the different PCR properties, such as which PCRs can have an auth value set.

idesai:
Fixes #3330


idesai:
TODO: 

- [ ] Need to add a test - Must implement #3333

AndreasFuchsTPM:
Hmmmmm... you know that pcrsetauthvalue is forbidden for all PTP compatible TPMs, right ?
@idesai What's the use case ?

AndreasFuchsTPM:
Please add big warnings to the manpage and also to the error printf's that this is not included in PC-Client compatible TPM; indicating that this is actually forbidden for PTP compliant TPMs.

PR #3345, opened 2024-01-31
Auth input: Check for NULL termination

AndreasFuchsTPM:
When reading auths from a file, add a check that they are NULL terminated.

### DIFF
--- a/lib/tpm2_auth_util.c
+++ b/lib/tpm2_auth_util.c
@@ -63,12 +63,14 @@ static tool_rc get_auth_for_file_param(const char* password, TPM2B_AUTH *auth) {
     const char* path = password;
     size_t size = (sizeof(auth->buffer) * 2) + HEX_PREFIX_LEN + 2;
     bool is_a_tty = isatty(STDIN_FILENO);
-    UINT8 *buffer = calloc(1, size);
 
+    /* Allocating one extra byte for \0 termination safety */
+    UINT8 *buffer = calloc(1, size + 1);
     if (!buffer) {
         LOG_ERR("oom");
         return tool_rc_general_error;
     }
+    buffer[size] = '\0';
 
     if (path) {
         path = strcmp("-", path) ? path : NULL;


JuergenReppSIT (changes requested):
Here it's possible that if the password is read from the file that there is no \0 char at the end of the buffer. It would be better to reallocate the buffer with size+1 after  files_load_bytes_from_buffer_or_file_or_stdin and add the \0.

AndreasFuchsTPM:
> Here it's possible that if the password is read from the file that there is no \0 char at the end of the buffer. It would be better to reallocate the buffer with size+1 after files_load_bytes_from_buffer_or_file_or_stdin and add the \0.

You mean like this ?

JuergenReppSIT:
> > Here it's possible that if the password is read from the file that there is no \0 char at the end of the buffer. It would be better to reallocate the buffer with size+1 after files_load_bytes_from_buffer_or_file_or_stdin and add the \0.
> 
> You mean like this ?

Yes, but I think the coverity error is a false positive. Now I saw: the buffer is already allocated with calloc an the size passed do `files_load_bytes_from_buffer_or_file_or_stdin` is decremented. So I would say we can close this PR. No change is needed.

AndreasFuchsTPM:
> > > Here it's possible that if the password is read from the file that there is no \0 char at the end of the buffer. It would be better to reallocate the buffer with size+1 after files_load_bytes_from_buffer_or_file_or_stdin and add the \0.
> > 
> > 
> > You mean like this ?
> 
> Yes, but I think the coverity error is a false positive. Now I saw: the buffer is already allocated with calloc an the size passed do `files_load_bytes_from_buffer_or_file_or_stdin` is decremented. So I would say we can close this PR. No change is needed.

Oh yes, I see this now.
However, calloc does not guarantee \0 initialization. So I guess the line `buffer[size-1] = '\0';` should stay, right ?

JuergenReppSIT (approved):

PR #3440, opened 2024-12-09
tpm2_getekcertificate: add support to high range NV indexes 

### LINKED ISSUE #3435: tpm2_getekcertificate should support values stored in high range (NVRAM)
By default, *tpm2_getekcertificate* search EK cert in NVRAM (if there is no search **ARGUMENT**).

The latest [TCG spec](https://trustedcomputinggroup.org/wp-content/uploads/PC-Client-Specific-Platform-TPM-Profile-for-TPM-2p0-v1p05p_r14_pub.pdf) encourages vendors to support ECC P384 and RSA 3072 EK certificates, as stated below:

```
1. If the TPM is pre-provisioned with EK Certificates, it SHALL be provisioned with EK Certificates for an RSA
EK and an ECC EK.
    a. The TPM MAY include an EK Certificate for an RSA 3072-bit key.
    b. The TPM SHALL be provisioned with either an ECC P384 EK Certificate or an RSA 3072-bit EK
    Certificate.
    c. All Certificates SHALL comply with the definition in the TCG EK Credential Profile for TPM Family
    2.0 [14]. 
```
_source: [4.5.2.1 Pre-provisioned EK Certificates](https://trustedcomputinggroup.org/wp-content/uploads/PC-Client-Specific-Platform-TPM-Profile-for-TPM-2p0-v1p05p_r14_pub.pdf#page=21)_

The latest *[TCG Credential Profile EK 2.0](https://trustedcomputinggroup.org/wp-content/uploads/TCG-EK-Credential-Profile-V-2.5-R2_published.pdf)* defines where the newest algorithms must be stored (ie. *High Range*):

```
The handle values are normative as of version 2.3 of this specification. Defining the NV
index handles as normative simplifies EK certificate validation because it eliminates parsing of
certificate content in order to identify the algorithm (RSA or ECC) and key/curve size.

      0x01c00012 RSA 2048 EK Certificate (H-1)
      0x01c00014 ECC NIST P256 EK Certificate (H-2)
      0x01c00016 ECC NIST P384 EK Certificate (H-3)
      0x01c00018 ECC NIST P521 EK Certificate (H-4)
      0x01c0001a ECC SM2_P256 EK Certificate (H-5)
      0x01c0001c RSA 3072 EK Certificate (H-6)
      0x01c0001e RSA 4096 EK Certificate (H-7)
```

_source: [2.2.1.5 High Range](https://trustedcomputinggroup.org/wp-content/uploads/TCG-EK-Credential-Profile-V-2.5-R2_published.pdf#page=14)_

It seems that the current implementation of `tpm2_getekcertificate` only search in [Low Range](https://trustedcomputinggroup.org/wp-content/uploads/TCG-EK-Credential-Profile-V-2.5-R2_published.pdf#page=13) (see code [here](https://github.com/tpm2-software/tpm2-tools/blob/master/tools/tpm2_getekcertificate.c#L99:L103)). The tool should also support *High Range* values in order to comply with the latest version of TCG spec.

PS: I'm willing to do a PR on this issue if this ticket is relevant to maintainers.

loicsikidi:
Fixes #3435.

_Note_:

* The test case uses self-signed certificates because I didn't found a valid one in `https://ekop.intel.com` (e.g. RSA 3072 or ECC P384)

### DIFF
--- a/test/integration/tests/getekcertificate.sh
+++ b/test/integration/tests/getekcertificate.sh
@@ -71,6 +71,7 @@ tpm2 loadexternal -C e  -u test_rsa_ek.pub -c rsa_key.ctx
 tpm2 readpublic -c rsa_key.ctx -f pem -o test_rsa_ek.pem
 openssl x509 -pubkey -in rsa_ek_cert.bin -noout -out test_ek.pem
 diff test_rsa_ek.pem test_ek.pem
+tpm2 flushcontext -t
 
 # Sample ECC ek public from a real platform
 echo "007a0023000b000300b20020837197674484b3f81a90cc8d46a5d724fd52
@@ -90,6 +91,7 @@ tpm2 loadexternal -C e  -u test_ecc_ek.pub -c ecc_key.ctx
 tpm2 readpublic -c ecc_key.ctx -f pem -o test_ecc_ek.pem
 openssl x509 -pubkey -in ecc_ek_cert.bin -noout -out test_ek.pem
 diff test_ecc_ek.pem test_ek.pem
+tpm2 flushcontext -t
 
 # Retrieve EK certificates from NV indices
 RSA_EK_CERT_NV_INDEX=0x01C00002
@@ -170,14 +172,15 @@ tpm2 getekcertificate -o nv_rsa_ek_cert.der -o nv_ecc_ek_cert.der
 diff nv_rsa_ek_cert.der rsa_ek_cert.der
 diff nv_ecc_ek_cert.der ecc_ek_cert.der
 
-rm nv_rsa_ek_cert.der rsa_ek_cert.der nv_ecc_ek_cert.der ecc_ek_cert.der priv_key.pem -f
+rm nv_rsa_ek_cert.der nv_ecc_ek_cert.der -f
 
 ## Make sure that if there are several certificates of the same type, then the one belonging to low range has priority
 openssl x509 -in ecc_ek_cert.bin -out ecc_low_range_ek_cert.der -outform DER
 define_ek_cert_nv_index ecc_low_range_ek_cert.der $ECC_EK_CERT_NV_INDEX
 
-tpm2 getekcertificate -o nv_ecc_ek_cert.der
+tpm2 getekcertificate -o nv_rsa_ek_cert.der -o nv_ecc_ek_cert.der
 
 diff nv_ecc_ek_cert.der ecc_low_range_ek_cert.der
+diff nv_rsa_ek_cert.der rsa_ek_cert.der
 
 exit 0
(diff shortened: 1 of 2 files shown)


JuergenReppSIT:
@loicsikidi your PR would change the behaviour of `tpm2_getekcertificate` for TPMs with certificates in the low and in the high range (e.g. Infineon SLB 9672). Instead of the low range certificates the high range certificates would be saved.
It would be better to implement the option `-u pubkey` which currently can't be used for certificates in the NV ram.

JuergenReppSIT (changes requested):
If you would change the lookup part as follows:
```
    ctx.rsa_ek_cert_nv_location = 0xffffffff;
    ctx.ecc_ek_cert_nv_location = 0xffffffff;

    for (i = 0; i < capability_data->data.handles.count; i++) {
        TPMI_RH_NV_INDEX index = capability_data->data.handles.handle[i];
        const ek_index_map *m = lookup_ek_index_map(index);
        if (!m) {
            continue;
        }

        if (m->key_type == KTYPE_RSA && index < ctx.rsa_ek_cert_nv_location) {
            LOG_INFO("Found pre-provisioned RSA EK certificate at %u [type=%s]", index, m->name);
            ctx.is_rsa_ek_cert_nv_location_defined = true;
            ctx.rsa_ek_cert_nv_location = m->index;
        }
        if (m->key_type == KTYPE_ECC && index < ctx.ecc_ek_cert_nv_location) {
            LOG_INFO("Found pre-provisioned ECC EK certificate at %u [type=%s]", index, m->name);
            ctx.is_ecc_ek_cert_nv_location_defined = true;
            ctx.ecc_ek_cert_nv_location = m->index;
        }
    }
```
it's ensured  that the certificate would be written as previously. The high range certificates would only be written if no low level certificates exist. After your PR is merges I would add a PR with the -u option to enable saving the EK certificate based on the public EK.
I think that's an better solution than allowing e.g. four -o options to save all certificates.

loicsikidi:
@JuergenReppSIT I've added your suggestion and one more test to attest that *if there are >= 2 certs of the same type => the low range one must be returned.*

Thanks for the feedback!

I wasn't aware that some TPMs have more than one EK cert from the same family :sweat_smile:

JuergenReppSIT (approved):
@loicsikidi Thank you for the PR. I have already prepared the -u change.

loicsikidi:
@JuergenReppSIT I think that I've found the issue :bulb: .

The last test is failing because there are 2 "eligible" EK certs in NVRAM but there is only one `-o`... 

```bash
## Make sure that if there are several certificates of the same type, then the one belonging to low range has priority
openssl x509 -in ecc_ek_cert.bin -out ecc_low_range_ek_cert.der -outform DER
define_ek_cert_nv_index ecc_low_range_ek_cert.der $ECC_EK_CERT_NV_INDEX

tpm2 getekcertificate -o nv_ecc_ek_cert.der

diff nv_ecc_ek_cert.der ecc_low_range_ek_cert.der
```

ERROR:

> WARN: Found 2 certficates on NV. Add another -o to save the ECC cert

My bad..., I'll push a fix :rocket:

loicsikidi:
I think that we can retry :muscle:

JuergenReppSIT:
@loicsikidi Thank you for the fix.

PR #3471, opened 2025-04-15
update rsa2048 and ecc_nist_p256 to policy_b

arianvp:
Pretty sure this was incorrect.  All the EK templates in the High Range should use Policy B; not Policy A

See B.4.5 Template H-2: ECC NIST P256 (Storage)

for source https://trustedcomputinggroup.org/wp-content/uploads/EK-Credential-Profile-For-TPM-Family-2.0-Level-0-V2.5-R1.0_28March2022.pdf

Maybe we should come up with better names for these

e.g.   `ecc-low` and `ecc-high` ?

### DIFF
--- a/tools/tpm2_createek.c
+++ b/tools/tpm2_createek.c
@@ -136,14 +136,14 @@ struct alg_map {
 
 static const alg_map alg_maps[] = {
     { "rsa",           "rsa2048:aes128cfb", "sha256", &policy_a_sha256, ATTRS_A },
-    { "rsa2048",       "rsa2048:aes128cfb", "sha256", &policy_a_sha256, ATTRS_A },
+    { "rsa2048",       "rsa2048:aes128cfb", "sha256", &policy_b_sha256, ATTRS_B },
     { "rsa3072",       "rsa3072:aes256cfb", "sha384", &policy_b_sha384, ATTRS_B },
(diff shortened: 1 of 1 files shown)


tools/tpm2_createek.c
```diff
@@ -136,14 +136,14 @@ struct alg_map {
 
 static const alg_map alg_maps[] = {
     { "rsa",           "rsa2048:aes128cfb", "sha256", &policy_a_sha256, ATTRS_A },
-    { "rsa2048",       "rsa2048:aes128cfb", "sha256", &policy_a_sha256, ATTRS_A },
+    { "rsa2048",       "rsa2048:aes128cfb", "sha256", &policy_b_sha256, ATTRS_B },
```
arianvp, line 139: Perhaps instead we want to introduce a `rsa2048_high` and `etc_nist_p256_high` alias?

JuergenReppSIT:
@arianvp your pr will not compile because policy_b_sha256 is currently not defined.  And as you wrote the best solution would be not to delete the old entries in alg_map but add `rsa2048_high` and etc_nist_p256_high` to alg_map.

JuergenReppSIT:
@arianvp Thank you for reporting the problem. I have added a PR based on your suggestion.

PR #3481, opened 2025-05-02
tpm2_getekcertificate: add option --x509-trunc

saurontech:
New option to handle x509 DER with trailing data

ST micro's ST33HTPHF2ENIST TPM creates NIST P-256 ECC and RSA EK certificates with 0xff data padded to a fixed 1600 bytes length. The option "--x509-trunc" or "-t" parses the output of "nv_read" with a X509 parser and truncates the trailing data. This makes the output formate more accessable to GOLANG's X509.ParseCertificate(). Details about this issue can be found at: tpm2-software#3474

Signed-off-by: Samuel Cheng <samuelogre@gmail.com>


JuergenReppSIT:
@saurontech lgtm. Could you please add the new option to ./man/tpm2_getekcertificate.1.md?

PR #3525, opened 2025-11-12
rc_decode.sh: Fix pattern matching for new tss2_tpm2_types.h.

JuergenReppSIT:
The format of tss2_tpm2_types.h had bin changed with clang-format. The pattern related to the RC defines had to be adapted.


AndreasFuchsTPM:
I wonder, how we can synchronize this with the tpm2-tss relase.
Do we make a tpm2-tools release first or release both at the same time ?

JuergenReppSIT:
The fix works for the current TSSmaster and also for older versions.  For the current test TSS master and  4.1.3 were tested.

PR #3547, opened 2026-02-10
Allocate the right amount of memory for auth_policy

AndreasFuchsTPM:


JuergenReppSIT (approved):

moritzbuhl:
There is also line 115: `(ctx.auth_policy)->size = UINT16_MAX;`.

PR #3557, opened 2026-02-19
tpm2_attr_util: tighten token parsing in handle_dispatch to prevent false prefix matches

hyperfinitism:
This PR fixes the token parsing logic in `tpm2_attr_util.c` to prevent false matches (e.g. `name="foo"` vs `token="foobar..."`) and spurious `dispatch_err` returns during dispatch table iteration.

- Inline `token_match()` into `handle_dispatch()`. This helper was file-local (no prototype in a public header), so this is not a breaking API change.
- Require an exact match for no-arg tokens (`strcmp(token, name)`). The previous implementation could incorrectly accept prefix matches.
- Require arg tokens to be in the strict `"name=arg"` form to avoid prefix matches. The previous implementation could incorrectly accept prefix matches.
- Split tokens on `'='` only after a confirmed match so non-matching table entries do not fail parsing.

This also silences `discarded-qualifiers` warnings/errors (related: #3556).

### DIFF
--- a/lib/tpm2_attr_util.c
+++ b/lib/tpm2_attr_util.c
@@ -381,21 +381,6 @@ static dispatch_table obj_attr_table[] = {         // Bit Index
         dispatch_reserved(31),                     // 31
 };
 
-static bool token_match(const char *name, char *token, bool has_arg,
-        char **sep) {
-
-    /* if it has an argument, we expect a separator */
-    size_t match_len = strlen(token);
-    if (has_arg) {
-        *sep = strchr(token, '=');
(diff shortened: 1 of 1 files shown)


JuergenReppSIT:
@hyperfinitism your change would produce a CLI break. 
e.g. the following command, which works in the current version , would produce an error:

` tpm2_createprimary -c primary.ctx -a "fixedt|fixedparent|sensitivedataorigin|userwithauth|restricted|decrypt"

( `fixedtpm` is replaced with `fixedt`)

PR #3573, opened 2026-04-15
tpm2_getekcertificate: Improve determination of the Intel certificate.

### LINKED ISSUE #3559: tpm2_getekcertificate: Intel EK cert parsing assumes JSON, but Intel endpoint returns XML
## Summary

The current Intel EK certificate parsing logic appears to be built around a JSON response format. However, the Intel EK cert service endpoint used by the [integration test](https://github.com/tpm2-software/tpm2-tools/blob/29e23a6e51d82434073797efece68500ad301024/test/integration/tests/getekcertificate.sh) returns XML. Additionally, the parser uses a hard-coded pointer offset *with a typo that changes the offset length*. Despite this, the integration test currently passes, seemingly due to accidental alignment rather than correct parsing.

### 1. Current implementation assumes a JSON response

The code path for Intel certificates is designed around a JSON payload. 

- It may check whether the response starts with the JSON prefix `{"pubhash`.
  https://github.com/tpm2-software/tpm2-tools/blob/29e23a6e51d82434073797efece68500ad301024/tools/tpm2_getekcertificate.c#L1200-L1203

- It locates the substring `"certificate"` using `strstr()` (which returns a pointer to the first 'c').
  https://github.com/tpm2-software/tpm2-tools/blob/29e23a6e51d82434073797efece68500ad301024/tools/tpm2_getekcertificate.c#L1217-L1218

- It then advances that pointer by a fixed offset corresponding to the expected JSON prefix `"certificate" : ` to reach the certificate value, and proceeds to decode/unescape it.
  https://github.com/tpm2-software/tpm2-tools/blob/29e23a6e51d82434073797efece68500ad301024/tools/tpm2_getekcertificate.c#L1159-L1161

### 2. The Intel EK cert service endpoint actually returns XML

The Intel endpoint used by the integration test `https://ekop.intel.com/ekcertservice/<pubhash>` returns an XML response of the form:

```xml
<RetrievalResponse>
  <pubhash>WVEG2rRwkQ7m3RpXlUphgo6Y2HLxl18h6ZZkkOAdnBE%3D</pubhash>
  <certificate>MIIE...</certificate>
</RetrievalResponse>
```

This does not match the JSON format the parser is written for.

Example URL: <https://ekop.intel.com/ekcertservice/WVEG2rRwkQ7m3RpXlUphgo6Y2HLxl18h6ZZkkOAdnBE%3D>

### 3. The hard-coded offset contains a typo (length mismatch)

The pointer is currently advanced by the length of `certficate\" : ` (note the missing 'i' in "certificate"). See the code quoted the previous section.

This offset length differs from what the code intends (a JSON prefix based on "certificate"). Nevertheless, the integration test passes today, apparently due to coincidence (i.e., the incorrect offset happens to land at a workable position for the XML response/whitespace layout).

## Why this matters

- The parsing is fragile and appears to rely on undefined/accidental behavior.
- Fixing the typo to the "correct" string changes behavior and causes the integration test to fail, indicating the parsing is not robust and likely not actually parsing the intended structure.
- Depending on response formatting (whitespace/newlines) or service changes, this could break unexpectedly.

## Suggested fix

- Stop using hard-coded `strlen()`-based pointer offsets.
- Detect/handle XML vs JSON explicitly (or always support XML for the Intel service), and extract the certificate payload using proper delimiters:
  - XML: extract between `<certificate>` and `</certificate>` (trim whitespace)
  - JSON: locate the `"certificate"` field and parse the value without fixed offsets
- Add NULL checks/error handling when expected fields are missing.

JuergenReppSIT:
Improve pointer determination for the Intel certificate field, which previously worked by coincidence despite a typo. Fixes: #3559

### DIFF
--- a/tools/tpm2_getekcertificate.c
+++ b/tools/tpm2_getekcertificate.c
@@ -1167,7 +1167,22 @@ static tool_rc process_input(ESYS_CONTEXT *ectx) {
 
 static char *base64_decode(char **split, unsigned int cert_length) {
 
-    *split += strlen("certficate\" : ");
+    *split += strlen("certificate\"");
+    while (*split && **split == ' ')
+        ++*split;
+
+    if (**split != ':')
+        return NULL;
+    ++*split;
+
+    while (*split && **split == ' ')
+        ++*split;
+
+    if (**split != '"')
+        return NULL;
+
+    ++*split;
+
     char *final_string = NULL;
     int outlen;
     CURL *curl = curl_easy_init();


JuergenReppSIT:
was replaced by #3574

PR #3585, opened 2026-05-22
Codechecker fixes

moritzbuhl:


tools/misc/tpm2_print.c
```diff
@@ -301,10 +301,12 @@ static bool print_TPMS_CONTEXT(FILE *fstream) {
      */
     UINT32 version;
```
moritzbuhl: `version = 0` could also work. Any preferences?

JuergenReppSIT: `version` should be initalized by `files_read_header` so wouldn't need to be initialized?

moritzbuhl: If `check_magic` in `files_read_header`fails, `version` is uninitialized.
Which would mean `files_read_32` and `readx` failed.

JuergenReppSIT: OK so perhaps we should initialize version with 0 and remove  ` if (!old_format) ` after `print_context:`?

JuergenReppSIT (approved):

Issue #3521, opened 2025-10-23
tpm2_unseal fails when both --pwd-session and -p pcr:sha256:<pcrs> are used


matthewruffell:
Hi @JuergenReppSIT 

I found a minor issue in tpm2_unseal with --pwd-session and -p pcr:sha256:<pcrs> when using both at the same time, for use on FIPS systems, but is also reproduces on normal non-FIPS systems.

On current master branch, works on normal non-FIPS systems:

Reproducer:

```
$ sudo apt install clevis clevis-tpm2 clevis-initramfs clevis-luks
$ mkfs.ext4 /dev/vdc
$ cryptsetup luksFormat --type luks2 /dev/vdc
$ sudo clevis luks bind -d /dev/vdc tpm2 '{"hash":"sha256","pcr_bank":"sha256","key":"rsa","pcr_ids":"0,2,4,7"}}'
$ clevis luks unlock -d /dev/vdc
...
+ pcr_ids=0,2,4,7
+ pcr_spec=
+ '[' -n 0,2,4,7 ']'
++ jose fmt -j- -Og clevis -g tpm2 -g pcr_bank -Su-
+ pcr_bank=sha256
+ pcr_spec=sha256:0,2,4,7
+ jose b64 dec -i- -O /tmp/tmp.WfP6ZPfRwk/jwk.pub
+ jose b64 dec -i- -O /tmp/tmp.WfP6ZPfRwk/jwk.priv
+ case "$TPM2TOOLS_VERSION" in
+ tpm2_createprimary --pwd-session -Q -C o -g sha256 -G rsa -c /tmp/tmp.WfP6ZPfRwk/primary.context
+ '[' -n '' ']'
+ case "$TPM2TOOLS_VERSION" in
+ tpm2_load --pwd-session -Q -C /tmp/tmp.WfP6ZPfRwk/primary.context -u /tmp/tmp.WfP6ZPfRwk/jwk.pub -r /tmp/tmp.WfP6ZPfRwk/jwk.priv -c /tmp/tmp.WfP6ZPfRwk/load.context
+ '[' -n '' ']'
+ case "$TPM2TOOLS_VERSION" in
++ tpm2_unseal --pwd-session -c /tmp/tmp.WfP6ZPfRwk/load.context -p pcr:sha256:0,2,4,7
ERROR: Cannot specify password type "pcr:"
ERROR: Invalid item handle authorization
ERROR: Unable to run tpm2_unseal
+ jwk=
[...]

PR #3587, opened 2026-06-10
Use password based sessions if no other session types specified

matthewruffell:
This PR fixes the problem where you currently can't specify authentication with ESYS_TR_PASSWORD by the --pwd-session (-z) parameter, and session / file / pcr based auth at the same time.

This is especially useful when --pwd-session is enabled by default in FIPS based environments for all tpm2-tools commands, but we want to disable it when users specify --auth=session: etc.

This lets combinations like the following work:

Before:

$ sudo tpm2_nvread --pwd-session --auth=session:session.ctx 0x1500016 
WARN: Reading full size of the NV index
ERROR: Cannot specify password type "session:"

After:

$ sudo tpm2_nvread --pwd-session --auth=session:session.ctx 0x1500016 
WARN: Reading full size of the NV index
ERROR: Could not open path "session.ctx", due to error: "No such file or directory"

This fixes issue #3521.

I know the code duplicates SESSION_PREFIX, FILE_PREFIX, PCR_PREFIX from lib/tpm2_auth_util.c, but I wanted to get your feedback on the idea before we start moving these definitions into common header files.

### DIFF
--- a/lib/object.c
+++ b/lib/object.c
@@ -1,4 +1,5 @@
 #include <stdio.h>
+#include <string.h>
 
 #include "files.h"
 #include "log.h"
@@ -10,6 +11,15 @@
 
 #define NULL_OBJECT "null"
 #define NULL_OBJECT_LEN (sizeof(NULL_OBJECT) - 1)
+
+#define SESSION_PREFIX "session:"
+#define SESSION_PREFIX_LEN sizeof(SESSION_PREFIX) - 1
+
+#define FILE_PREFIX "file:"
+#define FILE_PREFIX_LEN sizeof(FILE_PREFIX) - 1
+
+#define PCR_PREFIX "pcr:"
+#define PCR_PREFIX_LEN sizeof(PCR_PREFIX) - 1
 
 TPM2B_PRIVATE tpm2_util_object_tsspem_priv = { 0 };
 TPM2B_PUBLIC tpm2_util_object_tsspem_pub = { 0 };
@@ -411,6 +421,12 @@ tool_rc tpm2_util_object_load_auth(ESYS_CONTEXT *ctx, const char *objectstr,
         const char *auth, tpm2_loaded_object *outobject,
         bool is_restricted_pswd_session, tpm2_handle_flags flags) {
 
+    bool is_session = auth && !strncmp(auth, SESSION_PREFIX, SESSION_PREFIX_LEN);
+    bool is_file = auth && !strncmp(auth, FILE_PREFIX, FILE_PREFIX_LEN);
+    bool is_pcr = auth && !strncmp(auth, PCR_PREFIX, PCR_PREFIX_LEN);
+    bool use_restricted_pswd_session = is_restricted_pswd_session &&
+        !is_session && !is_file && !is_pcr;
+
     return tpm2_util_object_load2(ctx, objectstr, auth, true, outobject,
-            is_restricted_pswd_session, flags);
+        use_restricted_pswd_session, flags);
 }


lib/object.c
```diff
@@ -11,6 +12,15 @@
 #define NULL_OBJECT "null"
 #define NULL_OBJECT_LEN (sizeof(NULL_OBJECT) - 1)
 
+#define SESSION_PREFIX "session:"
+#define SESSION_PREFIX_LEN sizeof(SESSION_PREFIX) - 1
```
moritzbuhl, line 16: wrap it in braces! Otherwise something like `SESSION_PREFIX_LEN * 2` would result in `sizeof(SESSION_PREFIX) - 2`.
The others too.

JuergenReppSIT:
Currently, there is no way to enable --pwd-session globally for tpm2_tools (e.g., via an environment variable). Why not just omit the --pwd-session parameter?

matthewruffell:
> Currently, there is no way to enable --pwd-session globally for tpm2_tools (e.g., via an environment variable). Why not just omit the --pwd-session parameter?

When FIPS 140-3 is enabled, tpm2-tools really isn't all that useful, since every single command looks a lot like this:

```
# ./tools/tpm2 clear
ERROR:esys_crypto:src/tss2-esys/esys_crypto_ossl.c:414:iesys_cryptossl_hmac_start() OpenSSL reported: error:00000000:lib(0)::reason(0)
ERROR:esys_crypto:src/tss2-esys/esys_crypto_ossl.c:414:iesys_cryptossl_hmac_start() ErrorCode (0x00070001) DigestSignInit
ERROR:esys_crypto:src/tss2-esys/esys_crypto.c:452:iesys_crypto_authHmac() Error ErrorCode (0x00070001)
ERROR:esys:src/tss2-esys/esys_iutil.c:1352:iesys_compute_hmac() HMAC error ErrorCode (0x00070001)
ERROR:esys:src/tss2-esys/esys_iutil.c:1462:iesys_gen_auths() Error while computing hmacs ErrorCode (0x00070001)
ERROR:esys:src/tss2-esys/api/Esys_Clear.c:191:Esys_Clear_Async() Error in computation of auth values ErrorCode (0x00070001)
ERROR:esys:src/tss2-esys/api/Esys_Clear.c:77:Esys_Clear() Error in async function ErrorCode (0x00070001)
ERROR: Esys_Clear(0x70001) - esapi:Catch all for all errors not otherwise specified
ERROR: Unable to run clear
```

Its why I was very thankful for your `--pwd-session` patch, as it works really well.

```
# ./tools/tpm2 clear --pwd-session
```

So I went and wrote a vendor patch for Ubuntu FIPS packages that when the system is in FIPS mode, turn on `--pwd-session` by default.

```
From: Matthew Ruffell <matthew.ruffell@canonical.com>
Date: Mon, 30 Jun 2025 17:02:21 +1200
Subject: Add FIPS mode detection and set pwd session by default
 Add a new function to see if the system is booted into
 fips mode, and if it is, then use the new pwd session
 flag by default. This helps users avoid isses with
 HMAC authentication under FIPS 140-3.
 
 Users can disable if necessary with --disable-pwd-session.
 
 Signed-off-by: Matthew Ruffell <matthew.ruffell@canonical.com>
Bug-Ubuntu: https://bugs.launchpad.net/bugs/2074270
Bug: https://github.com/tpm2-software/tpm2-tss/issues/2889
Origin: vendor
Last-Update: 2025-06-30

Index: tpm2-tools-5.7/lib/tpm2_options.c
===================================================================
--- tpm2-tools-5.7.orig/lib/tpm2_options.c	2026-06-05 17:36:20.672600185 +1200
+++ tpm2-tools-5.7/lib/tpm2_options.c	2026-06-05 17:36:20.672216500 +1200
@@ -273,6 +273,27 @@
 	return TSS2_TCTI_RC_NOT_IMPLEMENTED;
 }
 
+static int is_fips_enabled(void) {
+    #define FIPS_MODE_SWITCH_FILE "/proc/sys/crypto/fips_enabled"
+    const char *fips_switch_path = FIPS_MODE_SWITCH_FILE;
+
+    int fd;
+    char c;
+    int fips_mode;
+
+    fd = open(fips_switch_path, O_RDONLY);
+    if (fd < 0) {
+       fips_mode = 0;
+       return fips_mode;
+    }
+
+    while (read(fd, &c, sizeof(c)) < 0 && errno == EINTR);
+    close(fd);
+
+    fips_mode = c == '1' ? 1 : 0;
+    return fips_mode;
+}
+
 tpm2_option_code tpm2_handle_options(int argc, char **argv,
     tpm2_options *tool_opts, tpm2_option_flags *flags,
     TSS2_TCTI_CONTEXT **tcti) {
@@ -289,11 +310,16 @@
         { "pwd-session",   no_argument,       NULL, 'z' },
+        { "disable-pwd-session",   no_argument,       NULL, 'D' },
     };
 
+    const int fips_mode = is_fips_enabled();
+    if (fips_mode) {
+        flags->restricted_pwd_session = 1;
+    }
 
     /* handle any options */
-    const char* common_short_opts = "T:h::vVQZz";
+    const char* common_short_opts = "T:h::vVQZzD";
     tpm2_options *opts = tpm2_options_new(common_short_opts,
             ARRAY_LEN(long_options), long_options, NULL, NULL, 0);
     if (!opts) {
@@ -377,6 +403,9 @@
         case 'z':
             flags->restricted_pwd_session = 1;
             break;
+        case 'D':
+            flags->restricted_pwd_session = 0;
+            break;
         case 'Q':
             flags->quiet = 1;
             break;
Index: tpm2-tools-5.7/man/common/options.md
===================================================================
--- tpm2-tools-5.7.orig/man/common/options.md	2026-06-05 17:36:20.672600185 +1200
+++ tpm2-tools-5.7/man/common/options.md	2026-06-05 17:36:20.672350290 +1200
@@ -32,6 +32,13 @@
     is passed to the TPM to authorize the action. This option can be used to avoid problems
     when unsalted sessions are used in OpenSSL FIPS mode. If auth values are used
     a salted session should be used for authentication.
+
+    Enabled by default in FIPS mode.
+
+  * **-D**, **\--disable-pwd-session**:
+    Disables password session and returns to using a HMAC session for authentication.
+    Use this option if you find something not working under FIPS mode.
+
   * **-R**, **\--autoflush**:
     Enable autoflush for transient objects created by the command. If a parent
     object is loaded from a context file also the transient parent object will
```

Since most users interact with tpm2-tools through scripts, like clevis etc, they don't want to go and update every single call to tpm2-tools to include --pwd-session. 

But at the same time, when we want to use `--session`, `--pcr`, and `--file`, we don't want `--pwd-session` to be enabled, hence this patch.

You don't have to merge the patch, but I just wanted to open this MR incase any other distro is interested in these patches if they ever implement FIPS SP800-131A. 

Thanks for your time.
Matthew

lib/object.c
matthewruffell: I think this is the better idea, I just wanted to copy https://github.com/tpm2-software/tpm2-tools/blob/844a611ec3be1c7ee76818d68217563cf61cf994/lib/tpm2_auth_util.c#L27-L28 incase you ever wanted to move this to a common header file.

AndreasFuchsTPM:
Hey @matthewruffell
Could you start a new PR with this vendor patch ?
I would however not merge it now, since it breaks compatibility.

However, the next release (for Q4 of 2026) is targeted to be a major version bump with breaking changes anyways, so we could merge it there.

Sounds good ?

matthewruffell:
I apologise for the delay. Here is the final vendor patch we ended up shipping to our users in Ubuntu that sets --pwd-session by default. 

Thanks for your time as always.

PR #3597, opened 2026-07-21
auth: use object nameAlg for implicit PCR sessions

chench246:
## Problem

Implicit `pcr:` authorization creates its policy session before the authorized object is loaded. The session therefore uses the default hash algorithm and fails with `TPM_RC_POLICY_FAIL` when the object policy uses a non-default `nameAlg`.

## Solution

Defer only implicit `pcr:` authorization until the object ESYS handle is available, then use the object `nameAlg` as the policy-session hash. Keep the existing fallback for invalid or non-hash name algorithms.

Other authorization forms are unchanged, including explicitly created sessions passed through `session:`.

## Tests

- Unit tests for object `nameAlg` selection and fallback behavior
- swtpm integration test with a SHA-384 object policy and SHA-256 PCR bank
- Explicit SHA-384 `startauthsession` path remains successful
- Implicit `pcr:sha256:...` succeeds without a new CLI parameter


chench246:
Hi maintainers, a gentle ping on this PR. Could someone take a look when you have time? Please let me know if any changes or additional tests are needed. Thanks!

chench246:
@JuergenReppSIT @AndreasFuchsTPM

AndreasFuchsTPM:
If I remember correctly, the session must always use the hash alg defined as nameAlg for the object being accessed.

So instead of getting this from a CLI parameter, it should be retrieved from the object being accessed.

P.S. Sorry it took so long

chench246:
> If I remember correctly, the session must always use the hash alg defined as nameAlg for the object being accessed.
> 
> So instead of getting this from a CLI parameter, it should be retrieved from the object being accessed.
> 
> P.S. Sorry it took so long

Updated the PR to derive the implicit PCR policy session hash from the accessed object's nameAlg, as suggested. The proposed @<policy-session-hash> CLI syntax has been removed.

The implementation now defers only pcr: authorization until the object ESYS handle is available, then uses the object's nameAlg when creating the implicit policy session. Other authorization forms remain unchanged, including an explicitly created policy session passed through session:.

The GitHub Actions runs are currently waiting for approval. Could you please take another look?

lib/object.c
```diff
@@ -354,7 +355,9 @@ static tool_rc tpm2_util_object_load2(ESYS_CONTEXT *ctx, const char *objectstr,
         bool is_restricted_pswd_session, tpm2_handle_flags flags) {
 
     tool_rc rc = tool_rc_success;
-    if (do_auth) {
+    bool defer_pcr_auth = do_auth && !is_restricted_pswd_session && auth &&
+            !strncmp(auth, "pcr:", sizeof("pcr:") - 1);
```
moritzbuhl, line 359: PCR_PREFIX and PCR_PREFIX_LEN

lib/tpm2_auth_util.c
moritzbuhl, line 445: this smells

lib/tpm2_auth_util.c
```diff
@@ -476,6 +482,23 @@ tool_rc tpm2_auth_util_from_optarg(ESYS_CONTEXT *ectx, const char *password,
     }
 }
 
+tool_rc tpm2_auth_util_from_optarg_with_auth_handle(ESYS_CONTEXT *ectx,
+        const char *password, tpm2_session **session, bool is_restricted,
+        ESYS_TR auth_handle) {
+
+    password = password ? password : "";
```
moritzbuhl, line 489: this smells too

moritzbuhl (changes requested):
I will look into this myself and get this in.

test/integration/tests/unseal.sh
```diff
@@ -89,6 +91,50 @@ unsealed=`tpm2 unseal -V --object-context $file_unseal_key_ctx \
 
 test "$unsealed" == "$secret"
 
+# Test that implicit PCR authorization uses the object's nameAlg. Also verify
+# that an explicitly created policy session keeps using its requested hash.
+if tpm2 getcap algorithms | grep -q '^sha384:'; then
+  tpm2 flushcontext -Q -t
+
+  rm $file_unseal_key_pub $file_unseal_key_priv $file_unseal_key_name \
+  $file_unseal_key_ctx
```
moritzbuhl, line 100: indentation is wrong

test/integration/tests/unseal.sh
moritzbuhl, line 45: indentation is wrong here too.

PR #3599, opened 2026-07-29
tpm2_checkquote: Fix incorrect formatting specifier

tpaukrt:
PRIu64 is not appropriate for sizeof() results.
Use %zu to correctly print size_t values on all architectures.


tools/misc/tpm2_checkquote.c
```diff
@@ -384,15 +384,15 @@ static bool parse_selection_data_from_file(FILE *pcr_input,
         // Convert TPML_DIGEST from little endian to host endian.
         pcrs->pcr_values[j].count = le32toh( pcrs->pcr_values[j].count);
         if (pcrs->pcr_values[j].count > ARRAY_LEN(pcrs->pcr_values[j].digests)) {
-            LOG_ERR("Malformed PCR file, TPML_DIGEST count cannot be greater than %" PRIu64,
+            LOG_ERR("Malformed PCR file, TPML_DIGEST count cannot be greater than %zu",
```
AndreasFuchsTPM, line 387: Should we maybe use `PRIuPTR` instead of `zu`?

JuergenReppSIT: > Should we maybe use `PRIuPTR` instead of `zu`?

"zu" is frequently used in the tools so far, so it could likely be used.

AndreasFuchsTPM: ok, let's stick to zu then...

Issue #1798, opened 2019-10-22
NV Attributes NT field should work without nt= specifier


williamcroberts:
Rather than use nt=<arg> format for specifying the NT field, it would be simpler if you could just "or" in counter or bits into the tpma nv string, like so:
`"-a "counter|ownerread|ownerwrite|authread|authwrite"

PR #3600, opened 2026-08-03
fix(attr_util): correct NV nt field decoding and show it as a friendly name

### LINKED ISSUE #1797: nt decoding scheme broken
The nt decoder goes bit by bit deciphering the string, and uses a field width to decipher bit sections when the width is greater than 1. It generates a mask using field width, which is decending when set in the array:
```C
    dispatch_arg_add(nt, 4),              //  4
    dispatch_arg_add(nt, 3),              //  5
    dispatch_arg_add(nt, 2),              //  6
    dispatch_arg_add(nt, 1),              //  7
```
The bitwise logic uses field width to grab the bit, and shift it, which causes NT to always be displayed as 1.

We need to re-work this logic so this is performed correctly and we can decode the friendly names as well.

katexochen:
Fixes https://github.com/tpm2-software/tpm2-tools/issues/1797, decided against implementing https://github.com/tpm2-software/tpm2-tools/issues/1798.

**fix(attr_util): decode NV nt field relative to its base**

tpm2_attr_util_nv_attrtostr renders the 4-bit TPM_NT field (TPMA_NV
bits 7:4) by walking to the lowest set bit and shifting the value by
that bit index instead of by the field base (bit 4). So any nt value
whose low bit is clear was mis-rendered: bits (0x2) and extend (0x4)
both printed as "nt=0x1", and pinfail (0x8) landed on the width-1
slot and printed a bare "nt" with no value. Only values with bit 4
set (counter 0x1, pinpass 0x9, ...) came out right, which is why the
existing tests never caught it.

Anchor extraction on the field base (bit_index & ~(w - 1)) and give
all four nt table slots the full field width (4), so the whole nibble
is read as (attrs >> 4) & 0xF regardless of which bit the scanner
lands on. The same expression reduces to bit_index for single-bit
attributes (w == 1), so their behavior is unchanged.

**feat(attr_util): render NV nt field as a friendly name**

The attribute parser already accepts friendly TPM_NT names
(nt=extend, nt=counter, ...), but tpm2_attr_util_nv_attrtostr only
ever emitted the raw field value (nt=0x4), so tpm2_nvreadpublic's
"friendly" output wasn't symmetric with the input it accepts.

Map the decoded nt field value back to its name, falling back to hex for
any value that isn't a defined type. Only the human-readable "friendly:"
string changes (e.g. nt=0x9 -> nt=pinpass); the raw "value:" field is
unchanged.

Keep the nt= prefix so the string still round-trips through
tpm2_nvdefine -a, which parses this same format.

### DIFF
--- a/lib/tpm2_attr_util.c
+++ b/lib/tpm2_attr_util.c
@@ -240,9 +240,9 @@ static dispatch_table nv_attr_table[] = { // Bit Index
     dispatch_no_arg_add(authwrite),       //  2
     dispatch_no_arg_add(policywrite),     //  3
     dispatch_arg_add(nt, 4),              //  4
-    dispatch_arg_add(nt, 3),              //  5
-    dispatch_arg_add(nt, 2),              //  6
-    dispatch_arg_add(nt, 1),              //  7
+    dispatch_arg_add(nt, 4),              //  5
+    dispatch_arg_add(nt, 4),              //  6
+    dispatch_arg_add(nt, 4),              //  7
     dispatch_reserved(8),                 //  8
     dispatch_reserved(9),                 //  9
     dispatch_no_arg_add(policydelete),    // 10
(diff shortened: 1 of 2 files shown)


AndreasFuchsTPM:
I added #3605 to completely rework this, so I'll close this one here for now.

@katexochen Would you need a backport ?

katexochen:
> katexochen Would you need a backport ?

No backport needed, thanks.

PR #3605, opened 2026-08-11
lib: simplify attr_util

### LINKED ISSUE #1797: nt decoding scheme broken

AndreasFuchsTPM:
Instead of a table driven dispatch logic, use a simple linear switch case like approach.

So I did break API here.
Old output: "authread|authwrite|nt=0x2"
New output: "authread|authwrtie|extend"
Parsing still accepts all variants though...

I like this much more; oppinions?

Also question: Should the tools return a bunch of "<reserved(8)>" reports or should the library's to text rather fail here ?

Fixes: #1797
Implements: #1798
Superseeds: #3600


moritzbuhl:
The following things are still needed to match the old functionality:

reserved bits are no longer allowed, old code wrote `<reserved(...)>`:
```
#define dispatch_reserved(pos) \
    { .name = "<reserved("xstr(pos)")>", .callback=NULL, .width = 1 }
 ```
 
 
nt=0x... and formatting values other than the named ones from the enum.


boolean tokens accept an equals sign, e.g. `authread=garbage`

File: lib/tpm2_attr_util.c [hunk 1]
```diff
@@ -8,631 +8,198 @@
 #include "log.h"
 #include "tpm2_attr_util.h"
 
-#define dispatch_no_arg_add(x) \
-    { .name = str(x), .callback=(action)x, .width = 1 }
+typedef struct _attrstrmap {
```
moritzbuhl, line 11: `typedef struct {`

lib/tpm2_attr_util.c
```diff
@@ -8,631 +8,198 @@
...
+    *attr = 0;
+
+    /* Unfortunately the value for TPM2_NT_ORDINARY is zero.
+       As such we cannot just check if the masked out *attr value is non-zero below
+       for conflict detection, but we have to do some special handling. */
+    if (strcasestr(str, "ordinary")) {
+        tpm_nt_ordinary = true;
+    }
+
+    while ((token = strtok_r((save)? NULL : str, "|", &save))) {
```
moritzbuhl: for (token = strtok_r(str, "|", &save); token != NULL; token = strtok_r(NULL, "|", &save)) {

moritzbuhl: So FreeBSD's libc sets `save = NULL` (https://github.com/freebsd/freebsd-src/blob/main/lib/libc/string/strtok.c#L67), glibc does not: https://github.com/lattera/glibc/blob/master/string/strtok_r.c#L59

moritzbuhl (approved):

PR #3609, opened 2026-08-17
Add FIPS mode detection and set pwd session by default

matthewruffell:
This is a vendor patch carried in Ubuntu FIPS 140-3 packages.

This is the actual patch we went with, and it is a refactor of the older version I posted in https://github.com/tpm2-software/tpm2-tools/pull/3587#issuecomment-4748721716

This patch has a nice function that asks openssl if the system has FIPS mode enabled, and it returns the result.

The patch uses the FIPS detection logic to set the --pwd-session flag by default for FIPS users.

It still needs the changes from #3587 to really work seamlessly for FIPS users.

I am providing this in the chance it happens to be helpful, but you don't have to merge it. This would be very useful for distros needing to support FIPS SP800-131A.

### DIFF
--- a/lib/tpm2_options.c
+++ b/lib/tpm2_options.c
@@ -17,6 +17,9 @@
 #include <sys/wait.h>
 #include <sys/stat.h>
 
+#include <openssl/evp.h>
+#include <openssl/provider.h>
+
 #include "config.h"
 #include "log.h"
 #include "tpm2_options.h"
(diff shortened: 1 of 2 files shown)


moritzbuhl:
I think we need to rerun the CI once #3602 is in.

moritzbuhl:
Sorry, I just updated the CI and removed EOL distributions. Could you rebase? I hope this also helps you with the OpenSSL providers.

lib/tpm2_options.c
```diff
@@ -290,6 +311,7 @@ tpm2_option_code tpm2_handle_options(int argc, char **argv,
         { "version",       no_argument,       NULL, 'v' },
         { "enable-errata", no_argument,       NULL, 'Z' },
         { "pwd-session",   no_argument,       NULL, 'z' },
```
AndreasFuchsTPM, line 313: I think, I would prefer `--pwd-session=disabled` by introducing an opt_arg here.
Would that work for you ?

PR #3622, opened 2026-09-02
tpm2_import: zeroize plaintext private key material before return

### LINKED ISSUE #3621: tpm2_import: plaintext private key and duplication keys never zeroed in process_input_key()
## Summary

`process_input_key()` in `tools/tpm2_import.c` holds plaintext private-key material and duplication keys in stack-local `TPM2B_*` buffers that are never cleared before the function returns. The buffers are left populated with real key material on every code path, including the multiple `goto out;` early-return paths.

## Affected buffers

In `process_input_key()` (`tools/tpm2_import.c`):

- `private_sensitive` (`TPM2B_SENSITIVE`, declared ~line 308) — populated by `tpm2_openssl_import_keys()` with the plaintext private key/seal data being imported, plus the generated duplication seed (`private_sensitive.sensitiveArea.seedValue`).
- `hmac_key` / `enc_key` (`TPM2B_MAX_BUFFER`, declared ~line 333-334) — the outer-integrity HMAC key and duplicate-sensitive encryption key derived from that seed via `tpm2_identity_util_calc_outer_integrity_hmac_key_and_dupsensitive_enc_key()`.

None of these are cleared. The function falls through to:

```c
out:
    if (free_ppub) {
        Esys_Free(parent_pub);
    }

    return rc;
```

with no `memset`/zeroing step, so the plaintext private key and derived duplication keys remain resident in the process's stack memory for the remaining lifetime of the (short-lived, but not instantaneous) CLI invocation — recoverable via a core dump, `ptrace`, swapped memory, or a cold-boot style attack during that window.

## Suggested fix

Clear these buffers before every return from `process_input_key()`, ideally with a zeroing primitive immune to dead-store elimination by the compiler (a plain `memset` can be optimized away since the buffers aren't read again afterward). tpm2-tss recently added exactly this kind of helper (`secure_mem_zero` in `src/util/aux_util.h`, added in tpm2-software/tpm2-tss#3072, addressing tpm2-software/tpm2-tss#2994 — the same class of issue for ESAPI session/auth material) — tpm2-tools could either vendor an equivalent helper or, if tpm2-tools depends on a new-enough tpm2-tss, use that one directly once #3072 merges.

## Verification

Read the full body of `process_input_key()` end-to-end in the current `master` branch and confirmed:
- all three buffers are stack-allocated (not heap, so `free()`-based zeroing patterns don't apply directly — the fix needs to run at every `out:`/return site instead)
- none of them are cleared on any path, including the early `goto out` error paths
- this is distinct from, and not covered by, tpm2-tss#3072 (which only touches the tpm2-tss library, not tpm2-tools)

Happy to submit a PR for this if useful — wanted to raise the issue first in case there's a preferred zeroing helper/convention for tpm2-tools I should follow.

Johnsonajibi:
## Summary

Fixes #3621. `process_input_ossl_import()` in `tools/tpm2_import.c` (referred to as `process_input_key()` in the original issue text — the actual function name is `process_input_ossl_import`) holds plaintext private-key material and derived duplication keys in stack-local `TPM2B_*` buffers that were never cleared before the function returns, on any path — including the early `goto out` error paths:

- `private_sensitive` (`TPM2B_SENSITIVE`) — populated by `tpm2_openssl_import_keys()` with the plaintext private key/seal data being imported, plus the generated duplication seed (`private_sensitive.sensitiveArea.seedValue`).
- `hmac_key` / `enc_key` (`TPM2B_MAX_BUFFER`) — the outer-integrity HMAC key and duplicate-sensitive encryption key derived from that seed.

## Fix

Clear all three with `OPENSSL_cleanse()` at the `out:` label, so they're wiped on every return path. `OPENSSL_cleanse` is used specifically instead of `memset`/`bzero` because it's guaranteed not to be optimized away by the compiler as a dead store (the buffers aren't read again after this point) — this project already links OpenSSL and uses it a few lines above (`RAND_bytes`), so no new dependency is introduced.

All three variables are declared at the top level of the function body (not inside a nested block), so they remain in scope at the `out:` label reached by every earlier `goto out;`, including the ones before their declaration point — legal in C, and safe here since `OPENSSL_cleanse` only writes through the pointer, it never reads the (possibly not-yet-assigned) prior value.

## Verification

Built `tools/tpm2_import.o` against a from-source tpm2-tss (ESAPI 4.x-equivalent) with the project's full strict warning set (`-Wall -Wextra -Wformat -Wformat-security -Wstack-protector -Wstrict-overflow=5 -Wstringop-overflow=4 -Wstringop-truncation -Wduplicated-branches -Wduplicated-cond -Wbool-compare -Werror`, `_FORTIFY_SOURCE=2`) — compiles clean, no new warnings.

I wasn't able to link the full `tpm2-tools` binary suite in my local environment (the `fapi`/`tss2` tool fails to link only because `libcurl` in this sandbox pulls in the system's OpenSSL 3.0 while I built a private OpenSSL 3.5 for the `tss2-esys` version requirement — an environment/toolchain mismatch unrelated to this change, not something the patch touches). `tpm2_import.c` itself has no dependency on `libcurl`/`fapi`, and its object file builds and links warning-free in isolation.


moritzbuhl:
I love the idea behind this change and I would love to see this get into tpm2-tools.
A few remarks: `OPENSSL_cleanse` is not the right function in my opinion, we should try to align this with what https://github.com/tpm2-software/tpm2-tss/pull/3072 is doing.

Further I would prefer a test that tries to validate that the cleared memory is in fact clear and that no new compiler optimizations have removed it for some reason.

Finally, I am not familiar enough with tpm2-tools yet but I would be very surprised if this is the only location in code that could benefit from zeroing secrets before resource deallocation.

moritzbuhl (changes requested):
Please rework according to the comments on the PR and issue.

Johnsonajibi:
Thanks for the review -- addressed:

- Switched from `OPENSSL_cleanse` to a new `tpm2_util_secure_mem_zero()` macro in `lib/tpm2_util.h`, using the same portable approach tpm2-tss took in #3072's `secure_mem_zero` (`explicit_bzero` on glibc >= 2.25 / recent FreeBSD, `SecureZeroMemory` on Windows, a volatile-pointer byte loop elsewhere) rather than depending on OpenSSL specifically for this.
- Added `test/unit/test_tpm2_util.c` coverage that fills a buffer with a non-zero pattern, calls the macro, and asserts via `assert_memory_equal()` against an all-zero buffer that it's actually been cleared -- plus a NULL/zero-size no-op case.
- On the third point (other locations that could benefit from this) -- agreed, and now that there's a reusable `tpm2_util_secure_mem_zero()` in place it should make future passes over the rest of the codebase for the same pattern much easier. I kept this PR scoped to the specific site I verified (`tpm2_import.c`) rather than expanding it, but happy to look at other candidates in a follow-up if useful.

AndreasFuchsTPM:
This is the failure on opensuse-leap, but only there

```
spawn sh -c tss2 sign --keyPath=HS/SRK/mySignKey1 --digest=digest.file --signature=test_signature.file --force 2> /tmp/tpm2_test_I6a9n5/tss2_fapi.bp2LHJ/log.file
PolicySigned: Authorize usage of P_ECCP256SHA256/HS/SRK/mySignKey1 by signing the nonce with the key corresponding to the fingerprint "SHA256:Gf8QGWT2MOy8tOfpe6m+6MUUm2GUijGPkvCTjtJPOJz//XMom+k+7OaWmA==" and the hash algorithm "sha1".
** (tpm2-abrmd:57610): WARNING **: 10:01:21.589: read_data: read on istream produced error: Error receiving data: Connection reset by peer
Please enter a valid file path
ERROR:fapi:src/tss2-fapi/ifapi_policy_callbacks.c:837:ifapi_sign_buffer() ErrorCode (0x00060001) Execute policy signature callback. 
ERROR:fapi:src/tss2-fapi/ifapi_policy_execute.c:446:execute_policy_signed() ErrorCode (0x00060001) Execute policy signature callback. 
ERROR:fapi:src/tss2-fapi/ifapi_policy_execute.c:1528:execute_policy_element() ErrorCode (0x00060001) Execute policy signed 
ERROR:fapi:src/tss2-fapi/ifapi_policy_execute.c:1737:ifapi_policyeval_execute() Execute policy. ErrorCode (0x00060001) 
ERROR:fapi:src/tss2-fapi/ifapi_policyutil_execute.c:301:ifapi_policyutil_execute() Execute policy. ErrorCode (0x00060001) 
ERROR:fapi:src/tss2-fapi/fapi_util.c:2116:ifapi_authorize_object() Execute policy. ErrorCode (0x00060001) 
ERROR:fapi:src/tss2-fapi/fapi_util.c:2883:ifapi_key_sign() Authorize signing key ErrorCode (0x00060001) 
```

@JuergenReppSIT Do you see the reason immediately ?

AndreasFuchsTPM:
So, this error should be fixed with https://github.com/tpm2-software/tpm2-tools/pull/3620 
Could you please rebase onto master for testing ?

Johnsonajibi:
Rebased onto master (24eb435), picking up #3620. Thanks for confirming the opensuse failure was unrelated.

